PR #2, opened 2019-10-08
(DOCSP-7048): Data Model - Realms

nlarew:
## Jira

- [(DOCSP-7048): Data Model - Realms](https://jira.mongodb.org/browse/DOCSP-7048)

## Staged

- [Data Model: Realms](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nick/data-model/data-model/realms.html)

### DIFF
--- a/source/data-model/realms.txt
+++ b/source/data-model/realms.txt
@@ -1,0 +1,205 @@
+======
+Realms
+======
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+A **realm** is a set of related objects that conform to a pre-defined
+schema and share user-level access permissions. Realms may contain more
+than one type of data as long as a schema exists for each type.
+
+Realms allow you to partition data according to who uses it and when
+they use it. Every realm stores data in a separate realm file that
+contains a binary encoding of each object in the realm. You can
+automatically synchronize realms across multiple devices and set up
+reactive event handlers that call a function any time an object in a
+realm is created, modified, or deleted.
+
+Comparison with Other Databases
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The MongoDB Realm data model is similar to both relational and document
+databases but has distinct differences from both. To underscore these
+differences, it's helpful to highlight what a realm **is not**:
+
+A realm is not a single, application-wide database.
+    Unlike other applications, which store all of their data in a single
+    database, Realm apps often split data across multiple realms to
+    organize data more efficiently and to enforce access controls.
+
+A realm is not a table.
+    Tables typically only store one kind of information, such as street
+    addresses or items in a store inventory, whereas a realm can contain
+    any number of object types.
+
+A realm is not a collection of schemaless documents.
+    Application objects are similar to documents, but every object in a
+    realm conforms to a defined schema for its object type. An object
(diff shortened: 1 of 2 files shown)


source/data-model/realms.txt
```diff
@@ -0,0 +1,173 @@
...
+
+A **realm** is a set of related objects that conform to a pre-defined
+schema and share user-level access permissions. Realms may contain more
+than one type of data as long as a schema exists for each type.
+
+Realms allow you to partition data according to who uses it and when
+they use it. Every realm stores data in a separate realm file that
+contains a binary encoding of each object in the realm. You can
+automatically synchronize realms across multiple devices and listen for
+notification events any time an object in a realm is created, modified,
```
cbush: Nitpick: "listen for notification events" - could this phrasing be tightened up?

source/data-model/realms.txt
```diff
@@ -0,0 +1,173 @@
...
+
+Realm Schema
+~~~~~~~~~~~~
+
+A **realm schema** is a list of valid :doc:`object schemas
+</data-model/objects>` that each define an object type that the realm
+may contain. Object schemas specify relationships between objects in the
+realm as well as field-level constraints properties such as the data
+type of each field, whether or not a field is required, and the default
+value for optional fields.
```
cbush: Wouldn't it be nice at this point to illustrate what a schema looks like?

>Here's an example of a schema in Swift: `blah`.

Doesn't have to be detailed but it might help break up the text.

source/data-model/realms.txt
```diff
@@ -0,0 +1,173 @@
...
+
+Client applications provide a realm schema with a specific schema
+version number when they first open a realm. If a realm already contains
```
cbush: First open? Or just open?

source/data-model/realms.txt
```diff
@@ -0,0 +1,173 @@
...
+       Write access requires read access, so the ``mayWrite``
+       permission also sets ``mayRead`` to ``true`` for the user.
+
+   * - ``mayManage``
+     - If ``true``, the user has read and write access to all objects in
+       the realm and can set permissions on the realm for other users.
+       
+       Manage access requires write access, so the ``mayManage``
+       permission also sets ``mayWrite`` and ``mayRead`` to ``true`` for
+       the user.
```
cbush: Of course, we will need links to a section on how to actually set permissions.

source/data-model/realms.txt
nlarew: Yeah we could include that here. I didn't initially because this section is for realm schemas which are just arrays of object schemas. Definitely going to include an example for the object schema section.

Do you think the example here should be tabbed across languages or would it suffice to have a basic JSON schema?

source/data-model/realms.txt
nlarew: Yeah definitely. I'm thinking this will link to a couple of By Example pages/sections:
- **Configure the Default Permissions for a Realm**
- **Manage a User's Permissions for a Realm**

source/data-model/realms.txt
nlarew: Buffed this up a bit:

```rst
You can automatically synchronize realms across multiple devices and set
up reactive event handlers that call a function any time an object in a
realm is created, modified, or deleted.
```

source/data-model/realms.txt
cbush: I think it would suffice to have a basic JSON schema. Really, just looking for an illustration to break up the text and show readers what you're talking about before getting too deep.

PR #3, opened 2019-10-15
Data Model - Objects & Relationships

nlarew:
## Jira

- [(DOCSP-7049): Data Model - Objects](https://jira.mongodb.org/browse/DOCSP-7049)
- [(DOCSP-7050): Data Model - Relationships](https://jira.mongodb.org/browse/DOCSP-7050)

## Staged

- [Objects](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nick/data-model/data-model/objects.html)
- [Relationships](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nick/data-model/data-model/relationships.html)

### DIFF
--- a/source/data-model/objects.txt
+++ b/source/data-model/objects.txt
@@ -16,21 +16,22 @@ Overview
 Realm applications model data as objects composed of field-value pairs
 that each contain one or more primitive data types or other Realm
 objects. Realm objects are essentially the same as regular objects in
-most programming languages but they also include additional features
+most programming languages, but they also include additional features
 like :ref:`real-time updating data views <live-object>` and reactive
 change event handlers.
 
-Every object in a realm has a *type* which is analagous to an object
-class. Object types configure the properties and relationships of a
-class of objects with an :ref:`object schema <object-schema>`. Realm
-client applications typically use native classes in their respective
-language as the schema for an object type.
+Every Realm object has an *object type* that refers to the object's
+class. Objects of the same type share an :ref:`object schema
+<object-schema>` that defines the properties and relationships of those
+objects. In most languages you define object schemas using the native
+class implementation.
 
 .. example::
 
    The following code block shows an object schema that describes a Dog.
-   In most most languages Realm uses native classes to define the
-   schema.
+   Every Dog object must include a ``name`` and ``age`` and may
+   optionally include the dog's ``breed`` and a reference to a Person
+   object that represents the dog's ``owner``.
 
    .. include:: /examples/DogSchema/DogSchema.rst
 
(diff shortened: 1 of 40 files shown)


source/data-model/objects.txt
```diff
@@ -0,0 +1,161 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Realm applications model data as objects composed of field-value pairs
+that each contain one or more primitive data types or other Realm
+objects. Realm objects are essentially the same as regular objects in
+most programming languages but they also include additional features
```
cbush: languages but -> languages, but

source/data-model/objects.txt
```diff
@@ -0,0 +1,161 @@
...
+like :ref:`real-time updating data views <live-object>` and reactive
+change event handlers.
+
+Every object in a realm has a *type* which is analagous to an object
```
cbush: I'm not sure of the value of the "which is analagous to an object class" part. If the intention is to compare to OOP, maybe we can frame it more explicitly that way:

>As in OOP, an object has a class. In Realm, this is called a Type.

But even then, I don't know if we need to belabor this point, especially since you connect it to "classes" a few sentences later.

source/data-model/objects.txt
```diff
@@ -0,0 +1,161 @@
...
+work with object-oriented data natively without an :wikipedia:`ORM
+<Object-relational_mapping>` tool. Live objects update automatically to
+reflect data changes, including synced remote changes, and emit
+notification events that you can subscribe to whenever their underlying
+data changes.
+
+Live objects are direct proxies to the underlying stored data, which
+means that a live object doesn't directly contain data. Instead, a live
+object :wikipedia:`lazy loads <Lazy_loading>` data when a read operation
+accesses it. This means that a realm can contain many objects but only
```
cbush: Is this accurate? I'm surprised. I thought Live Objects were memory mapped to the data on disk, which means they always point to the most up-to-date data. The point about them being live is that they always reflect the current state even if they were modified through a different handle.

source/data-model/objects.txt
```diff
@@ -0,0 +1,161 @@
...
+Primary Key
+~~~~~~~~~~~
+
+A **primary key** is a String or Integer property that uniquely
+identifies an object. You may optionally define a primary key for an
+object type as part of the :ref:`object schema <object-schema>`.
+Realm automatically indexes primary key properties, which allows you to
+efficiently read and modify objects based on their primary key.
+
+Primary key properties are always required and the value of an object's
```
cbush: There's always required by you may optionally define them? Can you clarify?

source/data-model/objects.txt
```diff
@@ -0,0 +1,161 @@
...
+
+.. example::
+
+   The following code block shows an object schema that describes a Dog.
+   In most most languages Realm uses native classes to define the
+   schema.
+
+   .. include:: /examples/DogSchema/DogSchema.rst
+
+Key Concepts
```
cbush: I do wonder why this should be a main heading with the rest of the page under it? What's the harm in having the rest of the items be top level items? Not feeling that strongly about it, just curious.

source/data-model/objects.txt
```diff
@@ -0,0 +1,161 @@
...
+.. _property-schema:
+
+Property Schema
+~~~~~~~~~~~~~~~
+
+A **property schema** is a field-level configuration that defines and
+constrains a specific property in an object schema. Every object schema
+must include a property schema for each property in the object. At
+minimum, a property schema defines a property's data type and indicates
+whether or not the property is required
```
cbush: required -> required.

source/data-model/relationships.txt
```diff
@@ -0,0 +1,107 @@
...
+
+.. _to-one-relationship:
+
+To-One Relationship
+~~~~~~~~~~~~~~~~~~~
+
+A **to-one** relationship means that an object is related in a specific
+way to no more than one other object. You define a to-one relationship
+for an object type in its :ref:`object schema <object-schema>` by
+specifying a property where the type is the related Realm object type.
```
cbush: It might be nice to add a written example after these. Something like, "For example, suppose you have a Person type and a Dog type in your realm. If each person can have one dog, the Person schema would have a Dog property."

source/examples/DogSchema/DogSchema.js
```diff
@@ -0,0 +1,9 @@
+{
```
cbush: I wonder if it would be nice to do something like `const Dog = {` on the first line here.

source/data-model/objects.txt
nlarew: "Analagous" comes from JS being weird about using classes. Technically every other language just uses the class name but we need to have a generic "type" to cover the JS object schemas that are defined with JSON instead of classes.

I may be overthinking this. I'll play around and remove the "analagous"

source/data-model/objects.txt
nlarew: My goal here is to give the page some structure, both for readers and writers.

- Readers should be able to quickly parse whether a section has the content that they need.
- Writers should have guidance on what does and does not go in a given section.

I think it works on this page because the **Data Model** section is "referencey". More guided or beginner-focused content can probably be less rigid since it doesn't need to be nearly as exhaustive.

I can also see us adding other sections to this page beyond just key concepts (though I'm not sure exactly what right now) and this gives us the space to do it.

source/data-model/objects.txt
nlarew: I do believe that the lazy load info is correct but agreed that the focus should be more on the instant updates, which just happen to be performant thanks to lazy load.

nlarew:
@cbush Updated per your comments. Mind taking another look when you get a chance?

PR #5, opened 2019-10-28
(DOCSP-7055): Realm notifications

cbush:
- Adds the notifications page, first of the Read & Write section
- Ignores kotlin examples for now as they are not readily available
and will have to be added later
- Added "Silent write" examples, but they only apply to iOS SDKs.
So they will be useful later in the SDK-specific guides section.

### Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/notifications/read-and-write/notifications.html

### JIRA
https://jira.mongodb.org/browse/DOCSP-7055


source/examples/Notifications/CollectionNotification.cs
```diff
@@ -0,0 +1,7 @@
+var token = realm.All<Person>().SubscribeForNotifications ((sender, changes, error) =>
+{
+    // Access changes.InsertedIndices, changes.DeletedIndices, and changes.ModifiedIndices
```
nlarew: Would be nice to have some code the demonstrates how to iterate over each set of changes like we do in other language examples

source/examples/Notifications/RealmNotification.cs
```diff
@@ -0,0 +1,4 @@
+realm.RealmChanged += (s, e) =>
+{
+    // Update UI
+}
```
nlarew: This could use more details (what are `s` and `e`?)

source/examples/Notifications/RealmNotification.js
```diff
@@ -0,0 +1,12 @@
+function updateUI() {
```
nlarew: Does Realm pass any arguments to this callback?

source/examples/Notifications/RealmNotification.js
```diff
@@ -0,0 +1,12 @@
...
+}
+
+// Observe Realm Notifications
+realm.addListener('change', updateUI);
+
+// ..later remove the listener
+realm.removeListener('change', updateUI);
+
+// ..or unregister all listeners
+realm.removeAllListeners();
```
nlarew: nit: add a newline

source/examples/Notifications/RealmNotification.swift
```diff
@@ -0,0 +1,8 @@
+// Observe realm notifications. Keep a strong reference to the notification token
+// or the observation will stop.
+let token = realm.observe { notification, realm in
```
nlarew: What's in the `notification` arg?

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,69 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Realm's notification system allows you to watch for and
+react to changes in your data.
+
+Realm emits three kinds of notifications:
```
nlarew: Could probably add a bit more flavor to this overview section.
- Why would I want to watch for and react to changes?
  - dynamic/reactive UI updates
  - logging?
  - additional validation?
- How do I watch for and react to changes?
  - add a listener to the realm/collection/object. Realm sends notification events to registered listeners.
  - this is sort of already covered by the other sections on the page but a more generic wording in the overview might be good for skimmers.
- Maybe we should include a note that talks about the importance of closing listeners that you don't need?

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,69 @@
...
+
+- :ref:`realm notifications <realm-notifications>` when the realm as a whole is changed;
+- :ref:`collection notifications <collection-notifications>` when objects are changed, added, or removed; and
+- :ref:`object notifications <object-notifications>` when a specific object is changed or removed.
```
nlarew: nit: my personal preference is to avoid bullet points as a linear continuation of a sentence because it leads to some extra boilerplate (i.e. semicolons and conjunctions). Not a big deal but thought I'd mention it.

nlarew: - "...when the realm as a whole is changed" what kind of changes are realm-level? Is it only write operations?
  - schema version? (probably not)
  - permissions? (maybe)
- "...whenever **one or more** objects **in a collection** are changed, added, or removed"

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,69 @@
...
+
+.. _realm-notifications:
+
+Realm Notifications
+-------------------
+You can register a notification handler on an entire realm.
+The handler will react whenever any write transaction
+involving that Realm is committed.
```
nlarew: - What does it mean for the handler to react?
- What kind of info does Stitch pass to the listener?
- When is it useful to use realm notifications vs others?

The text in the **Collection Notifications** section is a pretty good example of what we could buff up here.

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,69 @@
...
+You can register a notification handler on a specific
+collection within a realm. The handler receives a
+description of changes since the last notification.
+Specifically, this description consists of the indices of
+objects that have been added, removed, or modified.
+
+Realm emits an initial notification after retrieving the
+collection. After that, Realm delivers collection
+notifications asynchronously whenever a write transaction
+adds, changes, or removes objects in the collection.
```
nlarew: - When is it useful to use collection notifications vs others?

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,69 @@
...
+.. _object-notifications:
+
+Object Notifications
+--------------------
+
+You can register a notification handler on a specific object
+within a realm. Realm will notify your handler:
+
+- when the object is deleted, or
+- when any of its managed properties change.
```
nlarew: What do you mean by "managed" properties here?

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,69 @@
...
+
+.. _collection-notifications:
+
+Collection Notifications
+------------------------
+You can register a notification handler on a specific
+collection within a realm. The handler receives a
+description of changes since the last notification.
+Specifically, this description consists of the indices of
+objects that have been added, removed, or modified.
```
nlarew: Maybe mention that the notification groups changes of the same type? (i.e. additions, modifications, deletions)

source/read-and-write/notifications.txt
cbush: To your first comment, I would kindly point out that the style guide has an answer:
>When listing three or more items, use a bullet list instead of embedding the items in a paragraph.
https://docs.mongodb.com/meta/style-guide/writing/write-clear-concise-sentences-paragraphs/#create-short-paragraphs

This is in the interest of readability when the listed clauses are parallel anyway. Arguably this could be three full sentence bullet points, but then we'd be repeating a phrase like "Realm emits... when..." over and over. What for?

source/read-and-write/notifications.txt
cbush: Good call. I had an entire section about how to subscribe, but it's not consistent across SDKs so the text was hard to make generic to the point that it felt meaningless. I'll give it another go.

source/read-and-write/notifications.txt
cbush: I did the same for realm and collection, but object notifications may be best left to the imagination. My example seemed redundant, but I can add it back if you think it's still lacking in the object notification section.

source/read-and-write/notifications.txt
cbush: I just removed the word. It was stated in the original docs, but without going into SDK specifics it's just more confusing than it's worth.

source/read-and-write/notifications.txt
nlarew: Ah sorry I was definitely unclear here. I like the bullet points, I meant specifically the use of semicolons and the "and" in the second to last bullet point.

cbush: Oh I see. Yeah, it looks more legalese this way. How do we capitalize and punctuate it exactly?

source/examples/Notifications/CollectionNotification.m
```diff
@@ -0,0 +1,35 @@
...
+
+        // Query results have changed, so apply them to the UITableView
+        [tableView beginUpdates];
+        [tableView deleteRowsAtIndexPaths:[changes deletionsInSection:0]
+                         withRowAnimation:UITableViewRowAnimationAutomatic];
+        [tableView insertRowsAtIndexPaths:[changes insertionsInSection:0]
+                         withRowAnimation:UITableViewRowAnimationAutomatic];
+        [tableView reloadRowsAtIndexPaths:[changes modificationsInSection:0]
+                         withRowAnimation:UITableViewRowAnimationAutomatic];
+        [tableView endUpdates];
```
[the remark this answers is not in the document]
cbush: That's for realm notifications vs. collection notifications. Every collection notification snippet now runs through the index collections in the proper order. It's handy to have the standard UI list update procedure here for Android and iOS, but on .NET and JS I just left it generic since there might be totally different ways to handle it.

source/examples/Notifications/RealmNotification.js
cbush: It passes the realm itself, which a user can see in the SDK docs. In the world of this example it wouldn't necessarily make sense to accept the realm as a parameter to a function simply called "UpdateUI", but it could be changed to be a more specific-to-realm callback. Anyway, it's now explained in the expanded Realm notification section that these notifications don't get a lot of information about what actually changed.

source/examples/Notifications/RealmNotification.swift
cbush: It's basically an enum saying what type of notification just occurred, but the use case is more complicated than I'd like to get into here. Added a comment referring to the SDK docs.

source/examples/Notifications/ObjectNotification.js
```diff
@@ -0,0 +1,26 @@
+async function example() {
```
nlarew: For some reason this isn't syntax highlighting on the staged version. Not sure why 😕

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,124 @@
...
+.. _object-notifications:
+
+Object Notifications
+--------------------
+
+You can register a notification handler on a specific object
+within a realm. Realm will notify your handler:
+
+- When the object is deleted.
+- When any of its properties change.
```
nlarew: nit: **the object's** properties

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,124 @@
...
+2. Add the notification handler to the realm, collection, or object that you want to observe.
+3. Receive a notification token from the call to add the handler. Retain this token as long as you want to observe.
+4. When you are done observing, invalidate the token.
+
+.. _realm-notifications:
+
+Realm Notifications
+-------------------
+You can register a notification handler on an entire realm.
+Realm will notify the notification handler whenever any
```
nlarew: nit: Realm will notify -> Realm calls
- Avoids future tense
- "Call" feels more closely connected to the function than "notify"

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,124 @@
...
+handler receives no information about the change.
+
+This is useful when you want to know that there has been a
+change but do not care to know specifically what changed.
+For example, suppose you are writing a real-time
+collaborative app. To give the sense that your app is
+buzzing with collaborative activity, you want to have an
+indicator that lights up when any change is made. In that
+case, a realm notification handler would be a great way to
+drive the code that controls the indicator.
```
nlarew: IMO the "For example" text (line 55-60) would make sense as part of the example block. Then we have a connection between a concrete example and the code that would/could make that happen.

Not sure if this is something we want to do now / prioritize but a thought.

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,124 @@
...
+notifications asynchronously whenever a write transaction
+adds, changes, or removes objects in the collection.
+
+Unlike realm notifications, collection notifications contain
+detailed information about the change. This enables
+sophisticated and selective reactions to changes. Collection
+notifications provide all the information needed to manage a
+list or other view that represents the collection in the UI.
+
+.. example::
```
nlarew: For this one it might make sense to add a bit of text that frames this as a developer listening for changes in some app to update the UI.

source/read-and-write/notifications.txt
```diff
@@ -0,0 +1,124 @@
...
+- When any of its properties change.
+
+The handler receives information about what fields changed
+and whether the object was deleted.
+
+.. example::
+
+   The following example shows how to open a default realm,
+   create a new instance of a class, and observe that
+   instance for changes.
```
nlarew: This text is helpful, but a little generic (as is the example currently) and redundant ("The following example" in an example block). Per my comments on the other two example blocks, maybe makes sense to have this be in the context of a some made up application.

cbush: Changed to "the following code". It is probably fine otherwise. Concrete examples are fine, but so are generic examples.

PR #8, opened 2019-11-14
(DOCSP-7054): Add Collections page

cbush:
## JIRA
https://jira.mongodb.org/browse/DOCSP-7054

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/collections/crud/collections.html

### DIFF
--- a/source/crud/collections.txt
+++ b/source/crud/collections.txt
@@ -34,10 +34,11 @@ Realm has two kinds of collections: **lists** and **results**.
 List Collections
 ~~~~~~~~~~~~~~~~
 
-A **list collection** represents a to-many relationship
-between two Realm types. Lists are mutable: within a write
-transaction, you can add and remove elements on a list.
-Lists are not associated with a query.
+A **list collection** represents a :ref:`to-many
+relationship <to-many-relationship>` between two Realm
+types. Lists are mutable: within a write transaction, you
+can add and remove elements on a list. Lists are not
+associated with a query.
 
 .. _results-collections:
 
@@ -55,10 +56,11 @@ contents.
 Collections are Live
 --------------------
 
-Collections are usually :ref:`live objects <live-object>`:
+Like :ref:`live objects <live-object>`, Realm collections
+are usually **live**:
 
 - Live results collections always reflect the current results of the associated query.
-- Live lists always reflect the current state of the list on the realm.
+- Live lists always reflect the current state of the relationship on the realm.
 
 There are only two cases when a collection is **not** live:
 
@@ -68,18 +70,19 @@ There are only two cases when a collection is **not** live:
 Combined with :ref:`collection notifications
 <collection-notifications>`, live collections enable clean,
 reactive code. For example, suppose your view displays the
-results of a query. You can store the results collection as
-a member of the view and read the results collection as
-needed without having to refresh it or validate that it is
-up-to-date.
+results of a query. You can keep a reference to the results
+collection in your view class, then read the results
+collection as needed without having to refresh it or
+validate that it is up-to-date.
 
 .. admonition:: Indexes may change
    :class: important
 
-   Since results update themselves automatically, avoid
-   storing collection indexes and element counts. The stored
-   index or count value could be outdated by the time you
-   use them.
+   Since results update themselves automatically, do not
+   store the positional index of an object in the collection
+   or the count of objects in a collection. The stored index
+   or count value could be outdated by the time you use
+   them.
 
 .. _lazy-evaluated-results:
 
--- a/source/crud/collections.txt
+++ b/source/crud/collections.txt
@@ -1,0 +1,153 @@
+===========
+Collections
+===========
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+One of the fundamental building blocks when working with any
+database is the **collection**. Generally, a collection is
+an object that contains zero or more other objects. With
+Realm, a collection is an object that contains zero or more
+instances of one :doc:`Realm type </data-model/objects>`.
+Collections are homogenous, i.e. all objects in a collection
+are of the same type.
+
+You can filter and sort any collection using Realm's
+:doc:`query syntax </crud/query-language>`. Collections are
+:ref:`live <live-object>`, so they always reflect the
+current state of the realm on the current thread. You can
+also listen for changes in the collection by subscribing to
+:ref:`collection notifications <collection-notifications>`.
+
+Realm has two kinds of collection:
+
+- A **list** collection, which represents a one-to-many relationship.
+- A **results** collection, which represents the live, lazily-evaluated results of a query operation.
+
+.. _list-vs-results:
+
+List vs. Results
+----------------
+
+When deciding which collection type to use to interact with
+a realm, follow these rules of thumb:
+
+- Use lists in your Realm type definitions to define a
+  one-to-many relationship, except for :ref:`implicit
+  inverse relationships
+  <collection-type-for-implicit-inverse-relationships>`.
+- Use results collections everywhere else.
+
+To understand the difference between list and result
+collections -- and when to use which -- consider whether you
+need to add or remove objects directly. Lists allow you to
+add and remove objects directly, while results collections
+do not because their contents are determined by a query.
+
+.. example::
+
+   Consider a Realm type called Person with a field called
+   ``emails`` that is a collection of strings representing
+   email addresses. You control this data. Your application
+   needs to add and remove email addresses from your Person
+   instances. Therefore, use a **list** to define the field
+   type of ``emails``.
+
+   On the other hand, when you query the realm for all
+   Persons over the age of 25, it would not make sense for
+   you to add or remove Persons directly to the resulting
+   collection. The contents of that collection only change
+   when the results of the query change. Therefore, Realm
+   gives you a **results** collection.
+
+.. _collection-type-for-implicit-inverse-relationships:
+
+.. admonition:: Explaining the exception to the rule for inverse one-to-many relationship fields
+   :class: note
+
+   Since Realm automatically determines the contents of
+   :doc:`implicit inverse relationship
+   </data-model/relationships>` collections, you may not add
+   or remove objects from such a collection directly.
+   Therefore, the type of such a one-to-many relationship
+   field is actually a results collection, not a list.
+
+.. _properties-of-results-collections:
+
+Properties of Results Collections
+---------------------------------
+
+There are two major properties you need to know about
+results collections:
+
+- Results are **live**.
+- Results are **lazily evaluated**.
+
+These properties and their implications are detailed in the
+following sections.
+
+.. _live-results:
+
+Results are Live
+~~~~~~~~~~~~~~~~
+
+Results are :ref:`live objects <live-object>`, so they
+always reflect the current results of the associated query.
+Combined with :ref:`collection notifications
+<collection-notifications>`, this feature enables clean,
+reactive code. For example, suppose your view displays the
+results of a query. You can store the results collection as
+a member of the view and read the results collection as
+needed without having to refresh it or validate that it is
+up-to-date. As a live object, it is always up-to-date.
+
+.. admonition:: Indices may change
+   :class: important
+
+   Since results update themselves automatically, avoid
+   relying on indices and collection element counts. These
+   can change unexpectedly.
+
+.. _lazy-evaluation:
+
+Results are Lazily Evaluated
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Realm only runs a query when you actually request the
+results of that query, e.g. by accessing elements of the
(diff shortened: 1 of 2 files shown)


source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+are of the same type.
+
+You can filter and sort any collection using Realm's
+:doc:`query syntax </crud/query-language>`. Collections are
+:ref:`live <live-object>`, so they always reflect the
+current state of the realm on the current thread. You can
+also listen for changes in the collection by subscribing to
+:ref:`collection notifications <collection-notifications>`.
+
+Realm has two kinds of collection:
```
jeff-allen-mongo:
```suggestion
Realm has two kinds of collections:
```

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+
+- A **list** collection, which represents a one-to-many relationship.
```
jeff-allen-mongo: A one-to-many relationship between what and what exactly? Maybe an example would help here?

jeff-allen-mongo: You could consider using a definition list here:

```
List Collection
  Represents a one-to-many relationship.

Results Collection
  Represents the live, lazily-evaluated results of a query operation.
```

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+  <collection-type-for-implicit-inverse-relationships>`.
+- Use results collections everywhere else.
+
+To understand the difference between list and result
+collections -- and when to use which -- consider whether you
+need to add or remove objects directly. Lists allow you to
+add and remove objects directly, while results collections
+do not because their contents are determined by a query.
+
+.. example::
```
jeff-allen-mongo: This example really helped me understand the difference between the two. So do I set my collection type in my application code? If so, is there any way we can put in simple example code blocks defining each type?

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+   On the other hand, when you query the realm for all
+   Persons over the age of 25, it would not make sense for
+   you to add or remove Persons directly to the resulting
+   collection. The contents of that collection only change
+   when the results of the query change. Therefore, Realm
+   gives you a **results** collection.
+
+.. _collection-type-for-implicit-inverse-relationships:
+
+.. admonition:: Explaining the exception to the rule for inverse one-to-many relationship fields
```
jeff-allen-mongo: I think we can shorten this admonition title to something like:

"Inverse one-to-many relationship fields exception"

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+.. _list-vs-results:
+
+List vs. Results
+----------------
+
+When deciding which collection type to use to interact with
+a realm, follow these rules of thumb:
+
+- Use lists in your Realm type definitions to define a
+  one-to-many relationship, except for :ref:`implicit
```
jeff-allen-mongo: When I clicked on this ref, I expected to be taken somewhere that explained what an "implicit inverse relationship" was. The admonition didn't really provide the info I was looking for, because I had to click again to a new page from the admonition to find out what an implicit inverse relationship actually was. Maybe we can put the definition in the admonition itself somehow, or up here where we introduce the term.

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+always reflect the current results of the associated query.
+Combined with :ref:`collection notifications
+<collection-notifications>`, this feature enables clean,
+reactive code. For example, suppose your view displays the
+results of a query. You can store the results collection as
+a member of the view and read the results collection as
+needed without having to refresh it or validate that it is
+up-to-date. As a live object, it is always up-to-date.
+
+.. admonition:: Indices may change
```
jeff-allen-mongo: In the Manual docs we use `Indexes`, not sure if you have a different standard here.

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+Pagination
+^^^^^^^^^^
+
+Thanks to lazy evaluation, the common task of pagination
+becomes quite simple. For example, suppose you have a
+results collection associated with a query that matches
+thousands of objects in your realm. You display one hundred
+objects per page. To advance to any page, simply access the
+elements of the results collection starting at the index
+that corresponds to the target page.
```
jeff-allen-mongo: Can we provide some generic (or pseudo-code) example here? Do you think that would be helpful?

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+results collection. This lazy evaluation enables you to
+write elegant, highly performant code for handling large
+data sets and complex queries.
+
+.. _limiting-results:
+
+Limiting Results
+^^^^^^^^^^^^^^^^
+
+The lazy evaluation of queries means there is no special
```
jeff-allen-mongo: optional wording tweak suggestion:

"As a result of lazy evaluation, you do not need any special mechanism to limit query results with realm."

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+
+Realm only runs a query when you actually request the
+results of that query, e.g. by accessing elements of the
+results collection. This lazy evaluation enables you to
+write elegant, highly performant code for handling large
+data sets and complex queries.
+
+.. _limiting-results:
+
+Limiting Results
```
jeff-allen-mongo: Maybe `Limiting Query Results`? Doesn't hurt to be explicit here I think.

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+   relying on indices and collection element counts. These
+   can change unexpectedly.
+
+.. _lazy-evaluation:
+
+Results are Lazily Evaluated
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Realm only runs a query when you actually request the
+results of that query, e.g. by accessing elements of the
```
jeff-allen-mongo: [optional / nit] Do we need `e.g.`? This might be better worded as something like:

Realm only runs a query when you actually request the
results of that query by accessing elements of the results
collection.

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+results of a query. You can store the results collection as
+a member of the view and read the results collection as
+needed without having to refresh it or validate that it is
+up-to-date. As a live object, it is always up-to-date.
+
+.. admonition:: Indices may change
+   :class: important
+
+   Since results update themselves automatically, avoid
+   relying on indices and collection element counts. These
```
jeff-allen-mongo: I'm a little confused why this means I shouldn't rely on indexes.

source/crud/collections.txt
```diff
@@ -0,0 +1,153 @@
...
+
+Results are :ref:`live objects <live-object>`, so they
+always reflect the current results of the associated query.
+Combined with :ref:`collection notifications
+<collection-notifications>`, this feature enables clean,
+reactive code. For example, suppose your view displays the
+results of a query. You can store the results collection as
+a member of the view and read the results collection as
+needed without having to refresh it or validate that it is
+up-to-date. As a live object, it is always up-to-date.
```
jeff-allen-mongo: I don't think we need this final sentence. It doesn't really offer anything that hasn't been said already.

source/crud/collections.txt
cbush: We don't have a different standard afaik. Is that referring to DB indexes™ vs. array indexes? Does that make a difference?

jeff-allen-mongo: DB Indexes: https://docs.mongodb.com/manual/indexes/. I've just never seen `indices` anywhere in our docs

source/crud/collections.txt
cbush: It could also run the query by checking the length of a results collection.

source/crud/collections.txt
cbush: relying on -> storing?

source/crud/collections.txt
cbush: I will be adding code examples as I develop the Query syntax page.

source/crud/collections.txt
jeff-allen-mongo: I'm good with e.g. in that case.

source/crud/collections.txt
jeff-allen-mongo: Sorry I don't think my original comment was clear. I was more confused why indexes would affect the results (in this context)?

cbush: If you store the index, and the collection changes, your stored index will be out of date.

source/crud/collections.txt
```diff
@@ -0,0 +1,156 @@
...
+instances of one :doc:`Realm type </data-model/objects>`.
+Collections are homogenous, i.e. all objects in a collection
+are of the same type.
+
+You can filter and sort any collection using Realm's
+:doc:`query syntax </crud/query-language>`. Collections are
+:ref:`live <live-object>`, so they always reflect the
+current state of the realm on the current thread. You can
+also listen for changes in the collection by subscribing to
+:ref:`collection notifications <collection-notifications>`.
```
nlarew: Might be worthwhile to note that adding a filter to a List returns a Result? (I think)

source/crud/collections.txt
```diff
@@ -0,0 +1,156 @@
...
+List Collection
+  Represents a one-to-many between two Realm types.
+
+Results Collection
+  Represents the live, lazily-evaluated results of a query operation.
+
+.. _list-vs-results:
+
+List vs. Results
+----------------
```
nlarew: This page focuses a lot on the distinction between Lists and Results, but to me that makes it harder to understand why they're even lumped together. We have some stuff about how they're similar in the Overview but I wonder if that approach should be more central to the page? What is common between these two types such that they both implement the same base data structure?

source/crud/collections.txt
```diff
@@ -0,0 +1,156 @@
...
+also listen for changes in the collection by subscribing to
+:ref:`collection notifications <collection-notifications>`.
+
+Realm has two kinds of collections:
+
+List Collection
+  Represents a one-to-many between two Realm types.
+
+Results Collection
+  Represents the live, lazily-evaluated results of a query operation.
```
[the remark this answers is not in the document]
cbush: I'm laying it out here as part of the overview.

cbush:
@nlarew Restructured a bit. Please re-review.

source/crud/collections.txt
```diff
@@ -0,0 +1,164 @@
...
+
+.. _live-collections:
+
+Collections are Live
+--------------------
+
+Collections are usually :ref:`live objects <live-object>`:
+
+- Live results collections always reflect the current results of the associated query.
+- Live lists always reflect the current state of the list on the realm.
```
nlarew: nit: Since lists are to-many relationships, "state of the list" -> "state of the relationship"

source/crud/collections.txt
```diff
@@ -0,0 +1,164 @@
...
+:ref:`collection notifications <collection-notifications>`.
+
+Realm has two kinds of collections: **lists** and **results**.
+
+.. _list-collections:
+
+List Collections
+~~~~~~~~~~~~~~~~
+
+A **list collection** represents a to-many relationship
```
nlarew: Should we link to the to-many relationship section here?

```rst
:ref:`to-many relationship <to-many-relationship>`
```

source/crud/collections.txt
```diff
@@ -0,0 +1,164 @@
...
+cannot add or remove elements on the results collection.
+Results have an associated query that determines their
+contents.
+
+.. _live-collections:
+
+Collections are Live
+--------------------
+
+Collections are usually :ref:`live objects <live-object>`:
```
nlarew: Not a big issue, but I hesitate on this since "live object" in my head refers specifically to Realm objects (as opposed to a collection of Realm objects). At some point we might want to revise that definition formally, but for now not a blocker or anything.

source/crud/collections.txt
```diff
@@ -0,0 +1,164 @@
...
+There are only two cases when a collection is **not** live:
+
+- The collection is part of a snapshot (JavaScript and Java SDKs only).
+- The collection is unmanaged, e.g. a List property of a Realm object that has not been added to a realm yet.
+
+Combined with :ref:`collection notifications
+<collection-notifications>`, live collections enable clean,
+reactive code. For example, suppose your view displays the
+results of a query. You can store the results collection as
+a member of the view and read the results collection as
```
nlarew: Perhaps it's been a minute since I've done MV* but I'm not totally clear on what "store ... as a member of the view" means

source/crud/collections.txt
```diff
@@ -0,0 +1,164 @@
...
+needed without having to refresh it or validate that it is
+up-to-date.
+
+.. admonition:: Indexes may change
+   :class: important
+
+   Since results update themselves automatically, avoid
+   storing collection indexes and element counts. The stored
```
nlarew: Per the comment around confusion between array indexes and db indexes, should we be a little more explicit here?

- Should we say "avoid" or "do not"?
- avoid storing a count of objects in the collection
- avoid storing the positional index of an object in the collection

PR #11, opened 2019-11-29
(DOCSP-7041): Install iOS

cbush:
## JIRA
https://jira.mongodb.org/browse/DOCSP-7041

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/install-ios/get-started/install-ios.html


source/get-started/install-ios.txt
```diff
@@ -0,0 +1,58 @@
...
+
+Installation
+------------
+
+Follow these steps to add the Realm Apple platform SDK to
+your project.
+
+.. note::
+   Not sure which dependency manager to use? We recommend
+   CocoaPods.
```
nlarew: Is there a particular reason for this recommendation?

source/includes/steps-install-cocoapods.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+   .. code-block:: text
+      :emphasize-lines: 9, 14
+
+      # Uncomment the next line to define a global platform for your project
+      # platform :ios, '9.0'
+
+      target 'MyRealmProject' do
+        # Comment the next line if you don't want to use dynamic frameworks
+        use_frameworks!
```
nlarew: Should we emphasize `use_frameworks!` as well since we called it out above?

source/includes/steps-install-cocoapods.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+---
+title: Use the CocoaPods-Generated xcworkspace
+ref: use-xcworkspace
+content: |
+
+   CocoaPods generates an ``xcworkspace`` file for you. This
+   file has all of the dependencies configured. From now on,
+   open this file -- not the ``xcodeproj`` file -- to work
+   on your project.
```
nlarew: Should these have the dot in front of them, e.g. ``` ``.xcworkspace`` ```?

source/includes/steps-install-carthage.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+---
+title: Enable the Copy Frameworks Script (iOS, tvOS, and watchOS only)
+ref: enable-copy-frameworks
+content: |
+
+   To work around an `App Store submission bug
+   <http://www.openradar.me/radar?id=6409498411401216>`__
+   for iOS, tvOS, and watchOS apps, you must add a script to
+   copy the frameworks as a build phase.
```
nlarew: Oof

source/get-started/install-ios.txt
```diff
@@ -0,0 +1,58 @@
...
+
+   .. tab:: Carthage
+      :tabid: carthage
+
+      .. include:: /includes/steps/install-carthage.rst
+
+   .. tab:: Dynamic Framework
+      :tabid: dynamic-framework
+
+      .. include:: /includes/steps/install-dynamic-framework.rst
```
nlarew: This just sort of ends, unlike the C# procedure which has a **That's It!** step to signal completion. Should we unify?

source/get-started/install-ios.txt
cbush: There are 3 options. If you don't know what you're doing, wouldn't you want a push in one direction?

source/includes/steps-install-cocoapods.yaml
cbush: It seems this is automatically added in recent versions of CocoaPods, so new users probably won't have to change this.

source/get-started/install-ios.txt
cbush: I will remove "That's it" steps and add "Import Realm" snippets to the pages.

source/get-started/install-ios.txt
nlarew: Yeah, no issue with us having an opinion here. Mainly interested in why our opinion is CocoaPods vs the other options.

PR #13, opened 2019-12-10
(DOCSP-7042, 7043): Add install JS/TS pages

cbush:
- Adds the Install pages for both RN and Node.
- Went back and forth on whether to merge into a single page.
Ultimately, I decided they are different enough and serve different
enough purposes to merit separate pages.
- I considered adding more detail about how to set up a Babel workflow
and TypeScript. Currently, I do provide links for how to set up TypeScript,
but it felt out of scope to expand more on that and could get out of date
quickly. I would be willing to either cut the TypeScript setup parts
completely, or expand on them and even include how to set up an es9
babel-node, eslint, and prettier development workflow. Thoughts?

## JIRA
https://jira.mongodb.org/browse/DOCSP-7042
https://jira.mongodb.org/browse/DOCSP-7043

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/install-node/get-started/install-node.html

https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/install-node/get-started/install-react-native.html


source/get-started/install-node.txt
```diff
@@ -0,0 +1,67 @@
...
+Node.js is best suited for writing server-side applications.
+Your Node.js application could be one component in a wider
+stack that also includes iOS and Android apps.
+
+For front-end web application development, the Realm Node.js
+SDK is not suitable due to limitations of the browser
+environment. Instead, see the :doc:`GraphQL installation
+page </get-started/graphql>`. Realm also does not officially
+support other environments, such as `Electron
+<https://electronjs.org/>`__.
```
nlarew: nit: Do you consider Electron to be a web application? We may want to scope this to "frontend application development" without mentioning web, i.e "DON'T USE THIS FOR ANY FRONTEND CODE"

nlarew: Is it "limitations of the browser environment" or "differences between browser and server environments" (or both)?

nlarew: Rather than "see the GraphQL installation page" could we more directly say "Use GraphQL for frontend JavaScript applications. You can find out how to set that up here."

source/get-started/install-node.txt
```diff
@@ -0,0 +1,67 @@
...
+
+
+Prerequisites
+-------------
+
+Before getting started, ensure your development environment
+meets the following prerequisites:
```
nlarew: Do you think "development environment" might be confusing when contrasted with a "production environment"?

source/get-started/install-node.txt
```diff
@@ -0,0 +1,67 @@
...
+
+- `Node.js <https://nodejs.org/en/>`__ version 8.3.0 (and later versions in 8.x) or Node.js 10.x.
+- Linux, macOS, or Windows 8 (and later).
```
nlarew: Perhaps a can of worms we don't want to open here, but are there specific macOS versions and Linux distros/versions?

source/includes/react-native-enable-typescript.rst
```diff
@@ -0,0 +1,6 @@
+TypeScript is a superset of JavaScript that adds static type
+checking and other features intended to make
+application-scale development more robust. If you'd like to
+use TypeScript in your project, follow the React Native
+team's official `TypeScript and React Native guide
+<https://facebook.github.io/react-native/docs/typescript#adding-typescript-to-an-existing-project>`__.
```
nlarew: Maybe include something about how Realm natively supports TS out of the box and that there's nothing special that you need to do that you wouldn't for any other TS app?

source/includes/steps-install-node.yaml
```diff
@@ -0,0 +1,37 @@
...
+title: Install Realm with NPM
+ref: install-realm-with-npm
+content: |
+
+   In your Node.js project directory, use the following command
+   to add Realm to your project:
+
+   .. code-block:: bash
+
+      npm install --save realm
```
nlarew: Do we also want to provide a yarn example?

source/includes/react-native-enable-typescript.rst
nlarew: (Non-blocker) IMO we should keep single-sourced RST like this as close to the include locations as possible / in a semantically meaningful location (i.e. in the `get-started` directory). I believe the platform team disagrees and prefers the `includes` directory for all included content.

Interested in your opinion on it and whether or not we should push for local/semantic includes in nextgen

source/get-started/install-react-native.txt
```diff
@@ -0,0 +1,79 @@
...
+   .. tab:: Older React Native Versions
+      :tabid: rn-pre-v-60
+
+      .. include:: /includes/steps/install-react-native-pre-v60.rst
+
+Import Realm
+------------
+
+Add the following line to the top of your source files to
+use Realm:
```
nlarew: nit: you only need to import in files that you use realm (as opposed to ALL files). Maybe that's obvious enough already?

source/get-started/install-react-native.txt
```diff
@@ -0,0 +1,79 @@
...
+
+- React Native v0.31.0 or later. Follow the `official React Native CLI Quickstart instructions <https://facebook.github.io/react-native/docs/getting-started.html>`__ to set up your environment.
+- `CocoaPods <https://cocoapods.org/>`__ (recommended for building an iOS app with React Native v.60+).
+
+.. admonition:: Expo does not support Realm
+   :class: important
+
+   Unfortunately, `Expo does not support Realm
+   <https://expo.canny.io/feature-requests/p/support-for-realm>`__,
+   so ``create-react-native-app`` does not work with Realm.
```
nlarew: Could we have some context here as to why?
- Realm requires you to install a [native module](https://facebook.github.io/react-native/docs/native-modules-ios)
- Expo does not currently support native modules
- ``create-react-native-app`` uses Expo

source/includes/steps-install-react-native-v60-plus.yaml
```diff
@@ -0,0 +1,60 @@
...
+ref: cocoapods-install
+content: |
+
+   For the iOS app, fetch the CocoaPods dependencies with
+   the following commands from your React Native project
+   directory:
+
+   .. code-block:: bash
+
+      cd ios && pod install && cd ..
```
nlarew: Does `react-native init` automatically take care of `pod init` for you?

source/includes/steps-install-react-native-v60-plus.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+   This downloads the Realm libraries and regenerates the
+   project ``.xcworkspace`` file that can work with in
+   Xcode to run your application.
```
nlarew: that **you** can work with

source/includes/steps-install-react-native-v60-plus.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+   In your React Native project directory, add Realm to your
+   project with the following command:
+
+   .. code-block:: bash
+
+      npm install --save realm
+
+---
+title: Resolve CocoaPods Dependencies
```
nlarew: Is there a similar step for dependencies on the Android side?

source/get-started/install-node.txt
cbush: It's limitations of the browser environment, i.e. Realm does not work with Local Storage etc. I just didn't want to get into specifics here.

I would consider Electron to be a desktop application built with web frameworks (??). Overall, this encompasses all "frontend development", though React Native is for frontend development and _is_ in fact supported...

source/includes/steps-install-react-native-v60-plus.yaml
cbush: Thankfully, no. But see the RN <v.60 install steps for the issues with the link step that you may need to manually resolve.

source/get-started/install-node.txt
cbush: Reworking this 

> as a bullet list, something more like:
> 
> The Realm Node.js SDK is not suitable for front-end
> application development.
> 
> - Due to limitations of the browser environment, it is not possible to build a browser-based web app with this SDK. For front-end web applications, use the :doc:`Realm GraphQL SDK </get-started/graphql>`.
> - For cross-platform mobile app development using web frameworks, use the :doc:`Realm React Native SDK </get-started/install-react-native>`.
> - Realm does not officially support other front-end frameworks, such as `Electron <https://electronjs.org/>`__, at this time.
> 

Thoughts?

source/get-started/install-node.txt
cbush: I will ask about this.

source/includes/steps-install-node.yaml
cbush: This may be controversial but I don't see a reason to start a new project with yarn since npm 5. If a person's existing project is using yarn then they probably won't bother switching to npm 5, but they would presumably know how to convert an npm command snippet to the yarn equivalent. So I'm going to say no at this time, but I'm open to being convinced otherwise.

source/get-started/install-react-native.txt
cbush: Haha. Clarified. Thanks.

source/includes/steps-install-react-native-v60-plus.yaml
cbush: As of v.60, yes.

source/get-started/install-node.txt
nlarew: I like this update 👍

source/includes/steps-install-node.yaml
nlarew: SGTM

cbush:
Merging and will update Linux versions if/when we learn specific information around that.

PR #33, opened 2020-02-07
(DOCSP-8762): Port auth provider: API Key

nathan-contino-mongo:
[JIRA](https://jira.mongodb.org/browse/DOCSP-8762)

[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-8762/authentication/api-key.html)

### DIFF
--- a/source/authentication/api-key.txt
+++ b/source/authentication/api-key.txt
@@ -17,19 +17,19 @@ API Key Authentication
 Overview
 --------
 
-The API Key authentication provider allows users to log in using
-generated keys. There are two types of API keys in MongoDB Realm:
-server keys and user keys.
+The API Key :ref:`authentication provider <authentication-providers>`
+allows users to log in using generated keys. There are two types of API
+keys in MongoDB Realm: server keys and user keys.
 
 .. _api-key-authentication-server-api-key:
 
 Server API Keys
 ~~~~~~~~~~~~~~~
 
-Server API keys are generated centrally in the Realm UI and are associated
-with automatically created Realm server users. Provide a server key to
-external applications and services to allow them to authenticate
-directly with Realm.
+Server API keys are generated in the Realm UI. Creating a server API key
+associates that API key with an automatically created Realm server user.
+Provide a server key to external applications and services to allow them
+to authenticate directly with Realm.
 
 .. _api-key-authentication-user-api-key:
 
@@ -38,7 +38,7 @@ User API Keys
 
 User API keys are generated for specific application users by the client
 SDKs. You can allow devices or services to communicate with Realm on
-behalf of a user by associating a unique user key with each device.
+behalf of a user by associating a unique user API key with each device.
 
 User keys are always associated with a user object created by another
 authentication provider. Each user can associate up to 20 user keys with
@@ -152,7 +152,7 @@ Key`. Enter a unique name for the key and click :guilabel:`Save`.
 
    Remember to copy the server key's value as soon as you create it.
    Once you leave the provider configuration page or disable the key you
-   will not be able to find the value again in the Realm UI.
+   cannot retrieve the value from the Realm UI.
 
 .. _api-key-authentication-usage-look-up-user-key:
 
@@ -210,3 +210,14 @@ Delete an API Key
      - id: ios
        content: |
          TBD
+
+Summary
+-------
+
+- The Realm API Key authentication provider allows users and services
+  to connect to a Realm application using API keys that look like a
+  string of characters.
+- User API Keys allow a user to interact with services via the Realm
+  SDK. They are automatically generated in the client SDK.
+- Server API Keys allow external services to interact with your Realm
+  application. These keys can only be generated via the Realm UI.
--- a/source/authentication/api-key.txt
+++ b/source/authentication/api-key.txt
@@ -211,6 +211,8 @@ Delete an API Key
        content: |
          TBD
 
+.. _api-key-authentication-summary:
+
 Summary
 -------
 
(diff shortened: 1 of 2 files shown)


source/authentication/api-key.txt
```diff
@@ -0,0 +1,212 @@
...
+The API Key authentication provider allows users to log in using
+generated keys. There are two types of API keys in MongoDB Realm:
+server keys and user keys.
+
+.. _api-key-authentication-server-api-key:
+
+Server API Keys
+~~~~~~~~~~~~~~~
+
+Server API keys are generated centrally in the Realm UI and are associated
```
ccho-mongodb: Nitpick: does "centrally" identify a specific condition or could it be omitted?

source/authentication/api-key.txt
```diff
@@ -0,0 +1,212 @@
...
+
+To create a new server API key, navigate to the API Key authentication
+configuration page in the Realm UI and click :guilabel:`Create API
+Key`. Enter a unique name for the key and click :guilabel:`Save`.
+
+.. important::
+
+   Remember to copy the server key's value as soon as you create it.
+   Once you leave the provider configuration page or disable the key you
+   will not be able to find the value again in the Realm UI.
```
ccho-mongodb: Nitpick: instead of future tense, you can reword as:
"Once you leave the provider configuration page or disable the key you cannot retrieve the value from the Realm UI."

source/authentication/api-key.txt
```diff
@@ -0,0 +1,212 @@
...
+Look up an Existing User Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. tabs-stitch-sdks::
+
+   hidden: true
+   tabs:
+     - id: javascript
+       content: |
+         TBD
```
ccho-mongodb: Is this part of the "directly copy" item from the JIRA task, or did you mean to fill this out?

nathan-contino-mongo: Basically these APIs aren't finalized for the realm transition, so until then we're leaving direct code examples out. Higher level stuff is fine, but direct code snippets are generally omitted in this round of migration.

source/authentication/api-key.txt
```diff
@@ -0,0 +1,212 @@
...
+         </deploy/stitch-cli-reference>` by importing an
+         application directory that contains a :ref:`configuration file
+         <appschema-authentication-providers>` for the provider.
+
+         The configuration file must be named ``api-key.json``
+         and stored in the ``/auth_providers`` directory. Configuration
+         files for the API Key authentication provider have the
+         following form:
+
+         .. cssclass:: wide-container
```
ccho-mongodb: Cool, never knew about this. I can't tell from looking at it whether it is wide, but should the code-block directive be nested or at the same level (as it currently is)?

source/authentication/api-key.txt
```diff
@@ -0,0 +1,212 @@
...
+     - id: import-export
+       content: |
+         You can enable and configure the API Key authentication
+         provider with :doc:`stitch-cli
+         </deploy/stitch-cli-reference>` by importing an
+         application directory that contains a :ref:`configuration file
+         <appschema-authentication-providers>` for the provider.
+
+         The configuration file must be named ``api-key.json``
+         and stored in the ``/auth_providers`` directory. Configuration
```
ccho-mongodb: Should this be "and be stored" to be parallel in structure to "be named"?

source/authentication/api-key.txt
```diff
@@ -0,0 +1,212 @@
...
+generated keys. There are two types of API keys in MongoDB Realm:
+server keys and user keys.
+
+.. _api-key-authentication-server-api-key:
+
+Server API Keys
+~~~~~~~~~~~~~~~
+
+Server API keys are generated centrally in the Realm UI and are associated
+with automatically created Realm server users. Provide a server key to
```
ccho-mongodb: Question: I wasn't sure what this part of the sentence means. Does it mean:
1) when you create a Realm server user, a server API key is generated for that user.
or
2) anytime you create a new user, they are associated with existing server API keys.

source/authentication/api-key.txt
```diff
@@ -0,0 +1,212 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. _api-key-authentication-overview:
+
+Overview
+--------
+
+The API Key authentication provider allows users to log in using
```
ccho-mongodb: I wonder if it might be useful to introduce the term "authentication provider" or at least mentioning it is a feature of Realm.

source/authentication/api-key.txt
```diff
@@ -0,0 +1,212 @@
...
+directly with Realm.
+
+.. _api-key-authentication-user-api-key:
+
+User API Keys
+~~~~~~~~~~~~~
+
+User API keys are generated for specific application users by the client
+SDKs. You can allow devices or services to communicate with Realm on
+behalf of a user by associating a unique user key with each device.
```
ccho-mongodb: is "user key" a "user API key" here?

source/authentication/api-key.txt
nathan-contino-mongo: it appears that the format is strictly as it appears here. Also not a directive I knew about!

source/authentication/api-key.txt
nathan-contino-mongo: Every new server API key generates a new user that is uniquely associated with that key for traceability reasons. I've reworded this to make this clearer, good point!

nathan-contino-mongo:
Added a recap section and responded to chris' feedback.

PR #35, opened 2020-02-07
(DOCSP-8763): Port Auth Provider: Apple ID

nlarew:
## Jira

- [(DOCSP-8763): Port Auth Provider: Apple ID](https://jira.mongodb.org/browse/DOCSP-8763)

## Staged Changes

- [Apple ID Authentication](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker/auth-apple/authentication/apple)


source/authentication/apple.txt
```diff
@@ -0,0 +1,52 @@
...
+
+The Apple ID authentication provider allows users to log in with their
+Apple ID credentials through `Sign in with Apple
+<https://developer.apple.com/sign-in-with-apple/>`_. This authentication
+method uses the industry-standard `OAuth 2.0 <https://oauth.net/2/>`_
+authorization protocol.
+
+When a user successfully logs in through Sign in with Apple, Apple
+returns a credential object that includes a :jwt-io:`JSON Web Token
+<introduction>` that the provider uses to authenticate the user. The
```
MongoCaleb: Which provider?

source/authentication/apple.txt
```diff
@@ -0,0 +1,52 @@
...
+credential may also include the user's name and email address if they
```
MongoCaleb: Consider: If the user has granted permissions to your app, the credential object may also contain the user's name and email address.

source/authentication/apple.txt
```diff
@@ -0,0 +1,52 @@
...
+For additional information on how to implement Sign in with Apple, check
+out the `Introducing Sign In with Apple
+<https://developer.apple.com/videos/play/wwdc2019/706>`_ session from
+WWDC 2019 and the associated `reference application
+<https://developer.apple.com/documentation/authenticationservices/adding_the_sign_in_with_apple_flow_to_your_app>`_.
+
+.. admonition:: Only Available on iOS
+   :class: note
+
+   Apple ID authentication is currently only available through the iOS
```
MongoCaleb: (active voice): You can only use Apple ID authentication via the iOS SDK in apps that target iOS 13 or newer.

source/includes/steps-apple-auth-configure.yaml
```diff
@@ -0,0 +1,309 @@
+title: Create an App ID
+ref: create-an-app-id
+content: |
+   An Apple **App ID** represents your application and allows you to
```
MongoCaleb: Somewhere, either in this list or the containing doc, we need to point out that one needs an Apple Developer account, too. In fact, the UI doesn't show any of these options until one joins the Apple Developer program. I don't know -- perhaps it's a _given_, but I don't think we should assume it.

source/includes/steps-apple-auth-configure.yaml
```diff
@@ -0,0 +1,309 @@
...
+   
+   6. Scroll down the :guilabel:`Register an App ID` page until you see
+      the :guilabel:`Sign in with Apple` capability. Check the checkbox
+      to enable the capability.
+---
+title: Create a Services ID
+ref: create-a-services-id
+content: |
+   An Apple **Services ID** represents a single application and allows
+   you to configure authorization callback URL and define a private key
```
MongoCaleb: you to configure **an** application callback...

source/includes/steps-apple-auth-configure.yaml
```diff
@@ -0,0 +1,309 @@
...
+   
+   6. Note down the :guilabel:`Key ID` and then click
+      :guilabel:`Download` to download the key as a ``.p8`` text file.
+      You will use these to generate the client secret.
+      
+      .. admonition:: Save the Key
+         :class: important
+         
+         You can only download the key one time. Make sure that you save
+         the key someplace safe in case you need it again. If you lose
```
MongoCaleb: technically someplace should be two words: "some place"

source/includes/steps-apple-auth-configure.yaml
```diff
@@ -0,0 +1,309 @@
...
+   4. On the :guilabel:`Configure Key` page, select the App ID that you
+      created as the :guilabel:`Primary App ID` and then click
+      :guilabel:`Save`.
+   
+   5. Click :guilabel:`Continue` to review your key configuration. When
+      you're sure that you've configured the key correctly, click
+      :guilabel:`Register`.
+      :guilabel:`Save`.
+   
+   6. Note down the :guilabel:`Key ID` and then click
```
MongoCaleb: "note down" is a bit weird. "Write down", "make note of", or just "note" are better.

source/includes/steps-apple-auth-configure.yaml
```diff
@@ -0,0 +1,309 @@
...
+      Sign in with Apple and then click :guilabel:`Configure`.
+   
+   4. On the :guilabel:`Configure Key` page, select the App ID that you
+      created as the :guilabel:`Primary App ID` and then click
+      :guilabel:`Save`.
+   
+   5. Click :guilabel:`Continue` to review your key configuration. When
+      you're sure that you've configured the key correctly, click
+      :guilabel:`Register`.
+      :guilabel:`Save`.
```
MongoCaleb: ooops
Register
Save

source/includes/steps-apple-auth-configure.yaml
```diff
@@ -0,0 +1,309 @@
...
+      
+      .. admonition:: Save the Key
+         :class: important
+         
+         You can only download the key one time. Make sure that you save
+         the key someplace safe in case you need it again. If you lose
+         the key, you will need to generate a new one.
+      
+      .. cssclass:: bordered-figure
+      .. figure:: /images/apple-auth-configure-key.png
```
MongoCaleb: This image confuses me a bit. I think it belongs above the note, perhaps? Also, it shows  Back and Save buttons, but the steps say to click Continue. Looking back, I believe what you are showing here is step 4, so it should be moved up there.

source/includes/steps-apple-auth-configure.yaml
```diff
@@ -0,0 +1,309 @@
...
+     
+     .. cssclass:: bordered-figure
+     .. figure:: /images/apple-team-id.png
+        :alt: An Apple team ID in the Apple Developer Portal
+   
+   Once you've confirmed that you have all the required information,
+   create a new file called ``generate_client_secret.rb`` and copy the
+   following code block into the file.
+
+   .. code-block:: ruby
```
MongoCaleb: None of the code block is showing up in the staged page!

source/includes/steps-apple-auth-configure.yaml
```diff
@@ -0,0 +1,309 @@
...
+         3. For :guilabel:`Client ID`, enter the :guilabel:`Services ID`
+            that you created.
+         
+         4. For :guilabel:`Client Secret`, create a new :ref:`secret
+            <app-secret>` with a descriptive name and set the
+            :guilabel:`Client Secret Value` to the JWT string that you
+            generated. Alternatively, you can select a pre-existing
+            secret that contains the JWT.
+         
+         5. For :guilabel:`Redirect URIs`, click :guilabel:`Add Redirect
```
MongoCaleb: This is confusing. if I'm building an iOS app and want to use Apple auth, what web page would I redirect to? Seems like I don't need or want a redirect. I think this is the most confusing part of the Stitch/Realm JWT auth story and needs a bit more attention.

source/includes/steps-apple-auth-setup-client.yaml
```diff
@@ -0,0 +1,113 @@
+title: Add the Sign in with Apple Capability
+ref: add-the-sign-in-with-apple-capability
+content: |
+   You must grant your client application permission to use Sign in with
+   Apple by defining a `capability
+   <https://developer.apple.com/documentation/xcode/adding_capabilities_to_your_app>`_.
+
+   In XCode, select your application and then click :guilabel:`Signing &
+   Capabilities`. Click :guilabel:`+ Capability` and add the
+   :guilabel:`Sign in with Apple` capability. You can now import the
```
MongoCaleb: Break this step up. After ":guilabel:`Sign in with Apple` capability. ", insert the image showing that, and then have the "You can now import..." with the code block below it.

source/includes/steps-apple-auth-configure.yaml
nlarew: Good call! Definitely not a given for some substantial subset of developers. Added a **What You'll Need** section to cover this.

source/includes/steps-apple-auth-configure.yaml
nlarew: Good catch, thanks!

source/includes/steps-apple-auth-configure.yaml
nlarew: Reworded this to be "copy the Key ID value somewhere that you can access it later"

source/includes/steps-apple-auth-configure.yaml
nlarew: You're right - changed this to "somewhere" since it reads better IMO

source/includes/steps-apple-auth-configure.yaml
nlarew: Weird, not sure why but neither of the code blocks in this stepfile render. I've marked it down as something to follow up on but for now I'm not sure what the issue is... :/

source/includes/steps-apple-auth-configure.yaml
nlarew: In the next paragraph we mention universal links which is the URL you redirect to. I'll shuffle this step around to emphasize that.

source/includes/steps-apple-auth-setup-client.yaml
nlarew: Good call this does feel a little out of order / jumbled up as is

source/authentication/apple.txt
```diff
@@ -0,0 +1,66 @@
...
+What You'll Need
+----------------
+
+Before you can implement Apple ID authentication you will need the
+following:
+
+- An active `Apple Developer Program
+  <https://developer.apple.com/programs/>`_ account.
+
+- `Xcode 11 <https://developer.apple.com/xcode/>` or newer.
```
MongoCaleb: This is not rendering on the staged page -- just says "or newer"

source/includes/steps-apple-auth-configure.yaml
MongoCaleb: Yeah, still not seeing it. Are you?

source/includes/steps-apple-auth-configure.yaml
MongoCaleb: Better, for sure. Perhaps it's just a brainfart on my end, but it's still not clear if I need to add a redirect URL, and if so, what is expected at the other end of the line.

source/authentication/apple.txt
nlarew: Good catch! The link is malformed (it should have a trailing `_`)

PR #39, opened 2020-02-11
DOCSP-8702): added comment+code for upserts - updateMode="modified"

mongomoe:
JIRA:
-----
https://jira.mongodb.org/browse/DOCSP-8702

Stage:
------
* https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-8702-updateMode-upsert/by-example/read-and-write.html#upsert-an-object
* https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-8702-updateMode-upsert/crud/writes.html#upsert-an-object

Old Version:
------------
* https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/master/by-example/read-and-write.html#upsert-an-object
* https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/master/crud/writes.html#upsert-an-object

### DIFF
--- a/source/crud/writes.txt
+++ b/source/crud/writes.txt
@@ -124,14 +124,13 @@ This automatically inserts the instance into the realm.
 Upsert an Object
 ~~~~~~~~~~~~~~~~
 
-An **upsert** is a write operation that either inserts a new
-object with a given primary key or updates an existing
-object that already has that primary key. We call this an
-upsert because it is an "**update** or **insert**"
-operation. This is useful when an object may or may not
-already exist, such as when bulk importing a dataset into an
-existing realm. Upserting is an elegant way to update
-existing entries while adding any new entries.
+An **upsert** is a write operation that either inserts a new  object
+with a given primary key or updates an existing object that already has
+that primary key. We call this an upsert because it is an "**update** or
+**insert**" operation. This is useful when an object may or may not
+already exist, such as when bulk importing a dataset into an existing
+realm. Upserting is an elegant way to update existing entries while
+adding any new entries.
 
 .. example::
 
--- a/source/crud/writes.txt
+++ b/source/crud/writes.txt
@@ -124,7 +124,7 @@ This automatically inserts the instance into the realm.
 Upsert an Object
 ~~~~~~~~~~~~~~~~
 
-An **upsert** is a write operation that either inserts a new  object
+An **upsert** is a write operation that either inserts a new object
 with a given primary key or updates an existing object that already has
 that primary key. We call this an upsert because it is an "**update** or
 **insert**" operation. This is useful when an object may or may not
--- a/source/examples/CRUD/Upsert.js
+++ b/source/examples/CRUD/Upsert.js
@@ -7,6 +7,6 @@ realm.write(() => {
 
   // If an object exists, setting the third parameter (`updateMode`) to
   // "modified" only updates properties that have changed, resulting in
-  // faster operations
+  // faster operations.
   const andy = realm.create("Person", { id: 1234, name: "Andy" }, "modified");
 });
--- a/source/examples/CRUD/Upsert.js
+++ b/source/examples/CRUD/Upsert.js
@@ -6,8 +6,7 @@ realm.write(() => {
   // Judging by the ID, it's the same person, just with a different name.
 
   // If an object exists, setting the third parameter (`updateMode`) to
-  // "modified" only updates properties that have changed. This results in
-  // faster operations and changes the way that Realm resolves conflicts
-  // between multiple write operations.
+  // "modified" only updates properties that have changed, resulting in
+  // faster operations
   const andy = realm.create("Person", { id: 1234, name: "Andy" }, "modified");
 });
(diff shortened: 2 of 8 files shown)


source/examples/CRUD/Upsert.js
```diff
@@ -1,10 +1,13 @@
...
   // Judging by the ID, it's the same person, just with a different name.
-  // When the third parameter (`update`) is true, you overwrite the original
-  // entry (i.e. Drew -> Andy).
-  const andy = realm.create("Person", { id: 1234, name: "Andy" }, true);
+
+  // If an object exists, setting the third parameter (`update`) to
+  // "updateMode", will only update properties where the value has
+  // changed, decreasing notifications and improving server-side
+  // performance
+  const andy = realm.create("Person", { id: 1234, name: "Andy" }, "updateMode");
```
nathan-contino-mongo: Looking at the JIRA ticket, it looks like the third parameter ought to be the string "modified" (https://realm.io/docs/javascript/latest/api/Realm.html#create) as ``updateMode`` is just the name of this parameter (does this work? I'm guessing it's getting cast to a boolean, which means it's interpreted as ``true`` which is equivalent to "default"?)

source/examples/CRUD/Upsert.js
```diff
@@ -1,10 +1,13 @@
 realm.write(() => {
   // Add a new person to the realm. Since nobody with ID 1234
   // has been added yet, this adds the instance to the realm.
-  const drew = realm.create("Person", { id: 1234, name: "Drew" }, true);
+  const drew = realm.create("Person", { id: 1234, name: "Drew" }, "updateMode");
```
nathan-contino-mongo: should also change this to "modified"

source/examples/CRUD/Upsert.js
```diff
@@ -1,10 +1,13 @@
...
 
   // Judging by the ID, it's the same person, just with a different name.
-  // When the third parameter (`update`) is true, you overwrite the original
-  // entry (i.e. Drew -> Andy).
-  const andy = realm.create("Person", { id: 1234, name: "Andy" }, true);
+
+  // If an object exists, setting the third parameter (`update`) to
+  // "updateMode", will only update properties where the value has
+  // changed, decreasing notifications and improving server-side
+  // performance
```
nathan-contino-mongo: Close, but I think this focuses on some ideas that aren't necessarily relevant to the end user (notably "decreasing notifications"). I'd suggest something like this:
```
// If an object exists, setting the third parameter (`updateMode`) to
// "modified" only updates properties that have changed. This results
// in faster operations and changes the way that Realm resolves
// conflicts between multiple write operations.
```

source/examples/CRUD/Upsert.js
mongomoe: YES This is why we have reviewers. Not sure how I misinterpreted that. Fixed now

source/examples/CRUD/Upsert.js
mongomoe: Fixed

source/examples/CRUD/Upsert.js
mongomoe: Well put - replaced verbatim

source/crud/writes.txt
```diff
@@ -124,14 +124,13 @@ This automatically inserts the instance into the realm.
...
 
-An **upsert** is a write operation that either inserts a new
-object with a given primary key or updates an existing
-object that already has that primary key. We call this an
-upsert because it is an "**update** or **insert**"
-operation. This is useful when an object may or may not
-already exist, such as when bulk importing a dataset into an
-existing realm. Upserting is an elegant way to update
-existing entries while adding any new entries.
+An **upsert** is a write operation that either inserts a new  object
```
nathan-contino-mongo: nite: extra space here:
```suggestion
An **upsert** is a write operation that either inserts a new object
```

source/examples/CRUD/Upsert.js
```diff
@@ -1,10 +1,13 @@
...
 
   // Judging by the ID, it's the same person, just with a different name.
-  // When the third parameter (`update`) is true, you overwrite the original
-  // entry (i.e. Drew -> Andy).
-  const andy = realm.create("Person", { id: 1234, name: "Andy" }, true);
+
+  // If an object exists, setting the third parameter (`updateMode`) to
+  // "modified" only updates properties that have changed. This results in
+  // faster operations and changes the way that Realm resolves conflicts
+  // between multiple write operations.
```
cbush: Kind of begs the question though, @nathan-contino-mongo -- "changes" from what to what?

nathan-contino-mongo: yeah i was torn on this. On one hand: I want to give users full information here. On the other hand, explaining a concept this complicated in a code comment is... problematic. I figure if they really want to know how it changes those differences they'll consult the reference material: https://realm.io/docs/javascript/latest/api/Realm.html#create

mongomoe: We can omit the second part of the second sentence. It would make sense for a code comment to be short and omitting it would still satisfy the requirement. 

```  // If an object exists, setting the third parameter (`updateMode`) to "modified" only updates properties that have changed, resulting in faster operations.```

nathan-contino-mongo: This seems like a good solution to me. SGTM

PR #41, opened 2020-02-11
(DOCSP-8478): Clarify User Context in System Functions (Port)

nlarew:
Port of the updated example included in https://github.com/10gen/baas-docs/pull/445

### DIFF
--- a/source/authentication.txt
+++ b/source/authentication.txt
@@ -155,25 +155,21 @@ CRUD and aggregation operations.
 .. admonition:: Security Warning
    :class: important
 
-   Rules do not apply to the system user, which means that
-   functions and webhooks that run as the system user can be
-   a security liability. Ensure that you do not expose these
-   functions to unauthorized users.
+   Rules do not apply to the system user, which means that functions and
+   webhooks that run as the system user can be a security liability.
+   Ensure that you do not expose these functions to unauthorized users.
    
-   For example, instead of calling a system function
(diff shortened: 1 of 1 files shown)


source/authentication.txt
```diff
@@ -155,25 +155,21 @@ CRUD and aggregation operations.
...
    
-   For example, instead of calling a system function
-   directly from a client application, you could call it
-   inside of a second, non-system function that you call
-   from the client. The non-system Function can use
-   :ref:`function context <context-user>` to check if the
-   active user is allowed to call the system function, e.g.:
+   For example, you can use :ref:`function context <context-user>` to
+   check if the active user is allowed to call the system function based
+   on a condition that you define, e.g.:
```
nathan-contino-mongo: I feel like the previous version makes the pattern of a system-check function -> system function very clear, but I worry that some users might misunderstand what you're saying here and create a single system function that contains the following lines of code. I think retaining a single sentence referring to that kind of structure might help reduce that potential confusion. Maybe:
```
You can use the following pattern inside of a non-system function to check if a user has privileges to call a second, system-level function.
```

nlarew: Your worry is actually the intended interpretation! I based the previous example on a misunderstanding of how user context worked in system functions - turns out you don't need to have a separate non-system function to check `context.user`.

nathan-contino-mongo: :raised_hands:

PR #43, opened 2020-02-12
(DOCSP-8912): Port Values & Secrets Overview

cbush:
## JIRA
https://jira.mongodb.org/browse/DOCSP-8912

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker/values-secrets/values-and-secrets


source/includes/steps-define-a-secret-realm-ui.yaml
```diff
@@ -0,0 +1,46 @@
...
+  configuration screen in the Realm UI. To navigate to the
+  configuration screen, click :guilabel:`Values & Secrets` in the
+  left-hand navigation and then select the :guilabel:`Secrets` tab.
+  Click :guilabel:`Add a Secret`.
+---
+title: Name the Secret
+ref: name-the-secret
+level: 4
+content: |
+  Enter a unique :guilabel:`Secret Name`. This name is used to link to
```
ccho-mongodb: Nitpick: I wonder if "associate" might be a better word in this sentence since "link" might be mistaken for a hyperlink. Feel free to disregard if association that does not seem likely.

source/includes/steps-define-a-secret-realm-ui.yaml
```diff
@@ -0,0 +1,46 @@
...
+     Secret names must not exceed 64 characters and may only contain
+     ASCII letters, numbers, underscores, and hyphens. The first
+     character must be a letter or number.
+---
+title: Define the Secret
+ref: define-the-secret
+level: 4
+content: |
+  Enter the value that the Secret refers to in the :guilabel:`Secret
+  Value` input and then enter the same value in the :guilabel:`Confirm
```
ccho-mongodb: Question: is there enough clarity to help a user distinguish between Values of "secret" type and "Secret Values"? I read through the "Values & Secrets" and "Define a Value" sections and it seems like they might be different (Secret requires a corresponding Value of secret type to reference it) because the creation flow involves a different UI navigation. Perhaps an admonition or sentence describing the difference (or equivalence) would be helpful.

source/values-and-secrets/access-a-value.txt
```diff
@@ -0,0 +1,62 @@
...
+Access a Value
+==============
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+You can reference a :ref:`Value <app-value>` from a
+:doc:`service rule expression </services/json-expressions>`
```
ccho-mongodb: Nitpick: should this be "JSON expression" since that's how you refer to it in the subheader a couple lines later? Or at least mention that a service rule expression is the same as "JSON Expression"? Or should "Reference a Value in a JSON Expression" be "Reference a Value in a Service Rule Expression"

source/values-and-secrets.txt
```diff
@@ -0,0 +1,110 @@
...
+   * - :doc:`Access a Value </values-and-secrets/access-a-value>`
+     - Learn how to reference the data stored in a Value from a Function
+       or rule expression.
+
+Summary
+-------
+
+- Values and Secrets are server-side constants you can use in your Realm app.
+- Values allow you to separate deployment-specific configuration data from the business logic of your app.
+- Secrets store sensitive data. You cannot export, import, or directly read a Secret. Instead, you access a Secret indirectly via a named Value.
```
ccho-mongodb: I like this summary!

source/values-and-secrets.txt
```diff
@@ -0,0 +1,110 @@
...
+--------
+
+.. _app-value:
+
+Value
+~~~~~
+
+A **value** is a named reference to a piece of static data
+stored by Realm that you can access in :doc:`Functions
+</functions>` and :doc:`rules </services/json-expressions>`.
```
ccho-mongodb: Question: if you refer to `Functions` and other Realm types with caps, shouldn't `Rules` also be capitalized?

source/includes/steps-define-a-secret-realm-ui.yaml
cbush: Good point. I will use the term "definition", though the UI still says "secret value".

source/values-and-secrets/access-a-value.txt
cbush: I considered this since we know "services" are going away. I will go ahead and do this now.

source/includes/steps-define-a-secret-realm-ui.yaml
```diff
@@ -12,9 +12,9 @@ title: Name the Secret
 ref: name-the-secret
 level: 4
 content: |
-  Enter a unique :guilabel:`Secret Name`. This name is used to link to
-  the Secret from authentication providers, service configurations, and
-  :ref:`Values <app-value>`.
+  Enter a unique :guilabel:`Secret Name`. This name is used
+  to refer to the Secret from authentication providers,
```
ccho-mongodb: 👍

source/includes/steps-define-a-secret-realm-ui.yaml
```diff
@@ -27,10 +27,10 @@ title: Define the Secret
 ref: define-the-secret
 level: 4
 content: |
-  Enter the value that the Secret refers to in the :guilabel:`Secret
-  Value` input and then enter the same value in the :guilabel:`Confirm
-  Secret Value` input. The values must match in order to save the
-  Secret.
+  Enter the Secret's definition in the :guilabel:`Secret
```
ccho-mongodb: Is there any explanation of which usage of "definition" this is? The first meaning I think of when I see "definition" is "explanation of a term". Would "passphrase" also be accurate?

cbush: I'm using definition as in definition of a variable or constant.

This will have to be reworked anyway, as they have added the ability to create secrets with the cli, so we'll come back to it.

PR #46, opened 2020-02-21
Docsp 8771 linking users

mongomoe:
JIRA:
-----
https://jira.mongodb.org/browse/DOCSP-8925

Stage:
------
https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-8771-LinkingUsers/authentication/linking.html


source/authentication/linking.txt
```diff
@@ -0,0 +1,63 @@
+=====================
+Linking User Accounts
+=====================
```
nlarew: Can we update the title: ** Link a New Identity to a User**

source/authentication/linking.txt
```diff
@@ -0,0 +1,63 @@
...
+
+.. default-domain:: mongodb
+
```
nlarew: Let's add an On This Page summary:

```rst
.. contents:: On this page
   :local:
   :backlinks: none
   :depth: 2
   :class: singlecol
```

source/authentication/linking.txt
```diff
@@ -0,0 +1,63 @@
...
+MongoDB Realm provides multiple ways for you to integrate user
+authentication into your apps. For this reason, you may want to provide
+a means for your users to link their accounts from one provider with
+another.
```
nlarew: At this point it isn't clear what it actually means to "link" accounts. I have to read the example paragraph to start to figure it out. Can we say upfront the purpose of linking: i.e. **allow a user to log in using any auth provider that has an identity linked to their account**

It could also be confusing to say "link accounts" - As you probably encountered in testing, if you try to link the credential of a fully confirmed user that already has an account it won't work. For linking there is only ever one user account: **you link an additional, new identity to an existing account.**

source/authentication/linking.txt
```diff
@@ -0,0 +1,63 @@
...
+
+For example, suppose you have an app that enables users to register and
+authenticate using a Username and Password. The application also allows
+users to log in anonymously, allowing new users to explore the app
+without signing up. After working with the app for a while, a user
+decides to create an account. The user now has two accounts with
+separate data on each account. The user would like to consolidate this
+into one account. They can do this by linking the accounts with Realm's
+``linkWithCredential()`` method that is called on the ``StitchUser`` instance.
```
nlarew: As I noted earlier, this isn't actually linking two existing user accounts so this example isn't quite right. Rather than "The user now has two accounts with separate data...." we could frame it as "the user would like to combine a new email/password account with their existing anonymous data."

source/authentication/linking.txt
```diff
@@ -0,0 +1,63 @@
...
+
+
+Usage
+-----
+
+The following usage example simulates an anonymous user login with the
+``loginWithCredential()`` method. This method returns the anonymous
+user. The anonymous user is then linked to an existing user created with
+the username/password authentication provider by calling
+``linkWithCredential()``.
```
nlarew: Not an existing user,  a _confirmed identity pending user login_

source/authentication/linking.txt
```diff
@@ -0,0 +1,63 @@
...
+:guilabel:`General Settings` and then use the
+:guilabel:`linkWithCredential()` method of the :guilabel:`StitchUser`
+class. Find documentation on the ``linkWithCredential()`` method in the
+following SDKs: 
+
+- :android-sdk:`Android SDK <com/mongodb/stitch/android/core/auth/StitchUser.html#linkWithCredential-com.mongodb.stitch.core.auth.StitchCredential->`
+
+- :ios-sdk:`iOS SDK <Protocols/StitchUser.html#/s:10StitchCore0A4UserP4linky0aB3SDK0A10Credential_p04withF0_yAA0A6ResultOyAaB_pGctF>`
+
+- :js-sdk:`JavaScript SDK <interfaces/stitchuser.html#linkwithcredential>`
```
nlarew: - Can we separate the Allowed Request Origins into an admonition?
- `linkWithCredential()` is a method, not a label in the UI, so we shouldn't use `guilabel`
- I think we can incorporate these SDK links in the **Usage** text rather than up here

source/authentication/linking.txt
nlarew: We can link to the API docs for these methods inline:

```rst
:js-sdk:`loginWithCredential() <interfaces/stitchuser.html#loginwithcredential>`
:js-sdk:`linkWithCredential() <interfaces/stitchuser.html#linkwithcredential>`
```

You'll need to put the text inside the tabs to achieve this for each SDK

source/authentication/usage-examples/linking/js.rst
```diff
@@ -0,0 +1,24 @@
...
+    UserPasswordCredential
+  } = require("mongodb-stitch-server-sdk");
+  
+  Stitch.defaultAppClient.auth
+  .loginWithCredential(new AnonymousCredential()) // log the user in anonymously
+  .then(user => {
+    console.log(`Logged in as anonymous user with id: ${user.id}`);
+    user
+      .linkWithCredential(
+        new UserPasswordCredential("<username>", "<password>") // call user.linkWithCredential to link the anonymous user with an already registered user/password user
```
nlarew: We could probably still be clearer on what "already registered" means. Maybe add more to the code comments here?

source/authentication/usage-examples/linking/android.rst
```diff
@@ -0,0 +1,17 @@
+.. code-block:: java
+
+  // Called when an anonymous user decides to create an account with username and password.
+  private void linkWithNewUserPasswordUser(String username, String password) {
+    // Get the currently logged-in (anonymous) user from Stitch.
+    final StitchAuth auth = Stitch.getDefaultAppClient().getAuth();
+    final StitchUser user = auth.getUser();
+
+    // Create and link a new account with the given username and password.
```
nlarew: Technically not a new account, a new **identity**

source/authentication/linking.txt
mongomoe: Fixed

source/authentication/linking.txt
mongomoe: Fixed

source/authentication/linking.txt
mongomoe: Updated

source/authentication/linking.txt
mongomoe: Updated

mongomoe: Updated to be inline links

source/authentication/linking.txt
mongomoe: * I really don't think the Allowed Request Origins should be an admonition, because admonitions are sometimes be overlooked - especially when there are more than one per page (we saw that with the CSFLE research). Since this is a core thing that needs to be done in order to accomplish linking, I think it should remain here.
* Fixed
* Fixed inline

source/authentication/usage-examples/linking/android.rst
mongomoe: Fixed

source/authentication/usage-examples/linking/js.rst
mongomoe: altered to: "// call user.linkWithCredential to link the anonymous identity (and it's data) to the account created with that username and password"

mongomoe:
@nlarew all comments addressed except for your request for a table or documentation explaining the different statuses required to link per auth provider.

The reason for this is because of the current warning for email/password should suffice. With all other auth providers (except email/pass and anonymous), an identity and user account are created upon first login, so they can be linked without any special status. This is unlike email/pass which is the only identity to require the specific status of "Pending User Login".

Martin explained it well in this slack comment: https://mongodb.slack.com/archives/C41SX3C4D/p1582831096085800?thread_ts=1582828698.085400&cid=C41SX3C4D

source/authentication/linking.txt
mongomoe: updated

source/authentication/usage-examples/linking/js.rst
```diff
@@ -0,0 +1,24 @@
+.. code-block:: javascript
+
+  const {
+    Stitch,
+    AnonymousCredential,
+    UserPasswordCredential
+  } = require("mongodb-stitch-server-sdk");
```
nlarew: This should use `mongodb-stitch-browser-sdk`

source/authentication/usage-examples/linking/js.rst
```diff
@@ -0,0 +1,24 @@
...
+        new UserPasswordCredential("<username>", "<password>") // call user.linkWithCredential to link the anonymous identity (and it's data) to the account created with that username and password
+      )
+      .then(users => {
+        console.error("accounts linked:\t", users);
+      })
+      .catch(err => {
+        console.error("err occurred in linking:\t", err);
+      });
+  })
+  .catch(console.error);
```
nlarew: - Check the indentation on this promise chain
- Don't nest promise chains, just return the promise an handle it in the next `.then()`
- You should only call `console.error` if there's an error to report

```js
Stitch.defaultAppClient.auth
  .loginWithCredential(new AnonymousCredential()) // log the user in anonymously
  .then(user => {
    console.log(`Logged in as anonymous user with id: ${user.id}`);
    // Link the anonymous identity (and its data) to the new email/password user
    return user.linkWithCredential(new UserPasswordCredential("<username>", "<password>"))
  })
  .catch(err => {
    console.error("Failed to link user: ", err);
  });

```

source/authentication/linking.txt
```diff
@@ -0,0 +1,94 @@
...
+want to link an additional, new identity to an existing user account.
+Consider the following example:
+
+A developer using MongoDB Realm has an app that allows users to explore
+the app anonymously without signing up upon first viewing. After working
+with an app, a user decides to create an account with their
+Email/Password.  The user now has data associated with their anonymous
+identity, and they would like to link it to their newly created
+Email/Password user account. They can do this by calling Realm's
+``linkWithCredential()`` method on the ``StitchUser`` instance.
```
nlarew: - The multi-paragraph construction here makes it a little unclear that this is an example / jarring to read. I'd combine the statement that this is an example in the same paragraph as the example itself.

  ```
  Consider an example application that allows users to explore anonymously
  without first signing up for an account...
  ```
- This is an example so we want to be specific: s/an app/the app
  
  ```
  After using the app for a while...
  ```

- "their Email/Password" reads a bit funny. Maybe "a user decides to create a new email/password account"

- "the user now has data" implies that this is the result of them making the email/password account but it's not. It's also not totally clear what "data" is in this context. Maybe: "The user wants to associate their anonymous browsing history with their new account."

source/authentication/linking.txt
```diff
@@ -0,0 +1,94 @@
...
+      :guilabel:`StitchUser` class.
+   .. tab::
+      :tabid: android
+
+      To provide user account linking in your code, add your app's
+      hosted domain to the list of :guilabel:`Allowed Request Origins`
+      in :guilabel:`General Settings` and then use the
+      :android-sdk:`linkWithCredential() <com/mongodb/stitch/android/core/auth/StitchUser.html#linkWithCredential-com.mongodb.stitch.core.auth.StitchCredential->` method
+      of the :guilabel:`StitchUser` class.     
+   .. tab::
```
nlarew: nit: add an extra newline between these tabs so they have some breathing room

source/authentication/linking.txt
```diff
@@ -0,0 +1,94 @@
...
+         .. include:: ./usage-examples/linking/android.rst
+     - id: ios
+       content: |
+         .. include:: ./usage-examples/linking/ios.rst
+
+
+Summary
+-------
+- MongoDB Realm applications can have multiple authentication providers, and multiple identities from various authentication providers can be linked to one Realm user.
+- Identities are linked to a user using the ``linkWithCredential()`` method on a ``StitchUser`` instance.
```
nlarew: nit: can we wrap these so the lines aren't so long?

source/authentication/usage-examples/linking/js.rst
mongomoe: It works with either for email/pass and anon (but not oauth), but fixed

source/authentication/usage-examples/linking/js.rst
mongomoe: * Fixed
* Makes sense, fixed 
* Sorry, overlooked this - Ik you should only call `console.error` when there's an error. I probably just copied & pasted too quickly. Thanks for catching this - Fixed

source/authentication/linking.txt
mongomoe: odd nit - but sure, fixed

source/authentication/linking.txt
mongomoe: I wish we could but then they don't render as bullets

source/authentication/linking.txt
mongomoe: * Fixed verbatim
* Fixed 
* Reworded the Email/Password thing to: "After using the app for a while, a user decides to create a new account by providing their email and password."
* Fixed verbatim

source/authentication/usage-examples/linking/js.rst
nlarew: Yeah I'm sure it works. Main point here is that all of our examples use the browser SDK so this one should too.

source/authentication/linking.txt
nlarew: RIP add it to the NextGen render bug pile

PR #47, opened 2020-02-21
(DOCSP-8925): Enable Hosting PORT - switch to 'Realm',Fixed grammar &…

mongomoe:
JIRA:
-----
https://jira.mongodb.org/browse/DOCSP-8925

Stage:
------
https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-8925-Port-Enable-Hosting/hosting/enable-hosting.html

### DIFF
--- a/source/hosting/enable-hosting.txt
+++ b/source/hosting/enable-hosting.txt
@@ -1,0 +1,34 @@
+==============
+Enable Hosting
+==============
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+You need to enable static hosting for your application before you can
+upload and access content. To allow static hosting from the Realm UI or
+by :doc:`importing </deploy/deploy-changes-with-realm-cli>` an
+application directory that enables hosting in its configuration file.
+Select the tab below that corresponds to the method you want to use.
+
+
+Procedure
+---------
+
+.. tabs-stitch-interfaces::
+
+   tabs:
+     - id: stitch-ui
+       content: |
+         .. include:: /includes/steps/enable-hosting-realm-ui.rst
+     - id: import-export
+       content: |
+         .. include:: /includes/steps/enable-hosting-import-export.rst
--- a/source/hosting/enable-hosting.txt
+++ b/source/hosting/enable-hosting.txt
@@ -7,21 +7,24 @@ Enable Hosting
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 2
+   :depth: 1
    :class: singlecol
 
 Overview
 --------
 
 You need to enable static hosting for your application before you can
-upload and access content. To allow static hosting from the Realm UI or
-by :doc:`importing </deploy/deploy-changes-with-realm-cli>` an
+upload and access content. You can allow static hosting from the Stitch
+UI or by :doc:`importing </deploy/deploy-changes-with-realm-cli>` an
 application directory that enables hosting in its configuration file.
 Select the tab below that corresponds to the method you want to use.
 
 
 Procedure
 ---------
+
+Select a tab below to see instructions on enabling static hosting for
+the corresponding method.
 
 .. tabs-stitch-interfaces::
 
--- a/source/includes/steps-enable-hosting-import-export.yaml
+++ b/source/includes/steps-enable-hosting-import-export.yaml
@@ -10,7 +10,7 @@ content: |
   following command from an authenticated instance of :doc:`realm-cli
   </deploy/realm-cli-reference>`:
 
-  .. code-block:: shell
+  .. code-block:: none
 
      realm-cli export --app-id=<App ID>
 
@@ -34,7 +34,7 @@ content: |
   Navigate to the root of the application directory and run the
   following command:
 
-  .. code-block:: shell
+  .. code-block:: none
 
      realm-cli import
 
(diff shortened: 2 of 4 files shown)


source/hosting/enable-hosting.txt
```diff
@@ -0,0 +1,34 @@
+==============
+Enable Hosting
+==============
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
```
ccho-mongodb: Nitpick: you should only need depth of 1 since there are no subheaders.

source/hosting/enable-hosting.txt
```diff
@@ -0,0 +1,34 @@
...
+   :class: singlecol
+
+Overview
+--------
+
+You need to enable static hosting for your application before you can
+upload and access content. To allow static hosting from the Realm UI or
```
ccho-mongodb: I think this sentence is not complete. I think you maybe missed a couple words and meant something like this:
"To allow static hosting, enable it in the Realm UI or import an application directory..."

source/hosting/enable-hosting.txt
```diff
@@ -0,0 +1,34 @@
...
+by :doc:`importing </deploy/deploy-changes-with-realm-cli>` an
+application directory that enables hosting in its configuration file.
+Select the tab below that corresponds to the method you want to use.
```
ccho-mongodb: Would it be more clear to the user if you mentioned what they should see in the tabs below? E.g.
"Select a tab below to see instructions on enabling static hosting for the corresponding method."

source/includes/steps-enable-hosting-import-export.yaml
```diff
@@ -0,0 +1,50 @@
+title: Export Your Realm Application
+ref: export-your-application
+content: |
+  To enable static hosting for your  application with
+  :doc:`realm-cli </deploy/realm-cli-reference>` you need to
```
ccho-mongodb: This isn't showing up as a link for me. Is it a snooty bug?

Screen Shot 2020-02-21 at 2 30 08 PM

ccho-mongodb: Same comment for the rest of the links in this step file.

source/includes/steps-enable-hosting-import-export.yaml
```diff
@@ -0,0 +1,50 @@
...
+  export an :ref:`application directory <application-directory>`.
+  You can :doc:`export your application </deploy/export-realm-app>`
+  configuration from the :guilabel:`Export` tab of the
+  :guilabel:`Settings` page in the Realm UI, or by running the
+  following command from an authenticated instance of :doc:`realm-cli
+  </deploy/realm-cli-reference>`:
+
+  .. code-block:: shell
```
ccho-mongodb: Although you're correct that this is a `shell` command, you might consider using `none` to avoid seemingly-random highlighting of "export" and nothing else.

ccho-mongodb: Similar comment for the code-block in line 37.

source/includes/steps-enable-hosting-import-export.yaml
nlarew: We haven't ported the Realm CLI page yet so the link has nowhere to point

source/includes/steps-enable-hosting-import-export.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+     realm-cli export --app-id=<App ID>
+
+---
+title: Enable Hosting in the Application Configuration
+ref: enable-hosting-in-the-application-configuration
+content: |
+  In :ref:`realm.json <appschema-realm-config>`, set the value of
```
ccho-mongodb: Since the link is not rendering, it seems out of place to have this formatted as plain text and then in bold on line 32.

source/hosting/enable-hosting.txt
```diff
@@ -0,0 +1,34 @@
...
+
+
+Procedure
+---------
+
+.. tabs-stitch-interfaces::
+
+   tabs:
+     - id: stitch-ui
```
ccho-mongodb: Should this tab and the tab set use "realm" instead of "stitch"?

mongomoe: Ideally, it should, and eventually it will but currently, these are to be left as Stitch-UI

source/includes/steps-enable-hosting-import-export.yaml
mongomoe: Good idea - switched to `none`

source/hosting/enable-hosting.txt
mongomoe: That's a great suggestion! Taken verbatim

source/hosting/enable-hosting.txt
mongomoe: Thanks, fixed to:

"You can allow static hosting from the Stitch UI or by importing an application directory that enables hosting in its configuration file."

source/hosting/enable-hosting.txt
mongomoe: Fixed

source/hosting/enable-hosting.txt
ccho-mongodb: Ok, it was unclear since the JIRA ticket said:

> Convert all instances of "Stitch" to "Realm"/"MongoDB Realm"

PR #75, opened 2020-02-27
(DOCSP-8951): Port Function Context Reference

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-8951

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-8951/functions/context.html


source/functions/context.txt
```diff
@@ -0,0 +1,465 @@
...
+A property that contains the :ref:`user object <user-objects>` of the
+function's :ref:`active user <function-active-user>`.
+
+.. admonition:: System Functions
+   :class: note
+   
+   In a :ref:`system function <system-functions>`, ``context.user``
+   resolves to the authenticated application user that called a function
+   even though the function executes as the system user and bypasses all
+   rules. If the function was not called by an authenticated user, such
```
mongomoe: Not required for LGTM, and feel free to ignore/not comment on this if you disagree or like your sentence better:
---------------------------------------------------------------------------------------------------------------

The following sentence is a bit wordy & since it contains a lot of technical jargon it may be hard for readers to understand:

"If the function was not called by an authenticated user, such as in a webhook or trigger, then context.user resolves to a system user object that has no data." 

Suggested rephrasing:

"If the authenticated user does not call the function, then the context.user resolves to a system user object with no data. Webhooks or triggers are two cases where the authenticated user does not call the function. "

source/functions/context.txt
```diff
@@ -0,0 +1,465 @@
...
+
+   * - ``remoteIPAddress``
+     - string
+     - The IP address of the client that issued the Function request.
+
+   * - ``requestHeaders``
+     - object
+     - An object where each field maps to a type of :mdn:`HTTP Header
+       <Web/HTTP/Headers>` that was included in the request that caused
+       the function to execute. The value of each field is an array of
```
mongomoe: Not required for LGTM, and feel free to ignore/not comment on this if you disagree or like your sentence better:
-----------------------------------------------------------------------------------------------------------------
The sentence - 

"The value of each field is an array of strings where each string maps to a header of the specified type that was included in the request."

-is a bit run-on, A suggestion to break it up like this:

"The value of each field is an array of strings. Each string maps to a header of the specified type that was included in the request."

source/functions/context.txt
nathan-contino-mongo: Good point, this is kind of wordy and confusing. I've reworded it -- take a look to tell me what you think.

Paging @nlarew to make sure this is technically correct still (and to see if he has any wording suggestions).

nlarew: My only hangup is that even if "the authenticated user calls the function" it may still be a system function. Generally it's all dependent on the configuration.

Frankly I think we can just delete this whole sentence. Thoughts?

PR #79, opened 2020-02-28
Fix some warnings/links

cbush:
- Fixing some broken links and warnings


source/functions.txt
```diff
@@ -125,12 +125,12 @@ System Functions
 ~~~~~~~~~~~~~~~~
 
 A **System Function** is a function that runs as the :ref:`system user
-<system-users>` rather than an :ref:`application user <user-accounts>`.
```
nathan-contino-mongo: The anchor for this is plural, did you change the definition?

cbush: Yes, previously

nathan-contino-mongo: Ah I see. I was mixing this up with the `system-functions` anchor.

PR #90, opened 2020-03-02
(DOCSP-9186): Port GraphQL Types and Resolvers

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-9186

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-9186/graphql/types-and-resolvers.html


source/graphql/types-and-resolvers.txt
```diff
@@ -0,0 +1,672 @@
...
+.. _graphql-input-types:
+
+Input Types
+~~~~~~~~~~~
+
+GraphQL uses :graphql:`input types <learn/schema/#input-types>` to
+represent parameters that you pass to a mutation. This is a standard
+approach used by all GraphQL APIs. The GraphQL docs describe why:
+
+    Objects can contain fields that express circular references or
```
mongomoe: Is this how it's supposed to be rendered - it might look better represented as an admonition or with some other formatting:
Screen Shot 2020-03-02 at 1 41 25 PM

mongomoe: (ignore the highlighting)

nathan-contino-mongo: I see what you're saying but this is a direct quote from the docs. This style is pretty common for quotes, and I feel like putting this in an admonition would make it less obvious that these are not *our* words.

PR #93, opened 2020-03-02
(DOCSP-9146): Link a MongoDB Atlas Cluster

mongomoe:
JIRA:
-----
* https://jira.mongodb.org/browse/DOCSP-9146

Original/Adjacent Stitch Doc:
-----------------------------
* https://docs.mongodb.com/stitch/mongodb/link-a-cluster/

Stage:
-------
* https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCS-9146-LinkACluster/mongodb/link-a-cluster.html


source/index.txt
```diff
@@ -66,8 +66,10 @@ MongoDB Realm
 .. toctree::
    :titlesonly:
    :caption: MongoDB
-
+   
+   MongoDB Overview </mongodb>
```
nathan-contino-mongo: Why is this linked twice now?

source/includes/steps-link-a-cluster-import-export.yaml
```diff
@@ -0,0 +1,90 @@
...
+  Once you've added the MongoDB service configuration, all that's left
+  is to import the function.
+
+  A. Save the ``config.json``. 
+  #. Navigate to the root of the exported application directory.
+  #. Log in to MongoDB Atlas with ``realm-cli``:  
+
+     .. code-block:: shell
+
+        stitch-cli login --username="<MongoDB Cloud Username>" --api-key="<MongoDB Cloud API Key>"
```
nathan-contino-mongo: This uses the old username/api-key login method. We should probably change this to use a private API key instead. You can follow the example here:
https://github.com/mongodb/docs-realm/blob/master/source/includes/steps-deploy-cli.yaml#L31

source/includes/steps-link-a-cluster-realm-ui.yaml
```diff
@@ -0,0 +1,62 @@
...
+  which could take up to five minutes. Once Realm is finished linking
+  to the cluster, you can start reading and writing data.
+
+  .. see::
+
+     - :doc:`Add Data to Mongodb </mongodb/add-data-to-mongodb>`
+     - :doc:`Find Documents </mongodb/find-documents-in-mongodb>`
+     - :doc:`Update Documents </mongodb/update-documents-in-mongodb>`
+     - :doc:`Delete Documents </mongodb/delete-documents-from-mongodb>`
+     - :doc:`Run Aggregation Pipelines </mongodb/run-aggregation-pipelines>`
```
nathan-contino-mongo: These follow-up examples feel a little weird as-is. I'd recommend one of two choices:
1) Move the end of this outside of the YAML to the end of the txt page itself. Would look something like:
```
Once Realm is finished linking to the cluster, you can start reading and writing data. The following pages demonstrate how to start working with your linked cluster:
<list of pages>"
```
2) Delete the "see" directive and its list of follow-up pages from the UI. Then both yaml includes share the same endpoint.

source/includes/steps-link-a-cluster-realm-ui.yaml
```diff
@@ -0,0 +1,62 @@
...
+title: Link a New Cluster
+ref: link-a-new-cluster
+content: |
+  Click :guilabel:`Link a Cluster` and provide the following
+  configuration information on the :guilabel:`Cluster Configuration`
+  screen:
+
+  .. list-table::
+    :header-rows: 1
+    :widths: 10 30
```
nathan-contino-mongo: On my browser, the far right edge of the second column (including several letters) is cropped off on this page. I would suggest decreasing the width of either/both columns until that doesn't happen any more, since horizontal scrolling is a pain and we don't even show the scrollbar.

source/index.txt
mongomoe: It should not be - fixed now

source/includes/steps-link-a-cluster-import-export.yaml
mongomoe: How's this:

```
realm-cli login --api-key=<My API Key> --private-api-key=<My Private API Key>
```

(Updated the doc to reflect this ^ )

source/includes/steps-link-a-cluster-realm-ui.yaml
mongomoe: Fixed:
- I've abstracted the finale paragraph and the links from those yaml files and put them in the main doc.

PR #95, opened 2020-03-04
(DOCSP-9192): Port Trigger Snippets

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-9192

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-9192/triggers/trigger-snippets.html


nathan-contino-mongo:
The block is just there for parallelism with the database triggers section (so it isn't strictly necessary itself), but I do feel like it would be strange to remove it because it's nice to see a list of all the examples before you dive straight in since they're kind of long.

PR #128, opened 2020-03-10
(DOCSP-9345): Port HTTP Service Overview

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-9345

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-9345/services/http.html

### DIFF
--- a/source/services/http.txt
+++ b/source/services/http.txt
@@ -141,7 +141,7 @@ Configuration
 
        .. note::
 
-          Each incoming webhook in a Twilio service interface must
+          Each incoming webhook in an HTTP service interface must
           have a unique name.
 
    * - :guilabel:`Respond With Result`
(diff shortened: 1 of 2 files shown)


source/services/http.txt
```diff
@@ -0,0 +1,340 @@
...
+       .. tab-content:: tabs-realm-admin-interfaces
+          :tab-id: cli
+
+          | *name*
+
+     - Required. The name of the webhook.
+
+       .. note::
+
+          Each incoming webhook in a Twilio service interface must
```
cbush: Should this be "HTTP"? Same below

nathan-contino-mongo: good question -- while I'd guess that this one is a typo, the other one seems too twilio-specific to apply here. This is exactly the note on the original Stitch page so I'm going to reach out to @nlarew for answers.

nathan-contino-mongo: Seems like a copy/paste error when the page was originally made, changing this occurrence to HTTP and removing the other note altogether (since the HTTP service has no equivalent to the Twilio text message side effect)

PR #142, opened 2020-03-17
(DOCSP-7083): Glossary for Realm

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-7038

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-7038/get-started/glossary.html

### DIFF
--- a/source/get-started/glossary.txt
+++ b/source/get-started/glossary.txt
@@ -72,7 +72,7 @@ Glossary
    compaction
       Allocation and deallocation of data within a :term:`Realm Database`
       and/or exceeding the size of physical memory can lead to
-      fragmentation on disk. Calling the ``compact()`` method
+      fragmentation on disk. Calling the ``compact`` method
       reorganizes internal data structures to reduce the storage
       overhead of your local Realm instance in the event of
       fragmentation.
@@ -95,7 +95,7 @@ Glossary
       :term:`Zero-Copy` objects that access object attributes via
       shallow pointers into memory-mapped data. You can subscribe to
       changes to any live object or collection of live objects,
-      including realm instances.
+      including Realm instances.
 
    write transaction
       You can only write to a :term:`Realm` during a write
--- a/source/get-started/glossary.txt
+++ b/source/get-started/glossary.txt
@@ -194,7 +194,7 @@ Glossary
       you can access :term:`Realm app` components, including functions
       and :term:`MongoDB Atlas` data.
 
-   value
+   Realm value
       A named reference to a piece of static data stored by
       :term:`MongoDB Realm`, accessible through :term:`Realm functions`
       and :term:`Realm rules`.
@@ -203,7 +203,7 @@ Glossary
       True or false expressions that determine whether or not a user can
       modify, insert, or read a given piece of data.
 
-   secret
+   Realm secret
       A private value stored in :term:`MongoDB Realm` that cannot be
       directly read, but can be linked in authentication providers and
       service configurations.
--- a/source/get-started/glossary.txt
+++ b/source/get-started/glossary.txt
@@ -26,39 +26,38 @@ Glossary
       </core/document/>`.
 
    Partition Key
-      A field within a :term:`MongoDB Document` that determines which Realm the
-      corresponding Realm Object belongs to.
+      A field within a :term:`MongoDB Document` that determines which realm
+      contains the corresponding Realm Object.
 
    Realm
-      A collection of :term:`Realm Object`s sharing the same
-      :term:`Partition Key` value.
-      Client instances subscribe to individual Realms which then passively
-      synchronize changes as network availability allows.
+      A collection of :term:`Realm Objects <Realm Object>` sharing the same
+      :term:`Partition Key` value. Client instances subscribe to
+      individual realms which passively synchronize changes as network
+      availability allows.
+
+   Realm Client
+      Anything that interacts with a :term:`Realm App` using the
+      :term:`Realm SDK`. Examples include phone apps, IoT devices,
+      and server applications.
+
+   Realm App
+      An instance of the :term:`MongoDB Realm` Backend-as-a-Service.
 
    Realm Database
       An :ref:`ACID`-compliant offline-first object database for mobile devices.
       Using the Realm SDK, you can query a local instance of the Realm
       Database and synchronize your local instance with data stored
-      inside :term:`MongoDB Atlas`. Offline-first functionality ensures that
+      inside :term:`MongoDB Atlas`. Offline-first design ensures that
       reads and writes query the local database instance only. When Sync
       is enabled, a separate background thread handles synchronization
       of changes between the local instance and your instance of MongoDB
       Atlas.
    
    Realm Object
-      Object that you can store in a :term:`Realm`, defined by a
+      Object that you can store in a :term:`realm <Realm>`, defined by a
       :term:`Realm Schema` through either any Realm SDK or the
       Realm Schema editor on the MongoDB Realm admin site. Object schemas
       can define primitive fields as well as embedded lists and relations.
-
-   Realm Platform
-      The combination of :term:`MongoDB Atlas` with a :term:`MongoDB Realm`
-      sync-enabled mobile app.
-      This platform abstracts away the network, synchronization, and
-      conflict-resolution aspects of your application. Object data within
-      the mobile :term:`Realm Database` synchronizes with collections of
-      documents representing those objects in MongoDB Atlas. Realm
-      membership is determined using a :term:`Partition Key`.
 
    GraphQL
       Open-source data query and manipulation language for APIs. Realm
@@ -66,7 +65,7 @@ Glossary
       to data changes. An alternative to :term:`Realm Database` queries
       using the client's native query language.
 
-   Realm Compaction
+   Compaction
       Allocation and deallocation of data within a :term:`Realm Database`
       can lead to memory fragmentation. Calling the ``compact()`` method
       reorganizes internal data structures to reduce the storage
@@ -75,7 +74,13 @@ Glossary
 
    Realm Schema
       Definition of all :term:`Realm Objects` that you can read or write
-      into your local :term:`Realm Instance`.
+      into a client :term:`Realm Instance` or :term:`MongoDB Atlas`
+      instance.
+
+   Schema Validation
+      Rules for an instance of MongoDB that specify the form of documents
+      that you can insert into each collection, including both fields
+      and values.
 
    Live Objects
       Instances of objects from your :term:`Realm Schema` that
@@ -102,24 +107,26 @@ Glossary
       :term:`Event Loop` also refresh UI-thread read transactions in
       between instances of the event loop.
 
-   Event Loop
-      Also known as the **main loop** or **main event loop** on some
-      platforms. A programming paradigm in which a central loop waits
+   Loop Thread
+      A design pattern in which a central loop waits
       for signals from a UI or data source before updating underlying
-      user-facing components.
+      data structures. On iOS, known as a :apple-dev:`run loop
+      <foundation/runloop>`. On Android, known as a :android-dev:`Looper
+      <reference/android/os/Looper>`. Also known as the **main loop** or
+      **main event loop** on some platforms.
 
    Realm Instance
       An interface through which a client application can read and write
-      to a local :term:`Realm`.
+      to a local :term:`realm <Realm>`.
 
    Realm Version
-      A snapshot of data in a local :term:`Realm` at a given moment.
+      A snapshot of data in a local :term:`realm <Realm>` at a given moment.
 
    Realm File
       File containing all data for a local :term:`Realm` for an instance
       of :term:`Realm Database`. Because Realm Database reads and
       writes Realm data using :wikipedia:`memory mapping
-      </Memory-mapped_file`, Realm files should not exceed the amount of
+      </Memory-mapped_file>`, Realm files should not exceed the amount of
       RAM on a mobile device.
 
    ACID
--- a/source/get-started/glossary.txt
+++ b/source/get-started/glossary.txt
@@ -44,13 +44,18 @@ Glossary
       :term:`Realm SDK`. Examples include phone apps, IoT devices,
       and server applications.
 
+   Realm SDK
+      Library used by a :term:`Realm client` to connect to a
+      :term:`MongoDB Realm` application or a local
+      :term:`Realm Database`.
+
    Realm app
       An instance of the :term:`MongoDB Realm` Backend-as-a-Service.
 
    Realm Database
-      An :ref:`ACID`-compliant offline-first object database for mobile devices.
-      Using the Realm SDK, you can query a local instance of the Realm
-      Database and synchronize your local instance with data stored
+      An :term:`ACID`-compliant offline-first object database for mobile
+      devices. Using the Realm SDK, you can query a local instance of
+      the Realm Database and synchronize your local instance with data stored
       inside :term:`MongoDB Atlas`. Offline-first design ensures that
       reads and writes query the local database instance only. When Sync
       is enabled, a separate background thread handles synchronization
@@ -78,8 +83,8 @@ Glossary
       fragmentation.
 
    Realm Schema
-      Definition of all :term:`Realm objects` that you can read or write
-      into a client :term:`Realm instance` or linked
+      Definition of all :term:`Realm objects <Realm object>` that you
+      can read or write into a client :term:`Realm instance` or linked
       :term:`MongoDB Atlas` cluster.
 
    Schema Validation
@@ -197,8 +202,8 @@ Glossary
 
    Realm value
       A named reference to a piece of static data stored by
-      :term:`MongoDB Realm`, accessible through :term:`Realm functions`
-      and :term:`Realm rules`.
+      :term:`MongoDB Realm`, accessible through :term:`Realm functions
+      <Realm function>` and :term:`Realm rules`.
 
    Realm rules
       True or false expressions that determine whether or not a user can
--- a/source/get-started/glossary.txt
+++ b/source/get-started/glossary.txt
@@ -112,7 +112,8 @@ Glossary
       **main event loop** on some platforms.
 
    Realm instance
-      An interface through which a client application can read and write
+      The result returned by a call to ``realm.open()``. An interface
+      through which a :term:`Realm client` application can read and write
       to a local :term:`Realm`. Every Realm instance references a
       :term:`Realm version`. You can update the Realm version referenced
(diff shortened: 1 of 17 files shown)


source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+      the ``realm.open()`` method begins a read transaction and the
+      ``realm.close()`` method closes the read transaction. Read
+      transactions always use the :term:`Realm Version` that was most
+      current when the realm was opened unless the client explicitly
+      advances the version. You can explicitly advance the version of a
+      read transaction by starting or committing a write transaction on
+      the same thread as the read transaction, or by calling the
+      ``realm.refresh()`` method. Platforms with a built in
+      :term:`Event Loop` also refresh UI-thread read transactions in
+      between instances of the event loop.
```
cbush: Hmm... we had not been framing reads as "transactions" before. I would read the existing pages on reads, transactions, and threading. If you think we can see where to reframe those pages and mention "read transaction" then we can keep this. Otherwise, I would suggest reframing this entry in the glossary to be more in line with how reads are explained in the existing sections. Thoughts?

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+
+   Event Loop
+      Also known as the **main loop** or **main event loop** on some
+      platforms. A programming paradigm in which a central loop waits
```
cbush:
```suggestion
      platforms. A programming design pattern in which a central loop waits
```

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+      for signals from a UI or data source before updating underlying
+      user-facing components.
```
cbush: Not necessary user-facing though. A run loop can exist entirely on a background thread.

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+      current when the realm was opened unless the client explicitly
+      advances the version. You can explicitly advance the version of a
+      read transaction by starting or committing a write transaction on
+      the same thread as the read transaction, or by calling the
+      ``realm.refresh()`` method. Platforms with a built in
+      :term:`Event Loop` also refresh UI-thread read transactions in
+      between instances of the event loop.
+
+   Event Loop
+      Also known as the **main loop** or **main event loop** on some
```
cbush: I would inform this from the page on threading, which tried to generalize '"loop" threads'. Beyond that do we need to explain such a generic concept? Where else is this relevant to Realm outside of the context of refreshing realms?

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+      to a local :term:`Realm`.
+
+   Realm Version
+      A snapshot of data in a local :term:`Realm` at a given moment.
+
+   Realm File
+      File containing all data for a local :term:`Realm` for an instance
+      of :term:`Realm Database`. Because Realm Database reads and
+      writes Realm data using :wikipedia:`memory mapping
+      </Memory-mapped_file`, Realm files should not exceed the amount of
```
cbush:
```suggestion
      </Memory-mapped_file>`, Realm files should not exceed the amount of
```

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+
+   Event Loop
+      Also known as the **main loop** or **main event loop** on some
+      platforms. A programming paradigm in which a central loop waits
+      for signals from a UI or data source before updating underlying
+      user-facing components.
+
+   Realm Instance
+      An interface through which a client application can read and write
+      to a local :term:`Realm`.
```
cbush:
```suggestion
      to a local :term:`realm`.
```

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+      :term:`Event Loop` also refresh UI-thread read transactions in
+      between instances of the event loop.
+
+   Event Loop
+      Also known as the **main loop** or **main event loop** on some
+      platforms. A programming paradigm in which a central loop waits
+      for signals from a UI or data source before updating underlying
+      user-facing components.
+
+   Realm Instance
```
cbush: I think the distinction between this, `realm` AND `Realm App` could be more clear in your description - in other words, why do we need a distinction between "Realm Instance" and "realm"?

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+
+   MongoDB Document
+      See the MongoDB Manual's :manual:`definition of document
+      </core/document/>`.
+
+   Partition Key
+      A field within a :term:`MongoDB Document` that determines which Realm the
+      corresponding Realm Object belongs to.
+
+   Realm
```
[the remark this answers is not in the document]
nlarew: In this instance I feel like we _should_ capitalize to preserve title case

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+      of changes between the local instance and your instance of MongoDB
+      Atlas.
+   
+   Realm Object
+      Object that you can store in a :term:`Realm`, defined by a
+      :term:`Realm Schema` through either any Realm SDK or the
+      Realm Schema editor on the MongoDB Realm admin site. Object schemas
+      can define primitive fields as well as embedded lists and relations.
+
+   Realm Platform
```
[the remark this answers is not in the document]
nathan-contino-mongo: I was on the fence about this (it's essentially synonymous with "MongoDB Realm") but this is the nail in the coffin -- I'm gonna kill this entry off. Good point!

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+      documents representing those objects in MongoDB Atlas. Realm
+      membership is determined using a :term:`Partition Key`.
+
+   GraphQL
+      Open-source data query and manipulation language for APIs. Realm
+      GraphQL supports reading and writing data as well as subscribing
+      to data changes. An alternative to :term:`Realm Database` queries
+      using the client's native query language.
+
+   Realm Compaction
```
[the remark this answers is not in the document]
nathan-contino-mongo: Going to preserve title case but when I refer to this elsewhere I will use lowercase.

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+      Allocation and deallocation of data within a :term:`Realm Database`
+      can lead to memory fragmentation. Calling the ``compact()`` method
+      reorganizes internal data structures to reduce the storage
+      overhead of your local Realm instance in the event of
+      fragmentation.
+
+   Realm Schema
+      Definition of all :term:`Realm Objects` that you can read or write
+      into your local :term:`Realm Instance`.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Not sure if "remote" is really a meaningful term here, but I'm going to explicitly state "client Realms" and "MongoDB Atlas" because you make a good point -- this doesn't just apply locally! It is a little unfortunate that your schema != validation, though -- items can end up in Atlas that violate the schema and don't sync with Realm as a result. Because of this I'm going to add a Schema Validation glossary entry.

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+
+   Live Objects
+      Instances of objects from your :term:`Realm Schema` that
+      automatically update in value to match the value of the underlying
+      object in the enclosing :term:`Read Transaction`. Note that the
+      enclosing read transaction might not reference the latest
+      :term:`version <Realm Version>`.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Definitely a bit of a judgment call here. I personally think it's an important enough point to bring up -- Live Objects are "Live" in a sense of reflecting the latest version of the object *within their enclosing _read transaction_*, not "Live" in a sense of always reflecting the most-recent data in the DB.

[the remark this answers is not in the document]
cbush: I'm just saying: could this level of detail actually lead to more confusion? Is it possible to link out to the proper detail page instead?

source/get-started/glossary.txt
nathan-contino-mongo: I think it's very possible to restructure that page to reflect this, but I'll also throw this toward Christian to see if he has any suggestions of how to word it better/make it clearer. I think it's a really important concept so I definitely want to keep it!

source/get-started/glossary.txt
nathan-contino-mongo: I think it's valuable in some sense because Event Loop/Looper/etc. threads have a special privileged status for Read Transactions that could potentially confuse users. However I think the Threading page covers this enough that we can just get rid of this.

nathan-contino-mongo: Moving some of the Threading page description into an "Event Loop" entry. I suspect we're going to have to link to this in multiple places, so I think it's best to consolidate a definition here. Let me know if you have any other ideas!

source/get-started/glossary.txt
nathan-contino-mongo: "Realm Instance" -> programming interface
"Realm" -> collection of disparate objects that belong "together" in some sense
"realm" -> ???
"Realm App" -> an instance of mongodb realm? I think?

@nlarew @cbush pls halp

source/get-started/glossary.txt
```diff
@@ -0,0 +1,135 @@
...
+      Allocation and deallocation of data within a :term:`Realm Database`
+      can lead to memory fragmentation. Calling the ``compact()`` method
+      reorganizes internal data structures to reduce the storage
+      overhead of your local Realm instance in the event of
+      fragmentation.
+
+   Realm Schema
+      Definition of all :term:`Realm Objects` that you can read or write
+      into a client :term:`Realm Instance` or :term:`MongoDB Atlas`
+      instance.
```
nlarew: "MongoDB Atlas instance" -> "linked MongoDB Atlas cluster"

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+.. glossary::
+   :sorted:
+
+   MongoDB Atlas
+      Cloud-hosted managed instances of MongoDB.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Asked jon about this, now have a blessed definition!

source/get-started/glossary.txt
```diff
@@ -0,0 +1,128 @@
...
+
+   MongoDB Realm
+      Cloud-hosted Backend-as-a-Service (BaaS) providing client-side
+      functionality via the Realm SDK and an administrative interface
+      via a website.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Sent this over to Drew, Christian, and Kenneth. We'll see what comes back from each of them.

source/get-started/glossary.txt
```diff
@@ -0,0 +1,234 @@
...
+
+   GraphQL
+      Open-source data query and manipulation language for APIs. Realm
+      GraphQL supports reading and writing data as well as subscribing
+      to data changes. An alternative to :term:`Realm Database` queries
+      using the client's native query language.
+
+   compaction
+      Allocation and deallocation of data within a :term:`Realm Database`
+      and/or exceeding the size of physical memory can lead to
```
cmelchior: This sentence reads weird to me? Could it instead be something like `Allocation of data within a Realm Database that exceeds the size of physical memory can lead to fragmentation on disk`

source/get-started/glossary.txt
```diff
@@ -0,0 +1,234 @@
...
+   external dependencies
+      An uploaded collection of npm modules that you can import into and
+      call from any :term:`Realm function`.
+
+   function context
+      A variable available in any :term:`Realm function` through which
+      you can access :term:`Realm app` components, including functions
+      and :term:`MongoDB Atlas` data.
+
+   value
```
cmelchior: This term is incredibly broad and used throughout the stack. I'm not sure it is beneficial to pin this definition to it.

source/get-started/glossary.txt
nathan-contino-mongo: Hm, my understanding is that (as with most memory allocation) a lot of object creation/deletion events can lead to fragmentation even when well below the size of physical memory. And exceeding the size of physical memory can cause fragmentation just through normal (not a lot of) object creation/no deletions necessary. Is my understanding incorrect? Happy to change it, just want to make sure we're giving the right warnings.

source/get-started/glossary.txt
nathan-contino-mongo: I'll change this to Realm value (and also Realm secret) because this is a great point.

source/get-started/glossary.txt
cmelchior: Fair point. I guess It was just the use of `and/or` that threw me off a little bit. But after re-reading it again, I think it makes sense to keep as is.

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+   system function
+      :term:`Realm function` that runs as the :term:`system user`.
+   
+   system user
+      A user that bypasses collection rules and schema validation for
+      the :term:`linked cluster` CRUD and Aggregation APIs.
+   
+   custom user data
+      Mapped :term:`MongoDB document` that stores arbitrary fields for
+      each Realm user, such as preferred language or timezone.
```
nlarew: - stores arbitrary fields -> contains data about
- each -> a given

Can we break into two sentences? The second one can give more contextual flavor to the data:

"For example, you may store a user's preferred language, local timezone, or an ID from an external service."

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+   
+   external dependencies
+      An uploaded collection of npm modules that you can import into and
+      call from any :term:`Realm function`.
```
nlarew: One or more JavaScript modules from `npm` that you have uploaded to your Realm application. You can import and use external dependencies from any Realm function.

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+
+   function context
+      A variable available in any :term:`Realm function` through which
+      you can access :term:`Realm app` components, including functions
+      and :term:`MongoDB Atlas` data.
```
nlarew: A _global_ variable that allows you to interface with other components of your application from a Realm function

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+
+   Realm value
+      A named reference to a piece of static data stored by
+      :term:`MongoDB Realm`, accessible through :term:`Realm functions`
+      and :term:`Realm rules`.
```
nlarew: A static _JSON_ value or Realm Secret defined in MongoDB Realm that you can access from Realm functions and rule expressions.

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+
+   Realm rules
+      True or false expressions that determine whether or not a user can
+      modify, insert, or read a given piece of data.
```
nlarew: - **Boolean** expressions
- Not sure about this since rules are more than just the expressions. The [JSON Expressions](https://docs.mongodb.com/stitch/services/json-expressions/) page sorta makes it seem that way but it's also generic for service rules where they really are more or less the whole thing.

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+
+   Realm secret
+      A private value stored in :term:`MongoDB Realm` that cannot be
+      directly read, but can be linked in authentication providers and
+      service configurations.
```
nlarew: You _can_ directly read it if you assign it to a value

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+
+   Realm UI
+      A front-end from which an administrator can modify your
+      :term:`MongoDB Realm` configuration with tasks including defining
+      :term:`functions <Realm function>`, creating :term:`triggers
+      <Realm trigger>`, linking :term:`MongoDB Atlas` clusters, and
+      more.
```
nlarew: - maybe s/front-end/web interface
- Not just modify your Realm configuration, it's the default interface to create/update/manage any Realm app.
- "administrator" feels off here. The UI is for app developers and anyone else that wants to modify the app or see logs.

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+
+   realm-cli
+      A command line interface from which an administrator can modify
+      your :term:`MongoDB Realm` configuration with tasks including
+      defining :term:`functions <Realm function>`, creating
+      :term:`triggers <Realm trigger>`, linking :term:`MongoDB Atlas`
+      clusters, and more.
```
nlarew: - `realm-cli` is the binary name, probably better to call this **Realm CLI** for parity with **Realm UI**
- Same comments vis-a-vis "Realm configuration" and "administrator" as above

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+
+   Realm Administrative API
+      An REST API through which you can perform a number of
+      administrative tasks.
```
nlarew: - Realm **Administration** API
- An -> A
- nit: Technically not a REST API since it violates HATEOAS and has at least one RPC endpoint.
  - Maybe just "An HTTP API..."?

source/get-started/glossary.txt
```diff
@@ -144,3 +144,91 @@ Glossary
...
+      :term:`Realm app`.
+
+   scheduled trigger
+      Activate a :term:`Realm function` based on a schedule that you
+      define.
+   
+   linked cluster
+      A :term:`MongoDB Atlas` database cluster that :term:`Realm clients
+      <Realm client>` can query based on roles and filters or
+      synchronize data to and from using :term:`Realm database`.
```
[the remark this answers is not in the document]
nathan-contino-mongo: I went back and forth on this. QueryAnywhere still exists and is used by some Stitch users today (who won't be going away) so I think it's still worth mentioning this ability. I agree that it's a little frustrating though.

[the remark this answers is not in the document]
nathan-contino-mongo: used your points above + a note about how you can access the data using realm database OR the mongodb atlas service. I think that's good for now -- as services evolve and our approach changes, we can reconsider this one!

source/get-started/glossary.txt
nathan-contino-mongo: reworded since this applies to functions as well

source/get-started/glossary.txt
nathan-contino-mongo: reworded to cover all of these points, thanks!

PR #145, opened 2020-03-24
(DOCSP-9479): Sync Rules

cbush:
- Adds "Define Sync Rules" page
- NOTE: Procedure section is very tentative and must be revisited

## Source material / further reading
- "Product Description: Rules for MongoDB Realm Sync"
- Realm Rules InVision https://jira.mongodb.org/browse/REALMC-4181
- https://docs.mongodb.com/stitch/mongodb/define-roles-and-permissions/

## JIRA
https://jira.mongodb.org/browse/DOCSP-9479

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/sync-rules/sync/rules.html


source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+
+In order to control who can sync, read, and write sections
+of data, MongoDB Realm allows you to configure **rules**.
+Rules enforce two essential data access principles:
+
+- Users can only read the objects that they are allowed to see.
+- Users can only make changes to the objects that they are allowed to write to.
+
+You can edit rules in the Rules UI in MongoDB Realm. You can
+also import rules using the Realm CLI.
```
nathan-contino-mongo: I would link to the UI/CLI here

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In order to control who can sync, read, and write sections
+of data, MongoDB Realm allows you to configure **rules**.
```
nathan-contino-mongo:
```suggestion
of data, :term:`MongoDB Realm` allows you to configure **rules**.
```

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+instead design the app to use section IDs as the partition
+keys. How you intend to grant access should inform how you
+partition your data.
+
+.. _sync-roles:
+
+Users and Sync Roles
+--------------------
+
+A **sync role** is a set of partition-level permissions that
```
nathan-contino-mongo: Consider adding sync role to the glossary and linking to it with a :term: here.

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+Realm evaluates when determining whether a user may sync a
+given partition. You select a property of the :ref:`user
+object <user-objects>` that determines whether
```
nathan-contino-mongo: I'm in so much suspense about what the end of this sentence is going to

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+<expressions>`. In a sync rules expression, access the
+partition key value with the ``%%partition``
+:ref:`expression expansion <expansions>`.
+
+.. example::
+
+   Suppose the store management application uses
+   :ref:`custom JWT authentication
+   <custom-jwt-authentication>`. In order to control who can
+   read and write store data, we provide the user object
```
nathan-contino-mongo: probably a good idea to link to the user object documentation

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+   with two custom properties:
```
nathan-contino-mongo: and ditto here for custom user data

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+   
+   - ``Store_List``: an array of IDs of the stores that the user can read from.
+   - ``Managed_Stores``: an array of IDs of the store that the user manages and can modify.
```
nathan-contino-mongo:
```suggestion
   - ``Managed_Stores``: an array of IDs of the stores that the user manages and can modify.
```

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+
+   By adding store IDs to these arrays on the user object,
+   we can:
+   
+   - Grant an employee read-only access to their own store.
+   - Grant a manager read/write access to their own store, and read-only access to other stores.
+   - Grant a regional manager read/write access to many stores.
+
+   The roles part of the realm configuration file would
+   looks like this:
```
nathan-contino-mongo:
```suggestion
   look like this:
```

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+   - ``Managed_Stores``: an array of IDs of the store that the user manages and can modify.
+
+   By adding store IDs to these arrays on the user object,
+   we can:
+   
+   - Grant an employee read-only access to their own store.
+   - Grant a manager read/write access to their own store, and read-only access to other stores.
+   - Grant a regional manager read/write access to many stores.
+
+   The roles part of the realm configuration file would
```
nathan-contino-mongo: link to the realm configuration file? I am not altogether sure about what this is.

source/sync/rules.txt
```diff
@@ -0,0 +1,137 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In order to control who can sync, read, and write sections
```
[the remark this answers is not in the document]
cbush: Good point. Removed "sections of".

source/sync/rules.txt
cbush: Not sure. I have just linked user objects and JWT. Don't want to overload.

nathan-contino-mongo: I think since we're advising folks to use custom user data here, it's worth pointing them to a resource to learn about how that works -- it's a pretty natural question to have, and if we don't link a lot of users might think that you can just add custom properties directly to the user object instead of using a custom user data document.

source/sync/rules.txt
```diff
@@ -0,0 +1,138 @@
...
+
+You can edit rules in the MongoDB :term:`Realm UI`. You can
+also import rules using the :term:`Realm CLI`.
+
+Rules and Partitions
+--------------------
+
+Rules apply to entire :ref:`partitions <partitions>`. This
+level of access control is suitable for most applications
+without compromising sync performance.
```
nathan-contino-mongo: Mentioning sync performance here and never returning to the subject feels like a bit of a tease. I'd say we have two options:
1) Discuss performance concerns in a later paragraph in this section, specifically some examples to demonstrate when rules *can* compromise sync performance.
2) Remove this line entirely -- I'm not sure if we need any more elaboration than "suitable for most applications". Personally I'd say that we should avoid the performance consideration if it only matters in niche cases -- instead, maybe we can talk about rules vs. roles and when you'd pick one versus the other. ("is suitable for most applications in which partitions correspond to access permissions", maybe?)

source/get-started/glossary.txt
```diff
@@ -39,6 +39,9 @@ Glossary
       individual Realms which passively synchronize changes as network
       availability allows.
 
+   partition
+      See :term:`Realm`.
```
nathan-contino-mongo: It's unfortunate that we have to have two terms that mean the same thing, but I definitely see how the partition value makes sense when you talk about partition keys. I do wonder if we're going to confuse users who want to know the difference between a Realm and a partition since right now they almost seem like separate concepts (client-side and server-side, respectively, further complicated by the fact that Atlas doesn't actually use partitions)

source/get-started/glossary.txt
```diff
@@ -265,3 +268,8 @@ Glossary
       <linked cluster>` without defining a strict, inviolable schema.
       Intended for fast interation during early stages of development
       before defining a schema.
+
+   Sync role
```
nathan-contino-mongo: Might be a good idea to spruce up the Realm rules entry as well since this PR details it so much.

source/sync/rules.txt
cbush: Can you give me a specific reference you have in mind?

source/sync/rules.txt
cbush: The idea here was to preempt criticisms like "But why can't I have per-field/document/etc. permissions?" I'll rework it.

source/get-started/glossary.txt
cbush: I don't see "Realm" as being "client-side" as opposed to server-side. Your definition for Realm applies just as well to my understanding of "partition". It's just that in some contexts, "partition" is the clearer term amidst other uses of "MongoDB Realm" and "Realm".

cbush: That said, I can feel "partition" is not a great word to rely on. I will switch it over to "Realm".

source/sync/rules.txt
nathan-contino-mongo: I'm guessing that the custom properties you've talked about are stored using https://docs.mongodb.com/stitch/users/configure-custom-user-data/ ? That's what I had in mind. If I'm misunderstanding let me know!

cbush: Yes, I _think_ that's what I mean. Added the link, we will revisit it when we have a concrete technical implementation.

conf.py
```diff
@@ -62,6 +62,7 @@
     '.. |atlas| replace:: Atlas',
     '.. |compass| replace:: MongoDB Compass',
     '.. |realm| replace:: realm',
+    '.. |realms| replace:: realms',
```
nathan-contino-mongo: I think these should both be uppercase, based on the conversations we had around the glossary.

cbush: These should be uppercase, but we still have many non-replaced "realm" (i.e. `realm` not `|realm|`) throughout the codebase. This would leave the codebase in an inconsistent state. I will create this ticket to change it over: https://jira.mongodb.org/browse/DOCSP-9636

nathan-contino-mongo: I ported over as many occurrences of "realm" as I could find in the recently-merged glossary PR.

cbush: Oh, okay. Didn't know that.

PR #153, opened 2020-03-30
(DOCSP-9476): Partitioning

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-9476

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-9476/sync/partitioning.html


source/get-started/partitioning.txt
```diff
@@ -0,0 +1,81 @@
...
+
+Limitations of Partition Keys
+-----------------------------
+
+Partition keys are subject to certain limitations:
+
+- You can only define partition keys using values of the String,
+  ObjectID, Int, or Long types.
+
+- Once you've chosed a partition key and enabled sync, you cannot change
```
cbush, line 68:
```suggestion
- Once you have chosen a partition key and enabled sync, you cannot change
```

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,81 @@
...
+
+- You cannot change the field name for your partition key.
+
+- Partition values should only change rarely. If you change a partition
+  value from Atlas, Realm interprets the change as a deletion of the
+  corresponding :term:`Realm object` from its original Realm and an
+  insertion of a new Realm object into its new Realm.
+  
+- Changing a partition value from the Realm SDK is an error and will
+  trigger a client reset.
```
cbush, line 81: Is there a relevant link for client reset?

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
...
+
+Partitioning helps you turn the large amount of data stored by your
+application into the small amount of data needed by individual clients,
+called :term:`Realms <Realm>`.
+
+Partition Keys
+--------------
+
+MongoDB Realm uses the :term:`partition key` to sort
+:term:`MongoDB documents <MongoDB document>` in Atlas into
```
nlarew, line 34:
```suggestion
:term:`MongoDB documents <MongoDB document>` in a synced cluster to
```

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
...
+like user settings or personal information.
+
+Partitioning helps you turn the large amount of data stored by your
+application into the small amount of data needed by individual clients,
+called :term:`Realms <Realm>`.
+
+Partition Keys
+--------------
+
+MongoDB Realm uses the :term:`partition key` to sort
```
nlarew, line 33:
```suggestion
MongoDB Realm uses the :term:`partition key` to map
```

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
...
+Realm app. After selecting a cluster to sync, choose your partition key.
+You can either use an existing field or create a new field.
+
+Limitations of Partition Keys
+-----------------------------
+
+Partition keys are subject to certain limitations:
+
+- You can only define partition keys using values of the String,
+  ObjectID, Int, or Long types.
```
nlarew, line 70:
```rst
Partition keys must be one of the following types:

- String
- ObjectID
- Int
```

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
...
+
+- Once you have chosen a partition key and enabled sync, you cannot
+  change the partition key to a different field. To use a different
+  field as your partition key, disable Realm sync and enable Realm sync
+  again, selecting the new field as your partition key.
```
nlarew, line 75: What's the scary impact of this? I assume universal client resets?

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
...
+
+- You cannot change the field name for your partition key.
```
nlarew, line 77: This seems like the same as the previous bullet. I like the direct phrasing though - can we combine?

```rst
You cannot change the partition key field name after sync has been enabled. To use a different field...
```

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
...
+
+- Partition values should only change rarely. If you change a partition
+  value from Atlas, Realm interprets the change as a deletion of the
+  corresponding :term:`Realm object` from its original Realm and an
+  insertion of a new Realm object into its new Realm.
```
nlarew, line 82: "Partition values should only change rarely" seems less like a limitation and more like guidance for working around a deeper limitation.

The rest of the bullet seems more like behavior than a limitation. What's limiting about deleting and inserting to move an object between Realms?

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
...
+  
+- Changing a partition value from the Realm SDK is an error and will
+  trigger a :ref:`client reset <client-reset>`.
```
nlarew, line 85: How does this happen? Can I just straight up change the value of a partition key on any object like any other field?
I'm also interested in why we both throw an error AND make you client reset.

- "is an error" -> "throws an error"

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
+============
+Partitioning
+============
```
nlarew, line 3: Let's use a verb phrase for this page title:
- Partition Data into Realms
- Map Documents to Realm Objects
- Define a Partition Key

nathan-contino-mongo: sounds good, i was thinking about that first choice already. I'm not sure where a stepfile would be appropriate, though -- as I mentioned on slack before, the "define a partition key" process is really just a subset of the steps to enable sync. I'd like to avoid repeating those steps in a separate guide because it's not something you can even do in isolation so it really truly is just "part of the sync enable process". Rest of these changes look solid though, I'll.

source/get-started/partitioning.txt
```diff
@@ -0,0 +1,85 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Your :term:`Realm app's <Realm app>` :term:`linked MongoDB Atlas cluster
+<linked cluster>` contains all of the data stored by your application:
+data for every client, every user, and every object you chose to persist
+in a database.
```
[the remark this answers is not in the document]
nathan-contino-mongo: - Moving away from the "every" repetition, good point.
- In this case I'm referring more to personal use data ("my todos") or user config data, like you'd store in custom user data. There are definitely some use cases that don't need to store this kind of data but I think as a general rule every user ends up having some data stored about them, even if it's just the data they've generated.
- You might want to take a look at my conversations with drew/ian in docs-realm -- TL;DR synced data != data in your linked cluster. Because any collection can just omit a realm schema and still exist in atlas while not synced.

source/get-started/partitioning.txt
nathan-contino-mongo: Going to use the phrase "linked Atlas cluster" for now for consistency above, and I'll introduce the new term "synced cluster" later on (and add it to the glossary)

source/get-started/partitioning.txt
nathan-contino-mongo: indeed

source/get-started/partitioning.txt
nathan-contino-mongo: I think you may have missed the difference between the two (my fault!) -- the first bullet point states that you can't point the partition key at a different field. The second bullet point states that you can't change the name of the field pointed to by the partition key. They're definitely similar points, but I want to make both clear because I can see question cropping up about either!

source/get-started/partitioning.txt
nathan-contino-mongo: Good point -- added the concrete limitation here, and leaving out the general guidance, which shouldn't have come first anyway.

source/get-started/partitioning.txt
nathan-contino-mongo: Honestly I don't think anyone knows yet. The client SDK *shouldn't* let you change it, but I'm not so sure we can just assume that so I put that here to be safe...

As for the "is an error and will trigger" that's just because a client reset can only be triggered by an error showing up in the realm error handler.

Basically I'm tempted to leave this here until we know more about the SDK. Going to open a ticket to verify the behaviour and update this bullet point once it's finalized.

nathan-contino-mongo: Created https://jira.mongodb.org/browse/DOCSP-9751 to revisit this.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,379 @@
...
+
+Partition Keys
+--------------
+
+MongoDB Realm uses the :term:`partition key` to map
+:term:`MongoDB documents <MongoDB document>` in a
+:term:`synced MongoDB Atlas cluster <synced cluster>` into
+:term:`Realm objects <Realm object>` in Realms. The partition key is a
+field that must exist in every document in every synced collection of
+your synced cluster. Each unique :term:`partition value`, or value of a
```
ianpward, line 47: > The partition key is a field that must exist in every document in every synced collection of your synced cluster. 
this is not entirely true, see your paragraph below

source/sync/partitioning.txt
```diff
@@ -0,0 +1,379 @@
...
+partition key, maps to a different Realm. Documents that share the same
+partition value belong to the same Realm and, as a result, share the
+same permissions.
+
+Choose a Partition Key
+----------------------
+
+The field selected as your partition key must exist in every document in
+the synced collections of your linked cluster. You should
```
ianpward, line 56: > The field selected as your partition key must exist in every document in the synced collections of your linked cluster.

not entirely true, suggest adding to the end "that you want to sync" or something similar

source/sync/partitioning.txt
nathan-contino-mongo: I tried to clarify this with the definitions of "synced collection" and "synced cluster" (which are different from a collection or a linked cluster), do you have any ideas about how to make this less ambiguous?

source/sync/partitioning.txt
```diff
@@ -0,0 +1,379 @@
...
+define your partition key in the Realm schema for each collection of
+your linked cluster. However, you don't have to include the partition
+key in all collections; any collections of documents that lack the
+partition key just won't be synced.
```
ianpward, line 60: > any collections of documents that lack the partition key just won't be synced.

suggest a little more specificity here. Also, collection is a special word in MDB. Any collection in MDB that is not defined in the JSON schema will not be synced. Any Collection that does not have a partitionKey field in the schema will not be synced. Any document or group of documents that does not have the partitionKey field defined or if the value is invalid will not be synced

nathan-contino-mongo: Good details here -- I've added the branches of handling for these specific scenarios. Thanks!

source/sync/partitioning.txt
```diff
@@ -0,0 +1,379 @@
...
+
+.. example::
+
+   - **CASE 1**: If your application stores data privately for each
+     individual user, you could use the ``owner_id`` field to partition
+     a database into Realms corresponding to individual users.
+     
+   - **CASE 2**: If your application shares resources across an entire
+     company, but not across separate companies, you could define and
+     partition your data by ``company_id``.
```
ianpward, line 87: this is a little confusing to me. Suggest: If your backend MDB cluster shares data across the entire company but teams within the company do not need to share data than you can use team_id to partition the data

nathan-contino-mongo: In this case I'm referring to an application that contains data for multiple companies (clients, I guess) -- I'm going to change this to teams within a company because that's a better example that's easier to follow!

source/sync/partitioning.txt
```diff
@@ -0,0 +1,379 @@
...
+  field as your partition key, disable Realm sync and enable Realm sync
+  again, selecting the new field as your partition key. Disabling and
+  re-enabling sync will trigger a :ref:`client reset <client-reset>` on
+  all clients.
+
+- You cannot change the field name for your partition key. To alter the
+  field name of your partition key, disable Realm sync, change the field
+  name, and enable Realm sync again, selecting the updated field name as
+  your partition key. Disabling and re-enabling sync will trigger a
+  :ref:`client reset <client-reset>` on all clients.
```
ianpward, line 125: this paragraph is the same as the paragraph above

nathan-contino-mongo: Not quite! The first paragraph states that you can't change which field the partition key uses, and the second paragraph states that you can't change the name of the partition key field itself. Any suggestions on how to make this less confusing? I tried combining the two but I think it's tricky to do that without losing one of those two points, and both are pretty important.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+   - **CASE 2**: If your application shares resources across teams, but
+     not across your entire company, you could partition your
+     data using a ``team_id``.
+
+Define a Partition Key
+----------------------
+
+You can define a partition key when enabling :term:`Realm sync` for your
+Realm app. After selecting a cluster to sync, choose your partition key.
+You can either use an existing field or create a new field.
```
nlarew, line 95: This section is super small and feels buried. Can we expand it, make it an admonition/CTA that points to the full procedure, or remove it entirely?

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
+.. _partitioning:
+
+================================
+Partition Atlas Data into Realms
+================================
```
nlarew, line 5: So the verb phrase that I suggested was for a procedural page, but I see what you mean re: this being only part of the procedure for defining a schema. This page seems to primarily be about **Partition Keys** - what if we make that the title? It's more reference-y which IMO fits the vibe of this page.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+information, are only relevant to a single user.
+
+:term:`Partitioning <partitioning>` helps you turn the large amount of
+data stored by your application into the small amount of data needed by
+individual clients, called :term:`Realms <Realm>`.
+
+.. figure:: /images/partitioning.svg
+
+   Partition shape data in an Atlas cluster using a ``color``
+   partition key with partition values ``red``, ``blue``, and ``green``.
```
nlarew, line 35: I like the underlying idea here but we should definitely get a design pass.

Also I think we should replace "Realm Sync" with something more directly about the partition key and its role in mapping from MongoDB collections to Realms.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+same permissions.
+
+Choose a Partition Key
+----------------------
+
+The field selected as your partition key should exist in every document
+in the synced collections of your synced cluster. You should
+define your partition key in the Realm schema for each collection of
+your linked cluster. However, you don't have to include the partition
+key in all collections; Real sync will ignore any collections that lack
```
nlarew, line 59:
```suggestion
key in all collections; Realm sync will ignore any collections that lack
```

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+a Realm schema as well as any documents that lack a valid partition
+value.
+
+All documents that share the same partition key also share the same
+permissions, so you should select a key whose unique values correspond
+with permissions. Your partition key depends entirely on your use case:
```
nlarew, line 65: nit: can we add some context to what it means to "correspond with permissions"?

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+:term:`synced MongoDB Atlas cluster <synced cluster>` into
+:term:`Realm objects <Realm object>` in Realms. The partition key is a
+field that must exist in every document in every synced collection of
+your synced cluster. Each unique :term:`partition value`, or value of a
+partition key, maps to a different Realm. Documents that share the same
+partition value belong to the same Realm and, as a result, share the
+same permissions.
+
+Choose a Partition Key
+----------------------
```
nlarew, line 53: Does this explain how to choose the partition key or the value of a partition key in each document? Right now it reads like both. Can we make that more apparent?

Also, choosing the _value_ of a partition key for a given document seems more fundamentally like, e.g. **Map Documents to Realms**. There's probably a whole bunch of patterns we could document, enough that this deserves its own section if nots its own page.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+value.
+
+All documents that share the same partition key also share the same
+permissions, so you should select a key whose unique values correspond
+with permissions. Your partition key depends entirely on your use case:
+some applications only require a per-user permission scheme, but many
+require permissions based on collections of users.
+
+In addition to realms
+belonging to individual users or collections of users, it's often
```
nlarew, line 70: It's not entirely clear what it means for a Realm to "belong" to a user

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+a Realm schema as well as any documents that lack a valid partition
+value.
+
+All documents that share the same partition key also share the same
+permissions, so you should select a key whose unique values correspond
+with permissions. Your partition key depends entirely on your use case:
+some applications only require a per-user permission scheme, but many
+require permissions based on collections of users.
+
+In addition to realms
```
nlarew, line 69:
```suggestion
In addition to Realms
```

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+  same data into the object's new Realm. Any client updates to the
+  object that have not yet synced to Atlas will be lost.
+  
+- Changing a partition value from the Realm SDK throws an error and will
+  trigger a :ref:`client reset <client-reset>`.
+
+.. _partition-key-examples:
+
+Partition Key Examples
+----------------------
```
nlarew, line 140: I like these examples!

Some thoughts:

- These aren't just examples of partition keys, they're really examples of different **partition strategies**.
- There are two separate examples here - can we give each one a heading so they show up in the contents box and have nice, addressable names?

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+
+Because Realms correspond to collections of documents that share the
+same permissions, partitioning can vary between Realm apps.
+
+.. example::
```
nlarew, line 145: Title e.g. **Partition Data for a Single User**

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+      Realm OP: [
+         { owner_id: "OP", playlist_ids: [ ... ] }
+      ]
+
+      Realm PUBLIC: [
+         { name: "Disco Anthems", owner_id: "PUBLIC", song_ids: [ ... ] }
+         { name: "Deep Focus", owner_id: "PUBLIC", song_ids: [ ... ] }
+      ]
+
+.. example::
```
nlarew, line 227: Title e.g. **Partition Data for a Group of Users**

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+
+   Consider the following documents in MongoDB collections
+   describing inventory and sales at a paper company:
+
+   .. code-block:: javascript
+
+      collection branches: [
```
nlarew, line 234: The more I look at this `collection` keyword the more I like it.  We don't really use it elsewhere though. I think maybe the manual alternative would be `db.branches` or even `db.branches.find()` with results below.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+   - Because most users only read and write to one or two Realms, this
+     partitioning scheme only exposes data to those who need it. The
+     partitioning remains flexible enough to accommodate future
+     features.
+
+   
+   Partitioning this data by ``partitionKey`` results in the following
+   Realms:
+
+   .. code-block:: javascript
```
nlarew, line 324: It'd be really cool if we had a component like the [code blocks on this site](https://wattenberger.com/blog/react-hooks) that show how one set maps to another.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,381 @@
...
+
+.. example::
+
+   - **CASE 1**: If your application stores data privately for each
+     individual user, you could use the ``owner_id`` field to partition
+     a database into Realms corresponding to individual users.
+     
+   - **CASE 2**: If your application shares resources across teams, but
+     not across your entire company, you could partition your
+     data using a ``team_id``.
```
nlarew, line 88: It's not super clear that this is connected to the two examples later on the page. On their own these don't feel like fully satisfying examples. What is the context of "case"?

source/sync/partitioning.txt
nathan-contino-mongo: i'd like to point to a full procedure but the "enable sync" guide doesn't exist yet afaik. I'd like to add a "to learn more about, see <>" but that's tricky without a page.

source/sync/partitioning.txt
nathan-contino-mongo: I disagree, I think this is *the* page to explain the concept of partitioning data, and partition keys are just a part of that concept. Titling this page "partition keys" leads to the following questions:
- what is a partition?
- where is the page that explains what partitioning does? Is this it?
- why does it take multiple sections to get to the concept of partition keys?

source/sync/partitioning.txt
nathan-contino-mongo: Reached out to Deep about this since Gilberto is OOO this week. Interested in your thoughts on how to reword "realm sync", since that concept is to me the glue that holds together atlas and realms.

I do agree that it would be nice to put something about partition keys/partition values in here. I'll see if I can add something about that into the corner.

source/sync/partitioning.txt
nathan-contino-mongo: elaborated

source/sync/partitioning.txt
nathan-contino-mongo: Good point, I've broken this section into two sections (choose a partition key & map documents to realms) to focus on keys and values, respectively. I also cleared up some bits of sentences where I used "key" previously but really meant "value".

source/sync/partitioning.txt
nathan-contino-mongo: Changed to ownership, a concept I elaborate on further

source/sync/partitioning.txt
nathan-contino-mongo: My biggest concern is that this makes matters more difficult for users who aren't already familiar with mongodb. While I definitely prefer the idea of executable code, I lean more toward the "approachable" than the "technically correct". :smile:

source/sync/partitioning.txt
nathan-contino-mongo: Honestly this would be a really useful visual aid for users trying to learn about Realms, I'd be open to hacking on it a bit and seeing if we could make a two-column tool...

source/sync/partitioning.txt
nathan-contino-mongo: These are just shorthand examples that are actually disconnected from the second pair of examples -- I think I'm just going to break them up into two separate example blocks instead to remove the awkward "case" wording.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,431 @@
...
+
+Your partition key depends entirely on your use case: some applications
+only require a per-user permission scheme, but many require permissions
+based on collections of users.
+
+You should define your partition key in the Realm schema for each
+collection of your linked cluster. However, you don't have to include
+the partition key in all collections; Realm sync will ignore any
+collections that lack a Realm schema as well as any documents that lack
+a valid partition value.
```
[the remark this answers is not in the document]
nathan-contino-mongo: I think it makes more sense in "choose" since your partition key might be a field that's already in your schema -- really, I should have used the word "declare" here. Rephrased to just mention that schemas should include it.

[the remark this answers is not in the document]
nathan-contino-mongo: I've also moved the paragraph into the define section because I think it does actually make more sense there!

source/sync/partitioning.txt
```diff
@@ -0,0 +1,431 @@
...
+   data using a ``team_id``.
+
+However, there are many other ways to partition your data.
+
+It's often helpful to store read-only data viewable to all users; for
+this kind of use case, you can create a special partition value, such as
+``PUBLIC``, that all users sync with read permissions only. You can use
+this same idea to create a private Realm, only viewable by
+administrators, with a different special partition value, such as
+``PRIVATE``.
```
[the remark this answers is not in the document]
nathan-contino-mongo: This is brilliant. I'm still thinking about what we can do with the ``PRIVATE`` example -- I think it could be a useful pattern (for instance, place data in the ``PRIVATE`` Realm before you change
it to ``PUBLIC`` to create a kind of staging area) but it might be too much for the initial pass.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,431 @@
...
+  
+  1) disable Realm sync
+  
+  2) enable Realm sync again with the new field selected as your
+     partition key.
+     
+  .. warning::
+
+     Disabling and re-enabling sync will trigger a
+     :ref:`client reset <client-reset>` on all clients.
```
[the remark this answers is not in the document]
nathan-contino-mongo: well disabling sync would actually just kill sync entirely, so I don't think anyone really knows what would happen right now (probably just a connection failure?). You'd definitely end up with a client reset error if you tried to connect to the realm while using an older partition key.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,431 @@
...
+  
+  2) change the field name
+
+  3) enable Realm sync again with the updated field name selected as
+     your partition key.
+
+  .. warning::
+
+     Disabling and re-enabling sync will trigger a
+     :ref:`client reset <client-reset>` on all clients.
```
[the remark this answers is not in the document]
nathan-contino-mongo: the difference is

- "changing the partition key from ``user_id`` to ``team_id`` because you thought you needed individual partitions only" (consider a data model where these fields already both exist for other reasons)

vs.

- "fixing the typo you made when you named the field ``onwer_id`` by changing it to ``owner_id`` (or deciding you prefer the name ``owner_id`` to ``user_id``)

You are right that the action is the same: you disable sync, fix your schema, and then re-enable sync. I mostly want to be comprehensive so that users don't have to wonder about their particular case, since I'd bet money that both of these cases will happen for many users as they play around with schemas.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,431 @@
...
+  Realm. Next, MongoDB Realm inserts a new Realm object containing the
+  same data into the object's new Realm. 
+  
+  .. warning::
+
+     Changing a partition value in Atlas will lose any un-synced client
+     updates to the object.
+
+- Changing a partition value in a Realm object using the Realm SDK
+  will throw an error and trigger a :ref:`client reset <client-reset>`.
```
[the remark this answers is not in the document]
nathan-contino-mongo: I could leave it out, I'm just not 100% sure that the SDK is fully-baked enough to prevent this in all cases (maybe some weird query could manage it, like incrementing an int partition key?)

However, it's kind of covered by my new guidance to choosing a partition key, so i could remove it.

source/sync/partitioning.txt
```diff
@@ -0,0 +1,464 @@
...
+++++++++++++++++++++++
+
+Your partition key depends entirely on your use case: some applications
+only require a per-user permission scheme, but many require permissions
+based on collections of users. Generally, a few guidelines exist to
+choose partition keys:
+
+- Realm clients should never modify the partition value directly. Any
+  field that clients can modify cannot be used as a partition
+  key.
```
nlarew, line 63: Nice.

One nit: "cannot" or "should not"? Does Realm do anything actively to safeguard against this?

source/sync/partitioning.txt
```diff
@@ -0,0 +1,464 @@
...
+
+- Changing a partition value is an expensive operation, requiring a
+  deletion of the object from one Realm and an insertion of the same
+  object into another Realm. As a result, you should choose a partition
+  key that rarely, if ever, needs to change.
+
+- Partition keys use the same field name in every synced document. Pick
+  a field name that is unlikely to collide with an another field name
+  in any Realm object's schema. You can begin an underscore, (e.g.
+  ``_partitionKey`` to decrease the likelihood of collisions) and
```
nlarew, line 73:
```suggestion
  ``_partitionKey``) to decrease the likelihood of collisions and
```

source/sync/partitioning.txt
```diff
@@ -0,0 +1,464 @@
...
+
+Define a Partition Key
+++++++++++++++++++++++
+
+You can define a partition key when enabling :term:`Realm sync` for your
+Realm app. After selecting a cluster to sync, choose your partition key.
+You can either use an existing field or create a new field.
+
+The :term:`Realm schema` for each synced collection of your synced
+cluster should include the partition key for your application. However,
```
nlarew, line 92: - should include or must include?
- still not sure about "synced collection of your synced cluster" but maybe not a big deal

source/sync/partitioning.txt
```diff
@@ -0,0 +1,464 @@
...
+you don't have to include the partition key in all collections; Realm
+sync will ignore any collections that lack a Realm schema as well as any
+documents that lack a valid partition value.
```
nlarew, line 95: We haven't introduced partition values yet. Is this the right spot to mention them?

IMO this subsection (**Define a Partition Key**) feels more like a step out of the procedural guide that could link to the rest of this page

source/sync/partitioning.txt
```diff
@@ -0,0 +1,464 @@
...
+++++++++++++++++++++
+
+Most applications partition data into Realms based on individual users
+or groups of users:
+
+- **Individual Realms**: If your application stores data privately for each
+  individual user, you could create an ``owner_id`` field containing a
+  specific user ID on every document. Selecting the ``owner_id``
+  field as your partition key would partition your database into Realms
+  corresponding to individual users.
```
nlarew, line 128: This makes it seem sort of like a "pick one" scenario.

> If your application stores data privately...

- In reality many apps will use all of these partition strategies in some combination.

> would partition your database into Realms corresponding to individual users"

- "partition your database" isn't really the right idea here I think. It's more "assign documents to Realms".

Maybe: "For private application data that belongs to a specific user, e.g. their order history, you could..."

source/sync/partitioning.txt
```diff
@@ -0,0 +1,464 @@
...
+     
+- **Team Realms**: If your application shares resources across teams, but
+  not across your entire company, you could create a ``teams``
+  collection that maps user IDs to team IDs. You could then create a
+  ``team_id`` field containing a specific team ID on every document you
+  want to sync. Selecting the ``team_id`` field as your partition key
+  would partition your database into Realms containing data for
+  entire teams.
```
nlarew, line 136: Same comment as above + what if my app's domain isn't a company?

Maybe: "For application data that some group of users share, you could..."

source/sync/partitioning.txt
nlarew: An `owner_id` field sounds like a partition key field name but we're in the **Partition Values** section. If you used this strategy to name your partition key field then you'd be locked into only using a single partition strategy. This is example 1 below.

Really most apps will need flexibility so IMO we should prefer the more abstract form in example 2. Using a query string convention for complex partition keys could be cool (and matches with the query string convention of NSPredicate). Maybe we should add some docs on that?

source/sync/partitioning.txt
```diff
@@ -0,0 +1,464 @@
...
+
+- **Public Realms**: It's often helpful to store read-only data viewable to all users; for
+  this kind of use case, you can define a special partition value in
+  your permissions, such as ``PUBLIC``, that all users sync with read
+  permissions only.
+  
+- **Hidden Realms**: To create a Realm that is only viewable by administrators, you can
+  define a different special partition value, such as ``PRIVATE``. You
+  can use this strategy to verify new content for correctness before
+  moving it into an individual, team, or public Realm.
```
nlarew, line 146: How is this different than a public Realm? Does Realm do some permissions magic if the value is `PRIVATE`? I think we can be more clear and say that the developer needs to configure the permissions.

Also why not call it an "Admin Realm" and set the value to `ADMIN`? Seems closer to the intent here.

source/sync/partitioning.txt
nathan-contino-mongo: Unfortunately that's an open question. The way I'm phrasing it here leaves that open to interpretation but is at least conclusive that they *shouldn't* change.

source/sync/partitioning.txt
nathan-contino-mongo: - unfortunately it sounds like it's optional
- my effort to be specific, unfortunately it is distinct from "collection of your synced cluster"

source/sync/partitioning.txt
nathan-contino-mongo: Fair point, i'll change the wording to avoid it for now. As for the procedural bit... we'll revist that when we create the *Enable Sync* guide :wink:

source/sync/partitioning.txt
nathan-contino-mongo: Fair fair. I like the idea of creating a Realm for which no users have permission so that you can stage content and just alter the Realm assignment to release it to a team, user, or public realm, but I think you make a good point -- it's really no different from an admin team realm. I'll leave it out to make things simpler.

PR #154, opened 2020-03-31
(DOCSP-9478): Add mention to Realm Data Models in Document Schema Types Doc

mongomoe:
JIRA:
-----
https://jira.mongodb.org/browse/DOCSP-9478

Original Stitch Doc:
-------------------
https://docs.mongodb.com/stitch/mongodb/document-schemas/

Stage:
------
https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-9478-Schema-Types/mongodb/document-schemas.html

### DIFF
--- a/source/mongodb/document-schemas.txt
+++ b/source/mongodb/document-schemas.txt
@@ -38,10 +38,11 @@ schemas for the following common types:
 
 .. note::
 
-   MongoDB Realm can be used to automatically create :doc:`GraphQL
-   </graphql>` types or :doc:`Realm Data Models
-   </get-started/define-realm-schema>` for collections that have a
-   defined document schema.
+   You can automatically create GraphQL types or a Realm Data Model for
+   collections that have a defined document schema using the MongoDB
+   Realm UI. See the :doc:`GraphQL </graphql>` doc or the :doc:`Realm
+   Data Models </get-started/define-realm-schema>` doc to learn how.
+
 
 
 Schema Enforcement Process
--- a/source/mongodb/document-schemas.txt
+++ b/source/mongodb/document-schemas.txt
@@ -430,3 +430,13 @@ For more information, see the official JSON Schema `boolean guide
 .. code-block:: json
 
    { "bsonType": "bool" }
+
+
+
+
+.. note::
+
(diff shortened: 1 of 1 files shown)


source/mongodb/document-schemas.txt
```diff
@@ -36,6 +36,14 @@ schemas for the following common types:
 - :ref:`Numbers <schema-type-numbers>`
 - :ref:`Booleans <schema-type-booleans>`
 
+.. note::
+
+   MongoDB Realm can be used to automatically create :doc:`GraphQL
```
ccho-mongodb: Nitpick: could you omit "be used to" or is that essential to the explanation? Also, how does a user initiate the automatic creation?

mongomoe: "Be used to" is arguably essential because although the process is considered 'automatic', you do have to click some buttons in the UI to generate the data model

mongomoe: I chose not to mention how to initiate the automatic creation because that will be in the linked docs, though perhaps this should be more clear. What do you think about the phrasing: 

```You can create GraphQL types or Realm Data Models in the MongoDB Realm UI. See the `GraphQL</link>` doc or the `Define a Realm Schema</link>` doc to learn how.```

Comment (cbush) [inline source/mongodb/document-schemas.txt]: Avoid passive voice: who is doing the using?

```suggestion
   You can use MongoDB Realm to automatically create :doc:`GraphQL
```
or

```suggestion
   You can automatically create :doc:`GraphQL
```
or
```suggestion
   MongoDB Realm can automatically create :doc:`GraphQL
```

ccho-mongodb: That's more clear. Before, I wasn't sure whether the links were just explanations of what GraphQL types and Realm data models were or if those pages contained instructions on how to trigger the automatic generation.

mongomoe: Thanks and yup realm schema page will have instructions

mongomoe: resolving since the sentence has been rephrased - no passive voice in the new rendition.

PR #165, opened 2020-04-10
DOCSP-9636: Term substitutions across repo (Part 1 of N)

nathan-contino-mongo:
staging: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/cleanup/get-started/glossary.html

JIRA: https://jira.mongodb.org/browse/DOCSP-9636


conf.py
```diff
@@ -63,9 +64,56 @@
...
+    'instance-short': 'app',
+    'instances-short': 'apps',
+    'client-database': 'Realm Database',
+    'schema': 'Realm Data Model',
+    'backend-schema': 'Realm Schema',
+    'frontend-schema': 'Realm Object Model',
+    'cli': '``realm-cli``',
+    'ui': 'Realm UI',
+    'function': 'Realm function',
+    'functions': 'Realm functions',
```
cbush, line 98: Same for "T"riggers:
```suggestion
    'functions': 'Realm Functions',
```

nathan-contino-mongo: added!

nathan-contino-mongo:
closing at @nlarew 's suggestion

cbush:
Certainly we need "Realm" -> "realm". Where there's ambiguity (and therefore no easy way to search/replace), I'm into the source constants!

nathan-contino-mongo:
I would agree that a source constant for the client database is necessary (because our conventions could easily change in the future). I would also argue that triggers and functions could very easily change names in the future (or just capitalization conventions) so why not use source constants if we can? It's not like replacing Triggers with {+triggers-short+} is really that much of a pain -- you can always draft pages without the source constants and add them once you've figured out what you want to say/as you fill in refs and links anyway.

source/data-model/migrations.txt
```diff
@@ -94,33 +94,33 @@ Realm handles synced migrations using the following framework:
...
        
-       - Realm sets the value of required deleted properties to the
+       - {+service+} sets the value of required deleted properties to the
          empty value of the property type (e.g. ``0`` for numeric
          fields, ``""`` for string properties, etc.).
 
    * - **Modifications**
-     - Realm prevents modifications to an existing property other than
-       delete operations. Modifications that synced Realms do not
+     - {+service+] prevents modifications to an existing property other
```
cbush, line 122:
```suggestion
     - {+service+} prevents modifications to an existing property other
```

PR #176, opened 2020-04-21
finalizing source_constants substitutions

nathan-contino-mongo:
As we discussed in our team meeting today, trimmed some source constants that we don't need and added one that we decided we could take advantage of.

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/subsFinal/index.html


source/get-started/what-is-realm.txt
```diff
@@ -59,7 +59,7 @@ Realm Product Suite
 {+backend+}:
   - Enables developers to quickly build applications without having to set up server infrastructure.
   - Accelerates app development by providing out-of-the-box authentication, user management, and permission management.
-  - Supports serverless application architectures with :term:`{+triggers+} <{+trigger+}>` and :term:`{+functions+} <{+function+}>`.
+  - Supports serverless application architectures with :term:`{+service-short+} Triggers <{+service-short+} Trigger>` and :term:`{+service-short+} Functions <{+service-short+} Function>`.
```
cbush, line 62: Yikes. I wonder, do we change the term to Trigger and Function respectively, and stop with the service-short on the glossary?

nathan-contino-mongo: I like this. Going to update the term links and the glossary definitions.

nathan-contino-mongo: Actually... this is consistent with some of the other definitions and is really the proper name, so I'm going to leave it for now. It is definitely a mouthful though, so I'm happy to revisit it in the future if it becomes a pain!

PR #193, opened 2020-05-08
clarify limitations

terakilobyte:
In response to several tickets about inconsistent support for ops.

[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-10218/)

### DIFF
--- a/source/mongodb/crud-and-aggregation-apis.txt
+++ b/source/mongodb/crud-and-aggregation-apis.txt
@@ -496,7 +496,8 @@ Aggregation
 stages </reference/operator/aggregation-pipeline/>` when you :doc:`run
 an aggregation pipeline </mongodb/run-aggregation-pipelines>` in the
 context of an :ref:`application user <user-accounts>`. All aggregation
-pipeline stages are available to the system user.
+pipeline stages are available to the system user except ``$indexStats``
+and ``$currentOp``.
 
 Aggregation Pipeline Stage Availability
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -516,7 +517,7 @@ Aggregation Pipeline Stage Availability
 
    * - :manual:`$currentOp </reference/operator/aggregation/currentOp/>`
      - No
-     - Yes
+     - No
 
    * - :manual:`$lookup </reference/operator/aggregation/lookup/>`
      - No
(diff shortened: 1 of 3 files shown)


source/mongodb/service-limitations.txt
```diff
@@ -52,15 +53,17 @@ MongoDB through {+service-short+}, see :ref:`Query Options
 Aggregation
 -----------
 
-{+service-short+} supports all :manual:`aggregation pipeline
+{+service-short+} supports nearly all :manual:`aggregation pipeline
```
nathan-contino-mongo, line 56: As a user, I think that "nearly all" ... "except" is a confusing construction. Are there pipeline stages other than $currentOp and $indexStats that Realm does not support? If so: why aren't they listed here? If not: can we trim "nearly" from the sentence since "except" covers the two unsupported stages?

source/mongodb/service-limitations.txt
```diff
@@ -52,15 +53,17 @@ MongoDB through {+service-short+}, see :ref:`Query Options
...
-some aggregation pipeline stages when you :doc:`run
-an aggregation pipeline </mongodb/run-aggregation-pipelines>` in a
-:ref:`user function <function-active-user>`. For a list of
-pipeline stages that are available when you connect to MongoDB through
-{+service-short+}, see :ref:`Aggregation
-<mongodb-crud-and-aggregation-apis-aggregation>`.
+:ref:`system functions <function-system-user>` except for ``$currentOp``
+and ``$indexStats``.
+
+In :ref:`user functions<function-active-user>`, supported
```
nathan-contino-mongo, line 61: Missing space:

```suggestion
In :ref:`user functions <function-active-user>`, supported
```

source/mongodb/crud-and-aggregation-apis.txt
```diff
@@ -44,6 +44,10 @@ MongoDB feature availability in {+service+}, see
 CRUD Operations
 ---------------
 
+.. warning::
+
+   {+service+} does not support the ``distinct()`` method.
```
nathan-contino-mongo, line 49: Personally I feel like your addition to the query options section in Limitations is enough to let users know that the option isn't available in Realm. At the very least, can we make this an admonition instead of a warning?

PR #216, opened 2020-06-02
(DOCSP-10651): Call a Function - Node.js

mongomoe:
JIRA:
-----
https://jira.mongodb.org/browse/DOCSP-10651

Stage:
------
https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-10651-Functions-Node/node/call-a-function-from-node.html


source/node.txt
```diff
@@ -62,7 +62,7 @@ Webhooks, {+sync+}, and built-in third-party authentication.
 
 To learn how to connect to a {+backend+} app to call
 Functions, query data in an instance of {+atlas+}, and synchronize data
-in {+realms+}, see :doc:`Call a Function </functions/call-a-function>`.
+in {+realms+}, see :ref:`Call a Function <node-client-call-function>`.
```
nathan-contino-mongo, line 65: nice catch, i can finally throw away my note to fix this now!

source/node/call-a-function-from-node.txt
```diff
@@ -0,0 +1,110 @@
...
+   :class: singlecol
+
+Overview
+--------
+
+:doc:`MongoDB Realm Functions </functions>` allow you to run serverless
+snippets of code. Functions are useful for abstracting away
+implementation details from client applications by doing things such as
+reusing code-snippets to process data or dynamically performing actions
+based on the current user, processing data, and more. User You can call
```
nathan-contino-mongo, line 22: Can we convert this to a bulleted list with consistent phrase structure? I think these are all great examples but it feels a bit run-on to me.

source/node/call-a-function-from-node.txt
```diff
@@ -0,0 +1,110 @@
...
+
+   // sum: adds two numbers
+   exports = function(a, b) {
+     return a + b;
+   };
+
+Call the Function From a Node.js Application
+--------------------------------------------
+
+There are two alternate methods for executing a Function from the
```
nathan-contino-mongo, line 45: I would suggest using lowercase "function" for consistency across the page. Just using uppercase in one place stands out to me.

source/node/call-a-function-from-node.txt
```diff
@@ -0,0 +1,110 @@
...
+
+Pass the function parameters directly to the first method,
+``functions.<functionName>()``, which you call on the user instance. For
+example, running the script: "myUserInstance.functions.sum(2,3)" would
+return the result of 5. 
+
+The alternative method, ``callFunction()`` calls the MongoDB Realm
+Function on the user object with the first parameter being the Function
+name and the second parameter being an array of arguments that you pass
+to the Function.
```
nathan-contino-mongo, line 56: Numbering these 1. and 2. would probably make this easier to read. Or at least bullet them to make it clear that they're separate ideas, not subsequent paragraphs.

source/node/call-a-function-from-node.txt
```diff
@@ -0,0 +1,110 @@
...
+   const Realm = require("realm");
+
+   const appId = "<appId>"; // Set Realm app ID here.
+   const appConfig = {
+     id: appId,
+     url: "https://realm-dev.mongodb.com",
+     timeout: 1000,
+     app: {
+       name: "default",
+       version: "0",
```
nathan-contino-mongo, line 78: comment explanations for "default" and "0" might be helpful here -- this seems really obtuse to me! I am unfamiliar with the Node SDK though.

source/node/call-a-function-from-node.txt
```diff
@@ -0,0 +1,110 @@
...
+:doc:`Define a Function </functions/define-a-function>` doc. 
+
+.. code-block:: javascript
+
+   // sum: adds two numbers
+   exports = function(a, b) {
+     return a + b;
+   };
+
+Call the Function From a Node.js Application
```
cbush, line 42:
```suggestion
Call the Function from a Node.js Application
```

source/node/call-a-function-from-node.txt
```diff
@@ -0,0 +1,110 @@
...
+
+.. admonition:: Link a MongoDB Atlas Cluster
+   :class: note
+   
+   This example requires a MongoDB Realm Application with a :doc:`linked
+   Atlas cluster </mongodb/link-a-cluster>`. Replace the "<appId>" in the
+   code block with your app id, available in the lefthand corner of the
+   MongoDB Realm UI.
+
+.. code-block:: javascript
```
cbush, line 67: It would be great to add :emphasize-lines: and highlight the function call lines.

source/node.txt
mongomoe: :D

source/node/call-a-function-from-node.txt
mongomoe: The reason I'm against bullets in this instance is that these are examples of use-cases of MongoDB Realm Functions. There are basically infinite number of additional alternative use-cases for server functions.

source/node/call-a-function-from-node.txt
mongomoe: fixed

source/node/call-a-function-from-node.txt
mongomoe: removed - these were unneeded here

source/node/call-a-function-from-node.txt
mongomoe: I've replaced this with a table instead - I think that achieves the same purpose

source/node/call-a-function-from-node.txt
mongomoe: Fixed - I've replaced all shorthands of the word "function" to MongoDB Realm Function. The only place I use "function" now is when referring to actual methods, rather than the feature. In practice, it's a little ugly and sales-esque to use the full feature name "MongoDB Realm Function"  but it's better than confusing the users.

PR #218, opened 2020-06-04
(DOCSP-10665): Added Node Authentication Client Guide Page

mongomoe:
jira: https://jira.mongodb.org/browse/DOCSP-10665
stage: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker/DOCSP-10665-Node-Auth/node/authentication.html


source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+   :class: singlecol
+
+Overview
+--------
+
+.. _node-login:
+
+MongoDB Realm provides developers with an API to log application user's
+in and out using the same methods, regardless of the authentication
+provider the user was registered using.
```
nlarew, line 22: nit: "the same methods" breaks down a little when we add in OAuth (which uses `logInWithRedirect()`)

also can we mention credentials and split this into a couple of sentences?

- The Node SDK allows users to log in with any enabled authentication provider
- The SDK uses a ``Credentials`` object to specify which provider to use
- The ``Credentials`` object also includes info required to authenticate with the provider (like an email address and password)

source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+.. list-table::
+   :header-rows: 1
+   :widths: 10 90
+
+   * - Method 
+     - Usage
+   * - ``<Realm.App instance>.logIn()``
+     - Pass your credentials directly to the method, ``logIn()``, which you call on the Realm App instance. For example, running the script: "app.logIn(Realm.Credentials.emailPassword('<email>', '<password>'))" would log in a user created with an email/password authentication provider.
+   * - ``<userInstance>.logOut()``
+     - The method ``logOut()`` is called on the user instance, and will log any user out, regardless of the authentication provider the user was registered using.
```
nlarew, line 33: I think we can safely refer to the type/class names here instead of using angle bracket syntax. Should be a bit cleaner and closer to the final usage pattern. i.e.

- `App.logIn()`
- `User.logOut()`

source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+Overview
+--------
+
+.. _node-login:
+
+MongoDB Realm provides developers with an API to log application user's
+in and out using the same methods, regardless of the authentication
+provider the user was registered using.
+
+.. list-table::
```
nlarew, line 24: Can we introduce this table? What does it describe?

e.g.
```rst
The SDK provides the following methods for user authentication:

.. list-table::
   
   ...
```

source/node/authentication.txt
nlarew: - The descriptions here are a bit inconsistent ("Pass your credentials..." vs "The method is called...").
- "The method ... is called" is passive voice (called by _whom?_)
- Given that this is a quick overview table, I think we prefer shorter, snappier descriptions to more prescriptive ones.

What about something like this?

```rst
Call ``App.logIn()`` to log a user in with a ``Realm.Credentials`` object for an enabled authentication provider
```

source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+     - The method ``logOut()`` is called on the user instance, and will log any user out, regardless of the authentication provider the user was registered using.
+
+Log In
+------
+
+.. _node-login-anonymous:
+
+Anonymous
+~~~~~~~~~
+The Anonymous authentication provider allows users to log in to your
```
nlarew, line 42:
```suggestion
The :doc:`Anonymous </authentication/anonymous>` authentication provider allows users to log in to your
```

source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+application without providing credentials. To do this, create an
```
nlarew, line 43: "without providing credentials" is a bit iffy because they do still need to call `Realm.Credentials.anonymous()`

Maybe: "log in with ephemeral accounts that have no associated information"

source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+anonymous credential by calling "Realm.Credentials.anonymous()" and then
+pass in that credential to "<Realm.App instance>.logIn()".
```
nlarew, line 45:
```suggestion
pass in that credential to "App.logIn()".
```
Same deal here as above, we can use `Realm.App.logIn()` or even just `App.logIn()`

source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+
+
+.. admonition:: Anonymous Authentication
+   :class: note
+   
+   To learn more about the configuration and use-cases of anonymous
+   authentication read the :doc:`Anonymous Authentication
+   </authentication/anonymous>` doc.
```
nlarew, line 53: Good thought here, but what if we just link to the provider in the opening text when we first mention it?

I added a suggestion to this effect.

source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+      :tabid: typescript
+      
+      .. code-block:: typescript
+
+.. _node-login-email-password:
+
+Email/Password
+~~~~~~~~~~~~~~
+
+To log an existing email/password user into your application, create a
```
nlarew, line 95:
```suggestion
To log a registered and confirmed email/password user into your application, create a
```

nit but trying to get ahead of the inevitable confusion that goes hand in hand with email/password auth

source/node/authentication.txt
```diff
@@ -0,0 +1,257 @@
...
+credential by passing the email and password of the registered user to
+the method "Realm.Credentials.emailPassword()" as parameters, and then
+calling "<Realm.App instance>.logIn()" with that credential.
```
nlarew, line 98:
```suggestion
calling "App.logIn()" with that credential.
```

source/node/authentication.txt
mongomoe: Altered to: 
"MongoDB Realm provides developers with an API to log application users in and out with any enabled authentication provider. Pass in a credentials object to specify the authentication provider info to the login methods." 
which I think gets your points across. Let me know if not & if you have any specific phrasing you'd like

source/node/authentication.txt
mongomoe: Fixed, verbatim

source/node/authentication.txt
mongomoe: Fixed - almost verbatim. I did keep the 'for example' bit in and I think having a 'for example' doesn't hurt (and actually helps), especially because these aren't API docs.

source/node/authentication.txt
mongomoe: Altered it to: "The Anonymous Authentication provider allows users to log in to your application with short-term accounts that have no associated information." 
to avoid using the word ephemeral.

source/node/authentication.txt
mongomoe: Fair idea, fixed to your suggestion.

PR #225, opened 2020-06-05
grammar fixes for iOS guides

ccho-mongodb:
A few grammar fixes for iOS guides.

### DIFF
--- a/source/ios/collections.txt
+++ b/source/ios/collections.txt
@@ -38,7 +38,7 @@ List Collections
 A **list collection** represents a :ref:`to-many relationship
 <ios-to-many-relationship>` between two {+service-short+} types. Lists
 are mutable: within a write transaction, you can add and remove elements
-from a list. Lists are not associated with a query and are usually
+to and from a list. Lists are not associated with a query and are usually
 declared as a property of an :ref:`object model <ios-object-model>`.
 
 .. seealso:: :ref:`To-Many Relationships <ios-to-many-relationship>`
--- a/source/ios/collections.txt
+++ b/source/ios/collections.txt
@@ -50,7 +50,7 @@ Results Collections
 
 A **results collection** represents the lazily-evaluated results of a
 query operation. Results are immutable: you cannot add or remove
-elements from the results collection. Results have an associated query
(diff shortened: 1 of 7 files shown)


source/ios/collections.txt
```diff
@@ -38,7 +38,7 @@ List Collections
 A **list collection** represents a :ref:`to-many relationship
 <ios-to-many-relationship>` between two {+service-short+} types. Lists
 are mutable: within a write transaction, you can add and remove elements
-on a list. Lists are not associated with a query and are usually
+from a list. Lists are not associated with a query and are usually
```
cbush, line 41: The awkward thing here is that the preposition connects to both "add" AND "remove", hence "on" (as in, while working "on"). Technically, it should be "add and remove elements to and from a list", which is quite wordy.
```suggestion
to and from a list. Lists are not associated with a query and are usually
```

source/ios/collections.txt
```diff
@@ -50,7 +50,7 @@ Results Collections
 
 A **results collection** represents the lazily-evaluated results of a
 query operation. Results are immutable: you cannot add or remove
-elements on the results collection. Results have an associated query
+elements from the results collection. Results have an associated query
```
cbush, line 53:
```suggestion
elements to or from the results collection. Results have an associated query
```

source/ios/collections.txt
ccho-mongodb: I think you can use the way you mentioned, but it is also valid using the last item in the list of items (the way I suggested).

cbush: Do you have a source for that, for my edification?

ccho-mongodb: I wasn't able to find any definitive information, but we may both be incorrect as mixing prepositions in a parallel structure creates a clumsy sentence. See https://www.grammarly.com/blog/parallel-structure-prepositions/#:~:text=What%20to%20Avoid%3A%20Mixing%20Prepositions,result%20is%20a%20clumsy%20sentence.
It seems ok if you place the preposition next to the noun which they describe though, so the following should work:
within a write transaction, you can add elements to the list and remove them from it.

PR #226, opened 2020-06-05
authentication first pass: android

nathan-contino-mongo:
Staging: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/AndroidLoginAnon/android/authenticate.html

You'll have to tab over to kotlin to see the code samples, unfortunately.

### DIFF
--- a/source/android/authenticate.txt
+++ b/source/android/authenticate.txt
@@ -15,15 +15,81 @@ Authenticate a User
 Overview
 --------
 
+MongoDB Realm provides an API for authenticating users using any enabled
+authentication provider. Instantiate a ``Credentials`` object and pass
+it to either of the  ``login`` or ``loginAsync`` method to authenticate
+a user login and create a ``User`` object. Each authentication provider
+corresponds to a static helper method used to instantiate
+``Credentials`` objects using that authentication providers. See the
+table below for a mapping of authentication providers to helper methods:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 50 50
+
+   * - Authentication Provider
+     - Credentials Generation Method
+
+   * - :ref:`Anonymous <android-login-anonymous>`
+     - ``Credentials.anonymous()``
+
+   * - :ref:`Email/Password <android-login-email-password>`
+     - ``Credentials.emailPassword(String email, String password)``
+
+   * - :ref:`API Key <android-login-api-key>`
+     - ``Credentials.apiKey(String key)``
+
+   * - :ref:`Custom Function <android-login-custom-function>`
+     - ``Credentials.customFunction(String functionName, Object... arguments)``
+
+   * - :ref:`Custom JWT <android-login-custom-jwt>`
+     - ``Credentials.jwt(String jwtToken)``
+
+   * - :ref:`Google OAuth <android-login-google>`
+     - ``Credentials.google(String googleToken)``
+
+   * - :ref:`Facebook OAuth <android-login-facebook>`
+     - ``Credentials.facebook(String accessToken)``
+   
+   * - :ref:`Sign-in With Apple <android-login-apple>`
+     - ``Credentials.apple(String idToken)``
+
 .. _android-login:
 
 Log In
 ------
 
+You can authenticate users with either the ``app.login()`` or
+``app.loginAsync()`` methods. While the ``app.login()`` method blocks
+code execution in the calling thread until the supplied credentials have
+either succeeded or failed to authenticate a user, the
+``app.loginAsync()`` method allows execution to continue, handling
+success or failure with a callback function that is guaranteed to
+execute on the same thread that called ``app.loginAsync()``.
+
+If successful, the ``app.login()`` method returns a ``User`` object.
+In the event of a failure, the ``app.login()`` method throws an
+exception of type ``ObjectServerError``.
+
+Pass a callback to the ``app.loginAsync()`` method to handle success or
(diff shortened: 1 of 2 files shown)


nathan-contino-mongo:
@cbush @nlarew do y'all know if there's any way to set the tabs to kotlin by default here? I'm guessing that's a global setting (or something defined in docs-tools) but i think it would be a good change to make for our initial realm release, since we're not going to have much java documentation for some time.

source/android/authenticate.txt
```diff
@@ -15,22 +15,107 @@ Authenticate a User
 Overview
 --------
 
+MongoDB Realm provides an API for authenticating users using any enabled
+authentication provider. Instantiate a ``Credentials`` object and pass
+it to either of the  ``login`` or ``loginAsync`` method to authenticate
+a user login and create a ``User`` object. Each authentication provider
+corresponds to a static helper method used to instantiate
```
ccho-mongodb, line 22: While this description seems correct, it might be easier and more concise to combine this sentence and the next with something like the following:
"See the table below to find the method that instantiates the ``Credentials`` instance for your authentication provider:"

source/android/authenticate.txt
```diff
@@ -15,22 +15,107 @@ Authenticate a User
...
+   
+   * - :ref:`Sign-in With Apple <android-login-apple>`
+     - ``Credentials.apple(String idToken)``
+
 .. _android-login:
 
 Log In
 ------
 
+You can authenticate users with either the ``app.login()`` or
```
ccho-mongodb, line 62: Nitpick: could be helpful to either describe what type of object "app" is or omit it and mention the class that defines those functions:
"You can call the ``login()`` or ``loginAsync()`` method on your ``App` instance."

source/android/authenticate.txt
```diff
@@ -15,22 +15,107 @@ Authenticate a User
...
+``app.loginAsync()`` method allows execution to continue, handling
+success or failure with a callback function that is guaranteed to
+execute on the same thread that called ``app.loginAsync()``.
+
+If successful, the ``app.login()`` method returns a ``User`` object.
+In the event of a failure, the ``app.login()`` method throws an
+exception of type ``ObjectServerError``.
+
+Pass a callback to the ``app.loginAsync()`` method to handle success or
+failure. This callback should accept a single parameter of type
```
ccho-mongodb, line 75: Nitpick: I think "should" indicates there is another possibility. Instead, you can omit it:
"This callback accepts a single parameter..."

source/android/authenticate.txt
```diff
@@ -15,22 +15,107 @@ Authenticate a User
...
+``App.Result``. If successful, the ``isSuccess()`` method of the
+``App.Result`` object passed to the callback should return true. In the
```
ccho-mongodb, line 77: I think you can make these two sentences more concise, e.g.:
"The `isSuccess()`` method on the ``App.Result`` object in the callback returns a boolean that indicates whether the operation succeeded. If it returns false, you can retrieve the error from ``App.Result`` by calling the ``getError()`` method.

source/android/authenticate.txt
```diff
@@ -15,22 +15,107 @@ Authenticate a User
...
+event of a failure, the ``isSuccess()`` method of the ``App.Result``
+object passed to the callback should return false, and you can view the
+error that caused the failure using the ``getError()`` method.
+
 .. _android-login-anonymous:
 
 Anonymous
 ~~~~~~~~~
 
+The :doc:`Anonymous </authentication/anonymous>` provider enables users
```
ccho-mongodb, line 87: Nitpick: should "Anonymous" be capitalized? Also, should the link be on both words of the noun "anonymous provider"?

source/android/authenticate.txt
```diff
@@ -15,22 +15,107 @@ Authenticate a User
 Overview
 --------
 
+MongoDB Realm provides an API for authenticating users using any enabled
+authentication provider. Instantiate a ``Credentials`` object and pass
+it to either of the  ``login`` or ``loginAsync`` method to authenticate
```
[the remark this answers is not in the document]
ccho-mongodb: Another nitpick: stay consistent with whether you include "()" with your methods.

source/android/authenticate.txt
```diff
@@ -161,12 +272,32 @@ Sign-in with Apple
 Log Out
 -------
 
+You can log out any user, regardless of the authentication provider used
+to log in, using the ``logOut()`` or ``logOutAsync()`` methods. This:
```
ccho-mongodb, line 276: Is this called on your instance of App? I could be helpful to include that information.

source/android/authenticate.txt
```diff
@@ -161,12 +272,32 @@ Sign-in with Apple
...
+
+- deletes locally stored user credentials from the device
+
+- immediately halts any synchronization to and from the user's {+realm+}s
+
+- marks the user's {+realm+}s for deletion the next time the app restarts
+
+Because logging out halts synchronization, you should only log out after
```
ccho-mongodb, line 284: Question: is there a function they can check to determine this? That could be useful to include if so.

nathan-contino-mongo: I actually reached out to Christian about this, but I haven't heard anything back yet. I'm recommending the async method because I *hope* that it'll wait for some of those syncs to finish, but once I get more info I will supplement this :)

PR #231, opened 2020-06-06
Clarify note about Node.js/RN incompatibility

cbush:
- The Node.js and RN SDKs are the same. However, the documentation is different
in case we want more RN-specific code examples. This commit clarifies the note
that previously implied the two SDKs were not the same.

https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/rn-text-update/node.html

### DIFF
--- a/source/node.txt
+++ b/source/node.txt
@@ -12,10 +12,11 @@ TypeScript applications.
 
 .. note::
 
-   The Node.js SDK does not support JavaScript or TypeScript applications
-   written for web browsers or React Native mobile apps. For those use cases,
-   you should consider the :ref:`Web SDK <web-intro>` and the :ref:`React Native
-   SDK <react-native-intro>`.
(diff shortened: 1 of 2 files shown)


source/node.txt
```diff
@@ -12,10 +12,11 @@ TypeScript applications.
 
 .. note::
 
-   The Node.js SDK does not support JavaScript or TypeScript applications
-   written for web browsers or React Native mobile apps. For those use cases,
-   you should consider the :ref:`Web SDK <web-intro>` and the :ref:`React Native
-   SDK <react-native-intro>`.
+   The Node.js SDK does not support JavaScript or TypeScript
+   applications written for web browsers. For that use case, you should
+   consider the :ref:`Web SDK <web-intro>`. For development on React
```
nlarew, line 17:
```suggestion
   consider the :ref:`Web SDK <web-intro>`. For development in React
```
nit

PR #295, opened 2020-06-17
(DOCSP-10587): Update measurements endpoint docs

cbush:
- Returned dates are actually RFC 3339
- Mem usage not returned in group result
- App ID and app name are returned in app result
- MILLIS unit removed

## Jira
https://jira.mongodb.org/browse/DOCSP-10587

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/measurements/admin/api/v3.html#billing-apis

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -1763,7 +1763,54 @@ paths:
             enum: ["P31D", "PT1H"]
           required: False
       responses:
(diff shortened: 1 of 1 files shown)


cbush:
@ccho-mongodb it is odd. Apparently they wanted to match the API of Atlas while using the built-in types of Go. Practically speaking, there should be no difference to the end user.

PR #299, opened 2020-06-17
TOC Cleanup + Hide Legacy Code

nlarew:
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/toc-cleanup/index.html)


source/services.txt
```diff
@@ -1,9 +1,10 @@
 .. _services:
 .. _external-services:
+.. _third-party-services:
 
-=================
-External Services
-=================
+==================
+3rd Party Services
```
nathan-contino-mongo, line 6: Can we change this to "Third"? And should we also update "external dependencies" to "third/3rd party dependencies"?

source/services/call-a-service-action.txt
```diff
@@ -54,154 +51,154 @@ To call a service action from a :doc:`function </functions>`:
...
+..             )
+
+..             http.execute(request) { result in
+..               switch result {
+..               case .success:
+..                 print("Successfully sent POST request!")
+..               case .failure(let error):
+..                 print("Error sending POST request: \(error)")
+..               }
+..             }
```
nathan-contino-mongo, line 204: I would say that we ought to just remove this instead of commenting it out -- we can always dig it out of the git history if we need it!

source/services.txt
nlarew: - +1 for 3rd => Third
- You can publish your own npm packages, so they're not necessarily third party

PR #315, opened 2020-06-30
(DOCSP-10802): Add Work with Multiple Users to Client Guides

nlarew:
## Jira

- [(DOCSP-10802): Add Work with Multiple Users to Client Guides](https://jira.mongodb.org/browse/DOCSP-10802)

## Staged Changes

- [Work with Multiple Users (Node)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/multi-user/node/work-with-multiple-users.html)
- [Work with Multiple Users (React Native)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/multi-user/react-native/work-with-multiple-users.html)
- [Work with Multiple Users (Web)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/multi-user/web/work-with-multiple-users.html)

### Currently Hidden

Not sure if these are fully supported yet. Either way, this PR scaffolds out the pages (including a couple of best effort code examples) for Android/iOS

- [Work with Multiple Users (Android)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/multi-user/android/work-with-multiple-users.html)
- [Work with Multiple Users (iOS)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/multi-user/ios/work-with-multiple-users.html)


source/android/work-with-multiple-users.txt
```diff
@@ -0,0 +1,147 @@
...
+--------
+
+Multiple :ref:`users <user-accounts>` can log in to a :term:`{+service-short+}
+client` simultaneously. You can quickly switch between authenticated users
+without requiring them to log in again.
+
+.. important::
+
+   {+service-short+} clients run in the context of a single active user even if
+   multiple users are logged in simultaneously.
```
ccho-mongodb, line 25: Question: is it apparent what the dangers are of running on the same context? Otherwise, would it be more clear/useful to provide an example scenario in which it poses a problem?

source/android/work-with-multiple-users.txt
```diff
@@ -0,0 +1,147 @@
...
+      
+      .. literalinclude:: /examples/MultiUser/ListUsers/ListUsers.java
+         :language: java
+
+.. _android-switch-user:
+
+Switch the Active User
+~~~~~~~~~~~~~~~~~~~~~~
+
+You can quickly switch an app's :ref:`active user <active-user>` to another
```
ccho-mongodb, line 110: Question: is a client the same as an app? I see it referred to as "client app" on line 88, but "client" prior to that (if they are the same thing).

source/examples/MultiUser/AddUser/AddUser.kt
```diff
@@ -0,0 +1,25 @@
...
+  } else {
+    Log.e(TAG, it.getError().toString());
+  }
+}
+
+// Log in as Emma
+val emmaCredentials: Credentials = Credentials.emailPassword("emma@example.com", "pa55word")
+app.loginAsync(emmaCredentials) {
+  if (it.isSuccess()) {
+    // The active user is now Emma
```
ccho-mongodb, line 19: Question: but is Joe still logged in? Should the comments be the same as the js example or is the behavior different?

source/examples/MultiUser/LogoutUser/LogoutUser.ts
```diff
@@ -0,0 +1,12 @@
+const app: Realm.App = new Realm.App({ id: "myapp-abcde" });
+
+// Remove the current user from the device
+const user: Realm.User = app.currentUser;
+await app.removeUser(user);
+
+// The user is no longer the active user
+// The active user is now another logged in user, if there is one
```
ccho-mongodb, line 8: Question: How does it select the other logged in user?

source/examples/MultiUser/AddUser/AddUser.kt
nlarew: Tough to order these since they're both async methods, not sure if there's a race condition. There's probably a way to await one but I'm not a kotlin pro :(

PR #338, opened 2020-07-15
(DOCSP-10023): IoT SDK installation instructions

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-10023

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-10023iot/node/install.html#iot-installation

We could also look into creating a separate tab for the IoT SDK, but since it doesn't deviate anywhere else I'm not sure if it's necessary. Open to the idea though if you think this layout is confusing or just too long though.

### DIFF
--- a/source/includes/steps-install-node-iot.yaml
+++ b/source/includes/steps-install-node-iot.yaml
@@ -35,7 +35,7 @@ content: |
 
    .. code-block:: bash
 
-      npm install --save realm@beta
+      npm install --save realm@10.0.0-beta.6
 
 ---
 title: Enable TypeScript (optional)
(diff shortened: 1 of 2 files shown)


source/includes/steps-install-node-iot.yaml
```diff
@@ -0,0 +1,53 @@
...
+title: Install Realm with NPM
+ref: install-realm-with-npm
+content: |
+
+   In your Node.js project directory, use the following command
+   to add {+service-short+} to your project:
+
+   .. code-block:: bash
+
+      npm install --save realm@beta
```
cbush, line 38: Question: Does that find the latest beta?

nathan-contino-mongo, line 38:
```suggestion
      npm install --save realm@10.0.0-beta.6
```

PR #351, opened 2020-07-21
(DOCSP-11291): Port conflict resolution page from old docs

cbush:
Based on https://docs.realm.io/sync/v/3.x/self-hosted/customize/conflict-resolution

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/bush/sync-details/sync/conflict-resolution.html

## Jira
https://jira.mongodb.org/browse/DOCSP-11291

### DIFF
--- a/source/sync/conflict-resolution.txt
+++ b/source/sync/conflict-resolution.txt
@@ -1,0 +1,123 @@
+.. _conflict-resolution:
+
+===================
+Conflict Resolution
+===================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+One of the defining features of mobile app development is the fact that
+you can never count on being online. :ref:`Loss of connectivity
+<the-connectivity-problem>`, slow networks, and choppy connections are
+facts of life. But people still expect their apps to work.
+
+This means that you may end up having two or more users
+making changes to the same piece of data independently, creating
+conflicts. Note that this can happen even with perfect connectivity, as
+there may be enough latency between device and server that they can
+create conflicting changes at the same time. 
+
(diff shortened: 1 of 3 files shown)


source/sync/conflict-resolution.txt
```diff
@@ -0,0 +1,123 @@
...
+
+Inserts in lists are ordered by time.
+  If two items are inserted at the same position, the item that was
+  inserted first will end up before the other item. This means that if
+  both sides append items to the end of a list, will include both items
+  in order of insertion time.
+
+Primary keys designate object identity.
+  If two sides both create objects of the same class with identical
+  primary keys, they will be treated as instances of the same object.
```
nlarew, line 66: Makes me wonder - if I have two simultaneous CREATE operations for the same primary key (**CONFLICT!**), does one get transformed into an UPDATE that follows _last update wins_ or are they merged some other way? If so, which one becomes the update and does it matter?"

source/sync/conflict-resolution.txt
```diff
@@ -0,0 +1,123 @@
...
+
+.. _custom-conflict-resolution:
+
+Custom Conflict Resolution
+--------------------------
+
+Generally speaking, the conflict resolution of {+service-short+} should
+work for most purposes, and you should not need to customize it. That
+said, the standard way to do custom conflict resolution is to change a
+property type from string to list. Each side can then add its updates to
```
nlarew, line 112: Does this apply broadly or is it limited to custom resolution for `string` fields?

source/sync/conflict-resolution.txt
```diff
@@ -0,0 +1,123 @@
...
+the list and apply any conflict resolution rules it wants directly in
+the data model. You can use this technique to implement max, min, first
```
nlarew, line 114: Maybe suggest using a trigger to handle custom conflict resolution?

PR #353, opened 2020-07-22
(DOCSP-11025): Android SDK MongoDB Data Access

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-11025

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-11025/android/mongodb.html

Also corrected the java-sdk link, which preserved the sidebar but apparently didn't correctly scroll down to method definitions.


source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+
+Update a Single Document
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can update a single document using
+:java-sdk:`collection.updateOne()
+<io/realm/mongodb/mongo/MongoCollection.html#updateOne-Bson-Bson->`.
+
+The following snippet updates a single document in a
+:ref:`collection of documents that describe plants for sale in a store
```
ccho-mongodb, line 413: I wonder if it would be useful to mention 1) you're searching all the different stores in the collection since in other examples you filtered on a single store, 2) the updateOne will update only the first match even if there are multiple documents that match.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+                    Log.v("EXAMPLE", "Did not update a document.");
+                }
+            } else {
+                Log.e("EXAMPLE", "failed to update document with: ", task.getException());
+            }
+         });
+
+.. _android-mongodb-updateMany:
+
+Update One or More Documents
```
ccho-mongodb, line 476: Similar comment to Insert One or More Documents section: perhaps this could be "update multiple documents". Not sure the "One" is necessary since that's subsumed by "multiple".

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+
+You can update a single document using
+:java-sdk:`collection.updateMany()
+<io/realm/mongodb/mongo/MongoCollection.html#updateMany-Bson-Bson->`.
+
+The following snippet updates multiple documents in a
+:ref:`collection of documents that describe plants for sale in a store
+<android-mongodb-example-dataset>`. This operation queries for documents
+where the ``_partition`` field contains the value "Store 47" and changes
+the value of the ``_partition`` field of each document to "Store 51":
```
ccho-mongodb, line 487: Nitpick: "matching document to Store 51" just to be very clear

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+queries for documents where:
+
+- the ``sunlight`` field has a value of "full"
+- the ``type`` field has a value of "perennial"
+- the ``color`` field has a value of "green"
+- the ``partition`` field has a value of "Store 47"
+
+Because this snippet sets the ``upsert`` option to ``true``, if no
+document matches the query, MongoDB creates a new document matching the
+literal values provided to the query and updates that new document with
```
ccho-mongodb, line 567: Nitpick: since this is the first time you use the term "literal values", perhaps you should clarify. Also, "matching literal values provided" is confusing. How about "MongoDB creates a new document that includes both the query filter and specified updates".

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+
+.. _android-mongodb-delete-documents:
+
+Delete Documents
+----------------
+
+These code snippets demonstrate how to delete documents that
+are stored in a MongoDB collection from a mobile application.
+Delete operations use a query filter to specify which documents to delete
+and return a results object containing information about the operation.
```
ccho-mongodb, line 639: Nitpick: could be helpful to specify what type of information that is. Possible rewording: "and returns an object that contains the results of the execution of the operation"

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+
+You can delete a single document from a collection using
+:java-sdk:`collection.deleteOne()
+<io/realm/mongodb/mongo/MongoCollection.html#deleteOne-Bson->`.
+
+The following snippet deletes one document in a
+:ref:`collection of documents that describe plants for sale in a store
+<android-mongodb-example-dataset>`. This operation queries for a
+document where the ``color`` field has a value of "green" and deletes
+the first document that matches the query:
```
ccho-mongodb, line 654: Perhaps I missed it, but I haven't seen any info in this guide that mentions how the first result is determined if there are multiple matches. Perhaps there's a server docs link you could point to?

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+documents as well as collect summary data about groups of related
+documents.
+
+.. _android-mongodb-aggregate:
+
+Aggregate Documents in a Collection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can execute an aggregation pipeline using
+:java-sdk:`collection.aggregate()
```
ccho-mongodb, line 799: Nitpick: I wonder if it could be more clear by describing the aggregation pipeline as the parameter of the call to this method as "using" does not describe how. I also wonder if it could be helpful to mention here that the "series of data aggregation stages" mentioned in the previous section are represented as a List.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+
+.. _android-mongodb-create-documents:
+
+Create Documents
+----------------
+
+These code snippets demonstrate how to insert one or more
+documents into a MongoDB collection from a mobile application. Insert
+operations take a document or documents to add to MongoDB as an argument
+and return a results object that describes the effect of the operation.
```
ccho-mongodb, line 42: I wonder if it would be worthwhile mentioning that in these code examples, you're chaining `addOnCompleteListener` to the results object to access the values when they operation completes. I haven't been able to find any mention of that here which could be confusing to the user.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+The following snippet updates a document in a
+:ref:`collection of documents that describe plants for sale in a store
+<android-mongodb-example-dataset>` or inserts a
+new document if no document matches the query filter. This operation
+queries for documents where:
+
+- the ``sunlight`` field has a value of "full"
+- the ``type`` field has a value of "perennial"
+- the ``color`` field has a value of "green"
+- the ``partition`` field has a value of "Store 47"
```
ccho-mongodb, line 563: Question: should this be "_partition"? I haven't seen any explanation of the `_partition` field syntax. Would it be useful to add?

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+         });
+
+.. _android-mongodb-aggregation-pipelines:
+
+Aggregation Pipelines
+---------------------
+
+These code snippets demonstrate how to configure and run
+:manual:`aggregation pipelines </core/aggregation-pipeline>` against a
+collection. Aggregation operations run all documents in a collection
```
ccho-mongodb, line 787: It could be helpful to connect these three concepts a little more clearly:
1) "aggregation pipelines"
2) "aggregation operations"
3) "aggregation stages"
Right now, I think there is a missing link between aggregation pipelines and an aggregation stage (multiple stages make up a pipeline).

source/android/mongodb.txt
nathan-contino-mongo: I debated this a lot, but I figure that users can just check the reference documentation for specifics (added bonus: we don't have to update this sentence if the SDK changes again)

nathan-contino-mongo: Rewording as per your suggestion though.

source/android/mongodb.txt
nathan-contino-mongo: I debated linking into the natural order definition in the glossary of the manual or [here](https://docs.mongodb.com/manual/reference/method/cursor.sort/#return-natural-order), but I think I lean in the direction of not linking at all since natural order is a big concept that basically boils down to "undefined implementation detail." I could also add some details about sorting results / a small discussion on cursors that explains more.

source/android/mongodb.txt
nathan-contino-mongo: It's suggested as a nice way to define your partition key in realm since it's agnostic (but also meaningless) across collections, but I'm open to changing it to `store`. I thought it would be nice to bleed some of the current realm sync standards into these examples to breed familiarity.

source/android/mongodb.txt
nathan-contino-mongo: Added a note about this to the preceding section.

source/android/mongodb.txt
nathan-contino-mongo: This is a really good point -- all of these methods actually return a `Task` which either resolves to an error or the described output. I'm going to revise the "return" values for each of the CRUDA sections to mention this explicitly.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+Insert a Single Document
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can insert a single document using
+:java-sdk:`collection.insertOne()
+<io/realm/mongodb/mongo/MongoCollection.html#insertOne-DocumentT->`.
+
+The following snippet inserts a single document describing a "lily of the
+valley" plant into a
+:ref:`collection of documents that describe plants for sale in a store
```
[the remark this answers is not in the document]
nathan-contino-mongo: Changed this to "group of stores" (since the partition value represents the individual store). Nice catch!

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+                    Log.v("EXAMPLE", results.next().toString());
+                }
+            } else {
+                Log.e("EXAMPLE", "failed to aggregate documents with: ", task.getException());
+            }
+         });
+
+.. _android-mongodb-aggregation-stages:
+
+Aggregation Stages
```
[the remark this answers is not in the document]
nathan-contino-mongo: This is a really good point. I'm also going to add output sections for each of the crud operations.

source/android/mongodb.txt
nathan-contino-mongo: Also changed the title of this section to "Aggregate Documents" to bring it in line with the other CRUD section headings.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+You can use the :manual:`$project
+</reference/operator/aggregation/project/>` stage to include or omit
+specific fields from documents or to calculate new fields using
+:manual:`aggregation operators </reference/operator/aggregation>`.
+To include a field, set its value to ``1``. To omit a field, set its
+value to ``0``.
+
+.. note::
+
+   You cannot simultaneously omit and include fields other than ``_id``.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Agreed that this could use some improvement -- rewording this entire note.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+   .. code-block:: javascript
+
+      {
+        "$unwind": {
+          "path": "$items",
+          "includeArrayIndex": "itemIndex"
+         }
+      }
+
+Consider the following document from the ``purchases`` collection:
```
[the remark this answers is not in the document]
nathan-contino-mongo: I think I'm going to leave the sample document where it is because otherwise the reader has to scroll way up the page to compare the sample data and the output of $unwind.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+        "$addFields": {
+          "numItems": { "$sum": { "$size": "$items" } }
+        }
+      }
+
+Unwind Array Values
+~~~~~~~~~~~~~~~~~~~
+
+You can use the :manual:`$unwind
+</reference/operator/aggregation/unwind/>` stage to aggregate
```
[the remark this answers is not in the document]
nathan-contino-mongo: Ditching the "aggregate" verb here -- I agree, it's a cop-out.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+
+Group Documents
+~~~~~~~~~~~~~~~
+
+You can use the :manual:`$group
+</reference/operator/aggregation/group/>` stage to aggregate summary
+data for groups of one or more documents. MongoDB groups documents based
+on the ``_id`` expression.
+
+.. note::
```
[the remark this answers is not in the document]
nathan-contino-mongo: Moving this into the "group documents" paragraph, because I think it's vital to understanding how this stage works.

source/android/mongodb.txt
```diff
@@ -0,0 +1,1082 @@
...
+.. _android-mongodb-aggregation-stages:
+
+Aggregation Stages
+------------------
+
+Filter Documents
+~~~~~~~~~~~~~~~~
+
+You can use the :manual:`$match
+</reference/operator/aggregation/match/>` stage to filter incoming
```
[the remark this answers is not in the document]
nathan-contino-mongo: "incoming" was definitely confusing -- got rid of it.

nathan-contino-mongo:
@ccho-mongodb Would appreciate it if you could take another look. Changes in my new commit:
- fixes for most of your comments
- moved the language-agnostic Aggregation Stages section to an includes file (for future MDB data access PRs in other SDKs)
- Aggregation Stages now use the same dataset as the rest of the examples
- output for every example below the example (output should be the same across the two languages)

Please take a look and let me know if you have any more suggestions or if this needs more changes! Thank you.

source/includes/aggregation-stages.rst
```diff
@@ -0,0 +1,230 @@
...
+        }
+      }
+
+Group Documents
+~~~~~~~~~~~~~~~
+
+You can use the :manual:`$group
+</reference/operator/aggregation/group/>` stage to aggregate summary
+data for one or more documents. MongoDB groups documents based
+on the ``_id`` expression. You can reference a specific document field
```
ccho-mongodb, line 39: Nitpick: While "_id expression" may be accurate, "expression defined in the _id field" may be more clear. The former makes me think "_id expression" is a term that I should know, but has never been introduced.

source/android/mongodb.txt
ccho-mongodb: How about "Note: the documents may be matched in an indeterminate order unless you specify a sorting parameter."

source/android/mongodb.txt
ccho-mongodb: Could you point me to these changes?

source/android/mongodb.txt
ccho-mongodb: I wasn't able to find any changes related to this. It's not that clear that "_partition" is the same as "partition", so it doesn't match the pattern of describing the fields.

source/android/mongodb.txt
```diff
@@ -0,0 +1,953 @@
...
+The following snippet updates a document in a
+:ref:`collection of documents that describe plants for sale in a group of stores
+<android-mongodb-example-dataset>` or inserts a
+new document if no document matches the query filter. This operation
+queries for documents where:
+
+- the ``sunlight`` field has a value of "full"
+- the ``type`` field has a value of "perennial"
+- the ``color`` field has a value of "green"
+- the ``_partition`` field has a value of "Store 47"
```
nathan-contino-mongo, line 615: @ccho-mongodb I'm not sure why github is struggling to show you this, but I've had enough github issues in the last week that it's hardly a surprise.

source/android/mongodb.txt
nathan-contino-mongo: I can add this note, but I should probably add it in multiple places (on almost every method). Do you think there's a better way?

ccho-mongodb: I think it's important enough to mention. Perhaps you could place it once in the CRUD intro. Something like:
"When a filter matches multiple documents in a collection, they are returned in an indeterminate order unless you specify a sorting parameter. For more information on sorting, see <>"

nathan-contino-mongo: Great idea, adding this!

PR #362, opened 2020-07-23
(DOCSP-11353): flesh out anonymous auth enable in UI steps

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-11353

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-11353/authentication/anonymous.html

Also took the liberty of converting the tabs on the anonymous auth page from legacy tabs to modern tabs. So shiny and new!

### DIFF
--- a/source/authentication/anonymous.txt
+++ b/source/authentication/anonymous.txt
@@ -60,35 +60,56 @@ Configuration
 
 .. tabs-realm-admin-interfaces::
 
-   tabs:
-     - id: ui
-       content: |
-         You can enable the Anonymous authentication provider from the
-         {+ui+} by selecting :guilabel:`Anonymous` from the
-         :guilabel:`Users > Providers` page.
 
-     - id: cli
-       content: |
-         You can enable the Anonymous authentication provider with
-         :doc:`realm-cli </deploy/realm-cli-reference>` by
-         importing an application directory that contains a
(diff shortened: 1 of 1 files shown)


source/authentication/anonymous.txt
```diff
@@ -60,35 +60,56 @@ Configuration
...
+         that appears at the top of the page.
+
+      #. Review the changes to your app's configuration and click
+         :guilabel:`Deploy` in the lower right of the dialog to
+         make your changes available to application users.
+
+   .. tab::
+      :tabid: cli
+
+      You can enable the Anonymous authentication provider with
```
ccho-mongodb, line 94: Question: does the user have to run anything with the realm-cli once they have the config file?

source/authentication/anonymous.txt
```diff
@@ -60,35 +60,56 @@ Configuration
...
+      :doc:`realm-cli </deploy/realm-cli-reference>` by
+      importing an application directory that contains a
```
ccho-mongodb, line 96: Related to my prior question, it looks like this is the answer, but I wasn't able to find how to import the application directory. If that's something mentioned on another page, would it be worthwhile adding a quick how-to here?

PR #363, opened 2020-07-23
Node.js CLI tutorial

MongoCaleb:
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/realm/caleb/node_cli/tutorial/nodejs-cli.html)
[Jira](https://jira.mongodb.org/browse/DOCSP-11230)


source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+``realmAppId``, which is currently set to "TODO". Change this value to your 
+Realm App ID. 
+
+.. note::
+
+   To learn how to find your MongoDB Realm ``appId``, read the :doc:`Find
+   Your App Id </get-started/find-your-app-id>` doc.
+
+Once you have made that change, you now need to complete the code needed to 
+open a Realm. In ``index.js``, find the ``openRealm`` function. It will look
```
cbush, line 160:
```suggestion
open a {+realm+}. In ``index.js``, find the ``openRealm`` function. It will look 
```

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+   async function openRealm() {
+      const config = {
+         schema: [schemas.TaskSchema, schemas.UserSchema, schemas.ProjectSchema],
+         sync: {
+            user: users.getAuthedUser(),
+            partitionValue: "myPartition",
+         },
+      };
+      // TODO: open a Realm with these configuration settings.
```
cbush, line 173: Lowercase refers to realm instance
```suggestion
      // TODO: open a realm with these configuration settings.
```

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+   }
+
+Replace the ``TODO`` line with a line of code that opens a Realm and assigns
```
cbush, line 176:
```suggestion
Replace the ``TODO`` line with a line of code that opens a {+realm+} and assigns 
```

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+D. Enable authentication
+------------------------
+In the ``users.js`` file, we have a ``logIn`` function that prompts the user for 
+an email address and password, and then, within a try-catch block, creates an 
+:js-sdk:`emailPassword <Realm.App.html#Realm.Credentials.html#.emailPassword>` 
+credential and passes it to the Realm 
+:js-sdk:`logIn() <Realm.App.html#logIn>` method. 
+
+Find the the following comment:
```
cbush, line 195:
```suggestion
Find the following comment:
```

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+    const user = await app.logIn(credentials);
+
+
+E. Implement the CRUD methods
+-----------------------------
+In the ``tasks.js`` file, there are a number of functions to handle typical  
+CRUD functionality: ``getTasks``, ``getTask``, ``createTask``, ``deleteTask``, 
+``editTask``, and ``changeStatus``, which is essentially a specific case of 
+``editTask``. Each of these functions (except ``getTasks``) prompts the user,
```
cbush, line 223: No comma because the subject didn't change in the next clause
```suggestion
``editTask``. Each of these functions (except ``getTasks``) prompts the user
```

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+
+E. Implement the CRUD methods
+-----------------------------
+In the ``tasks.js`` file, there are a number of functions to handle typical  
+CRUD functionality: ``getTasks``, ``getTask``, ``createTask``, ``deleteTask``, 
+``editTask``, and ``changeStatus``, which is essentially a specific case of 
+``editTask``. Each of these functions (except ``getTasks``) prompts the user, 
+and then makes the appropriate call to Realm. Your task is to implement the 
+calls to Realm. The following list shows each function and provides
```
cbush, line 225: Replacement here would be:
```suggestion
calls to {+service-short+}. The following list shows each function and provides 
```

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+CRUD functionality: ``getTasks``, ``getTask``, ``createTask``, ``deleteTask``, 
+``editTask``, and ``changeStatus``, which is essentially a specific case of 
+``editTask``. Each of these functions (except ``getTasks``) prompts the user, 
+and then makes the appropriate call to Realm. Your task is to implement the 
+calls to Realm. The following list shows each function and provides 
+guidance on how to complete the task.
+
+* ``getTasks``
+
+  To get all objects, call the :js-sdk:`obects() <Realm.html#objects>` method
```
cbush, line 230:
```suggestion
  To get all objects, call the :js-sdk:`objects() <Realm.html#objects>` method 
```

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+guidance on how to complete the task.
+
+* ``getTasks``
+
+  To get all objects, call the :js-sdk:`obects() <Realm.html#objects>` method 
+  and pass in the name of the collection:
+  
+  .. code-block:: js
+     
+     const tasks = await realm.objects("Task");
```
cbush, line 235: I don't think this should be awaited (may need to update the tutorial source)

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+* ``getTask``
+  
+  In the Tasks collection, a task's ``id`` field is the primary key, so 
+  we call the :js-sdk:`objectForPrimaryKey() <Realm.html#objectForPrimaryKey>`  
+  function to get a task by its Id.
+
+  .. code-block:: js
+     
+     let result = await realm.objectForPrimaryKey(
```
cbush, line 246: It's misleading to `await` this non-async function call

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+
+* ``createTask``
+
+  Whenever we modify an object in realm, we must do so within a 
+  :ref:`transaction <node-open-a-transaction>`. 
+  While you can do this by calling the 
+  :js-sdk:`beginTransaction() <Realm.html#beginTransaction>` and 
+  :js-sdk:`commitTransaction() <Realm.html#commitTransaction>`
+  methods, the
```
cbush, line 259: Is it worth mentioning begin/commit? Should we prefer to lead with the recommended approach (write)?

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+------------------------------------
+A very handy feature in Realm is the ability to add a listener to a collection. 
+The listener notifies your app when a change occurs in the collection. For our 
+CLI, we want to listen for changes on the ``Tasks`` collection, and when one 
+occurs, display a notification in the console window. To do this:
+
+1. In your text editor, open the ``watch.js`` file. 
+
+#. Our internal ``listener`` function is a callback function that takes two 
+   parameters: the collection of tasks and a ``changes`` dictionary that is sent
```
cbush, line 346: I would probably use "object" instead of "dictionary" in JS land.

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+   * **modifications** -- returns the index of the modified object.
+
+#. In the ``watchForChanges`` function, you will get all of the objects in 
+   the **Tasks** collection (just as we did in the ``getTasks`` function),  
+   and then call the :js-sdk:`addListener() <Realm.Collection.html#addListener>` 
+   method. Under the ``TODO`` comment, add the following code:
+
+.. code-block:: js
+   
+   const tasks = await realm.objects("Task");
```
cbush, line 362: No need to await.

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+2. Open a second terminal window and run the app again. In this second window, 
+   try creating, modifying, and deleting tasks. Each time you perform an action, 
+   the first terminal window will update with a notification. 
+
+.. note::
+   
+   If you do not want to open 2 terminal windows, you can modify the code in 
+   ``main.js``, within the ``case Choices.WatchForChanges`` handler. The 
+   inline code comment explains the change you can make.
```
cbush, line 378: I don't know. Why wouldn't someone want to open another terminal window?

source/tutorial/nodejs-cli.txt
andf-mongodb: Tiny thought: prompts the user for what? Guessing each function expects different input parameters. Maybe something like "Each ... prompts the user for input / a response / a necessary parameter", etc. Very minor.

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+E. Implement the CRUD methods
+-----------------------------
+In the ``tasks.js`` file, there are a number of functions to handle typical  
+CRUD functionality: ``getTasks``, ``getTask``, ``createTask``, ``deleteTask``, 
+``editTask``, and ``changeStatus``, which is essentially a specific case of 
+``editTask``. Each of these functions (except ``getTasks``) prompts the user, 
+and then makes the appropriate call to Realm. Your task is to implement the 
+calls to Realm. The following list shows each function and provides 
+guidance on how to complete the task.
```
andf-mongodb, line 226: Slight wordsmithing:

"The following list provides guidance on how to complete this task for each function:"

source/tutorial/nodejs-cli.txt
andf-mongodb: Hmm, good point here. I do think it could be OK to include this context: it's useful to know that, maybe, if I wanted to do similar elsewhere, but couldn't use `write()` for some reason, I now know I have options. Also, maybe users already know about these two transaction functions, but not that the `write()` function includes them. But, if so, I do think we need to slightly-more-strongly encourage use of `write()`. Maybe something like:

"While you could perform this task using the X and Y functions, the Z function handles these tasks automatically for us as part of its operation, so we'll use that here.

Of course, a perfectly valid alternative is not mentioning it at all, as per Bush's suggestion.

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+          _partition: "myPartition",
+          name: task.name,
+          status: task.status,
+        });
+     });
+
+
+* ``deleteTask``
+    
+  As with creating an object, you delete an object within a transaction. Again,
```
andf-mongodb, line 279: Previous point reference modifying an object, let's reuse that here for consistency (since you use "as with" to link back). Also struggled with "As with creating ... you delete". How about:

"Deleting objects must also take place within a transaction. As with modifying an object, we'll use the `write()` function to handle the transaction for us, and then call the the `delete()` function within it:"

Or similar. Still not quite happy with that rearrangement.

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+1. Open a terminal window and change to your app's directory.
+
+#. Run the following commands to install all of the dependencies and start the 
+   app:
+
+   .. code-block:: sh
+      
+      ``npm install``
+      ``node index.js``
```
andf-mongodb, line 323: No need for double-backticks while within `code-block`s

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+
+#. Our internal ``listener`` function is a callback function that takes two 
+   parameters: the collection of tasks and a ``changes`` dictionary that is sent 
+   from Realm. The ``changes`` dictionary has the following keys:
+
+   * **deletions** -- returns the index of the deleted item before the deletion.
+
+   * **insertions** -- returns the index of the new object.
+
+   * **modifications** -- returns the index of the modified object.
```
andf-mongodb, line 353: Each of these items is plural (insinuates containing multiple items), but described as containing only one item. Perhaps "returns the index of any deleted / modified objects / items"?

Probably minor.

source/tutorial/nodejs-cli.txt
MongoCaleb: I tested without the `await` and it failed.

source/tutorial/nodejs-cli.txt
MongoCaleb: I think it's still worth mentioning them, as its a valid approach, but I have demoted them to a note after the code.

source/tutorial/nodejs-cli.txt
MongoCaleb: The SDK uses "dictionary", but I agree for most js readers "object" is clearer.

source/tutorial/nodejs-cli.txt
MongoCaleb: apparently there is, based on my testing.

source/tutorial/nodejs-cli.txt
MongoCaleb: I guess the point is more "if you want to keep working and just have alerts pop up...", so I'll rewrite.

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+   tasks.addListener(listener);
+
+With this code update, you can now re-run the app and watch for changes. 
+
+1. In the main menu, choose "Watch for changes". You will see that the UI 
+   stays in a "waiting" state. 
+
+2. Open a second terminal window and run the app again. In this second window, 
+   try creating, modifying, and deleting tasks. Each time you perform an action, 
+   the first terminal window will update with a notification.
```
andf-mongodb, line 372: It might be nice to present one example create|modify|delete task that is copy and pasteable to showcase this behavior immediately, such that even a user who had not, at this point, completed the tutorial for those commands would be able to see the behavior of this watch feature in full.

source/tutorial/nodejs-cli.txt
cbush: It's kind of a legacy approach with its own problems. Not sure if we want to recommend it at all.

source/tutorial/nodejs-cli.txt
cbush: Failed how? This is not an async function. We need to dig into this, because something is wrong here.

source/tutorial/nodejs-cli.txt
```diff
@@ -0,0 +1,386 @@
...
+   GitHub. Just follow the instructions in ``README.md`` to get started. Don't
+   forget to update the ``config.js`` file with your App ID, which you can
+   find in the {+ui+}.
+
+Prerequisites
+-------------
+
+Before you begin, ensure you have:
+
+- `Node.js <https://nodejs.org/en/download/>`_ installed.
```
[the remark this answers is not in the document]
MongoCaleb: Weird! It renders correctly for me. @cbush can you confirm?

source/tutorial/nodejs-cli.txt
MongoCaleb: Yeah, that's an annoying part of the SDK. That's what they're called, but they fire with each individual change/addition/deletion.

source/tutorial/nodejs-cli.txt
MongoCaleb: Since it's all done through the UI, there's no code that can demonstrate this.

andf-mongodb: This section says: "in the second (terminal) window, try creating ..."

Terminal window = UI?

Oh! It's the ASCII art menu selection process! Ah I see, sorry, I understand now.

source/tutorial/nodejs-cli.txt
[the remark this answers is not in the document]
andf-mongodb: Looks good now: https://docs-mongodbcom-staging.corp.mongodb.com/realm/caleb/node_cli/tutorial/nodejs-cli.html#prerequisites

source/tutorial/nodejs-cli.txt
```diff
@@ -76,6 +75,16 @@ directly from GitHub:
...
+.. important::
+   
+   The ``main`` branch is a finished version of the app as it should look *after*
+   you complete this tutorial. To walk through this tutorial, please check 
+   out the ``todo`` branch:
+
+   .. code-block:: shell
+    
+      git checkout todo
+
```
andf-mongodb, line 87: Nice, good call.

source/tutorial/nodejs-cli.txt
```diff
@@ -273,12 +272,21 @@ guidance on how to complete the task.
         });
      });
 
+  .. note::
+    
+    The ``write`` function replaces the need to call the 
+    :js-sdk:`beginTransaction() <Realm.html#beginTransaction>`,  
+    :js-sdk:`commitTransaction() <Realm.html#commitTransaction>`, and 
+    :js-sdk:`cancelTransaction() <Realm.html#cancelTransaction>`
+    methods.
```
andf-mongodb, line 281: I think this, as a follow-up -- informational content outside the actionable path -- works well.

source/tutorial/nodejs-cli.txt
```diff
@@ -333,6 +341,11 @@ Once you have completed the code, you should run the app and check functionality
    the main "menu" of choices. All of the options should now work for you except 
    the "watch" functionality, which we'll enable in the next section.
 
+.. admonition:: Reminder
+   
+   If something isn't working for you, you can check out the ``main`` branch of 
+   this repo to compare your code with our finished solution. 
+
```
andf-mongodb, line 348: 👍

PR #370, opened 2020-07-27
add 'distinct' info

MongoCaleb:
Add info about using `distinct` in realm functions. 

[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/realm/caleb/distinct/mongodb/actions/collection.distinct.html#collection.distinct)

[Jira](https://jira.mongodb.org/browse/DOCSP-11076)

### DIFF
--- a/source/mongodb/actions/collection.distinct.txt
+++ b/source/mongodb/actions/collection.distinct.txt
@@ -1,0 +1,84 @@
+.. _mongodb-service-collection-distinct:
+
+=====================
+collection.distinct()
+=====================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Definition
+----------
+
+.. method:: collection.distinct()
+
+Return a list of distinct values for the given key within a collection or within 
+a find result set. 
+
+Usage
+-----
+
+Example
+~~~~~~~
+
+To call the ``collection.distinct()`` action from a :doc:`Function
+</functions>`, get a collection handle with
+:method:`database.collection()`, and then call the handle's
+``distinct()`` method, passing in the field from which you want to get distinct 
+values, and the query object for finding the documents. The following code 
+returns all unique values of the field "status" for all documents within the 
+"tasks" collection:
+
+.. code-block:: javascript
+   :emphasize-lines: 4-8
+   :linenos:
+
+   var taskCollection = context.services.get("mongodb-atlas")
+     .db("tracker").collection("tasks");
+
+   taskCollection.distinct("status", {})
+     .then(results => {
+         console.log(JSON.stringify(results));
+         console.log(results.length);
+     })
+     .catch(err => console.error(err))
+
+Parameters
+~~~~~~~~~~
+
+The ``collection.distinct()`` action has the following form:
+
+``distinct(key, query, session)``
+
+.. list-table::
+   :header-rows: 1
+
+   * - Parameter
+     - Description
+   * - ``key``
+     - The name of the field in the document in which to find distinct values.
+   * - ``query``
+     - A query for filtering the set of documents *before* the distinct filter 
+       is applied.
+   * - ``session``
+     - An optional `ClientSession <https://mongodb.github.io/node-mongodb-native/3.6/api/ClientSession.html>`_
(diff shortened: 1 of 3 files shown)


source/mongodb/actions/collection.distinct.txt
```diff
@@ -0,0 +1,84 @@
...
+   :header-rows: 1
+
+   * - Parameter
+     - Description
+   * - ``key``
+     - The name of the field in the document in which to find distinct values.
+   * - ``query``
+     - A query for filtering the set of documents *before* the distinct filter 
+       is applied.
+   * - ``session``
```
arahmanan, line 68: The third param is the `options` object, which contains the `session`. See https://mongodb.github.io/node-mongodb-native/3.6/api/Collection.html#distinct for some more details. That's the only option we support at the moment.

MongoCaleb: @arahmanan 
So does the call look like this?
`.distinct("foo", {}, {session:{id:123}}).then(...`

arahmanan: Yes, but you would never create the session like that. The session must be created by the `startSession` function.  Here's an example:

```
exports = async function(){
  const client = context.services.get("mongodb");
  const session = client.startSession();
  session.startTransaction();
  const coll = client.db('db1').collection('coll1');
  await coll.insertOne({_id: 4, z: "d", "group": "new"}, {session});
  await coll.insertMany([{_id: 5, z: "e", "group": "new"}, {_id: 6, z: "f", "group": "new"}], {session});
  // perform some other operations...

  const res = await coll.distinct('z', {}, {session});
  await session.commitTransaction();
  session.endSession();
  return res;
}
```

MongoCaleb: Solid! Thanks!

MongoCaleb: @arahmanan updated and restaged.

source/mongodb/actions/collection.distinct.txt
```diff
@@ -0,0 +1,111 @@
...
+
+.. list-table::
+   :header-rows: 1
+
+   * - Parameter
+     - Description
+   * - ``key``
+     - Required. The name of the field in the document in which to find distinct values.
+   * - ``query``
+     - Required. A query for filtering the set of documents *before* the distinct filter
```
arahmanan, line 92: This field is optional.

source/mongodb/actions/collection.distinct.txt
```diff
@@ -0,0 +1,111 @@
...
+``distinct()`` method, passing in the field from which you want to get distinct 
+values, and the query object for finding the documents. The following code 
+returns all unique values of the field "status" for all documents within the 
+"tasks" collection:
+
+.. code-block:: javascript
+   :emphasize-lines: 4-8
+   :linenos:
+
+   var taskCollection = context.services.get("mongodb-atlas")
```
mongomoe, line 41:
```suggestion
   const taskCollection = context.services.get("mongodb-atlas")
```

mongomoe: Const should be used here, since taskCollection is a variable that will not be changed. Var isn't used in our docs often anymore. Typically we adhere to es6 syntax, where we use "const" if a value is unchanged, and let if it will be changed (see: https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/#:~:text=Hoisting%20of%20const&text=var%20declarations%20are%20globally%20scoped%20or%20function%20scoped%20while%20let,be%20updated%20nor%20re%2Ddeclared.)

source/mongodb/actions/collection.distinct.txt
```diff
@@ -0,0 +1,111 @@
...
+
+.. code-block:: javascript
+   :emphasize-lines: 2, 7
+   :linenos:
+
+   const client = context.services.get("mongodb-atlas");
+   const session = client.startSession();
+   session.startTransaction(); 
+
+   var taskCollection = client.db("tracker").collection("tasks");
```
mongomoe, line 63:
```suggestion
   const taskCollection = client.db("tracker").collection("tasks");
```

MongoCaleb:
@mongomoe back to you

PR #378, opened 2020-07-29
(DOCSP-13680): Android quick-start CRUD operations and Notifications

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCS-13780

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/DOCSP-13780/android/quick-start.html

### DIFF
--- a/source/android/code-samples/quick-start.kt
+++ b/source/android/code-samples/quick-start.kt
@@ -1,0 +1,135 @@
+package com.mongodb.realm.quickstart
+
+import io.realm.RealmObject
+import io.realm.annotations.PrimaryKey
+import io.realm.annotations.Required
+import org.bson.types.ObjectId
+
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import android.util.Log
+import io.realm.OrderedRealmCollectionChangeListener
+
+import io.realm.Realm
+import io.realm.RealmResults
+import io.realm.kotlin.where
+import io.realm.mongodb.App
+import io.realm.mongodb.AppConfiguration
+
+import io.realm.mongodb.Credentials
+import io.realm.mongodb.User
+import io.realm.mongodb.sync.SyncConfiguration
+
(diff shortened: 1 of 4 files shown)


source/android/code-samples/quick-start.kt
```diff
@@ -0,0 +1,135 @@
...
+                val tasks : RealmResults<Task> = realm.where<Task>().findAll()
+
+                // you can also filter a collection
+                val tasksThatBeginWithN : List<Task> = tasks.where().beginsWith("name", "N").findAll()
+                val openTasks : List<Task> = tasks.where().equalTo("status", TaskStatus.Open.name).findAll()
+
+                val otherTask: Task = tasks[0]!!
+
+                tasks.addChangeListener(OrderedRealmCollectionChangeListener<RealmResults<Task>> { collection, changeSet ->
+                    // if keeping parallel arrays in sync, handle deletions in reverse to avoid changing indices after removals
```
mongomoe, line 63: This is a bit of a confusing sentence, can you rephrase it to be simpler?

How about: "If parallel arrays are kept in sync, deletions are handled in reverse. This avoids changing the index after the task is removed".


Alternatively, since this is a fairly important comment, maybe it should be an admonition. That way you don't have to be so careful on the word count (compared to an inline code-comment) and can thoroughly explain the sentence.

nathan-contino-mongo: This one is a little tricky to me for a couple of reasons:
1) There isn't an easy way to reverse iterate this in java (that I can find)
2) Reverse iteration is only required in VERY select circumstances

I left it in there because I wanted to demonstrate that you can reverse iterate for cases where you're trying to match notifications in Realm to your own data changes (like implementing your own RecyclerView, where you need to remove items in a list when a deletion occurs)... but I'm wondering if it might just be confusing for end users.

PR #382, opened 2020-07-31
(DOCSP-11256): Access Custom User Data Node/RN

mongomoe:
Jira:
-----
* https://jira.mongodb.org/browse/DOCSP-11256

Stage:
-------
* Node: https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-11256-AccessCustomUserData-NodeRN/node/access-custom-user-data.html
* React Native: https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-11256-AccessCustomUserData-NodeRN/node/access-custom-user-data.html
* Link to this page from the "Enable Custom User Data" (formerly: "Define Custom Data") page:  https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-11256-AccessCustomUserData-NodeRN/users/enable-custom-user-data.html

### DIFF
--- a/source/node/access-custom-user-data.txt
+++ b/source/node/access-custom-user-data.txt
@@ -1,0 +1,171 @@
+.. _node-access-custom-user-data:
+
+=======================
+Access Custom User Data
+=======================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+You can read arbitrary data about your application users, known as
+custom user data, directly within your Node application. For example,
+you might store a user's preferred language, date of birth, or local
+timezone. You can write to the Custom User Data object via a
+:ref:`RemoteMongoClient <node-mongodb-data-access>`. 
+
+
+.. note::
+
+   You must enable Custom User Data within the {+service+} UI.  To learn
+   how to enable it, read the :doc:`Define Custom User Data
+   </users/define-custom-user-data>` doc.
+
+
+
+The code examples in this page use the following :ref:`user object
+<user-objects>` in which the custom_data field has not been set yet.
+
+.. code-block:: console
+
+   {
+     id: '5f1f216e82df4a7979f9da93',
+     type: 'normal',
+     custom_data: {
+       _id: '5f20d083a37057d55edbdd57',
+       userID: '5f1f216e82df4a7979f9da93',
+       description: 'a test document for user: test@test.com',
+       custom_data: { }
+     },
+     data: { email: 'test@test.com' },
+     identities: [
+       { id: '5f1f216e82df4a7979f9da90', provider_type: 'local-userpass' }
+     ]
+   }
+
+
+Write to Custom User Data with a RemoteMongoClient
+--------------------------------------------------
+Using standard CRUD operations through the :doc:`MongoDB Atlas service
+</mongodb>`, a user's custom data can be accessed. The following example
+updates the user's custom data to alter the user's ``favoriteColor`` to
+pink.
+
+.. code-block:: javascript
+   :emphasize-lines: 3, 16
+
+   // a user must be logged in to use a remoteMongoClient
+   const user = await app.logIn(credentials);
+   const mongo = user.remoteMongoClient("<atlas service name>");
+   const collection = mongo.db("<database name>").collection("<collection name>");
+
+   const filter = {
+     userID: user.id, // Query for the user object of the logged in user
+   };
+
+   const updateDoc = {
+     $set: {
+       favoriteColor: "pink", // set the logged in user's favorite color to pink
+     },
+   };
+
+   const result = await collection.updateOne(filter, updateDoc);
+   console.log(result);
+
+Output
+~~~~~~
+.. code-block:: console
+
+   { matchedCount: 1, modifiedCount: 1 }
+
+Read Custom User Data
+---------------------
+If you have updated your custom user data within the last 30 minutes,
+use the :js-sdk:`refreshCustomData()
+<Realm.User.html#refreshCustomData>` method on the user object.
+
+.. code-block:: javascript
+
+   const customUserData = await user.refreshCustomData()
+   console.log(customUserData);
+
+If you have not recently updated your custom user data, use the user object's :js-sdk:`customData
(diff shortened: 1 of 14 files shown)


source/node/access-custom-user-data.txt
```diff
@@ -0,0 +1,171 @@
...
+you might store a user's preferred language, date of birth, or local
+timezone. You can write to the Custom User Data object via a
+:ref:`RemoteMongoClient <node-mongodb-data-access>`. 
+
+
+.. note::
+
+   You must enable Custom User Data within the {+service+} UI.  To learn
+   how to enable it, read the :doc:`Define Custom User Data
+   </users/define-custom-user-data>` doc.
```
cbush, line 29: "Doc" is a bit informal. Why not bake the link into what you're trying to say:

>To use custom user data, you must :doc:`enable custom user data </users/define-custom-user-data>` first.

source/node/access-custom-user-data.txt
```diff
@@ -0,0 +1,171 @@
...
+
+
+
+The code examples in this page use the following :ref:`user object
+<user-objects>` in which the custom_data field has not been set yet.
```
cbush, line 34: That reads a bit awkwardly. The field is "on" the user object, not "in". How about:
```suggestion
<user-objects>` on which the custom_data field has not yet been set.
```

source/node/access-custom-user-data.txt
```diff
@@ -0,0 +1,171 @@
...
+.. code-block:: console
+
+   {
+     id: '5f1f216e82df4a7979f9da93',
+     type: 'normal',
+     custom_data: {
+       _id: '5f20d083a37057d55edbdd57',
+       userID: '5f1f216e82df4a7979f9da93',
+       description: 'a test document for user: test@test.com',
+       custom_data: { }
```
cbush, line 45: Why is there custom_data within the custom_data? This is confusing.

source/node/access-custom-user-data.txt
```diff
@@ -0,0 +1,171 @@
...
+     identities: [
+       { id: '5f1f216e82df4a7979f9da90', provider_type: 'local-userpass' }
+     ]
+   }
+
+
+Write to Custom User Data with a RemoteMongoClient
+--------------------------------------------------
+Using standard CRUD operations through the :doc:`MongoDB Atlas service
+</mongodb>`, a user's custom data can be accessed. The following example
```
cbush, line 57: Blocking question: Does this only work if the user has permission to access the custom user data collection? If so, we should specify that.

source/node/access-custom-user-data.txt
```diff
@@ -0,0 +1,171 @@
...
+
+Output
+~~~~~~
+.. code-block:: console
+
+   { matchedCount: 1, modifiedCount: 1 }
+
+Read Custom User Data
+---------------------
+If you have updated your custom user data within the last 30 minutes,
```
cbush, line 90: I feel like this sentence could be rewritten. It's like you're hinting at information without being explicit. Why 30 minutes? _Only_ if I have updated recently? Let's try to rethink what information you're actually trying to convey here.

source/node/access-custom-user-data.txt
```diff
@@ -0,0 +1,171 @@
...
+       const result = await collection.updateOne(filter, updateDoc);
+       console.log(result);
+ 
+       const customUserData = await user.refreshCustomData();
+       console.log(customUserData);
+     } finally {
+       user.logOut();
+     }
+   }
+   run().catch(console.dir);
```
cbush, line 171: Perhaps we could use a Summary section?

source/node/access-custom-user-data.txt
mongomoe: 👍

source/node/access-custom-user-data.txt
mongomoe: removed

source/node/access-custom-user-data.txt
mongomoe: It does require custom user data to be enabled. There was a note just under the overview, I've moved it to this section now.

source/node/access-custom-user-data.txt
mongomoe: There's actually a note that explains it in the Enable Custom User Data (formerly Define Custom User Data) doc. See: https://docs.mongodb.com/realm/users/define-custom-user-data/ (admonition - "CUSTOM DATA MAY BE STALE")

Given that it's explained in that doc maybe we could do one of the following:
a. Copy and paste that admonition
b. Rephrase this sentence to explain the contents of that admonition
c. something else?

What do you think?

source/node/access-custom-user-data.txt
mongomoe: 👍

source/node/access-custom-user-data.txt
cbush: I think either of those options is good. (Rather than copy and paste, I would move the admonition to an include and include it in both places).

source/node/access-custom-user-data.txt
cbush: The question was not about whether custom data needs to be enabled. It was whether the user needs access to the actual custom data collection according to the Rules of the app (they do). This information needs to be included.

source/node/access-custom-user-data.txt
mongomoe: updated to an includes

source/node/access-custom-user-data.txt
mongomoe: Gotcha, sorry I misunderstood. So I've adjusted the old **note** that said something like 'you need to enable custom user data' to the following:

**"To use custom user data, you must Enable Custom User Data and configure your rules to give users access to the custom data collection."**

with links to the appropriate docs:

https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-11256-AccessCustomUserData-NodeRN/users/enable-custom-user-data.html

and

https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-11256-AccessCustomUserData-NodeRN/mongodb/configure-advanced-rules.html

PR #398, opened 2020-08-07
Add auth methods for iOS

MongoCaleb:
[Jira](https://jira.mongodb.org/browse/DOCSP-11178)

[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/11178/ios/authenticate.html)

### DIFF
--- a/source/ios/authenticate.txt
+++ b/source/ios/authenticate.txt
@@ -15,23 +15,67 @@ Authenticate a User
 Overview
 --------
 
+{+service+} provides an API for authenticating users using any enabled
+authentication provider. Instantiate a ``AppCredentials`` object and pass
+it to the  ``app.login()`` method to authenticate a user and create a ``User`` 
+object. Each authentication provider corresponds to a method used to
+instantiate ``AppCredentials`` objects using that authentication provider.
+See the table below to find the method that instantiates the
+``AppCredentials`` instance for your authentication provider:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 50 50
+
+   * - Authentication Provider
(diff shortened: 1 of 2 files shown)


source/ios/authenticate.txt
```diff
@@ -67,6 +116,148 @@ you can log in:
...
+API Key
+~~~~~~~
+If you have enabled :ref:`API Key authentication <api-key-authentication>`,
+you can log in using the following code:
+
+
+.. code-block:: swift
+   :emphasize-lines: 1-3
+
+   let apiKey = "c8t0numjwVR9WyJS1k4YMxE7elRFDOKv7fCGi5NuJfXhonTUQB5iWYxnLotzAAlF"
```
cbush, line 130: Should this be left in here?

source/ios/authenticate.txt
```diff
@@ -67,6 +116,148 @@ you can log in:
...
+Custom Function
+~~~~~~~~~~~~~~~
+If you have enabled the 
+:doc:`Custom Function authentication provider </authentication/custom-function>`,
+you can log in using the following code:
+
+.. code-block:: swift
+   :emphasize-lines: 1-7
+
+   let params: [AnyHashable: Any] = [
```
cbush, line 153: Prefer type inference:
```suggestion
   let params = [
```

source/ios/authenticate.txt
MongoCaleb: It's fake, so it's safe to keep in there, but I suppose replacing it with a placeholder makes ore sense.

PR #412, opened 2020-08-18
11085 - add info on background refresh

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-11085

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/11085/ios/sync-data.html#background-data-sync

### DIFF
--- a/source/ios/sync-data.txt
+++ b/source/ios/sync-data.txt
@@ -71,8 +71,9 @@ Sync Data
 
 The syntax to :ref:`read <ios-realm-database-reads>`, :ref:`write
 <ios-realm-database-writes>`, and :ref:`watch for changes <ios-client-notifications>` on a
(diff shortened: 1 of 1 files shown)


source/ios/sync-data.txt
```diff
@@ -107,9 +108,22 @@ changes, then writes a new ``Task`` to the {+realm+}:
 
 .. seealso:: :ref:`Threading <ios-client-threading>`
 
+
+Background Data Sync
+~~~~~~~~~~~~~~~~~~~~
+If you want your app to update data in the background, while the app is minimized,
```
nathan-contino-mongo, line 114:
```suggestion
If you want your app to update data in the background while the app is minimized, 
```
I don't think this comma is necessary. That being said, it's not a big deal at all.

PR #418, opened 2020-08-19
(DOCSP-11871): Remove 'Remotely Access MongoDB' [Web, Node, RN]

mongomoe:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-11871

https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/docsp-11875-remove-access-mongodb-remotely/


config/redirects
```diff
@@ -2,3 +2,7 @@ define: base https://docs.mongodb.com/realm
 
 raw: realm/tutorials -> ${base}/tutorial
 raw: realm/deploy/application-schema -> ${base}/deploy/application-configuration-files
+
+raw: realm/node/remotely-access-mongodb -> ${base}/node/mongodb
```
nathan-contino-mongo, line 6: Question: do we really need these redirects? Certainly makes sense from a CYA perspective, do we know of any significant external links into these pages that would require a redirect like this? Admittedly this is something I haven't bothered with myself when changing page names etc.

mongomoe: I don't imagine anyones linking to it externally but this is more of a "just in case anyone has this page bookmarked"

PR #420, opened 2020-08-20
(DOCSP-8291): Identify common Node API version for functions modules

cbush:
## Pull Request Info
This adds an admonition specifying the Node API version and adds links to the corresponding Node API docs.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-8291

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/node-version/functions/upload-external-dependencies.html#built-in-module-support

### DIFF
--- a/source/functions/upload-external-dependencies.txt
+++ b/source/functions/upload-external-dependencies.txt
@@ -36,47 +36,54 @@ the {+ui+} under the :guilabel:`Dependencies` tab.
 Built-In Module Support
 -----------------------
 
-.. list-table::
-   :header-rows: 1
-   :widths: 50 50
+You can use the following built-in modules in your {+service-short+}
+functions:
 
-   * - Supported
-     - Unsupported
+.. note::
+   
+   The supported modules are compatible with :nodejs:`Node API version
+   10.18.1 <docs/v10.18.1/api/>`. Avoid using APIs in these modules
+   introduced after or deprecated since Node 10.18.1.
 
-   * - .. hlist::
-          :columns: 2
-          
-          - ``assert``
-          - ``buffer``
-          - ``crypto``
-          - ``events``
-          - ``fs``
-          - ``http``
-          - ``https``
-          - ``net``
-          - ``os``
-          - ``path``
-          - ``querystring``
-          - ``stream``
-          - ``string_decoder``
-          - ``timers``
-          - ``tls``
-          - ``tty``
-          - ``url``
-          - ``util``
-          - ``zlib``
-     - .. hlist::
-          :columns: 1
-          
-          - ``child_process``
-          - ``cluster``
-          - ``dgram``
-          - ``dns``
-          - ``domain``
-          - ``punycode``
-          - ``readline``
-          - ``v8``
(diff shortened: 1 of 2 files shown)


source/functions/upload-external-dependencies.txt
```diff
@@ -36,47 +36,54 @@ the {+ui+} under the :guilabel:`Dependencies` tab.
...
+   - :nodejs:`buffer <docs/v10.18.1/api/buffer.html>`
+   - :nodejs:`crypto <docs/v10.18.1/api/crypto.html>`
+   - :nodejs:`events <docs/v10.18.1/api/events.html>`
+   - :nodejs:`fs <docs/v10.18.1/api/fs.html>`
+   - :nodejs:`http <docs/v10.18.1/api/http.html>`
+   - :nodejs:`https <docs/v10.18.1/api/https.html>`
+   - :nodejs:`net <docs/v10.18.1/api/net.html>`
+   - :nodejs:`os <docs/v10.18.1/api/os.html>`
+   - :nodejs:`path <docs/v10.18.1/api/path.html>`
+   - process (supports :nodejs:`nextTick <docs/latest-v10.x/api/process.html#process_process_nexttick_callback_args>` only)
```
tsedgwick, line 61: process we added support for:
hrTime
env
argv
version
versions
stderr
sdin

source/functions/upload-external-dependencies.txt
```diff
@@ -36,47 +36,54 @@ the {+ui+} under the :guilabel:`Dependencies` tab.
...
+   
+   - ``child_process``
+   - ``cluster``
+   - ``dgram``
+   - ``dns``
+   - ``domain``
+   - ``punycode``
+   - ``readline``
+   - ``v8``
+   - ``vm``
```
tsedgwick, line 86: dns - we will support in upcoming release 

domain - might be worth calling out node urges users not to use this as it's deprecated:
from https://nodejs.org/docs/latest-v10.x/api/domain.html: 
```
This module is pending deprecation. Once a replacement API has been finalized, this module will be fully deprecated. Most end users should not have cause to use this module. Users who absolutely must have the functionality that domains provide may rely on it for the time being but should expect to have to migrate to a different solution in the future.
```

dgram: we're releasing a forked version of  "browser-node-dgram": "^1.0.4",

source/functions/upload-external-dependencies.txt
cbush: Interesting. Do stderr/stdin go to Logs? And what can one use env or argv for?

tsedgwick: under the covers, they are just set to objects. on second thought of docs, we might not want to add them to the docs since it won't operate 1:1 with how users will expect


I would say you could keep hrTime and version(s)

source/functions/upload-external-dependencies.txt
cbush: Are those releases coming soon? I'm guessing we can raise a new DOCSP at that point and we'll update this page again.

PR #435, opened 2020-09-02
DOCSP-11664: Clarify Custom Function Auth Example

llazala000:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-11664

I'm not able to stage at the moment (still resolving AWS access key request).

### DIFF
--- a/source/includes/steps-configure-custom-function-auth-realm-ui.yaml
+++ b/source/includes/steps-configure-custom-function-auth-realm-ui.yaml
@@ -33,16 +33,16 @@ content: |
   existing Realm user and automatically creates a new user if it does
   not match an existing user. If the function does not return a string,
   Realm throws an error and does not create or authenticate a user.
-  
+
   .. admonition:: Realm Generates New User ID Values
      :class: important
-     
+
      The ID value that you return from the authentication function is
      **not** the internal Realm user id (i.e. the value that
      :json-expansion:`%%user` and :ref:`context.user <context-user>`
      resolve to). Realm automatically generates a unique id for Custom
      Function users when it creates them.
-  
+
   To define a new authentication function, click the
   :guilabel:`Function` dropdown and select :guilabel:`New Function`.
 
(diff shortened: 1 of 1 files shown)


source/includes/steps-configure-custom-function-auth-realm-ui.yaml
```diff
@@ -51,36 +51,42 @@ content: |
...
      collection for an existing user with the specified username. If the
      user already exists, the function returns their stored ``id`` value. If
      the user does not exist, the function stores a new user document in
      the collection and returns that document's ``id`` value.
-     
+
+     .. note::
+       The collection ``app.users`` is not created by Realm. You can use
+       your custom function with an external authentication system and/or
+       use data that you store in a MongoDB collection to identify and
```
nathan-contino-mongo, line 71:
```suggestion
       data that you store in a MongoDB collection to identify and
```
no need for an extra verb here!

source/includes/steps-configure-custom-function-auth-realm-ui.yaml
```diff
@@ -33,16 +33,16 @@ content: |
   existing Realm user and automatically creates a new user if it does
   not match an existing user. If the function does not return a string,
   Realm throws an error and does not create or authenticate a user.
-  
+
   .. admonition:: Realm Generates New User ID Values
      :class: important
-     
+
```
nathan-contino-mongo, line 39: it always makes me happy to see whitespace trimmed

source/includes/steps-configure-custom-function-auth-realm-ui.yaml
```diff
@@ -51,36 +51,42 @@ content: |
...
+     The application's authentication function queries the ``app.users``
      collection for an existing user with the specified username. If the
      user already exists, the function returns their stored ``id`` value. If
      the user does not exist, the function stores a new user document in
      the collection and returns that document's ``id`` value.
-     
+
+     .. note::
+       The collection ``app.users`` is not created by Realm. You can use
```
nathan-contino-mongo, line 69: nit: avoid passive voice ("is not created") when possible. Active voice, in comparison, helps users feel more engaged with text because the objects you describe actually "act" on other objects instead of just potentially acting.
If you decide you'd like to reword this, you could use something like this (feel free to customize this as much as you'd like; this is just a suggestion!):
```suggestion
       {+service-short+} does not automatically create an ``app.users`` collection for your application. You can use
```

nathan-contino-mongo: PS: after we renamed "Stitch" into "Realm" (and realized that "Realm" also includes the concept of "realms"), we decided that we should save ourselves some work if the product ever gets renamed again, so we now use [source constants](https://docs.mongodb.com/meta/reference/source-constants/) to replace the a few key terms in the documentation. You can see the full list of replacements [here](https://github.com/mongodb/docs-realm/blob/master/snooty.toml#L4) and [here](https://github.com/mongodb/docs-realm/blob/master/conf.py#L72) (and yes, you do have to update both places to manually keep them in sync -- I know it sucks, but it's an artifact of our transition from oldgen to nextgen).

source/includes/steps-configure-custom-function-auth-realm-ui.yaml
nathan-contino-mongo: Suggestion: I'm not sure where might be the best place for this, but [custom user data](https://docs.mongodb.com/realm/users/enable-custom-user-data/) is a pretty natural way to manage per-user data like this. Consider adding a link if you can find a way to bring it up organically!

PR #445, opened 2020-09-09
(DOCSP-11835): Clarify Partition Key handling on Client-side vs. Server-side

llazala000:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-11835

https://docs-mongodbcom-staging.corp.mongodb.com/realm/lisbethlazala/clarify-partition-key-handling/sync/rules.html
https://docs-mongodbcom-staging.corp.mongodb.com/realm/lisbethlazala/clarify-partition-key-handling/sync/partitioning.html#choose-a-partition-key

### DIFF
--- a/source/get-started/glossary.txt
+++ b/source/get-started/glossary.txt
@@ -438,3 +438,7 @@ Glossary
    relationship
      A relationship is an object property that references another object rather
      than a scalar or embedded object.
+     
+   null partition
+      A null partition is the default :term:`{+service-short+}` on the client 
+      device.
--- a/source/get-started/glossary.txt
+++ b/source/get-started/glossary.txt
@@ -440,5 +440,5 @@ Glossary
      than a scalar or embedded object.
      
    null partition
-      A null partition is the default :term:`{+service-short+}` on the client 
+      A null partition is the default :term:`{+realm+}` on the client 
       device.
--- a/source/includes/steps-define-sync-rules-ui.yaml
+++ b/source/includes/steps-define-sync-rules-ui.yaml
@@ -35,12 +35,13 @@ content: |
   .. note::
     
     The partition key can either be required or optional. If the partition key 
-    field is optional, then all documents that either exclude the partition 
-    key or have a null value for the partition key will be sent to the 
+    field is optional, then all MongoDB documents that either exclude the 
+    partition key or have a null value for the partition key will be sent to the 
     :term:`null partition`. If the partition key field is required, {+sync+} 
-    will ignore any documents that lack a valid value for 
-    the partition key. We recommend a required partition key unless your 
-    application specifically needs access to unpartitioned documents. 
+    ignores any MongoDB documents that lack a valid value for the partition 
+    key. We recommend a required partition key unless you want to sync
+    pre-existing data with invalid or missing partition values from a
+    MongoDB collection.
 
   Once you have decided which field to use, select it from the :guilabel:`Choose
   A Partition Key` dropdown menu.
--- a/source/includes/steps-define-sync-rules-ui.yaml
+++ b/source/includes/steps-define-sync-rules-ui.yaml
@@ -27,9 +27,10 @@ title: Choose a Partition Key
 ref: choose-a-partition-key
 content: |
   The sync :term:`partition key` is a field in every synced document that maps
-  each document to a client-side realm. If the partition key field is optional, 
-  then all documents that either don't contain the partition key or have a null 
-  value for the partition key will be sent to the null partition by default. 
+  each document to a client-side realm. Partition key fields should be made 
+  required through the {+Realm+} UI. If the partition key field is optional, 
+  then all documents that either exclude the partition key or have a null 
+  value for the partition key will be sent to the null partition. 
   Sync rules apply at the partition level, so it's particularly important to 
   consider your data model and access patterns. For more information on 
   partition keys and how to choose one, see
--- a/source/includes/steps-define-sync-rules-ui.yaml
+++ b/source/includes/steps-define-sync-rules-ui.yaml
@@ -27,9 +27,12 @@ title: Choose a Partition Key
 ref: choose-a-partition-key
 content: |
   The sync :term:`partition key` is a field in every synced document that maps
-  each document to a client-side realm. Sync rules apply at the partition level,
-  so it's particularly important to consider your data model and access
-  patterns. For more information on partition keys and how to choose one, see
+  each document to a client-side realm. If the partition key field is optional, 
+  then all documents that either don't contain the partition key or have a null 
+  value for the partition key will be sent to the null partition by default. 
+  Sync rules apply at the partition level, so it's particularly important to 
+  consider your data model and access patterns. For more information on 
+  partition keys and how to choose one, see
   :ref:`Partition Atlas Data into Realms <partitioning>`.
 
   Once you have decided which field to use, select it from the :guilabel:`Choose
--- a/source/includes/steps-define-sync-rules-ui.yaml
+++ b/source/includes/steps-define-sync-rules-ui.yaml
@@ -27,14 +27,18 @@ title: Choose a Partition Key
 ref: choose-a-partition-key
 content: |
   The sync :term:`partition key` is a field in every synced document that maps
-  each document to a client-side realm. Partition key fields should be made 
-  required through the {+Realm+} UI. If the partition key field is optional, 
-  then all documents that either exclude the partition key or have a null 
-  value for the partition key will be sent to the null partition. 
-  Sync rules apply at the partition level, so it's particularly important to 
-  consider your data model and access patterns. For more information on 
-  partition keys and how to choose one, see
+  each document to a client-side realm. Sync rules apply at the partition 
+  level, so it's particularly important to consider your data model and access 
+  patterns. For more information on partition keys and how to choose one, see
   :ref:`Partition Atlas Data into Realms <partitioning>`.
+  
+  .. note::
+    
+    Partition key fields should be made required. If the partition key field is 
+    optional, then all documents that either exclude the partition key or have 
+    a null value for the partition key will be sent to the 
+    :term:`null partition`. If the partition key field is required, {+sync+} 
+    will ignore any documents that lack a valid value for the partition key. 
 
   Once you have decided which field to use, select it from the :guilabel:`Choose
   A Partition Key` dropdown menu.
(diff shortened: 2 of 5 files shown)


source/sync/partitioning.txt
```diff
@@ -79,6 +79,10 @@ choose partition keys:
 - Partition keys must have a value of one of the
   :ref:`supported types for partition keys
   <partition-limitations-types>`.
+  
+- If the partition key field is designated as optional, all documents that 
+  don't contain the partition key or have a null value for the partition key 
+  will be sent to the null partition by default.
```
ccho-mongodb, line 85: Nitpick: when qualifying with "by default", I might expect to also see a case in which the behavior could be overridden. Is that the case here? Otherwise, you might consider omitting "by default".

ccho-mongodb (approved):
LGTM, a couple nitpicks.

source/sync/partitioning.txt
```diff
@@ -79,6 +79,11 @@ choose partition keys:
 - Partition keys must have a value of one of the
   :ref:`supported types for partition keys
   <partition-limitations-types>`.
+  
+- Partition key fields should be made required through the {+Realm+} UI. If 
+  the partition key field is designated as optional, all documents that 
+  exclude the partition key or have a null value for the partition key will be 
+  sent to the null partition.
```
nlarew, line 86: A few nitpicks here:

- > Partition key fields should be made required...

  In general our style prefers [active voice](https://docs.mongodb.com/meta/style-guide/writing/use-active-voice/), so let's rephrase a bit to answer "made required _by whom?_"

  ```rst
  You can configure your app to treat partition keys as either optional or required.
  ```

- We should also describe what happens to documents that exclude or have a null partition key when partition keys are required

- A reader might ask "what is the null partition?" Can we either clear that up here or add a glossary entry (in `source/get-started/glossary.txt`) and reference it here (`` :term:`null partition` ``). e.g. it's the default realm on the client device

source/includes/steps-define-sync-rules-ui.yaml
```diff
@@ -27,9 +27,13 @@ title: Choose a Partition Key
...
-  each document to a client-side realm. Sync rules apply at the partition level,
-  so it's particularly important to consider your data model and access
-  patterns. For more information on partition keys and how to choose one, see
+  each document to a client-side realm. Partition key fields should be made 
+  required through the {+Realm+} UI. If the partition key field is optional, 
+  then all documents that either exclude the partition key or have a null 
+  value for the partition key will be sent to the null partition. 
+  Sync rules apply at the partition level, so it's particularly important to 
+  consider your data model and access patterns. For more information on 
+  partition keys and how to choose one, see
```
nlarew, line 36: Can we make it a bit clearer that "should" here is just our default/general suggestion? Maybe we could move this out of the paragraph and into a `.. note::` right below the paragraph.

nlarew (changes requested):
A few nitpicks on the partitioning page + let's mirror this to the other versions of the Define Sync Rule procedure

source/includes/steps-define-sync-rules-ui.yaml
llazala000: Should I move just the general suggestion out to a note? Or should I move the description of the partition key handling out too?

nlarew: Not totally sure what you mean by "partition key handling" but I was referring just to the new text, i.e.

"Partition key fields should be made required through the {+Realm+} UI. If the partition key field is optional, then all documents that either exclude the partition key or have a null value for the partition key will be sent to the null partition."

llazala000: Gotcha

source/includes/steps-define-sync-rules-ui.yaml
```diff
@@ -27,10 +27,18 @@ title: Choose a Partition Key
...
-  each document to a client-side realm. Sync rules apply at the partition level,
-  so it's particularly important to consider your data model and access
+  each document to a client-side realm. Sync rules apply at the partition 
+  level, so it's particularly important to consider your data model and access 
   patterns. For more information on partition keys and how to choose one, see
   :ref:`Partition Atlas Data into Realms <partitioning>`.
+  
+  .. note::
+    
+    Partition key fields should be made required. If the partition key field is
```
nlarew, line 37: This is passive voice - partition key fields should be made required _by whom?_ Let's make it a bit more active! 

Also, I think we can (1) more explicitly call out that this is our recommendation (2) give a bit more context on why we recommend this (3) move the recommendation to the end of the note so that readers have some context. Thoughts?

"The partition key field can either be required or optional. [If optional... If required...] We recommend a required partition key unless your application specifically needs access to unpartitioned documents."

source/sync/partitioning.txt
```diff
@@ -79,6 +79,12 @@ choose partition keys:
 - Partition keys must have a value of one of the
   :ref:`supported types for partition keys
   <partition-limitations-types>`.
+  
+- You can configure your app to treat partition keys as either optional or 
+  required. If the partition key field is designated as optional, all documents 
+  that exclude the partition key or have a null value for the partition key 
+  will be sent to the :term:`null partition`. If the partition key field is 
+  required,
```
nlarew, line 87: Looks like we're missing the end of the sentence here

source/includes/steps-define-sync-rules-ui.yaml
```diff
@@ -27,10 +27,20 @@ title: Choose a Partition Key
...
+  
+  .. note::
+    
+    The partition key can either be required or optional. If the partition key 
+    field is optional, then all documents that either exclude the partition 
+    key or have a null value for the partition key will be sent to the 
+    :term:`null partition`. If the partition key field is required, {+sync+} 
+    will ignore any documents that lack a valid value for 
+    the partition key. We recommend a required partition key unless your 
+    application specifically needs access to unpartitioned documents.
```
nlarew, line 43: Can we add this note to the step files for the other interfaces (i.e. CLI & API)?

- source/includes/steps-define-sync-rules-cli.yaml
- source/includes/steps-define-sync-rules-api.yaml

nlarew (approved):
LGTM % one last update to include

Let's get a review from Ian on this too

source/includes/steps-define-sync-rules-ui.yaml
ianpward, line 43: > We recommend a required partition key unless your 
    application specifically needs access to unpartitioned documents. 

I'm not sure I would use this phrasing - since any documents will be partitioned just in the nil partition. The use case for nil partitions is more for the case where a user has pre-existing data in their MongoDB and wants to sync data on documents that may not have that field in it or may not have a valid value in it. Asking them to refactor and change their data might be a big ask

source/get-started/glossary.txt
```diff
@@ -438,3 +438,7 @@ Glossary
    relationship
      A relationship is an object property that references another object rather
      than a scalar or embedded object.
+     
+   null partition
+      A null partition is the default :term:`{+service-short+}` on the client
```
nathan-contino-mongo, line 443:
```suggestion
      A null partition is the default :term:`{+realm+}` on the client 
```

We actually use the substitution realm for the word realm just in case they change the name in the future :)

nathan-contino-mongo: service-short gives you capital R realm, which is the service itself. Far too many things are named Realm here!

source/includes/steps-define-sync-rules-ui.yaml
```diff
@@ -27,10 +27,20 @@ title: Choose a Partition Key
...
   :ref:`Partition Atlas Data into Realms <partitioning>`.
+  
+  .. note::
+    
+    The partition key can either be required or optional. If the partition key 
+    field is optional, then all documents that either exclude the partition 
+    key or have a null value for the partition key will be sent to the 
+    :term:`null partition`. If the partition key field is required, {+sync+} 
+    will ignore any documents that lack a valid value for 
+    the partition key. We recommend a required partition key unless your
```
nathan-contino-mongo, line 42: I think we need to clarify that this refers to documents in MongoDB, since some users might not recognize that Realm uses objects and MongoDB uses documents (even I did a short double-take here to make sure I understood). Suggestion: use the term "MongoDB documents" throughout this note?

source/sync/partitioning.txt
```diff
@@ -79,6 +79,14 @@ choose partition keys:
 - Partition keys must have a value of one of the
   :ref:`supported types for partition keys
   <partition-limitations-types>`.
+  
+- You can configure your app to treat partition keys as either optional or 
+  required. If the partition key field is designated as optional, all documents 
+  that exclude the partition key or have a null value for the partition key 
+  will be sent to the :term:`null partition`. If the partition key field is 
+  required, {+sync+} will ignore any documents that lack a valid value for
```
nathan-contino-mongo, line 87:
```suggestion
  required, {+sync+} ignores any MongoDB documents that lack a valid value for 
```
Both here and in the above usage of "document" I think it would help users to use the term "MongoDB documents" so they don't confuse this with client-side behavior. Also, prefer active voice when possible (which I know can be a huge pain when we're writing about theoretical scenarios like "if x is y, z..")

source/sync/partitioning.txt
```diff
@@ -79,6 +79,14 @@ choose partition keys:
 - Partition keys must have a value of one of the
   :ref:`supported types for partition keys
   <partition-limitations-types>`.
+  
+- You can configure your app to treat partition keys as either optional or 
+  required. If the partition key field is designated as optional, all documents
```
nathan-contino-mongo, line 84: documents -> MongoDB documents

source/includes/steps-define-sync-rules-ui.yaml
```diff
@@ -27,10 +27,20 @@ title: Choose a Partition Key
...
-  so it's particularly important to consider your data model and access
+  each document to a client-side realm. Sync rules apply at the partition 
+  level, so it's particularly important to consider your data model and access 
   patterns. For more information on partition keys and how to choose one, see
   :ref:`Partition Atlas Data into Realms <partitioning>`.
+  
+  .. note::
+    
+    The partition key can either be required or optional. If the partition key 
+    field is optional, then all documents that either exclude the partition
```
nathan-contino-mongo, line 38: documents -> MongoDB documents

nathan-contino-mongo (approved):
LGTM with one caveat: I think we should use the term "MongoDB documents" instead of "documents" because (particularly in the "partitioning" page) this will make it clearer that this refers to data originating in Atlas, rather than client-side applications.

PR #450, opened 2020-09-11
added warning to log out documentation about local realm access

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/RJAVA-669

https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/master/android/authenticate.html#log-out
https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/master/ios/authenticate.html#log-out
https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/master/node/authenticate.html#log-out
https://docs-mongodbcom-staging.corp.mongodb.com/realm/pegasus/master/react-native/authenticate.html#log-out

### DIFF
--- a/source/includes/log-out-queries-in-progress.rst
+++ b/source/includes/log-out-queries-in-progress.rst
@@ -1,0 +1,7 @@
+.. warning::
+
+   You can only access {+realm+}s accessed with a given user object
+   while that user is logged in. As a result, logging out a user while
+   operations are still executing can cause errors, prevent reads from
+   returning valid results, and prevent data from writing to local
+   {+realm+}s.
--- a/source/includes/log-out-queries-in-progress.rst
+++ b/source/includes/log-out-queries-in-progress.rst
@@ -1,7 +1,7 @@
 .. warning::
 
-   Realms accessed with a given user object only remain valid
-   while that user is logged in. As a result, any operation that has not
-   yet completed before the initiating user logs out cannot complete
-   successfully and will likely result in an error. Any data in a write
-   operation that fails in this way will be lost.
+   If a user logs out, you can no longer read or write data in any
+   synced {+realm+}s that the user opened. As a result, any operation
+   that has not yet completed before the initiating user logs out cannot
+   complete successfully and will likely result in an error. Any data in
+   a write operation that fails in this way will be lost.
--- a/source/includes/log-out-queries-in-progress.rst
+++ b/source/includes/log-out-queries-in-progress.rst
@@ -1,6 +1,6 @@
 .. warning::
 
-   If a user logs out, you can no longer read or write data in any
+   When a user logs out, you can no longer read or write data in any
    synced {+realm+}s that the user opened. As a result, any operation
    that has not yet completed before the initiating user logs out cannot
    complete successfully and will likely result in an error. Any data in
--- a/source/includes/log-out-queries-in-progress.rst
+++ b/source/includes/log-out-queries-in-progress.rst
@@ -1,7 +1,7 @@
 .. warning::
 
-   You can only access {+realm+}s accessed with a given user object
-   while that user is logged in. As a result, logging out a user while
-   operations are still executing can cause errors, prevent reads from
-   returning valid results, and prevent data from writing to local
-   {+realm+}s.
+   Realms accessed with a given user object only remain valid
+   while that user is logged in. As a result, any operation that has not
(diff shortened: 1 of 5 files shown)


source/react-native/authenticate.txt
```diff
@@ -681,7 +681,9 @@ the user's ID token to finish logging the user in to your app.
 Log Out
 -------
 
-To log any user out, call the ``User.logOut()`` on their user instance.
+To log any user out, call ``User.logOut()`` on their user instance.
```
ccho-mongodb, line 684: Nitpick: since you included the user instance in `User.logout()`, is the text that follows redundant? From looking at the snippets, maybe it would be better just to show `logOut()` without `User`?

source/includes/log-out-queries-in-progress.rst
```diff
@@ -0,0 +1,7 @@
+.. warning::
+
+   You can only access {+realm+}s accessed with a given user object
```
ccho-mongodb, line 3: Nitpick: is there another word that can be used to prevent using access twice here: "access realms accessed with", and does that sentence correctly portray what you meant? Did you mean "You can only access a realm that belongs to a user while that user is logged in"?

source/includes/log-out-queries-in-progress.rst
```diff
@@ -0,0 +1,7 @@
...
+   while that user is logged in. As a result, logging out a user while
+   operations are still executing can cause errors, prevent reads from
```
ccho-mongodb, line 5: Nitpick: I wonder if "operations are still executing" should be more specific here. Perhaps "while operations are still running on their associated realm" would work (if correct)?

source/includes/log-out-queries-in-progress.rst
```diff
@@ -0,0 +1,7 @@
...
+   returning valid results, and prevent data from writing to local
```
ccho-mongodb, line 6: Nitpick: for parallel structure with the previous item, this could be "prevent writes from being performed on local realms" or otherwise the prior item could be "prevent valid results from being being returned on reads"

ccho-mongodb (approved):
LGTM. A few nitpicks.

source/includes/log-out-queries-in-progress.rst
nathan-contino-mongo: Technically it's slightly more complicated than that -- no realm really "belongs" to a user, it's just that certain users can access certain realms. If two users both access partition "b", that realm is only stored once. You might say that certain in-memory realm objects belong to certain users, but "accessed with" is a bit more accurate.

Going to reword a bit so at least we don't have multiple uses of "access" in the same sentence.

source/includes/log-out-queries-in-progress.rst
nathan-contino-mongo: Love this!

source/includes/log-out-queries-in-progress.rst
nathan-contino-mongo: Reworded the last two sentences; take another look if you get a chance!

source/react-native/authenticate.txt
nathan-contino-mongo: Stealing the wording from the android snippet.

source/includes/log-out-queries-in-progress.rst
```diff
@@ -0,0 +1,7 @@
+.. warning::
+
+   Realms accessed with a given user object only remain valid
+   while that user is logged in. As a result, any operation that has not
+   yet completed before the initiating user logs out cannot complete
+   successfully and will likely result in an error. Any data in a write
+   operation that fails in this way will be lost.
```
nlarew, line 7: - Is this true only for synced realms? AFAIK local realms don't require you to specify a user.
- What does it mean for a realm to be "valid"?

Maybe - "If a user logs out, you can no longer read or write data in any synced realms that the user opened."

nathan-contino-mongo: Not just synced realms; all realms.

nathan-contino-mongo: The distinction between "local write to a synced realm" and "local write to a local realm" isn't one I was thinking of originally (I was just thinking of as "local write" vs "syncing")... i think this phrasing is even better.

PR #472, opened 2020-09-23
(DOCSP-12343): Specify $function operator support

cbush:
- Adds section to crud-and-aggregation-apis for aggregation operator support exceptions

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12343

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/function-agg-op/mongodb/crud-and-aggregation-apis.html#aggregation-pipeline-operator-availability

### DIFF
--- a/source/mongodb/crud-and-aggregation-apis.txt
+++ b/source/mongodb/crud-and-aggregation-apis.txt
@@ -499,7 +499,7 @@ Query Option Availability
    :header-rows: 1
    :widths: 25 10 10
 
-   * - Modifier
+   * - Option
      - User Context
      - System Context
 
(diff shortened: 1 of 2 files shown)


kpatel71716 (approved):
LGTM!

source/mongodb/crud-and-aggregation-apis.txt
```diff
@@ -520,22 +520,22 @@ Query Option Availability
 Aggregation
 -----------
 
+Aggregation Pipeline Stage Availability
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
nathan-contino-mongo, line 525: Correct me if I'm wrong, but I don't think I've ever seen a header with a subheader immediately following. It looks like the other sections on this page usually put the subheader right before the table, after the description paragraph. Could we do that here for consistency?

nathan-contino-mongo (approved):
LGTM with one suggested change for consistency.

source/mongodb/crud-and-aggregation-apis.txt
cbush: I moved the description for stage availability, which was above the stage availability section, down to underneath that header and added the operator availability header with its own description. I can combine both descriptions up top and leave the subheaders without descriptions. Thoughts?

nathan-contino-mongo: Ah, I see what you mean now. It's just that the aggregation section has an extra level of hierarchy compared to some of the other sections. This is fine then!

PR #476, opened 2020-09-24
Migrate phase 2 tutorial source code to docs-realm/tutorial

cbush:
- Please do not thoroughly review code at this time. The code has been reviewed when it went into the phase2 feature branch on the tutorial and will be reviewed again before we publish.
- This is the source. Bluehawk annotation will result in artifact output repos that students can download.
- This PR should be checked to ensure we aren't adding files we don't want to add and that we're okay with migrating this code here and closing down mongodb-university/realm-tutorial.
- As with the examples/ directory added earlier, your VS Code ignore settings will have to be updated when working with docs-realm source/ now. Create `docs-realm/.vscode/settings.json`:

```json
{
  "files.exclude": {
    "examples/": true,
    "tutorial/": true
  },
  "files.watcherExclude": {
    "examples/": true,
    "tutorial/": true
  },
  "search.exclude": {
    "examples/": true,
    "tutorial/": true
  }
}
```


tutorial/README.md
```diff
@@ -0,0 +1,139 @@
+# Task Tracker Tutorial
+
+The Task Tracker is a collaborative project management tool.
+
```
nlarew, line 4:
```suggestion

**NOTE:** These projects are internal source code for the task tracker applications used in the Realm tutorial for each platform/language. We use these to generate clean, ready-to-run repositories that users who follow the tutorial can access. For details on how to generate user-facing repositories from these projects, see Bluehawk **(SECTION COMING SOON!)**.
```

nlarew (approved):
LGTM w/ suggestion for a note on the root-level README (we should probably flesh it out more at some point)

PR #486, opened 2020-09-24
(DOCSP-11287, 10812): Document Sync Protocol + Clarify Billing

nlarew:
### Issue JIRA link:

- (DOCSP-11287): Document Sync Protocol Details
- (DOCSP-10812): Add more detail on billing for Realm Requests for Sync Operations

- [Realm Sync Protocol](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/sync-protocol/sync/protocol.html)

### DIFF
--- a/source/includes/steps-sync-protocol.yaml
+++ b/source/includes/steps-sync-protocol.yaml
@@ -1,0 +1,68 @@
+title: Realm SDK Initiates a Sync Session
+ref: realm-sdk-initiates-a-sync-session
+content: |
+  To begin a sync session, a Realm SDK sends a :sync-client-message:`BIND`
+  request to a MongoDB Realm application server. The request identifies a
+  specific local Realm file to sync and includes a WebSocket connection key that
+  the server will use to open a bidirectional connection to the SDK.
+
+  If the SDK is attempting to sync a particular Realm file for the first time,
+  it does not yet possess a server-generated client identifier for the file. In
+  this case, the :sync-client-message:`BIND` request also indicates that the
+  Realm server should allocate one.
+---
+title: Realm Allocates a New Client File Identifier
+ref: realm-allocates-a-new-client-file-identifier
+content: |
+  If a :sync-client-message:`BIND` request indicates that the SDK needs a client
+  file identifier, the Realm server generates a unique value for the specified
+  Realm file and sends it to the SDK in an :sync-server-message:`IDENT` message.
+  When the SDK receives the :sync-client-message:`IDENT`, it stores the new
+  client identifier persistently in the local Realm file.
+
+  An SDK only needs to request a client file identifier the first time it syncs
+  each Realm file. For subsequent sync sessions, the SDK can use the persisted
+  identifier.
+---
+title: Realm SDK Sends a Client Identifier
+ref: realm-sdk-sends-a-client-identifier
+content: |
+  Once an SDK has initiated a sync session with a :sync-client-message:`BIND`
+  request, it must identify the local Realm file that it wants to sync. To do
+  this, the SDK sends the application server an :sync-client-message:`IDENT`
+  message that contains the client file identifier. If the identifier is valid,
+  then the server establishes the session and begins to send and receieve sync
+  changesets to and from the SDK.
+---
+title: Realm SDK Uploads & Downloads Sync Changesets
+ref: realm-sdk-uploads-downloads-sync-changesets
+content: |
+  Once a sync session is established, the SDK can send and recieve
+  :sync-client-message:`UPLOAD` and :sync-client-message:`DOWNLOAD` messages to
+  and from the Realm server.
+
+  The SDK sends an :sync-client-message:`UPLOAD` message for every changeset it
+  applies except for those that were specified in a
+  :sync-client-message:`DOWNLOAD` message. When the server receives an
+  :sync-client-message:`UPLOAD` message, it applies operational transforms to
+  resolve any conflicts with other changesets and then applies the transformed
+  changeset to the server version of the realm. This persists the changeset to
+  the synced Atlas cluster and triggers the Realm server to send
+  :sync-client-message:`DOWNLOAD` messages to any other clients that are syncing
+  the same partition.
+
+  The server groups one or more transformed changesets that originated from
+  other clients into a :sync-client-message:`DOWNLOAD` message that it sends to
+  a synced SDK. The :sync-client-message:`DOWNLOAD` message specifies the
+  changesets in chronological order from oldest to most recent according to the
+  server's history and the SDK applies the changesets in the same order.
+---
+title: Realm SDK Terminates the Sync Session
+ref: realm-sdk-terminates-the-sync-session
+content: |
+  Once a sync session is established, Realm servers will continue to accept
+  :sync-client-message:`UPLOAD` messages and send
+  :sync-client-message:`DOWNLOAD` messages until the SDK terminates the session.
+  To terminate a sync session, an SDK sends an :sync-client-message:`UNBIND`
+  request to the Realm server.
+...
--- a/source/includes/steps-sync-protocol.yaml
+++ b/source/includes/steps-sync-protocol.yaml
@@ -3,10 +3,13 @@ ref: realm-sdk-connects-to-the-app-server
 content: |
   The sync protocol is primarily handled over a WebSocket connection between the
   SDK and the server. To establish a connection, the SDK sends a handshake HTTP
-  request that specifies a protocol version and a WebSocket key. The request
-  must include a valid access token for an authenticated Realm application user.
+  request that includes the following:
   
-  The server sends an :mdn`HTTP 101 Switching Protocols <Web/HTTP/Status/101>`
+  - a protocol version
+  - a WebSocket key
+  - a valid access token for an authenticated Realm application user
+  
+  The server sends an :mdn:`HTTP 101 Switching Protocols <Web/HTTP/Status/101>`
   response that specifies a WebSocket connection for the SDK. The rest of the
   sync protocol occurs over this connection.
 ---
@@ -28,9 +31,9 @@ ref: realm-allocates-a-new-client-file-identifier
 content: |
   If a :sync-client-message:`BIND` request indicates that the SDK needs a client
   file identifier, the Realm server generates a unique value for the specified
-  Realm file and sends it to the SDK in an :sync-server-message:`IDENT` message.
-  When the SDK receives the :sync-client-message:`IDENT`, it stores the new
-  client identifier persistently in the local Realm file.
+  Realm file and sends it to the SDK in an :sync-server-message:`IDENT`
+  response. When the SDK receives the :sync-client-message:`IDENT`, it stores
+  the new client identifier persistently in the local Realm file.
 
   An SDK only needs to request a client file identifier the first time it syncs
   each Realm file. For subsequent sync sessions, the SDK can use the persisted
@@ -42,32 +45,34 @@ content: |
   Once an SDK has initiated a sync session with a :sync-client-message:`BIND`
   request, it must identify the local Realm file that it wants to sync. To do
   this, the SDK sends the application server an :sync-client-message:`IDENT`
-  message that contains the client file identifier. If the identifier is valid,
-  then the server establishes the session and begins to send and receieve sync
-  changesets to and from the SDK.
+  message that contains the client file identifier. If the SDK has previously
+  synced the {+realm+} with the server, it can specify the most recently synced
+  server version to optimize the sync process.
+  
+  When it receives the :sync-client-message:`IDENT` message, the server
+  establishes the session. The SDK and server can can now freely send upload and
+  download sync changesets at any time.
 ---
 title: Realm SDK Uploads & Downloads Sync Changesets
 ref: realm-sdk-uploads-downloads-sync-changesets
 content: |
-  Once a sync session is established, the SDK can send and recieve
-  :sync-client-message:`UPLOAD` and :sync-client-message:`DOWNLOAD` messages to
-  and from the Realm server.
+  Once a sync session is established, the SDK and server can freely send and
+  receive :sync-client-message:`UPLOAD` and :sync-client-message:`DOWNLOAD`
+  messages to sync changes whenever they occur.
 
   The SDK sends an :sync-client-message:`UPLOAD` message for every changeset it
-  applies except for those that were specified in a
-  :sync-client-message:`DOWNLOAD` message. When the server receives an
-  :sync-client-message:`UPLOAD` message, it applies operational transforms to
-  resolve any conflicts with other changesets and then applies the transformed
-  changeset to the server version of the realm. This persists the changeset to
-  the synced Atlas cluster and triggers the Realm server to send
-  :sync-client-message:`DOWNLOAD` messages to any other clients that are syncing
-  the same partition.
-
-  The server groups one or more transformed changesets that originated from
-  other clients into a :sync-client-message:`DOWNLOAD` message that it sends to
-  a synced SDK. The :sync-client-message:`DOWNLOAD` message specifies the
+  applies except for those that it received in a :sync-client-message:`DOWNLOAD`
+  message from the server.
+  
+  When the server receives an :sync-client-message:`UPLOAD` message, it applies
+  :ref:`operational transformations <operational-transformations>` to resolve
+  any conflicts with other changesets and then applies the transformed changeset
+  to the server version of the realm. This triggers the server to send
+  :sync-client-message:`DOWNLOAD` messages to other connected clients, including
+  the synced Atlas cluster which mirrors the server realm. A
+  :sync-client-message:`DOWNLOAD` message groups one or more transformed
   changesets in chronological order from oldest to most recent according to the
-  server's history and the SDK applies the changesets in the same order.
+  server's history. The SDK applies the changesets in the same order.
 ---
 title: Realm SDK Terminates the Sync Session
 ref: realm-sdk-terminates-the-sync-session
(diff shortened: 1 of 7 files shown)


source/includes/steps-sync-protocol.yaml
```diff
@@ -0,0 +1,68 @@
+title: Realm SDK Initiates a Sync Session
+ref: realm-sdk-initiates-a-sync-session
+content: |
+  To begin a sync session, a Realm SDK sends a :sync-client-message:`BIND`
```
ianpward, line 4: so I know this is for the sync-protocol but the first thing a client must do is authenticate and get a valid token which can then be used open a websocket sync connection with a BIND message - should we call that out here or would that be better in another doc?

source/includes/steps-sync-protocol.yaml
```diff
@@ -0,0 +1,68 @@
...
+  request to a MongoDB Realm application server. The request identifies a
+  specific local Realm file to sync and includes a WebSocket connection key that
```
ianpward, line 6: should we spell out how the the websocket is actually established? something like this - https://docs.realm.io/server/-Lrk2at2AQQPwaKdjheq/configure/learn-realm-sync-and-integrate-with-a-proxy

source/sync/protocol.txt
```diff
@@ -0,0 +1,267 @@
...
+
+Changeset
+~~~~~~~~~
+
+A **changeset** is a list of instructions that describe granular modifications
+made to a known object state or version by one or more write operations.
+Changesets are the base unit of the sync protocol. Synced {+realm+} clients send
+changesets to the {+service-short+} server whenever they perfrom a write
+operation and the server sends each client changesets for operations executed by
+other clients. The {+service-short+} sync server accepts changesets from any
```
ianpward, line 45: including mongodb driver mutations

source/sync/protocol.txt
```diff
@@ -0,0 +1,267 @@
...
+connected sync client at any time and uses an operational transformation
+algorithm to serialize changes and resolve conflicting changesets.
+
+.. important::
+   
+   Within a single session, a client may upload a given changeset only once
+   and the client version of any uploaded changeset must be greater than all
+   previously uploaded changesets.
```
ianpward, line 53: I don't understand why we are mentioning this as important for the user to understand? do you mean the history version here? I think this is sync just making sure that changesets are only applied once, idempotent, across the wire

source/includes/steps-sync-protocol.yaml
nlarew: We should definitely call out that a sync session requires an authenticated user's valid access token, I'll find a spot on this page that feels good

nlarew: To be clear - does the sync protocol use the same auth process/access tokens as e.g. `app.logIn()` or is it a separate system? The [Proxy Integration](https://docs.realm.io/server/-Lrk2at2AQQPwaKdjheq/configure/learn-realm-sync-and-integrate-with-a-proxy) page you linked makes it seem like they're separate systems.

source/includes/steps-sync-protocol.yaml
nlarew: Ah yeah I think I initially missed the HTTP Request and HTTP Response since they have very generic names. I can add that the WS connection is established over HTTP before any of the other protocol messages can be sent

source/sync/protocol.txt
nlarew: Probably better to summarize as "idempotent", I'll update

source/includes/steps-sync-protocol.yaml
nlarew: Separately we may want to have another page on proxied requests, if it's still relevant

source/includes/steps-sync-protocol.yaml
ianpward: it uses the same system as app.LogIn() - which I believe is covered here:
https://docs.mongodb.com/realm/reference/authenticate-http-client-requests/index.html#get-a-client-api-access-token

source/includes/steps-sync-protocol.yaml
ianpward: I actually don't think we can integrate with a proxy in the new Realm Sync - id have to check with the server team

source/includes/steps-sync-protocol.yaml
```diff
@@ -0,0 +1,80 @@
...
+  each Realm file. For subsequent sync sessions, the SDK can use the persisted
+  identifier.
+---
+title: Realm SDK Sends a Client Identifier
+ref: realm-sdk-sends-a-client-identifier
+content: |
+  Once an SDK has initiated a sync session with a :sync-client-message:`BIND`
+  request, it must identify the local Realm file that it wants to sync. To do
+  this, the SDK sends the application server an :sync-client-message:`IDENT`
+  message that contains the client file identifier. If the identifier is valid,
```
ccho-mongodb, line 45: Nitpick: it could be useful to expand on what conditions qualify it to be valid. Is a Realm name supplied in the request and does that have to match?

source/includes/steps-sync-protocol.yaml
```diff
@@ -0,0 +1,80 @@
...
+  then the server establishes the session and begins to send and receieve sync
+  changesets to and from the SDK.
+---
+title: Realm SDK Uploads & Downloads Sync Changesets
+ref: realm-sdk-uploads-downloads-sync-changesets
+content: |
+  Once a sync session is established, the SDK can send and recieve
```
ccho-mongodb, line 52: recieve -> receive

source/includes/steps-sync-protocol.yaml
```diff
@@ -0,0 +1,80 @@
...
+  changeset to the server version of the realm. This persists the changeset to
+  the synced Atlas cluster and triggers the Realm server to send
+  :sync-client-message:`DOWNLOAD` messages to any other clients that are syncing
+  the same partition.
+
+  The server groups one or more transformed changesets that originated from
+  other clients into a :sync-client-message:`DOWNLOAD` message that it sends to
+  a synced SDK. The :sync-client-message:`DOWNLOAD` message specifies the
+  changesets in chronological order from oldest to most recent according to the
+  server's history and the SDK applies the changesets in the same order.
```
ccho-mongodb, line 70: Nitpick: I think a "," before the "and" could help break up this sentence. Alternatively, you could rewrite this as two sentences.

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+   
+   The {+service-short+} SDKs internally implement and manage the sync protocol,
+   so for most applications you don't need to understand the sync protocol to
+   use {+service-short+}. This page covers the protocol at a high level and is
+   not an implementation spec.
+
+Key Concepts
+------------
+
+.. _changeset:
```
ccho-mongodb, line 35: I think this may apply if you're prepping for nextgen (anchors shouldn't be named the same as titles):
https://jira.mongodb.org/browse/DOP-1532?focusedCommentId=3414748&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3414748

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+
+Changeset
+~~~~~~~~~
+
+A **changeset** is a list of instructions that describe granular modifications
+made to a known object state or version by one or more write operations.
+Changesets are the base unit of the sync protocol. Synced {+realm+} clients send
+changesets to the {+service-short+} server whenever they perfrom a write
+operation and the server sends each client changesets for operations executed by
```
ccho-mongodb, line 44: Question: does this server changeset sent to the client include anything other than write operations? If not, then maybe it would be more clear to refer to them as write operations instead.

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+other clients. The {+service-short+} sync server accepts changesets from any
+connected sync client (including changes in a synced MongoDB cluster) at any
+time and uses an operational transformation algorithm to serialize changes and
```
ccho-mongodb, line 47: Question: Is this "serialize" conversion of the data to another format or ordering of the data? I think it could be important to specify which one.

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+which means that any device may perform offline writes and upload the
+corresponding changesets later when network connectivity is re-established. The
+operational transformation algorithm is designed to gracefully handle changesets
+that arrive "out of order" with respect to the logical server clock such that
+every synced {+service-short+} file converges to the same version of each
+changed object.
+
+.. tip::
+   
+   An operational transformation on {+service-short+} changesets is analagous to
```
ccho-mongodb, line 71: analagous -> analogous

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+
+.. _client-file-identifier:
+
+Client File Identifier
+~~~~~~~~~~~~~~~~~~~~~~
+
+A **client file identifier** is a value that uniquely identifies a synced client
+{+service-short+} file and its corresponding server file. The server generates a
+client file identifier whenever an SDK requests one during its initial sync of a
+{+service-short+} file. Each identifier is a  64-bit, non-zero, positive signed
```
ccho-mongodb, line 82: There's an extra space here before "64-bit"

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+   particular server file are unique with respect to each other. The server is
+   free to generate identical identifiers for two client files if they are
+   associated with different server files.
+
+.. _sync-session-process:
+
+Sync Session Process
+--------------------
+
+To initiate, execute, and terminate a {+service-short+} Database sync session, a
```
ccho-mongodb, line 97: Question: Is Realm Database a product name or should Database be lowercase?

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+.. code-block:: text
+   :copyable: false
+   
+   Realm SDK                      Realm App Server
+       |                                  |
+       |  <---- 1. HTTP Handshake ----->  |
+       |                                  |
+       |  --------- 2. BIND ----------->  |
+       |                                  |
+       |  <-------- 3. IDENT* ----------  |
```
ccho-mongodb, line 118: Question: did you intend to explain why the Server's IDENT request has a "*"?

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+       not yet possess a :ref:`client file identifier <client-file-identifier>`
+       for the {+service-short+} file it wants to sync, this also indicates that
+       the server should generate one and send it back to the client.
+       
+       A client must send a BIND before it can send any other requests.
+   
+   * - .. sync-client-message:: IDENT
+     - Provides the :ref:`client file identifier <client-file-identifier>` that
+       indicates which {+service-short+} file will be synced and specifies the
+       client {+realm+}'s current version.
```
ccho-mongodb, line 160: Question 1: Is this true for the "IDENT*" request from the server as well?

Question 2: Is this the client's SDK version or the changeset version? I think it's the latter, but had to read further down to figure that out.

Question 3: Is there a description about the changeset versions somewhere? Is it monotonically increasing? could be good info in the Changeset section. Comment applies to both Client and Server message descriptions.

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+requests to the server over the WebSocket connections. Once the session is
+established, the SDK and server send synced changesets for a given
+{+service-short+} file to eachother via :sync-client-message:`UPLOAD` and
+:sync-client-message:`DOWNLOAD` messages. To end the session, the SDK sends an
+:sync-client-message:`UNBIND` request.
+
+.. code-block:: text
+   :copyable: false
+   
+   Realm SDK                      Realm App Server
```
ccho-mongodb, line 112: Nice diagram!

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+   * - .. sync-client-message:: STATE_REQUEST
+     - Requests that the server send one or more :sync-server-message:`STATE`
+       messages, which the client uses to download the current server version of
+       the {+service-short+} file. Clients issue state requests when they
+       asynchronously open a synced {+realm+}.
+   
+   * - .. sync-client-message:: CLIENT_VERSION_REQUEST
+     - Requests that the server send the client version of the latest changeset
+       that was sent by the client and processed by the server. This is most
+       commonly used to execute a client reset.
```
ccho-mongodb, line 195: Question: is a client reset when it requests the first changeset version?

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+   
+   * - .. sync-client-message:: PING
+     - Indicates that the client is still connected and that the server should
+       maintain the sync session. A client must send at least one PING to the
+       server every 10 minutes. The server acknowledges to each PING with a
+       :sync-server-message:`PONG`.
+       
+       If the server has not received a PING from a client in more than 10
+       minutes, it considers the client to be disconnected and may automatically
+       end the session.
```
ccho-mongodb, line 205: Question: does this happen regardless of any other type of request being sent by the client during the 10 minutes? Could you mention which one is the case here?

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+     - Specifies the client version of the latest changeset that was sent by the
+       client and processed by the server. Sent in response to a
+       :sync-client-message:`CLIENT_VERSION_REQUEST`.
+   
+   * - .. sync-server-message:: ERROR
+     - Indicates that the server encountered an issue that appears to have been
+       caused by the connected client.
+   
+   * - .. sync-server-message:: PONG
+     - Acknowledges a :sync-client-message:`PING`.
```
ccho-mongodb, line 265: Question: Does a client ever timeout its connection with a server if it doesn't receive the PONG?

source/includes/steps-sync-protocol.yaml
```diff
@@ -0,0 +1,80 @@
+title: Realm SDK Connects to the App Server
+ref: realm-sdk-connects-to-the-app-server
+content: |
+  The sync protocol is primarily handled over a WebSocket connection between the
+  SDK and the server. To establish a connection, the SDK sends a handshake HTTP
+  request that specifies a protocol version and a WebSocket key. The request
+  must include a valid access token for an authenticated Realm application user.
+  
+  The server sends an :mdn`HTTP 101 Switching Protocols <Web/HTTP/Status/101>`
```
ccho-mongodb, line 9: `:mdn` -> `:mdn:`

source/includes/steps-sync-protocol.yaml
```diff
@@ -0,0 +1,80 @@
...
+  identifier.
+---
+title: Realm SDK Sends a Client Identifier
+ref: realm-sdk-sends-a-client-identifier
+content: |
+  Once an SDK has initiated a sync session with a :sync-client-message:`BIND`
+  request, it must identify the local Realm file that it wants to sync. To do
+  this, the SDK sends the application server an :sync-client-message:`IDENT`
+  message that contains the client file identifier. If the identifier is valid,
+  then the server establishes the session and begins to send and receieve sync
```
ccho-mongodb, line 46: receieve -> receive

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+
+.. _changeset:
+
+Changeset
+~~~~~~~~~
+
+A **changeset** is a list of instructions that describe granular modifications
+made to a known object state or version by one or more write operations.
+Changesets are the base unit of the sync protocol. Synced {+realm+} clients send
+changesets to the {+service-short+} server whenever they perfrom a write
```
ccho-mongodb, line 43: perfrom -> perform

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+
+To initiate, execute, and terminate a {+service-short+} Database sync session, a
+{+service-short+} SDK and application server send and receive a set of
+protocol-specific requests.
+
+The SDK negotiates a WebSocket connection over HTTP and then establishes a sync
+session by sending :sync-client-message:`BIND` and :sync-client-message:`IDENT`
+requests to the server over the WebSocket connections. Once the session is
+established, the SDK and server send synced changesets for a given
+{+service-short+} file to eachother via :sync-client-message:`UPLOAD` and
```
ccho-mongodb, line 105: eachother -> each other

source/sync/protocol.txt
```diff
@@ -0,0 +1,265 @@
...
+       A client must send a BIND before it can send any other requests.
+   
+   * - .. sync-client-message:: IDENT
+     - Provides the :ref:`client file identifier <client-file-identifier>` that
+       indicates which {+service-short+} file will be synced and specifies the
+       client {+realm+}'s current version.
+   
+   * - .. sync-client-message:: UPLOAD
+     - Specifies one or more :ref:`changesets <changeset>` for operations that
+       occured on the client. The changesets are listed by client version in
```
ccho-mongodb, line 164: occured -> occurred
(multiple instances)

ccho-mongodb (changes requested):
Overall looks pretty good. I have a number of questions on the content as it wasn't completely clear after reading through it. Also, there are a few typos. You can use a tool like aspell or one of its variants to catch these.

source/sync/protocol.txt
nlarew: It's a product name afaik

source/sync/protocol.txt
nlarew: Lol I deleted some content to this effect in response to one of Ian's comments. Maybe I can re-introduce with a slightly different angle.

ccho-mongodb (approved):
LGTM!

PR #512, opened 2020-10-02
add auth page with code snippets

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12591

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/12591/dotnet/authenticate.html


source/dotnet/authenticate.txt
```diff
@@ -15,115 +15,150 @@ Authenticate a User
 Overview
 --------
 
-.. _dotnet-login:
+{+service+} provides an API for authenticating users using any enabled
+authentication provider. Instantiate a ``Credentials`` object and pass
+it to the  ``app.login()`` method to authenticate a user and create a ``User``
```
nirinchev, line 20: The .NET method is `LogInAsync` - I don't think anyone would get confused over it, but it's always better to use the exact method names.

"to authenticate a user and create a ``User`` object" sounds a bit repetitive. Additionally, the "create a User" wording is a bit imprecise as it implies constructing the instance as opposed to obtaining a reference to it by calling a method. Don't have suggestions for avoiding the repetition, but "create a `User` object" can probably be reworded as "obtain a `User` instance".

source/dotnet/authenticate.txt
```diff
@@ -15,115 +15,150 @@ Authenticate a User
 Overview
 --------
 
-.. _dotnet-login:
+{+service+} provides an API for authenticating users using any enabled
+authentication provider. Instantiate a ``Credentials`` object and pass
+it to the  ``app.login()`` method to authenticate a user and create a ``User`` 
+object. Each authentication provider corresponds to a method used to
+instantiate ``Credentials`` objects using that authentication provider.
```
nirinchev, line 22: This sounds a bit heavy - perhaps if we inverted the flow, it would be easier to follow. I.e. something like "The `Credentials` class exposes factory methods for each authentication provider."

source/dotnet/authenticate.txt
```diff
@@ -15,115 +15,150 @@ Authenticate a User
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 50 50
+
+   * - Authentication Provider
+     - Credentials Generation Method
+
+   * - :ref:`Anonymous <dotnet-login-anonymous>`
+     - ``Credentials.Anonymous()``
```
nirinchev, line 34: Would be nice if those linked to the API docs for that method as those typically contain valuable information about parameters, return types, etc.

I can send you a .zip containing the API docs for the v10 SDK if that would help.

examples/dotnet/Examples/RealmTests.cs
```diff
@@ -84,8 +86,79 @@ public async System.Threading.Tasks.Task ModifiesATask()
...
+            await apiUser.LogOutAsync();
+            // :code-block-start: logon_Function
+            var functionParameters = new Dictionary<string, string>()
+            {
+                { "username", "caleb" },
+                { "password", "shhhItsASektrit!" },
+                { "someOtherProperty", "cheesecake" }
+            };
+            Realms.Sync.User functionUser =
+                await app.LogInAsync(Credentials.Function(functionParameters));
```
nirinchev, line 115: This will work, but the code can be simplified by using anonymous objects:

```csharp
var functionParameters = new
{
    username = "caleb",
    password = "shhhItsASektrit",
    someOtherProperty = "cheesecake"
};
var functionUser = await app.LogInAsync(Credentials.Function(functionParameters));
```

Note that when using the anonymous object syntax, you can also hint that properties don't have to be strings only. Any serializable .NET type, including embedded objects will work:

```csharp
var functionParameters = new
{
    biometricsValidated = false,
    todaysDate = DateTime.UtcNow
}
```

Of course, don't have to go overboard with the extra properties, just FYI.

examples/dotnet/Examples/RealmTests.cs
```diff
@@ -84,8 +86,79 @@ public async System.Threading.Tasks.Task ModifiesATask()
...
+            }
+            catch (Exception e)
+            {
+                Assert.AreEqual("http error code considered fatal: Client Error: 401", e.Message);
+            }
+            try
+            {
+                // :code-block-start: logon_google
+                Realms.Sync.User googleUser =
+                    await app.LogInAsync(Credentials.Google("<google_token>"));
```
nirinchev, line 140: I think we're calling this google auth code now as using the access token flow doesn't work.

source/examples/generated/code/final/RealmTests.cs
```diff
@@ -70,8 +72,63 @@ public async System.Threading.Tasks.Task ModifiesATask()
...
+            Assert.AreEqual(UserState.LoggedIn, jwtUser.State);
+            await jwtUser.LogOutAsync();
+            try
+            {
+                Realms.Sync.User fbUser =
+                    await app.LogInAsync(Credentials.Facebook("<facebook_token>"));
+            }
+            catch (Exception e)
+            {
+                Assert.AreEqual("http error code considered fatal: Client Error: 401", e.Message);
```
nirinchev, line 108: FYI: this message will change when you upgrade to the latest v10 package - the generic wording was due to a bug that has now been addressed and the message should be more meaningful (I don't think that's exposed in the docs, but will break your test, so don't be surprised).

nirinchev (commented):
Some wording comments + code snippet suggestions. None of these are "must fix" or anything, so if you don't agree, feel free to disregard.

examples/dotnet/Examples/RealmTests.cs
```diff
@@ -84,8 +86,79 @@ public async System.Threading.Tasks.Task ModifiesATask()
             return;
         }
 
+        [Test]
+        public async System.Threading.Tasks.Task LogsOnManyWays()
+        {
+            // :code-block-start: logon_anon
+            Realms.Sync.User anonUser = await app.LogInAsync(Credentials.Anonymous());
```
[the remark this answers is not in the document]
MongoCaleb: @nirinchev - the reason I had the FQDN here was because we have a user-defined class called "User", so the compiler complained. I've updated the custom class names because they are semi-arbitrary.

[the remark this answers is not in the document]
nirinchev: My suggestion would be to always use `var` as the variable name should be indicative enough. E.g. `anonUser` is obviously a User, so explicitly specifying the type is redundant. This also lets you avoid some of the type collisions.

source/dotnet/authenticate.txt
MongoCaleb: Yes, please...or a link would be fine, too.

examples/dotnet/Examples/RealmTests.cs
MongoCaleb: niiiiice

examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        [SetUp]
+        public async System.Threading.Tasks.Task Setup()
+        {
+            // :code-block-start: initialize-realm
+            app = App.Create(myRealmAppId);
+            // :code-block-end:
+            user = app.LogInAsync(Credentials.EmailPassword("foo@foo.com", "foobar")).Result;
+            config = new SyncConfiguration("My Project", user);
+            Realm realm = await Realm.GetInstanceAsync(config);
```
nirinchev, line 27: Probably move `realm` to a class-level variable as well to avoid the `Realm realm = await Realm...` repetition.

examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,179 @@
...
+            // :code-block-stat: create
+            RealmTask testTask = new RealmTask()
```
nirinchev, line 29: Same here - `RealmTask testTask = new RealmTask` is fairly repetitive. Also, you can drop the `()` from the constructor when using the property initializer syntax - there are different opinions about it, but most high profile .NET people seem to omit them.

examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            // :code-block-start: anon-login
+            User user = app.LogInAsync(Credentials.Anonymous()).Result;
+            // :code-block-end:
+            // :code-block-start: config
+            config = new SyncConfiguration("My Project", user);
+            Realm realm = await Realm.GetInstanceAsync(config);
+            // :code-block-end:
+            // :code-block-start: read-all
+            var tasks = realm.All<RealmTask>().ToList();
```
nirinchev, line 55: We shouldn't encourage people to `.ToList()` query results as that removes the "liveness" from the collection.

examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,179 @@
...
+            // :code-block-end:
+            Assert.AreEqual(1, tasks.Count);
+            // :code-block-start: read-some
+            tasks = realm.All<RealmTask>().Where(t => t.Status == "Open").ToList();
```
nirinchev, line 59: Same here - materializing the collection is generally viewed as a bad practice unless you have very specific needs. Instead, you should operate on the query as much as possible.

examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,179 @@
...
+            User apiUser = await app.LogInAsync(Credentials.ApiKey("eRECwv1e6gkLEse99XokWOgegzoguEkwmvYvXk08zAucG4kXmZu7TTgV832SwFCv"));
+            // :code-block-end:
+            Assert.AreEqual(UserState.LoggedIn, apiUser.State);
+            await apiUser.LogOutAsync();
+            // :code-block-start: logon_Function
+            var functionParameters = new
+            {
+                username=  "caleb",
+                password = "shhhItsASektrit!",
+                IQ = 42,
```
nirinchev, line 111: Hahaha... I guess explains the password 😅

ccho-mongodb (approved):
Strange, my review seems to never have gone through. Sorry about that.
I'm certain that I typed a message, selected Approve, and submitted, but I don't see that message nor the approval.
Anyway, hope these are helpful.

examples/dotnet/Examples/Examples.cs
MongoCaleb: Since these code snippets live "alone" within the docs, and the user won't be seeing this unit test file as a whole, I think it's useful to have the `Realm` there to be clear what we are building.

examples/dotnet/Examples/Examples.cs
MongoCaleb: Can you tell I was getting punchy?

examples/dotnet/Examples/Examples.cs
nirinchev: Right, but it's in the variable name as well - i.e. knowing that I have a variable called `realm` of type `Realm` isn't that much more useful than just knowing that I have a variable called `realm`.

examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,189 @@
...
+            Assert.AreEqual(TaskStatus.InProgress.ToString(), allTasks.First().Status);
+
+            return;
+        }
+
+        [Test]
+        public async System.Threading.Tasks.Task LogsOnManyWays()
+        {
+            // :code-block-start: logon_anon
+            User anonUser = await app.LogInAsync(Credentials.Anonymous());
```
nirinchev, line 96: I don't think we should prefix the variable name with the provider type - in the iOS/Android docs, we're using `user` for all the examples. Since that will result in compilation errors here, you can add a scope around each snippet:

```csharp
public async System.Threading.Tasks.Task LogsOnManyWays()
{
    {
        // :code-block-start: logon_anon
        var user = await app.LogInAsync(CredentialAnonymous());
        // :code-block-end:
        Assert.AreEqual(UserState.LoggedIn, user.State);
        await user.LogOutAsync();
    }

    {
        // :code-block-start: logon_EP
        var user = await app.LogInAsync(
            Credentials.EmailPassword("caleb@mongodb.com", "shhhItsASektrit!"));
        // :code-block-end:
        Assert.AreEqual(UserState.LoggedIn, user.State);
        await user.LogOutAsync();
    }
    // ...
}
```

nirinchev (approved):
Final nitpick from me.

PR #519, opened 2020-10-06
Open a Realm, Call a Function

MongoCaleb:
### Issue JIRA links:
Open a Realm: https://jira.mongodb.org/browse/DOCSP-12487
Call a Function: https://jira.mongodb.org/browse/DOCSP-12476

OaR: https://docs-mongodbcom-staging.corp.mongodb.com/realm/caleb/clientguide/dotnet/open-a-realm.html
CaF: https://docs-mongodbcom-staging.corp.mongodb.com/realm/caleb/clientguide/dotnet/call-a-function.html


examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        [Test]
+        public async Task CallsAFunction()
+        {
+            // :code-block-start: callfunc
+            var result = await
+                user.Functions.CallAsync("sum", 2, 40);
+
+            // result.ToInt32() == 42
```
nathan-contino-mongo, line 201: To me, this comment looks a little out of place in the code-snippet on this page. Could we maybe store the result in a variable and print it out instead?

source/dotnet/open-a-realm.txt
```diff
@@ -0,0 +1,51 @@
...
+by calling ``GetInstanceAsync``. You can also open a {+realm+} *synchronously* 
+by calling ``GetInstance``, but this may lead to temporary data inconsistencies 
+while the remote data is downloaded, and is generally not recommended. 
+
+.. literalinclude:: /examples/generated/code/start/Examples.codeblock.open-synced-realm-sync.cs
+   :language: csharp
+
+Open a Local (Non-Synced) Realm
+-------------------------------
+When opening a local (non-synced) {+realm+}, you pass a ``RealmConfiguration``
```
nathan-contino-mongo, line 34:
```suggestion
When opening a local (non-synced) {+realm+}, pass a ``RealmConfiguration`` 
```
extraneous word, I would probably just trim this.

source/dotnet/open-a-realm.txt
```diff
@@ -0,0 +1,51 @@
...
+   :language: csharp
+
+.. note::
+
+   If you specify a local path, that path must already exist on the device.
+
+Close a Realm
+-------------
+There is no need to manually close a {+realm+} with the .NET SDK. When a {+realm+} 
+goes out of scope and is removed from memory, the {+realm+} is closed.
```
nathan-contino-mongo, line 50: This is great. I wish the other SDKs did this 😢

nathan-contino-mongo (approved):
LGTM with a couple of small nits.

source/dotnet/open-a-realm.txt
```diff
@@ -0,0 +1,51 @@
...
+To open a synced {+realm+}, call the ``GetInstanceAsync()`` method, 
+passing in a ``SyncConfiguration`` object that includes the partition name and 
+the user. The following code demonstrates this:
+
+.. literalinclude:: /examples/generated/code/start/Examples.codeblock.open-synced-realm.cs
+   :language: csharp
+
+In the above example, the code shows how to open the {+realm+} *asynchronously* 
+by calling ``GetInstanceAsync``. You can also open a {+realm+} *synchronously* 
+by calling ``GetInstance``, but this may lead to temporary data inconsistencies
```
nirinchev, line 26: I'm not sure that's very clear - there are no data inconsistencies associated with opening a Realm file synchronously and I don't believe we have strong recommendations against it. In fact, opening a Realm asynchronously is only possible if the device is online and it kind of defeats one of the major selling points for Realm.

I believe the general recommendation we have is to use `GetInstanceAsync` the first time a user logs in as it's likely that they're connected at this point, but switch to `GetInstance` after that to ensure offline capabilities.

examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,224 @@
...
+        [SetUp]
+        public async Task Setup()
+        {
+            // :code-block-start: initialize-realm
+            app = App.Create(myRealmAppId);
+            // :code-block-end:
+            user = app.LogInAsync(Credentials.EmailPassword("foo@foo.com", "foobar")).Result;
+            // :code-block-start: open-synced-realm
+            config = new SyncConfiguration("My Project", user);
+            Realm realm = await Realm.GetInstanceAsync(config);
```
nirinchev, line 31: As mentioned before, let's use `var` instead of the type name here as it improves readability. It's also consistent with the [Swift](https://docs.mongodb.com/realm/ios/open-a-realm/) and [Kotlin](https://docs.mongodb.com/realm/android/open-a-realm/) docs where `let` and `val` are used throughout the documentation.

source/dotnet/open-a-realm.txt
```diff
@@ -0,0 +1,51 @@
...
+creates a ``RealmConfiguration`` object with a local file path, sets the 
+``IsReadOnly`` property to ``true``, and then opens a local {+realm+}  with that 
+configuration information:
+
+.. literalinclude:: /examples/generated/code/start/Examples.codeblock.local-realm.cs
+   :language: csharp
+
+.. note::
+
+   If you specify a local path, that path must already exist on the device.
```
nirinchev, line 45: This is only the case when opening the realm with `IsReadOnly = true`, which is probably fairly rare. If you're opening a Realm in a read/write mode, the SDK will create the file if it didn't exist before that.

source/dotnet/open-a-realm.txt
nirinchev: It's not true unfortunately. The `Realm` instance implements `IDisposable` (see https://realm.io/docs/dotnet/latest/#closing-realm-instances) so the general recommendation is to open a Realm instance with a `using`, unless that instance is going to be long lived (e.g. on the main thread) and you need to continuously read/write data to it as the user interacts with the app. It's true that we have hooks into GC to do that for you, but the issue is that the majority of the memory associated with a Realm instance is native memory which GC doesn't know about. This means that it may take a long time between an instance going out of scope and it being collected because the GC sees it as a very lightweight object and may decide not to run.

examples/dotnet/Examples/Examples.cs
```diff
@@ -0,0 +1,224 @@
...
+                Assert.AreEqual("InvalidSession: authentication via 'oauth2-apple' is unsupported", e.Message);
+            }
+        }
+
+        [Test]
+        public async Task CallsAFunction()
+        {
+            // :code-block-start: callfunc
+            var result = await
+                user.Functions.CallAsync("sum", 2, 40);
```
nirinchev, line 199: It's probably a good idea to showcase the generic overload of `CallAsync` so that we don't have to work with the raw BsonValue. I believe this would look like `var result = await user.Function.CallAsync<int>("sum", 2, 40);`

nirinchev (commented):
Added a few small corrections, but looks good overall.

conf.py
```diff
@@ -109,6 +109,7 @@
     'swift-sdk': ('https://docs.mongodb.com/realm-sdks/swift/10.0.0-rc.1/%s', ''),
     'objc-sdk': ('https://docs.mongodb.com/realm-sdks/objc/10.0.0-rc.1/%s', ''),
     'js-sdk': ('https://docs.mongodb.com/realm-sdks/js/latest/%s', ''),
+    'dotnet-sdk': ('https://docs.mongodb.com/realm-sdks/js/latest/%s', ''),
```
nirinchev, line 112: Should this be updated to `.../dotnet/latest/...`?

examples/dotnet/Examples/Examples.cs
```diff
@@ -17,23 +19,24 @@ public class Examples
...
-            Realm synchronousRealm = Realm.GetInstance(config);
+            var synchronousRealm = Realm.GetInstance(config);
             // :code-block-end:
             // :code-block-start: create
-            RealmTask testTask = new RealmTask
+            var testTask = new RealmTask
             {
                 Name = "Do this thing",
-                Status = TaskStatus.Open.ToString()
+                Status = dotnet.TaskStatus.Open.ToString()
```
nirinchev, line 39: We probably don't want to use the namespace here?

examples/dotnet/Examples/Examples.cs
```diff
@@ -46,95 +49,147 @@ public async System.Threading.Tasks.Task Setup()
...
+            Assert.AreEqual(1, tasks.Count(), "Get Some");
             return;
         }
+       
+        [Test]
+        public async Task ScopesARealm()
+        {
+            // :code-block-start: scope
+            config = new SyncConfiguration("My Project", user);
+            using (var realm = await Realm.GetInstanceAsync(config))
```
nirinchev, line 107: You can showcase the new `using` syntax from C# 8 here and simplify the code a little:

```csharp
config = new SyncConfiguration(...);
using var realm = await Realm.GetInstanceAsync(...);
var allTasks = realm.All<RealmTask>();
```

examples/dotnet/Examples/Examples.cs
nirinchev: If it conflicts with something, you can add

```csharp
using TaskStatus = dotnet.TaskStatus;
```

next to your using statements to help the compiler resolve the correct type.

source/dotnet/call-a-function.txt
```diff
@@ -0,0 +1,47 @@
...
+To execute a function from the .NET Client SDK, use the ``user.Functions.CallAsync()`` 
+method, passing in the name of the function as the first parameter and the arguments
+as the remaining parameters:
+
+.. literalinclude:: /examples/generated/code/start/Examples.codeblock.callfunc.cs
+   :language: csharp
+
+.. note::
+   The ``CallAsync()`` method returns a raw ``BsonValue``, which must be cast to the 
+   expected type, unless you use one of the generic overloads. In the example above,
```
nirinchev, line 46: [super-nit] There's only one generic overload for `CallAsync`.

source/dotnet/open-a-realm.txt
```diff
@@ -0,0 +1,63 @@
...
+In the above example, the code shows how to open the {+realm+} *asynchronously* 
+by calling ``GetInstanceAsync``. You can also open a {+realm+} *synchronously* 
+by calling ``GetInstance``, which is necessary if the device is offline. 
+
+.. literalinclude:: /examples/generated/code/start/Examples.codeblock.open-synced-realm-sync.cs
+   :language: csharp
+
+.. note::
+
+   The first time a user logs on to your realm app, you should open the realm
```
nirinchev, line 33: I would use a less compelling verb - i.e. `The first time a user logs on to your realm app, it is recommended to open the realm *asynchronously*`. 

And on a separate note - before the acquisition, the branding instructions stated that we should capitalize Realm when talking about the files - e.g. "Can you send me your Realm file", "To open a synchronized Realm, just pass in a SyncConfiguration with a logged in user", etc. Not sure if that has changed, but might want to sync up with the docs/marketing teams to ensure we're consistent in the usage.

source/dotnet/open-a-realm.txt
```diff
@@ -0,0 +1,63 @@
...
+creates a ``RealmConfiguration`` object with a local file path, sets the 
+``IsReadOnly`` property to ``true``, and then opens a local {+realm+}  with that 
+configuration information:
+
+.. literalinclude:: /examples/generated/code/start/Examples.codeblock.local-realm.cs
+   :language: csharp
+
+Scoping the Realm
+-----------------
+The Realm instance implements ``IDisposable``, and you should dispose of a realm
```
nirinchev, line 51: I would reword this slightly to make it less forceful:

```
The Realm class implements ``IDisposable`` to clean up the native resources it uses. It is recommended
to always dispose of Realm instances after use, especially on background threads. The simplest way is to
declare the variable with a `using` or wrap the code that interacts with a Realm in a `using (...)` statement:
```

Also, might be worth adding a note that disposing of a Realm instance will invalidate all objects associated with that instance, so if the developer needs to use them throughout the lifetime of the app - e.g. for databinding purposes on the main thread - it's safe not to dispose of the Realm. The rule of thumb is "don't dispose on the main thread, dispose on background ones".

nirinchev (approved):
Final nits and clarifications! Thanks for bearing with me and my obsessiveness 🙏

source/dotnet/open-a-realm.txt
MongoCaleb: FYI: https://docs.google.com/document/d/1Yc30_-YfVziocquh5jNgBDmTQh0ynrPSyVGSRFma8Jw/edit

nirinchev: Thanks! I hadn't seen this, but super helpful - I just need to untrain my brain to uppercase the local file. But I guess the `realm` in `your realm app` needs to be uppercase then?

PR #531, opened 2020-10-08
(DOCSP-12650) Electron Quick Start Updates

mongomoe:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12650

https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-12650-Electron-QuickStart-Updates/node/electron.html

### DIFF
--- a/source/includes/steps-realm-with-electron-using-create-react-app.yaml
+++ b/source/includes/steps-realm-with-electron-using-create-react-app.yaml
@@ -11,7 +11,7 @@ content: |
 
   Your application file structure should look something like the
   following. This representation abstracts some files, such as CSS,
-  service worker and test files. 
+  service worker, and test files. 
 
   .. code-block:: text
      :emphasize-lines: 7, 8
--- a/source/includes/steps-realm-with-electron-using-create-react-app.yaml
+++ b/source/includes/steps-realm-with-electron-using-create-react-app.yaml
@@ -9,6 +9,33 @@ content: |
 
      npx create-react-app my_electron_react_application
 
+  Your application file structure should look something like the
+  following. Some files are abstracted from this representation, such
+  as CSS, service worker and test files. 
+
+  .. code-block:: text
+     :emphasize-lines: 7, 8
+
+     .
+     |-- package.json 
+     |-- yarn.lock
+     |-- public
+     |   |-- index.html
+     |-- src
+     |   |-- App.js // renderer process
+     |-- |-- index.js // renderer process
+
+  Any JS Files within the ``src`` directory executes on a ``renderer process``.
+
+  .. note::
+
+   Each Electron application can only have one **main process**. The
+   main process creates web pages. Each web page runs in its
+   own process, known as a ``renderer process``. To learn more about
+   this, read the official `Electron Application
+   Architecture <https://www.electronjs.org/docs/tutorial/application-architecture>`_
+   doc.
+
 ---
 title: Install CRACO to Alter Your Webpack Configuration
 ref: install-craco
@@ -80,6 +107,23 @@ content: |
 
   .. literalinclude:: /examples/QuickStarts/electron/quickstart-setup-react-electron.js
 
+  Your application file structure should now look something like the following. Some files are abstracted from this representation such as CSS, service worker, and test files. 
+
+  .. code-block:: text
+     :emphasize-lines: 7, 9, 10
+
+     .
+     |-- package.json 
+     |-- yarn.lock
+     |-- craco.config.js
+     |-- public
+     |   |-- index.html
+     |   |-- electron.js // main process
+     |-- src
+     |   |-- App.js // renderer process
+     |-- |-- index.js // renderer process
+
+  All JS Files within the ``src`` directory executes on the ``renderer`` process. The electron.js and any files required by it execute on the ``main process``.
 ---
 title: Run Your Application
 ref: run-your-app
--- a/source/includes/steps-realm-with-electron-using-create-react-app.yaml
+++ b/source/includes/steps-realm-with-electron-using-create-react-app.yaml
@@ -10,8 +10,8 @@ content: |
      npx create-react-app my_electron_react_application
 
   Your application file structure should look something like the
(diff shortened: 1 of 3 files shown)


source/includes/steps-realm-with-electron-using-create-react-app.yaml
```diff
@@ -9,6 +9,33 @@ content: |
...
+  .. code-block:: text
+     :emphasize-lines: 7, 8
+
+     .
+     |-- package.json 
+     |-- package-lock.lock
+     |-- public
+     |   |-- index.html
+     |-- src
+     |   |-- App.js // renderer process
```
ccho-mongodb, line 25: Question: since you mention that all files within src are executed on a renderer process, could you omit these? I've never seen "//" used in the context of a directory tree.
Also, I think it could be confusing to label files as "renderer process" because they run on the renderer process, but are actually files. Maybe "runs on renderer process" would be more descriptive.

source/includes/steps-realm-with-electron-using-create-react-app.yaml
```diff
@@ -9,6 +9,33 @@ content: |
 
      npx create-react-app my_electron_react_application
 
+  Your application file structure should look something like the
+  following. This representation abstracts some files, such as CSS,
```
ccho-mongodb, line 13: Question: do you mean "omits" instead of "abstracts"? I think the form of "abstracts" you are using implies that the representation is acting upon something to remove the files (action) vs describing the current state of the representation.

source/includes/steps-realm-with-electron-using-create-react-app.yaml
```diff
@@ -9,6 +9,33 @@ content: |
...
+  Any JS Files within the ``src`` directory executes on a ``renderer process``.
+
+  .. note::
+
+   Each Electron application can only have one **main process**. The
+   main process creates web pages. Each web page runs in its
+   own process, known as a ``renderer process``. To learn more about
+   this, read the official `Electron Application
+   Architecture <https://www.electronjs.org/docs/tutorial/application-architecture>`_
+   doc.
```
ccho-mongodb, line 37: "doc" -> "article" or "document" (which you used in the node/electron.txt file)

source/includes/steps-realm-with-electron.yaml
```diff
@@ -7,11 +7,34 @@ content: |
 
      mkdir myElectronApplication
      
-  Create an index.html and main.js file in the root directory of
-  your application. The code in the main.js file is the Electron main
-  process file, which can be thought of as the entry point into your
+  Create an index.html and main.js file in the root directory of your
+  application. The code in the main.js file is the Electron main process
+  file, which can be thought of as the entry point into your
```
ccho-mongodb, line 12: Nitpick: how about "which is the entry point into..."?

source/includes/steps-realm-with-electron.yaml
```diff
@@ -7,11 +7,34 @@ content: |
...
   application. The main.js file is responsible for loading your
-  index.html file into the BrowserWindow created by Electron.
+  index.html file into the BrowserWindow created by Electron. Any script
+  files that you require in this HTML file will execute on the renderer
+  process. Add the following to your ``index.html`` file:
```
ccho-mongodb, line 16: In the prior sentence, you mention "any script files", and then your example adds renderer.js. I wonder if it would be helpful to call that out in this sentence. E.g.
"Add the renderer.js script into your index.html file as follows"

source/node/electron.txt
```diff
@@ -32,6 +32,14 @@ Before you begin, ensure you have:
 - :ref:`Created a {+app+} <create-a-realm-app>`
 - :ref:`Enabled {+sync+} <enable-sync>`
 
+.. admonition:: Version Requirement
+   :class: note
+
+   {+service+} recommends that you use the latest stable Electron
```
ccho-mongodb, line 38: This paragraph seems to repeat "Electron version(s)" a lot. Instead, how about rearranging the sentence to something like this:
{+service+} works with any of the officially supported versions of Electron, but we recommend you use the latest stable release...

ccho-mongodb (approved):
LGTM! I learned a bit about Electron from this guide. A few questions/nitpicks.

source/includes/steps-realm-with-electron-using-create-react-app.yaml
mongomoe: Omits does work better here. Thanks for the suggestion, altered the doc to use the word 'omits' now!

source/includes/steps-realm-with-electron-using-create-react-app.yaml
mongomoe: There is precedent for having "//" within a representation of file structure within our docs (See: https://docs.mongodb.com/realm/tutorial/android-kotlin/#connect-to-your-mongodb-realm-app).

I have altered this (and all similar comments) to "runs on the renderer process"

source/includes/steps-realm-with-electron-using-create-react-app.yaml
mongomoe: Good call, replaced this w/ "document"

source/includes/steps-realm-with-electron.yaml
mongomoe: good idea, changed it to this

source/includes/steps-realm-with-electron.yaml
mongomoe: I don't think this is necessary, as a developer can add any number of script files to the index.html with any name and they would all execute on the renderer process. The name "renderer.js" is only used for clarity, it could really be called anything.

source/node/electron.txt
mongomoe: Good idea, the new paragraph reads: "{+service+} works with any of the officially supported Electron
   versions. We recommend using the latest stable release. To see the
   officially supported Electron versions, check out the `Electron
   Releases <https://www.electronjs.org/releases/>`_ document."

PR #534, opened 2020-10-09
(DOCSP-12414): [REALMC] Reorganize Realm Side Navigation

nlarew:
### Issue JIRA link:

- (DOCSP-12414): [REALMC] Reorganize Realm Side Navigation

- [Staged](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/new-sidebar/index.html)

### DIFF
--- a/source/android/access-custom-user-data.txt
+++ b/source/android/access-custom-user-data.txt
@@ -123,7 +123,7 @@ information from your custom user data configuration:
   to users (via :term:`user ID`)
 
 You can find this information in the the :guilabel:`Custom User Data`
-tab of the :guilabel:`Users` left sidebar entry of the {+ui+}.
+tab of the :guilabel:`App Users` left sidebar entry of the {+ui+}.
 
 .. _android-create-custom-user-data:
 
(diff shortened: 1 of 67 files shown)


source/android/access-custom-user-data.txt
```diff
@@ -123,7 +123,7 @@ information from your custom user data configuration:
   to users (via :term:`user ID`)
 
 You can find this information in the the :guilabel:`Custom User Data`
-tab of the :guilabel:`Users` left sidebar entry of the {+ui+}.
+tab of the :guilabel:`App Users` left sidebar entry of the {+ui+}.
```
ccho-mongodb, line 126: Nitpick: I wonder if it might be easier to follow if you order it chronologically:
"To view this information, click on the App Users left sidebar entry and then select the Custom User Data tab."
Looks like you do this in the step file, so it could be helpful to the user to be consistent.

source/users/find-and-view.txt
```diff
@@ -13,7 +13,7 @@ Find and View a User
 Find a User
 -----------
 
-To find a specific user, click :guilabel:`Users` in the left navigation
+To find a specific user, click :guilabel:`App Users` in the left navigation
```
ccho-mongodb, line 16: Nitpick: Looks like there are three variants:
"left navigation"
"left-hand navigation"
"left-side navigation"
Would it be helpful to consolidate for consistency, and maybe even create a custom role?

ccho-mongodb (approved):
LGTM! A couple nitpicks.
Also, I'm wondering if it's possible to source another file in a yaml file as it appears some of the text is repeated a few times as it could reduce time to update it in the future if necessary.

PR #538, opened 2020-10-12
(DOCSP-12110): Node example app

nlarew:
### Issue JIRA link:

- (DOCSP-12110): Node example app

- [Node.js SDK > Authenticate a User](https://docs-mongodbcom-staging.corp.mongodb.com/realm/nlarew/node-example-app/node/authenticate.html)


examples/node/.eslintrc.js
```diff
@@ -0,0 +1,21 @@
...
+    node: true,
+    es2021: true,
+    "jest/globals": true,
+  },
+  extends: [
+    "eslint:recommended",
+    // "plugin:prettier/recommended",
+    "plugin:@typescript-eslint/recommended",
+  ],
+  parser: "@typescript-eslint/parser",
```
cbush, line 12: I don't think you actually want to use the typescript parser on the JS examples. This could lead to inaccurate diagnostics on code examples that are meant for a different environment like non-transpiled node or browser. In the sample I showed you, you can use the "overrides" field to provide different configurations for other environments.

examples/node/.eslintrc.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  env: {
+    node: true,
+    es2021: true,
+    "jest/globals": true,
+  },
+  extends: [
+    "eslint:recommended",
+    // "plugin:prettier/recommended",
```
cbush, line 9: Why not use prettier/recommended?

cbush (approved):
Nice work! I can see future work around splitting and being more intentional about the different possible environments (browser, non-babel node, typescript, RN...) but we can deal with that later.

examples/node/.eslintrc.js
nlarew: Oops forgot to uncomment, was having trouble with it because I was missing a package

PR #541, opened 2020-10-13
12485: working with MongoDB

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12485

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/12485/dotnet/mongodb.html


source/examples/generated/code/start/Plant.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        [MapTo("_partition")]
+        public string Partition { get; set; }
+
+        public Plant()
+        { }
+    }
+    public enum Sunlight
+    {
+        full,
```
nirinchev, line 32: [nit] These should probably be PascalCase in the C# code, even if they're camelCase/lowercase in MongoDB.

examples/dotnet/Examples/MongoDBExamples.cs
```diff
@@ -109,7 +91,6 @@ public async Task InsertsMany()
                 Color = PlantColor.purple,
                 Type = PlantType.annual,
                 Partition = "Store 47"
-
             };
 
             var listofPlants = new List<Plant>
```
nirinchev, line 96: [nit] You can simplify that by passing in an array rather than a list:

```csharp
var plants = new[]
{
    sweetBasil,
    thaiBasil,
    // rest of the plants...
}
```

examples/dotnet/Examples/MongoDBExamples.cs
```diff
@@ -135,39 +115,91 @@ public async Task InsertsMany()
         public async Task ReadsDocuments()
         {
             // :code-block-start: mongo-find-one
-            var petunia = await plantsCollection.FindOneAsync<Plant>(
-                new BsonDocument("name", "petunia"), null);
+            var petunia = await plantsCollection.FindOneAsync(
+                new BsonDocument("Name", "Petunia"), null);
```
nirinchev, line 119: Once you update `MapTo` to be `BsonElement`, `Name` will have to become lowercase.

source/dotnet/mongodb.txt
```diff
@@ -15,1171 +15,301 @@ from a .NET application using the {+service-short+} .NET SDK.
...
-collection from a mobile application. Update operations use query filters
-to specify which documents to update and :manual:`update operators
-</reference/operator/update>` to describe how to mutate documents that
-match the query. Update operations return a :android:`Task
-<reference/com/google/android/play/core/tasks/Task>` that resolves to an object
-that contains the results of the execution of the operation.
+To update an existing document in the MongoDB datastore, you create a 
+``BsonDocument`` filter that defines the properties you want to search on, and 
+then create a second ``BsonDocument`` that defines the properties you want to 
+change. If you are updating only one document, you pass both ``BsonDocument``s to
```
nirinchev, line 153: There seems to be an issue with the rendering of ` ``BsonDocument``s ` - the ticks show up and are rendered as an anchor link.

source/examples/generated/code/start/MongoDBExamples.cs
```diff
@@ -0,0 +1,193 @@
...
+            {
+                var updateResult = await plantsCollection.UpdateOneAsync(
+                    new BsonDocument("Name", "Petunia"),
+                    new BsonDocument("Sunlight", Sunlight.partial));
+                Assert.AreEqual(1, updateResult.MatchedCount);
+                Assert.AreEqual(1, updateResult.ModifiedCount);
+            }
+            {
+                var foo = plantsCollection.FindAsync(
+                 new BsonDocument("Partition", "Store 47")).Result;
```
nirinchev, line 133: This should use `await` rather than .Result.

source/dotnet/mongodb.txt
```diff
@@ -15,1171 +15,301 @@ from a .NET application using the {+service-short+} .NET SDK.
...
-
-Running this snippet produces output resembling the following:
-
-.. code-block:: text
-
-   V/EXAMPLE: successfully upserted a document with id: BsonObjectId{value=5f19...}
+Both :dotnet-sdk:`UpdateOneAsync() <reference/Realms.Sync.MongoClient.Collection-1.html#Realms_Sync_MongoClient_Collection_1_UpdateOneAsync_System_Object_System_Object_System_Boolean_>` 
+and 
+:dotnet-sdk:`UpdateManyAsync() <reference/Realms.Sync.MongoClient.Collection-1.html#Realms_Sync_MongoClient_Collection_1_UpdateManyAsync_System_Object_System_Object_System_Boolean_>`
+have an option Boolean property that specifies whether the update should be an
```
nirinchev, line 185: `option` -> `optional`?

source/dotnet/mongodb.txt
```diff
@@ -15,1171 +15,301 @@ from a .NET application using the {+service-short+} .NET SDK.
...
+and 
+:dotnet-sdk:`UpdateManyAsync() <reference/Realms.Sync.MongoClient.Collection-1.html#Realms_Sync_MongoClient_Collection_1_UpdateManyAsync_System_Object_System_Object_System_Boolean_>`
+have an option Boolean property that specifies whether the update should be an 
+upsert (that is, if the document doesn't exist, it should be created). By default,
+no upsert is performed. 
+
+The following example looks for a plant whose "name" property is 
+"Pothos", "type" property is "perennial", and "sunlight" property is "full". 
+If a plant matches these criteria, the method updates the plant's "_partition" 
+value to "store 42". If no plant exists in the collection with that name, the
```
nirinchev, line 192: The quotes are a bit overwhelming here - should we use backticks for property names (so they're rendered as code) and quotes for property values?

nirinchev (commented):
I've made a couple of suggestions for cleanup/clarification + a few comments about fixing the serialization mapping. I'll take another look once the serialization issues have been addressed as I imagine there'll be quite some changes in response to that. 

On a separate note - for the `XxxOne` methods (find, update, etc.), might be worth adding an example that uses a `$sort` to ensure a stable ordering.

examples/dotnet/Examples/AggregationExamples.cs
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        [Test]
+        public async Task Filters()
+        {
+            // :code-block-start: agg_filter
+            var matchStage = new BsonDocument("$match",
+                    new BsonDocument("type",
+                        new BsonDocument("$eq",
+                            PlantType.Perennial)));
```
nirinchev, line 162: Hm... will this work though - in the Plant POCO we've annotated the property to serialize enums as strings and this will instead use an int. I guess we need to manually type it out, since using `PlantType.Perennial.ToString()` will give us the PascalCase version and we're supposed to be storing the lowercase one.

examples/dotnet/Examples/AggregationExamples.cs
```diff
@@ -0,0 +1,229 @@
...
+            var projectStage = new BsonDocument("$project",
+                new BsonDocument
+                {
+                    { "_id", 0 },
+                    { "_partition", 1 },
+                    { "type", 1 },
+                    { "name", 1 },
+                    { "storeNumber",
+                        new BsonDocument("$arrayElemAt",
+                        new BsonArray {
```
nirinchev, line 198: This line needs to be indented to show it's a continuation from the previous line. Would also be nice to have the equivalent of this stage in `BsonDocument.Parse` syntax for people who are familiar with the shell syntax.

examples/dotnet/Examples/MongoDBExamples.cs
```diff
@@ -121,53 +102,96 @@ public async Task InsertsMany()
...
         }
 
         [Test]
         public async Task ReadsDocuments()
         {
             // :code-block-start: mongo-find-one
-            var petunia = await plantsCollection.FindOneAsync<Plant>(
-                new BsonDocument("name", "petunia"), null);
+            var petunia = await plantsCollection.FindOneAsync(
+                new BsonDocument("name", "Petunia"), null);
```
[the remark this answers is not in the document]
nirinchev: Perhaps keep most examples using BsonDocument as that's the most versatile way to define complex structures, but mention that anonymous objects still work. The main issue with anon objects is that dollar-prefixed fields are invalid property names in C# so you can't do something like `var foo = new { $set = new { _partition = 5 }}`.

nirinchev (commented):
Starting to look great - I like how the examples have cleaned up nicely and are easier to read now. I have some followup suggestions/corrections, but we're getting close to the finish line.

source/dotnet/mongodb.txt
```diff
@@ -15,1171 +15,316 @@ from a .NET application using the {+service-short+} .NET SDK.
...
 
+
 Example Data
 ------------
 
 .. include:: /includes/mongodb-data-access-example-plant-documents.rst
 
+Mapping Classes
+~~~~~~~~~~~~~~~
+When working with objects in MongoDB, you should create .NET classes (POCOs) that
```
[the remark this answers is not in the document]
MongoCaleb: We tend to use "should" in the docs because we want to be prescriptive WRT best practices. We're not saying they must. :)

source/dotnet/mongodb.txt
```diff
@@ -15,1171 +15,316 @@ from a .NET application using the {+service-short+} .NET SDK.
...
-.. _dotnet-mongodb-aggregate:
-
-Aggregate Documents in a Collection
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. note::
 
-You can execute an aggregation pipeline using
-:java-sdk:`collection.aggregate()
-<io/realm/mongodb/mongo/MongoCollection.html#aggregate-java.util.List->`.
+   {+compass+} provides a utility for building aggregation pipelines and exporting
```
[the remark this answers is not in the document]
MongoCaleb: @nirinchev yes, it's 92% cool. ;-) It is copy-pastable, but some of the comments/nits you have suggested are a result of me doing just that. So, in other words, it _works_, but might not always be the best way.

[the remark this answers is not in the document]
nirinchev: Still, better than nothing! I'll try to find some time during skunkworks to give it a spin :)

examples/dotnet/Examples/Task.cs
```diff
@@ -1,32 +1,33 @@
...
 
 namespace dotnet
 {
-    public class RealmTask : RealmObject
+    public class Task : RealmObject
     {
         [PrimaryKey]
-        [MapTo("_id")]
-        public ObjectId Id { get; set; }
+        [BsonElement("_id")]
```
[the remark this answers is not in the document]
MongoCaleb: Thanks for the clarification!

examples/dotnet/Examples/Plant.cs
```diff
@@ -1,53 +1,55 @@
...
+        Green,
+        Yellow,
+        Purple
     }
-
     public enum PlantType
     {
-        perennial,
-        annual
+        Perennial,
```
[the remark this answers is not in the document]
MongoCaleb: I think for the enum values, this suffices...unless you feel strongly about it.

examples/dotnet/Examples/Plant.cs
```diff
@@ -1,53 +1,55 @@
...
         public PlantType Type { get; set; }
 
-        [MapTo("_partition")]
+        [BsonElement("_partition")]
         public string Partition { get; set; }
 
-
         public Plant()
         {
+            this.Id = ObjectId.GenerateNewId();
```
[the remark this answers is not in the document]
MongoCaleb: oooo! I knew about adding that to the `set`, but didn't know about this clever notation!

examples/dotnet/Examples/Plant.cs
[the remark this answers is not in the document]
nirinchev: Don't feel strongly - just we should probably update the example data to use PascalCase enum values then. Otherwise we're implying that casing doesn't matter for enum members.

source/dotnet/mongodb.txt
```diff
@@ -15,1171 +15,316 @@ from a .NET application using the {+service-short+} .NET SDK.
...
-   V/EXAMPLE: successfully found a document: Document{{_id=5f18..., name=venus flytrap, sunlight=full, color=white, type=perennial, _partition=Store 42}}
+.. literalinclude:: /examples/generated/code/start/MongoDBExamples.codeblock.mongo-find-one.cs
+   :language: csharp
 
 .. _dotnet-mongodb-findMany:
 
 Find Multiple Documents
 ~~~~~~~~~~~~~~~~~~~~~~~
+The following example shows how to find all plants where the "type" property is 
+"perennial":
```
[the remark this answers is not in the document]
MongoCaleb: That's covered in the next PR - https://github.com/mongodb/docs-realm/pull/542. :)

examples/dotnet/Examples/MongoDBExamples.cs
```diff
@@ -121,53 +102,96 @@ public async Task InsertsMany()
...
+                var updateResult = await plantsCollection.UpdateManyAsync(
+                    new BsonDocument("_partition", "Store 47"),
+                    new BsonDocument("$set", new BsonDocument("_partition", "Area 51")));
+                // :code-block-end:
+                Assert.AreEqual(1, updateResult.MatchedCount);
+                Assert.AreEqual(1, updateResult.ModifiedCount);
+            }
+            {
+                // :code-block-start: mongo-upsert
+                var filter = new BsonDocument("name", "Pothos")
```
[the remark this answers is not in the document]
MongoCaleb: @nirinchev FYI, this is an example of what Compass outputs.

examples/dotnet/Examples/AggregationExamples.cs
MongoCaleb: The tests all pass, so something about is right. ¯\_(ツ)_/¯ :-)

examples/dotnet/Examples/AggregationExamples.cs
MongoCaleb: Fixed and added another code example to call out the `Parse` approach.

source/dotnet/mongodb.txt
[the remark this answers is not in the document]
nirinchev: But that one said to only look at the custom data section, nothing about projections 😄 Anyway, if it's gonna get covered later, I'll just take a look then.

examples/dotnet/Examples/AggregationExamples.cs
nirinchev: That's somewhat suspicious 🤔 Can you use the shell/compass and see what the data looks like in MongoDB for that collection? I wonder if there's a serialization bug and the `BsonRepresentation` attribute gets ignored.

PR #542, opened 2020-10-13
12475

MongoCaleb:
# IMPORTANT
Because this PR also includes code in another PR, please **IGNORE ALL FILES** except `access-custom-user-data.txt`!
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12475

https://docs-mongodbcom-staging.corp.mongodb.com/realm/caleb/12475/dotnet/access-custom-user-data.html


examples/dotnet/Examples/CustomUserDataExamples.cs
```diff
@@ -0,0 +1,109 @@
...
+        MongoClient.Collection<CustomUserData> cudCollection;
+
+
+        [OneTimeSetUp]
+        public async Task Creates()
+        {
+            // :code-block-start: create
+            app = App.Create(myRealmAppId);
+            user = await app.LogInAsync(Credentials.EmailPassword("foo@foo.com", "foobar"));
+            var config = new SyncConfiguration("ufda", user);
```
nirinchev, line 30: Should these be in the example code block?

source/dotnet/access-custom-user-data.txt
```diff
@@ -0,0 +1,111 @@
...
+document containing the user ID of the currently logged in user and a
+``favoriteColor`` value into the custom user data collection:
+
+.. literalinclude:: /examples/generated/code/start/CustomUserDataExamples.codeblock.create.cs
+   :language: csharp
+
+You can add any number of arbitrary fields and values to the custom user
+data document when you create it. The user ID field is the only
+required fields. You may find it helpful to create a C# class (POCO) that 
+represents the custom user data. UYou can then serialize/deserialize this class
```
nirinchev, line 58:
```suggestion
represents the custom user data. You can then serialize/deserialize this class 
```

examples/dotnet/Examples/CustomUserDataExamples.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            await cudCollection.DeleteManyAsync();
+        }
+
+    }
+
+    // :code-block-start: cud
+    public class CustomUserData
+    {
+        [Required]
```
nirinchev, line 94: `Required` has no effect on this class as it's a RealmObject attribute.

examples/dotnet/Examples/CustomUserDataExamples.cs
```diff
@@ -0,0 +1,109 @@
...
+        public string _id { get; set; }
```
nirinchev, line 95: Probably make the setter private since we only want to set it in the constructor?

nirinchev (commented):
Looking good! On a non-docs related note, do you think I should change the `CustomData` property to be instead `GetCustomData` method with a generic overload:

```csharp
public BsonDocument GetCustomData() {}
public T GetCustomData<T>() {}
```

examples/dotnet/Examples/CustomUserDataExamples.cs
nirinchev: But the sync config should at least be removed, right? And if we're keeping the login part, let's use non-foo email/password :P

source/dotnet/access-custom-user-data.txt
MongoCaleb: oops

MongoCaleb:
> do you think I should change the `CustomData` property to be instead `GetCustomData` method with a generic overload:
> 
> ```cs
> public BsonDocument GetCustomData() {}
> public T GetCustomData<T>() {}
> ```

That would be sick! How it is now is also easy/great, but being able to deserialize in one call would be nice to have, for sure!

nirinchev (approved):
Looks good - only https://github.com/mongodb/docs-realm/pull/542#discussion_r504299181 remains to be addressed + will need a few code changes once I make `CustomData` into a method (sorry!)

ccho-mongodb (approved):
LGTM! Left some nitpicks and questions.

PR #544, opened 2020-10-14
Dotnet encrypt

cbush:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12481

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/dotnet-encrypt/dotnet/encrypt.html


source/dotnet/encrypt.txt
```diff
@@ -0,0 +1,60 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+You can the encrypt the {+realm+} database file on disk with AES-256 +
```
nirinchev, line 18:
```suggestion
You can encrypt the {+realm+} database file on disk with AES-256 +
```

source/dotnet/encrypt.txt
```diff
@@ -0,0 +1,60 @@
...
+
+Considerations
+--------------
+
+Storing & Reusing Keys
+~~~~~~~~~~~~~~~~~~~~~~
+
+You must pass the same encryption key when opening the encrypted {+realm+}
+again. Apps should store the encryption key in the :xamarin:`Xamarin Secure
+Storage <essentials/secure-storage>` or your platform's secure key/value pair
```
nirinchev, line 37: This is worded a bit imprecisely. Xamarin Secure Storage is a library for accessing a platform-specific secure storage API - it doesn't store anything, so "store the key in the Xamarin Secure Storage" is misleading. Since that library only supports a subsets of the platform the .NET SDK supports, I would probably reword that slightly to something like:

```
Apps should securely store the encryption key - typically in the target platform's secure key/value pair storage. There are some cross-platform secure storage libraries, such as Xamarin Secure Storage that can simplify that for mobile apps, but ultimately it's the developer's responsibility to ensure that the key cannot be easily extracted by an attacker.
```

source/dotnet/encrypt.txt
```diff
@@ -0,0 +1,60 @@
...
+storage.
+
+Performance Impact
+~~~~~~~~~~~~~~~~~~
+
+Typically, reads and writes on encrypted {+realm+}s can be up to 10%
+slower than unencrypted {+realm+}s.
```
nirinchev, line 44:
```suggestion
slower than on unencrypted {+realm+}s.
```

Perhaps?

source/dotnet/encrypt.txt
```diff
@@ -0,0 +1,60 @@
...
+
+Encryption and Realm Sync
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can encrypt a :ref:`synced realm <dotnet-sync-data>`. {+service+} only
+encrypts the data on the device and stores the data unencrypted in your
```
nirinchev, line 50: It's probably a bit redundant, but might be worth mentioning that the data is encrypted during transfer between the app and the server? To avoid ambiguity and to ease people's potential security concerns.

examples/dotnet/Examples/EncryptionExamples.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class EncryptionExamples
+    {
+        [Test]
+        public void EncryptRealm()
+        {
+            // :code-block-start: encrypt_a_realm
+            var config = new RealmConfiguration("Mine.realm");
+            config.EncryptionKey = new byte[64] // Key MUST be exactly this size
+            {
+                0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
```
nirinchev, line 16: Rather than show a fixed encryption key, which would be a bad practice, wouldn't it make more sense to show a realistic example where we generate a random key using [RNGCryptoServiceProvider](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rngcryptoserviceprovider.getbytes?view=netcore-3.1#System_Security_Cryptography_RNGCryptoServiceProvider_GetBytes_System_Byte___):

```csharp
// Check if we already have a key stored in the platform's secure storage.
// If we don't, generate a new one:

var encryptionKey = new byte[64];
using var rng = new RNGCryptoServiceProvider();
rng.GetBytes(encryptionKey);

// Store the key securely to be used next time we want to open the Realm.

var config = new RealmConfiguration
{
    EncryptionKey = encryptionKey
};

var realm = Realm.GetInstance(config);
```

examples/dotnet/README.md
```diff
@@ -31,19 +31,13 @@
 
    f. Back in the Manage Nuget Packages dialog, select your new source in the
       dropdown, and your should see your two packages. Select botyh, and confirm in the right
```
MongoCaleb, line 33: your ==> you

source/dotnet/encrypt.txt
```diff
@@ -0,0 +1,60 @@
...
+You must pass the same encryption key when opening the encrypted {+realm+}
+again. Apps should store the encryption key in the :xamarin:`Xamarin Secure
+Storage <essentials/secure-storage>` or your platform's secure key/value pair
+storage.
+
+Performance Impact
+~~~~~~~~~~~~~~~~~~
+
+Typically, reads and writes on encrypted {+realm+}s can be up to 10%
+slower than on unencrypted {+realm+}s.
```
MongoCaleb, line 44: consider if it's worth adding an "and so...:" sentence here. "So don't ever use encryption", for example... ;-)

MongoCaleb (approved):
LGTM with one note

PR #551, opened 2020-10-15
Client Resets

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12477

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/12477/dotnet/client-reset.html

### DIFF
--- a/source/dotnet/client-reset.txt
+++ b/source/dotnet/client-reset.txt
@@ -1,0 +1,58 @@
+.. _dotnet-client-resets:
+.. _dotnet-client-reset:
+
+=============
+Client Resets
+=============
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+When using :doc:`{+sync+} </sync>`, a **client reset** is a task that your client
+app must handle in the following situations:
+
+- The given synced {+realm+} on the server was restored from a backup.
+  For example, due to a {+service-short+} server crash.
+
+- The client app made changes to that {+realm+} since the backup was made,
+  but did not sync those changes back to the server before the server
+  was restored.
+
+In other words, the client app must perform a client reset
+on a given synced {+realm+} if the server is restored to a
+version older than the version on the client.
+
+.. warning::
+
+   A client reset erases all local data and downloads a new copy of the
+   data stored in MongoDB Atlas. Performing a client reset loses all local
+   changes made since the client last successfully synced.
+
+
+Example
+-------
+
+By default, the SDK performs client resets automatically on startup when
+instructed to do so by the server. When a client
+reset occurs, the SDK creates a backup of local data. By default,
+the SDK makes no effort to recover lost local changes from this backup.
+
+If a reset is required, {+realm+} sends an 
+:dotnet-sdk:`ClientResetException <reference/Realms.Sync.Exceptions.ClientResetException.html>`. 
+Your code should add a listener on the 
+:dotnet-sdk:`Session.Error <reference/Realms.Sync.Session.html#Realms_Sync_Session_Error>` 
+event and handle it by calling the 
+:dotnet-sdk:`InitiateClientReset() <reference/Realms.Sync.Exceptions.ClientResetException.html#Realms_Sync_Exceptions_ClientResetException_InitiateClientReset>` 
+method. The following example demonstrates this:
+
+.. literalinclude:: /examples/generated/code/start/ClientResetExamples.codeblock.handle.cs
+   :language: csharp
+
--- a/source/dotnet/client-reset.txt
+++ b/source/dotnet/client-reset.txt
@@ -16,19 +16,17 @@ Client Resets
 Overview
 --------
 
-When using :doc:`{+sync+} </sync>`, a **client reset** is a task that your client
-app must handle in the following situations:
+When using :doc:`{+sync+} </sync>`, a **client reset** is a possible issue that 
+your client app should handle. While not common, it can be caused by the following 
+situation:
 
-- The given synced {+realm+} on the server was restored from a backup.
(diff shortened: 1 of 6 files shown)


source/dotnet/client-reset.txt
```diff
@@ -0,0 +1,58 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+When using :doc:`{+sync+} </sync>`, a **client reset** is a task that your client
```
nirinchev, line 19: I wouldn't call it a task - more like "situation"?

source/dotnet/client-reset.txt
```diff
@@ -0,0 +1,58 @@
...
+app must handle in the following situations:
+
+- The given synced {+realm+} on the server was restored from a backup.
+  For example, due to a {+service-short+} server crash.
```
nirinchev, line 23: This is not applicable to the new server - I don't believe there's a way to restore it from backup.

source/dotnet/client-reset.txt
```diff
@@ -0,0 +1,58 @@
...
+
+- The client app made changes to that {+realm+} since the backup was made,
+  but did not sync those changes back to the server before the server
+  was restored.
+
+In other words, the client app must perform a client reset
+on a given synced {+realm+} if the server is restored to a
+version older than the version on the client.
```
nirinchev, line 31: I believe with the current sync architecture, the most likely causes of client reset are:
1. Sync was terminated and then reenabled - for example to change the schema in a destructive manner or because of a cluster upgrade (not sure if those terminate sync, so might want to check up with the Realm Cloud team).
2. A bug (but probably shouldn't mention it here 😅)

source/dotnet/client-reset.txt
```diff
@@ -0,0 +1,58 @@
...
+
+Example
+-------
+
+By default, the SDK performs client resets automatically on startup when
+instructed to do so by the server. When a client
+reset occurs, the SDK creates a backup of local data. By default,
+the SDK makes no effort to recover lost local changes from this backup.
+
+If a reset is required, {+realm+} sends an
```
nirinchev, line 48:
```suggestion
If a reset is required, {+realm+} sends a 
```

source/dotnet/client-reset.txt
```diff
@@ -0,0 +1,58 @@
...
+   data stored in MongoDB Atlas. Performing a client reset loses all local
+   changes made since the client last successfully synced.
+
+
+Example
+-------
+
+By default, the SDK performs client resets automatically on startup when
+instructed to do so by the server. When a client
+reset occurs, the SDK creates a backup of local data. By default,
```
nirinchev, line 45: I don't think we should mention this backup at all unless we explain how to access it and what can be done. Since I'm not even sure there's a way to read it or a way to figure out where it's stored, we might as well pretend it doesn't exist.

nirinchev (approved):
Just a few nits from me.

MongoCaleb:
@nirinchev back to you, although you 👍 it, because it was a hefty rewrite. Thanks!

source/dotnet/client-reset.txt
```diff
@@ -0,0 +1,50 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+When using :doc:`{+sync+} </sync>`, a **client reset** is a possible issue that 
+your client app should handle. While not common, it can be caused when Sync was 
+terminated on the server. For example, a destructive schema change might cause 
+Sync to stop. Meanwhile, the user made changes on the client while Sync was
```
nirinchev, line 22: I feel that is a bit misleading - a destructive schema change can't cause sync to stop on its own, because there are safeguards that prevent destructive schema changes. But let's say a developer wants to roll out a new feature and they accidentally mess up the schema - for example, they declare a field double when in reality, it should have been a decimal. Changing that would be a destructive schema change which is forbidden by sync, so if they really wanted to do it, they'd need to terminate sync, modify their schema, then reenable sync, causing existing clients to experience a client reset.

source/dotnet/client-reset.txt
```diff
@@ -0,0 +1,50 @@
...
+instructed to do so by the server and makes no effort to preserve local 
+changes.
+
+However, when a reset is required, {+realm+} sends a  
+:dotnet-sdk:`ClientResetException <reference/Realms.Sync.Exceptions.ClientResetException.html>`. 
+You can add a listener on the 
+:dotnet-sdk:`Session.Error <reference/Realms.Sync.Session.html#Realms_Sync_Session_Error>` 
+event and handle the client reset by calling the 
+:dotnet-sdk:`InitiateClientReset() <reference/Realms.Sync.Exceptions.ClientResetException.html#Realms_Sync_Exceptions_ClientResetException_InitiateClientReset>` 
+method. The following example demonstrates this:
```
nirinchev, line 46: Might be worth mentioning what's the point of doing that as it's not immediately clear if everything is automatic, why should the developer bother subscribing to the event and calling the `InitiateClientReset` method. This could also be done in pseudocode in the snippet. Something like:

```
// Prompt user to perform client reset immediately. If they don't do it,
// they won't receive any data from the server until they restart the app
// and all changes they make will be discarded when the app restarts.

var didUserConfirmReset = true;
if (didUserConfirmReset)
{
    // Close the Realm before doing the reset as it'll need
    // to be deleted and all objects obtained from it will be
    // invalidated.
    openedRealm.Dispose();
    var didReset = clientResetEx.InitiateClientReset();
    if (didReset)
    {
        // Navigate the user back to the main page or reopen the
        // the Realm and reinitialize the current page.
    }
    else 
    {
        // Reset failed - notify user that they'll need to restart the app.
    }
}
```

nirinchev (approved):

PR #552, opened 2020-10-16
Realm Quick Start with SwiftUI and Combine

cbush:
## Staged view
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/swiftui/ios/swiftui.html

## Jira
https://jira.mongodb.org/browse/DOCSP-11771
https://jira.mongodb.org/browse/DOCSP-11773

## Update 10/22 🌱

Additional refinement on top of @jsflax suggestions:

- Added a logout button to keep the AppState out of child views that don't really need it.
- Changed item to plain var (instead of @ObservedObject) in leaf views as they should not update upon change. This was causing a crash upon delete of an edited item.
- AppState does not retain the realm, but only keeps the group's items. The idea being "retain only what you need".
- Solidified the shouldIndicateActivity logic and ensuring no unwanted activity indicators.

## Overview

The app presents a user with a login screen to log in anonymously. Once logged in, the realm is opened asynchronously, simulating the architecture of a Realm Sync-based app. Realm Sync is not actually required. Once the user is logged in and the realm is ready, the app navigates to an items screen, where the user can add, update, delete, and rearrange items in a list. The items belong to a single group.

### Run the app
- Check out the PR branch.
- Go to examples/ios
- Run `pod install --repo-update`
- Open `Examples.xcworkspace`
- In Xcode, select the "QuickStartSwiftUI" target in the target selector (top left next to the Run button)
- Click the Run button.

### Log in
- When the app launches, you should be on the login screen.
- Click the button to log in anonymously.
- You should be taken to an items screen.

### Add and delete some items
- Click the "+" button in the bottom right corner to add random items.
- Swipe to delete some items.

### Rearrange items
- Click "Edit" in the top right to go into edit mode.
- Rearrange some items in the list.
- Click "Done" to exit edit mode.

### Edit an item
- Click an item in the list to go to the Item details page.
- Click the text field in the middle of the screen. The software keyboard should come up. Type something in.
- Press "return" and see the item change.

### Restart the app & log out
- Restart the app. You should go straight to the items screen since the same user is reused. 
- On the items screen, click log out to go back to the login screen.

## Known Issues / TODO (aka Monday's Problems)
- ~Click an item in the list to go to its details page. Edit its name and save. Go back to the list. Click the item again. Result: app does not go back into that item's details page until you select another item first.~
- "Group" and "items" are pretty unimaginative. Maybe rename to "Bags" and "Items" as if in an RPG inventory?
- TODO: Text writeup for actual quick start (this PR is code only)

## JIRA
https://jira.mongodb.org/browse/DOCSP-11773


ericjordanmossman (commented):
LGTM , except I know @jsflax has thoughts on not using wrappers for realm login/logout in Auth - the alternative being placing our login/logout directly in the views.

examples/ios/QuickStartSwiftUI/QuickStart.swift
```diff
@@ -0,0 +1,324 @@
...
+
+    init() {
+        let auth = Auth()
+        self.groupOwner = GroupOwner(auth: auth)
+        self.auth = auth
+    }
+
+    var body: some View {
+        // When `auth` or `groupOwner` change, this will be reevaluated.
+        if !auth.isLoggedIn || groupOwner.group == nil {
```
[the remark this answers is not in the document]
cbush: Tried that, but it does not work (on Xcode 11.3.1 anyway):
image

cbush:
Until Realm App becomes observable, I think wrapping the auth component is pretty reasonable.

examples/ios/QuickStartSwiftUI/QuickStart.swift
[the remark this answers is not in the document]
ericjordanmossman: You're right. I get the same in Xcode 11.3. Appears to work only in Swift >=5.3. Definitely keep the original that builds consistently.

examples/ios/QuickStartSwiftUI/QuickStart.swift
```diff
@@ -0,0 +1,354 @@
...
+        // for the given configuration.
+        if let user = app.currentUser {
+            openRealm(user: user)
+        }
+
+        // monitor log in state and open a realm on login
+        loginPublisher.receive(on: DispatchQueue.main).sink(
+            receiveCompletion: { _ in },
+            receiveValue: { user in
+                self.openRealm(user: user)
```
jsflax, line 88: We should remove `onRealmOpened` and `openRealm` in lieu of chaining the new `Realm.asyncOpen` Publisher.

examples/ios/QuickStartSwiftUI/QuickStart.swift
```diff
@@ -0,0 +1,354 @@
...
+/// Random noun for more interesting demo item names
+let randomNouns = [
+    "floor", "monitor", "hair tie", "puddle", "hair brush", "bread",
+    "cinder block", "glass", "ring", "twister", "coasters", "fridge",
+    "toe ring", "bracelet", "cabinet", "nail file", "plate", "lace",
+    "cork", "mouse pad"
+]
+
+/// An individual item. Part of a `Group`.
+final class Item: Object, ObjectKeyIdentifiable {
```
nirinchev, line 27: Out of curiosity, why are we modeling items as objects vs embedded objects? From reading the code, it doesn't seem like they need to be standalone objects from Realm's perspective. Is it to guard against inserting a billion items and hitting the 16 MB document limit?

nirinchev (commented):
Perhaps I'm late to the party and definitely don't want to sound like I'm questioning the fundamentals here, but I'm not sure I understand why we're making the sync part optional rather than creating an app/logging in a user altogether? At least to me, it sounds like for people using the local database, signing up for Atlas and creating a MongoDB Realm app would be a much bigger deterrent than enabling sync would be for people who already have an app. So instead of having `let USE_REALM_SYNC = false` in the middle of the code, I would place a `let appId = ""` at the top and if people set it to non-empty, go with the sync flow, otherwise demonstrate the non-sync flow. That being said, I can see how this will require a bit of restructuring the app and can understand if people are opposed to it.

examples/ios/QuickStartSwiftUI/QuickStart.swift
jsflax: I would agree that EmbeddedObjects seem like the right call– I imagine @cbush was considered that, if we actually fleshed out this app, we may want standalone `Items`. But in the interim, I agree.

jsflax:
I think that's a fair point @nirinchev . We already have a quick start on SwiftUI in the repo for the local DB– this quick start could be for Sync only (unless @cbush had other intentions).

examples/ios/QuickStartSwiftUI/QuickStart.swift
nirinchev: The main appeal of embedded objects for me is that they are much easier to reason about when inserted into MongoDB - it's more manageable to see how the list looks like without having to mentally map `_id`s. As an added benefit, it highlights a new feature in the local database that many people might not be aware exists.

examples/ios/QuickStartSwiftUI/QuickStart.swift
cbush: Ah, cool. Is it okay to always use asyncOpen(), then?

examples/ios/QuickStartSwiftUI/QuickStart.swift
cbush: That's exactly what I was thinking @jsflax, but I hadn't even considered using EmbeddedObjects here. I think even if someone expanded this app, then in the typical item/group (thing/container, employee/store, ingredient/recipe, whatever) use case they still might use EmbeddedObject here. So, I'm definitely in favor of making that change.

cbush:
I think you're right. Something was irking me during the write-up -- I realized that setting up Sync is not that much more of a stretch if you're already setting up an app. I will rework it on Monday, probably do as you suggest where if the app ID is blank then we skip the login/sync flow altogether.

cbush:
For some context this was originally intended to be a Sync-only demo and the "optionality" was a bonus in light of our effort to make things nicer for the local-only use case. Now I'm realizing we should go the whole way.

cbush:
@nirinchev @jsflax updated for the local-only use case as well as the asyncOpen publisher -- please take another look and let me know where it's getting messy (I think it's fairly clean). I'll be looking into EmbeddedObjects on Monday because it wasn't quite as straightforward as I thought.

source/examples/generated/code/start/QuickStart.codeblock.complete-swiftui-combine-quick-start.swift
```diff
@@ -0,0 +1,394 @@
...
+            VStack {
+                // The list shows the items in the realm.
+                List {
+                    // ⚠️ ALWAYS freeze a Realm list while iterating in ForEach().
+                    // Otherwise, unexpected behavior will occur (especially when
+                    // deleting object from the list).
+                    ForEach(items.freeze()) { frozenItem in
+                        // "Thaw" the item before passing it in, as ItemRow
+                        // may want to edit it, and cannot do so on a frozen object.
+                        ItemRow(item: items.realm!.resolve(ThreadSafeReference(to: frozenItem))!)
```
astigsen, line 279: This says "ALWAYS", but isn't there only a need to freeze the list if there is chance of modifications _while_ iterating over it?

That does not seem to be the case here (deletes and moves do not seem to happen during iteration), so won't the freezing and thawing just be overhead?

source/examples/generated/code/start/QuickStart.codeblock.complete-swiftui-combine-quick-start.swift
```diff
@@ -0,0 +1,394 @@
...
+/// Represents a screen where you can edit the item's name.
+struct ItemDetailsView: View {
+    var item: Item
+
+    // ⚠️ Beware using a Realm object or its properties directly in a @Binding.
+    // Writes to Realm objects MUST occur in a transaction (realm.write() block),
+    // but a default Binding will not do that for you. Therefore, either use a
+    // separate @State object to hold the data before writing (as we do here),
+    // or create a custom Binding that handles writes in a transaction.
+    @State var newItemName: String = ""
```
astigsen, line 355: Do we have any examples of how to do a custom binding that automatically does the transaction wrapping?

cbush: No, but I could definitely add one!

source/examples/generated/code/start/QuickStart.codeblock.complete-swiftui-combine-quick-start.swift
cbush: If we don't freeze while iterating, and then delete an object, we end up with an "index out of bounds" exception. I suspect it has to do with how SwiftUI handles re-rendering a list, but maybe @tgoyne or @jsflax can speak more to that.

I can clarify this "ALWAYS" applies very specifically to SwiftUI views.

nirinchev: Does Cocoa not expose snapshotting collections? That should be more seamless than freezing and thawing stuff.

astigsen: I'd love to understand how that can happen. From the above code there does not seem to be any way that the delete can happen during iteration. Does SwiftUI store the list you pass to `ForEach` and do its own iteration during deletes, or is there something else going on?

source/examples/generated/code/start/QuickStart.codeblock.complete-swiftui-combine-quick-start.swift
astigsen: Having to add custom commit methods and local variables to keep intermediate values seems like a lot of boilerplate to put on users. It would be nice if we could remove the need for that kind of busy work.

cbush: Agreed. Meanwhile, here's an example. https://github.com/mongodb/docs-realm/pull/552/commits/b376326d0f142153b1a69d3ccf4aa33206b408ab

jsflax: We've been chatting about doing this for users– it hasn't been prioritised yet until a few other features have fallen into place.

source/examples/generated/code/start/QuickStart.codeblock.complete-swiftui-combine-quick-start.swift
jsflax: SwiftUI stores several copies of the list depending on the state of the view, hence the need to freeze it. When you delete from a List, it still displays on the screen briefly, so the view still shows a previously cached copy of the list. Because Realm's List's are "live", the caching doesn't actually matter and it will try to display an element from the list that no longer exists.

astigsen: Ok, so it seems that they have basically copied React's way of doing rendering. The whole premise of table views in iOS was to be virtual views which can handle more data than fits in memory. To try to shoehorn a web paradigm based on having a limited DOM that you can cache and diff in memory, to a native setting where your UI may be backed with datasets far bigger than memory seems like several steps backwards, but I guess that is what we have to work with 🤷‍♂️.

That said, I am not even convinced that the above implementation is guaranteed to be correct. If it does a re-render after a delete, and uses the frozen (pre-cached) version, and then thaws the individual entries, won't it also try to thaw the deleted entry?

I any case, I think we have to find a way to hide all these mechanics from the user. There is no way we can expect them to remember when to freeze, thaw, manually cache and all that other busy work.

source/examples/generated/code/start/QuickStart.codeblock.complete-swiftui-combine-quick-start.swift
astigsen: I definitely think that this is something we have to handle for the user. Otherwise it will turn into a crazy amount of boilerplate when user have more than just a token number of properties that can be modified.

source/examples/generated/code/start/QuickStart.codeblock.complete-swiftui-combine-quick-start.swift
```diff
@@ -0,0 +1,394 @@
...
+
+    var body: some View {
+        NavigationView {
+            VStack {
+                // The list shows the items in the realm.
+                List {
+                    // ⚠️ ALWAYS freeze a Realm list while iterating in ForEach().
+                    // Otherwise, unexpected behavior will occur (especially when
+                    // deleting object from the list).
+                    ForEach(items.freeze()) { frozenItem in
```
astigsen: Is the `ForEach` needed here? Wouldn't it be more concise to use the `List` directly?:

```Swift
List(items.freeze()) { frozenItem in
    ItemRow(item: items.realm!.resolve(ThreadSafeReference(to: frozenItem))!)
}
```

cbush: It would be more concise, but unfortunately the onDelete and onMove actions only exist on the ForEach.

jsflax: Yeah, `List` and `ForEach` are not the same thing in SwiftUI.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+   and assumes the number of elements in the list has not changed. Consequently,
+   the view would try to display a list element that no longer exists. Freezing
+   the list removes its "live" property, allowing the list copy to behave in
+   line with SwiftUI's expectation.
+
+Before passing each item to the ItemRow and ItemDetailsView views, we "thaw" the
+frozen object by fetching it again from the {+realm+}. This is only necessary
+because the ItemDetailsView may edit the item, and you cannot modify a frozen
+item. The top of the iteration is a convenient place to thaw the item, because
+we have access to the {+realm+} via the items list there. However, we must be
```
ccho-mongodb, line 175: I'm a little confused since I thought you were talking about the item list being frozen. Do you need access to the realm that contains them to call thaw? Is the realm no longer in scope later in the code? Maybe if I looked at the code, I would understand this better?

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+careful to treat the item as frozen until we intend to modify it.
+
+Note that the ItemRow and ItemDetailsView views do not observe their items.
+That is, they do not declare the item as ``@ObservedObject var item: Item``
+but rather ``var item: Item``. Otherwise, these views would suddenly find
+themselves observing an invalid object when an item gets deleted, which could
+cause the app to crash.
```
ccho-mongodb, line 182:
```suggestion
but rather ``var item: Item``. Otherwise, if these views observed an invalid
object caused by deleting an item, the app could crash.
```

ccho-mongodb: I think this is conveys the same message.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+   and re-render its list of ItemRows, so you do not need to and should not
+   observe individual items in a list.
+
+
+Modifying Properties of Presented Realm Objects
+```````````````````````````````````````````````
+
+The ItemDetailsView allows users to edit the name of an item with a TextField.
+We use a ``@State`` variable to hold the new name value entered by a user. It's
+important not to use the item's ``name`` property directly here, because that
```
ccho-mongodb, line 197:
```suggestion
important to avoid directly updating the item's ``name`` property here, because that
```

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+   You may only update a {+database+} object's properties in a transaction. If
+   we used the object's properties directly here, any edits to the TextField
+   would automatically apply to the item's name outside of a transaction,
+   resulting in a fatal exception.
+
+There are two approaches to binding an editable Realm object property to a
+SwiftUI control:
+
+- Use another variable to hold the value until ready to commit it in a write transaction.
+- You could also create a custom :apple:`binding <documentation/swiftui/binding>` that handles writes in a transaction.
```
ccho-mongodb, line 211:
```suggestion
- Create a custom :apple:`binding <documentation/swiftui/binding>` that handles writes in a transaction.
```

ccho-mongodb (approved):
LGTM! A few suggestions and notes on descriptions that weren't that clear to me.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+
+If you opted in to using :ref:`{+sync+} <enable-sync>`, the first screen is the
+LoginView. Otherwise, you start at the ItemsView.
+
+On the LoginView, you could implement :ref:`email/password authentication
+<ios-manage-email-password-users>` or :ref:`another authentication provider
+<authentication-providers>` here. For simplicity's sake, this uses
+:ref:`Anonymous authentication <anonymous-authentication>`.
+
+When you press the "log in" button, the app navigates to the ItemsView, where
```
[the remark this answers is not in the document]
cbush: I considered that, but I think that applies to Cloud UI elements more than in-app elements. Thoughts? Happy to change.

source/ios/swiftui.txt
cbush: Much better, thanks.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+- A randomly generated name, which the user can edit.
+- An ``isFavorite`` boolean property, which indicates whether the user "favorited" the item.
+
+A group contains items. You could extend the group to have a name and an
+association with a specific user, but that is out of scope of this quick start.
+
+State Objects
+~~~~~~~~~~~~~
+
+It's a good idea to keep app State separate from Views as much as possible. A
```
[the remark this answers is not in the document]
cbush: The latter is not passive voice ("to be" has subjects on both sides of the verb), but I like "We recommend" too.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+SwiftUI and Combine that we'll discuss in the following sections.
+
+Models
+~~~~~~
+
+A common {+service-short+} data modeling use case is to have "things" and
+"containers of things". This app defines two related Realm object models: Item
+and Group.
+
+An item has two properties:
```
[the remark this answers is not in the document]
cbush: Would you be okay with de-capitalizing on the previous lines? I'm trying to avoid my tendency to over-capitalize.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+~~~~~~~~~~~~~
+
+It's a good idea to keep app State separate from Views as much as possible. A
+View should only be aware of what it needs to be aware of. We define the
+AppState class to centralize the state of the app. As the app grows in
+complexity, we might consider breaking functionality into smaller classes that
+become members of the AppState class. For now, this class reports authentication
+events and holds the list of items.
+
+The AppState opens a {+realm+} in any of the following situations:
```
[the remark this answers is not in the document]
cbush: "Open" is the correct terminology.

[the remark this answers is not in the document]
cbush: How about something like this:

>When the app launches, the AppState opens a {+realm+} if either of the following
>conditions are true:
>
>- A user from a previous session still logged in.
>- We are not using {+sync+} and authentication at all.
>
>Otherwise, the AppState opens a {+realm+} when a user logs in.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+before opening it. When using the local-only {+realm+}, we can use the Realm
+constructor directly.
+
+Once the {+realm+} opens, the AppState retrieves the group in the {+realm+} or
+creates one if one does not exist already. The AppState stores the list of items
+in the group. Any views observing this state object will refresh themselves in
+response.
+
+On logout, the AppState clears its reference to the items. Again, any views
+observing this state object will refresh themselves in response.
```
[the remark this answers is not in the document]
cbush: Changed to:
>On logout, the AppState clears its reference to the items, and any views
>observing this state object refresh themselves in response.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+
+Views
+~~~~~
+
+The Main View: Switching Views Depending on the App State
```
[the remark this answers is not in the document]
cbush: I'm not sure if there's a 1:1 for views and concept explanations, but point taken. For consistency, I'll remove "The Main View:" here.

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+it in a SwiftUI view, even if you do not modify the list while iterating.
+Otherwise, unexpected behavior will occur when modifying the list, especially
+when deleting objects.
+
+.. note::
+
+   To understand why freezing a list is necessary, consider that SwiftUI caches
+   several copies of a list depending on the state of the view. When you delete
+   from a list, the view still shows a cached copy of that list for a moment.
+   However, Realm lists are "live", so the cached Realm list actually reflects
```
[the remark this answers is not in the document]
cbush: Hm... trying to tweak it:
>However, Realm lists are "live", so the cached copy actually reflects the latest state of the Realm list.

Thoughts?

source/ios/swiftui.txt
```diff
@@ -0,0 +1,234 @@
...
+   the latest state. SwiftUI expects a cached list to reflect the previous state
+   and assumes the number of elements in the list has not changed. Consequently,
+   the view would try to display a list element that no longer exists. Freezing
+   the list removes its "live" property, allowing the list copy to behave in
```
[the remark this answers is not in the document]
cbush: I think I will put a reference to ios-live-collections here.

source/ios/swiftui.txt
cbush: I'll just delete this line. It's explained in a comment in the code but just confusing here.

cbush (commented):
@ccho-mongodb thanks for the review! Updated some text.

source/ios/swiftui.txt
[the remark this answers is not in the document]
ccho-mongodb: According to the style guide, it applies to several UI elements including buttons: https://docs.mongodb.com/meta/style-guide/style/text-formatting/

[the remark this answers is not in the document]
cbush: Ah, perfect. Updated.

source/examples/generated/code/start/QuickStart.codeblock.complete-swiftui-combine-quick-start.swift
cbush: Stepping through with a debugger, SwiftUI iterates over the correct number of items after a delete when the list is frozen. If the list is not frozen and you delete any item other than the last one, the ForEach() immediately tries to grab an index that's out of bounds (i.e. the formerly last element). It seems items.freeze() is called on every re-render. I suspect for the internal diffing to work it needs to be able to compare with the previous version of the list, so it must be frozen. But when actually iterating, it does indeed iterate over the latest version.

PR #556, opened 2020-10-16
add user linking

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12483

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/link/dotnet/link-identities.html

### DIFF
--- a/examples/dotnet/Examples/Examples.cs
+++ b/examples/dotnet/Examples/Examples.cs
@@ -257,18 +257,31 @@ namespace UnitTests
         [Test]
         public async System.Threading.Tasks.Task LinksAUser()
         {
-            // :code-block-start: link
-            // 1) A user logs on anonymously:
-            var anonUser = await app.LogInAsync(Credentials.Anonymous());
-            // 2) They create some data, and then decide they want to save
-            //    it, which requires creating an Email/Password account.
-            // 3) We prompt the user to log in, and then use that new
-            //    EmailPassword credential to link the existing anonymous
-            //    account:
-            var officialUser = await anonUser.LinkCredentialsAsync(
-                Credentials.EmailPassword("caleb@example.com","shhhItsASektrit!"));
-            // :code-block-end:
-            
+            {
+                // :code-block-start: link
+                // 1) A user logs on anonymously:
+                var anonUser = await app.LogInAsync(Credentials.Anonymous());
+                // 2) They create some data, and then decide they want to save
+                //    it, which requires creating an Email/Password account.
+                // 3) We prompt the user to log in, and then use that info to
+                //    register the new EmailPassword user, and then generate an
+                //    EmailPassword credential to link the existing anonymous
+                //    account:
+                var email = "caleb@example.com";
+                var password = "shhhItsASektrit!";
+                await app.EmailPasswordAuth.RegisterUserAsync(
+                    email, password);
+                var officialUser = await anonUser.LinkCredentialsAsync(
+                   Credentials.EmailPassword(email, password));
+                // :code-block-end:
+            }
+            {
+                // :code-block-start: link2
+                var anonUser = await app.LogInAsync(Credentials.Anonymous());
+                var officialUser = await anonUser.LinkCredentialsAsync(
+                   Credentials.Google("<google-token>"));
+                // :code-block-end:
+            }
             return;
         }
 
--- a/examples/dotnet/Examples/Examples.cs
+++ b/examples/dotnet/Examples/Examples.cs
@@ -151,7 +151,7 @@ namespace UnitTests
             {
                 // :code-block-start: logon_EP
                 var user = await app.LogInAsync(
-                    Credentials.EmailPassword("caleb@mongodb.com", "shhhItsASektrit!"));
+                    Credentials.EmailPassword("caleb@example.com", "shhhItsASektrit!"));
                 // :code-block-end:
                 Assert.AreEqual(UserState.LoggedIn, user.State);
                 await user.LogOutAsync();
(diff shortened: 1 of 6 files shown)


source/dotnet/link-identities.txt
```diff
@@ -0,0 +1,34 @@
...
+
+Example
+-------
+
+For instance, consider an application that enables users to authenticate
+via :ref:`anonymous authentication <anonymous-authentication>` so they
+may explore functionality without linking to an existing
+SSO provider or generating a new password. When that user decides to
+finally create a full account with an SSO provider or email/password
+authentication, you need some way of persisting the user's original
```
nirinchev, line 24: Maybe?

```suggestion
authentication, you need some way of associating the user's original
```

examples/dotnet/Examples/Examples.cs
```diff
@@ -254,6 +254,24 @@ public async System.Threading.Tasks.Task CallsAFunction()
...
         }
 
+        [Test]
+        public async System.Threading.Tasks.Task LinksAUser()
+        {
+            // :code-block-start: link
+            // 1) A user logs on anonymously:
+            var anonUser = await app.LogInAsync(Credentials.Anonymous());
+            // 2) They create some data, and then decide they want to save
+            //    it, which requires creating an Email/Password account.
```
nirinchev, line 264: This isn't very precise - technically, their anonymous user data is still saved and they can come back to it on the device. I guess a better example will be if they want to access their data on another device or if they want to ensure it's not lost when they change phones.

examples/dotnet/Examples/Examples.cs
```diff
@@ -254,6 +254,24 @@ public async System.Threading.Tasks.Task CallsAFunction()
...
+            // 3) We prompt the user to log in, and then use that new
+            //    EmailPassword credential to link the existing anonymous
+            //    account:
+            var officialUser = await anonUser.LinkCredentialsAsync(
+                Credentials.EmailPassword("caleb@example.com","shhhItsASektrit!"));
```
nirinchev, line 269: This is unfortunately a very weird design on the server, but the user actually needs to first register an email/password account and then link it. So this example would look something like:

```csharp
var email = "caleb@example.com";
var password = "..";
await app.EmailPasswordAuth.RegisterUserAsync(email, password);
var officialUser = await anonUer.LinkCredentialsAsync(Credentials.EmailPassword(email, password));
```

nirinchev: I filed https://jira.mongodb.org/browse/DOCSP-12594 to fix this for the other SDKs some time ago.

nirinchev: Also, note that this is only applicable for email/password auth - all the other credential types don't need this extra step.

source/dotnet/link-identities.txt
```diff
@@ -0,0 +1,34 @@
...
+anonymous identity with their new permanent identity.
+
+You can link identities using the 
+:dotnet-sdk:`LinkCredentialsAsync() <reference/Realms.Sync.User.html#Realms_Sync_User_LinkCredentialsAsync_Realms_Sync_Credentials_>` 
+method on the :dotnet-sdk:`User <reference/Realms.Sync.User.html>` object of a 
+logged in user.
+
+.. literalinclude:: /examples/generated/code/start/Examples.codeblock.link.cs
+   :language: csharp
+
```
nirinchev, line 34: Perhaps we should add a note at the bottom that you can't link two existing users - e.g. if you had already logged in with an email/password or Facebook token, you can't link an anonymous user with the SSO one.

Also, would be nice to note that you can't link multiple email/password users.

andf-mongodb: Hmm, can you clarify the valid link combinations then? So far we have:

✅   Anon --> Email / Password 
❌   Email / password --> Anon
❌   External SSO provider --> Anon
❌   Email / password --> Email / password

examples/dotnet/Examples/Examples.cs
MongoCaleb: @nirinchev I haven't tested this, but it *used to be* that if a user logs off and then logs on again as anonymous, the data is in accessible, thus essentially lost. Not the case now?

examples/dotnet/Examples/Examples.cs
MongoCaleb: Thanks for the explanation!

andf-mongodb (commented):
Some thoughts on the Overview, hope this helps!

source/dotnet/link-identities.txt
MongoCaleb: @nirinchev I need a bit of clarification here. If I log on as email/password, I can still link to a google or FB account, right?

MongoCaleb: Also, we're going to have a single page on linking identities that will list the rules, so it won't be in each SDK page.

examples/dotnet/Examples/Examples.cs
nirinchev: This is correct - logging out will remove the local data they created as anonymous user. But the way I read the code comment is that data created while they were anonymous won't be saved, which is not the case - it will still be synchronized to the server and persisted in MongoDB. Then, depending on the way their partitions are setup, it's possible that they still can access their original data.

source/dotnet/link-identities.txt
nirinchev: Yeah, nothing can link anonymous credentials (because that'd be nonsensical) and linking multiple email/password accounts is not allowed, even if the original user was not created via email-password. E.g. Anon -> email/password -> email/password is disallowed (and as implied, you can link multiple SSO credentials). And yes, email/password -> FB/Google/Apple is legal.

nirinchev (approved):

andf-mongodb (commented):
A few lingering little language suggestions, one of  which is my poor suggestion from earlier.

Also, I see you caught the unlinked "Link User Accounts" reference from paragraph 2 here: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/link/authentication/providers.html#authentication-providers (in DOCSP-12834) -- nice!

PR #565, opened 2020-10-19
12843

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12843

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/postreleasecleanup/dotnet/objects.html

### DIFF
--- a/examples/dotnet/Examples/Examples.cs
+++ b/examples/dotnet/Examples/Examples.cs
@@ -355,7 +355,7 @@ namespace UnitTests
     }
     /*  To add items to the IList<T>:
      
-        var dog = newDog();
+        var dog = new Dog();
         dog.Owners.Add("Caleb");
         
      */
--- a/examples/dotnet/Examples/Examples.cs
+++ b/examples/dotnet/Examples/Examples.cs
@@ -7,6 +7,7 @@ using Realms;
 using Realms.Sync;
 using TaskStatus = dotnet.TaskStatus;
 using Task = dotnet.Task;
+using System.Collections.Generic;
 
 namespace UnitTests
 {
@@ -45,6 +46,14 @@ namespace UnitTests
             });
             // :code-block-end:
             testTaskId = testTask._id;
+
+            var schemas = config.ObjectClasses;
+            foreach (var schema in schemas)
+            {
+                Console.WriteLine(schema.FullName);
+            }
+
+
             return;
         }
 
@@ -285,6 +294,7 @@ namespace UnitTests
             return;
         }
 
+
         
         [OneTimeTearDown]
         public async System.Threading.Tasks.Task TearDown()
@@ -317,10 +327,42 @@ namespace UnitTests
         [MapTo("name")]
         [Required]
         public string Name { get; set; }
-       
+
         public MyClass()
         {
             this.Id = ObjectId.GenerateNewId();
         }
     }
+
+    // :code-block-start: dog_class
+    public class Dog : RealmObject
+    {
+        [Required]
+        public string Name { get; set; }
+
+        public int Age { get; set; }
+        public string Breed { get; set; }
+        public IList<Person> Owners { get; }
+
+        public Dog()
+        {
+            this.Owners = new List<Person>();
+        }
+
+        public Dog(Person owner)
+        {
+            this.Owners = new List<Person>
+            {
+                owner
+            };
+        }
+    }
+
+    public class Person : RealmObject
+    {
+        [Required]
+        public string Name { get; set; }
+        //etc...
+    }
+    // :code-block-end:
 }
--- a/examples/dotnet/Examples/Examples.cs
+++ b/examples/dotnet/Examples/Examples.cs
@@ -344,18 +344,7 @@ namespace UnitTests
         public string Breed { get; set; }
         public IList<Person> Owners { get; }
 
-        public Dog()
-        {
-            this.Owners = new List<Person>();
-        }
-
-        public Dog(Person owner)
-        {
-            this.Owners = new List<Person>
-            {
-                owner
-            };
-        }
+        public Dog() { }
     }
 
     public class Person : RealmObject
@@ -364,5 +353,11 @@ namespace UnitTests
         public string Name { get; set; }
         //etc...
     }
+    /*  To add items to the IList<T>:
+     
+        var dog = newDog();
+        dog.Owners.Add("Caleb");
+        
+     */
     // :code-block-end:
 }
--- a/examples/dotnet/Examples/Examples.cs
+++ b/examples/dotnet/Examples/Examples.cs
@@ -343,8 +343,6 @@ namespace UnitTests
         public int Age { get; set; }
         public string Breed { get; set; }
         public IList<Person> Owners { get; }
-
-        public Dog() { }
     }
 
     public class Person : RealmObject
@@ -356,7 +354,8 @@ namespace UnitTests
     /*  To add items to the IList<T>:
      
         var dog = new Dog();
-        dog.Owners.Add("Caleb");
+        var caleb = new Person { Name = "Caleb" };
+        dog.Owners.Add(caleb);
         
      */
     // :code-block-end:
--- a/source/dotnet/objects.txt
+++ b/source/dotnet/objects.txt
@@ -32,12 +32,18 @@ class implementation.
 .. example::
 
    The following code block shows an object schema that describes a Dog.
-   Every Dog object must include a ``name`` and ``age`` and may
-   optionally include the dog's ``breed`` and a reference to a Person
-   object that represents the dog's ``owner``.
+   Every Dog object must include a ``name`` and may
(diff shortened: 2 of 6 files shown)


MongoCaleb:
@nirinchev 
I'm gonna need some guidance on ``ObjectClasses``. When I try to use it to discover the schemas, I always get a ``null`` value (ie, ``var schemas = config.ObjectClasses;``).

source/dotnet/objects.txt
```diff
@@ -137,38 +145,32 @@ You can configure the following constraints for a given property:
...
-   * - Default
-     - If a client application creates a new object that does not have a
-       value for a defined property, the object uses the default value
-       instead.
-       
-       Objects that are missing a required field without a defined
-       default value will fail validation when you attempt to create
-       them and will not persist to the {+realm+}.
+       omitted. You mark properties as required by using the 
+       :dotnet-sdk:`[Required] <reference/Realms.RequiredAttribute.html>` attribute.
```
nirinchev, line 164: I think this should be expanded a little. For non-nullable primitives, you don't need `[Required]` as they're by default non-nullable. If you want to make them optional, you should declare them with the optional type - `int?`, `float?`, etc. The only two types where `[Required]` makes sense is for `string` and `byte[]` because they are implicitly nullable.

Additionally, `[Required]` can be applied on a collection property to indicate that the elements in the list are required. This has always been a bit confusing, because it doesn't really apply to the list itself (the list is never null). And again, this only makes sense for `IList<string>` and `IList<byte[]>`. 

Also worth noting is that object properties cannot be `[Required]` because objects and embedded objects are always nullable.

source/examples/generated/code/start/Examples.cs
```diff
@@ -265,10 +275,35 @@ public class MyClass : RealmObject
...
+        public int Age { get; set; }
+        public string Breed { get; set; }
+        public IList<Person> Owners { get; }
+
+        public Dog() { }
+
+        public Dog (IList<Person> owners)
+        {
+            this.Owners = owners;
+        }
```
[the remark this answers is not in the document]
nirinchev: On one hand - because it's more verbose (i.e. you have to expose multiple constructors). On the other, it misleads people into thinking that they always need to assign the list, because the fact it's generated by the SDK is not immediately obvious. So when they see

```csharp
public Foo(IList<Bar> bars)
{
    Bars = bars;
}
```

They start copying it everywhere or start doing things like

```csharp
public Foo()
{
    this.Bars = new List<Bar>();
}
```

Which is not wrong, but is unnecessary. Additionally, it is misleading because it implies a referential equality for the list, which will break after the object is added to a Realm. If you have regular POCOs, you can do something like:

```csharp
public class Foo
{
    public IList<int> Ints { get; }

    public Foo(IList<int> ints)
    {
        Ints = ints;
    }
}

var list = new List<int>();
var foo = new Foo(list);

foo.Ints.Add(5);
list.Add(6);

var intCount = foo.Ints.Count; // 2 - both foo.Ints.Add and list.Add operate on the same list
```

This will not work after an object is added to the Realm:

```csharp
public class Foo : RealmObject
{
    public IList<int> Ints { get; }

    public Foo(IList<int> ints)
    {
        Ints = ints;
    }
}

var list = new List<int>();
var foo = new Foo(list);
realm.Write(() =>
{
    realm.Add(foo);
});

list.Add(5);
var intCount = foo.Ints.Count; // 0 - list != foo.Ints
```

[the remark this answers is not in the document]
MongoCaleb: @nirinchev thanks for the explanation. Updated the code snippet again.

source/dotnet/objects.txt
```diff
@@ -137,38 +147,37 @@ You can configure the following constraints for a given property:
...
-       - Float
-       - Double
-       - Decimal (see :ref:`Using Decimal Values <dotnet-decimal128>` below)
-       - String
-       - Date
-       - Data
+       {+client-database+} supports the following data types:
+
+       - bool
+       - byte/short/int/long
```
nirinchev, line 153: We don't support `byte`, but we do support `char`.

source/dotnet/objects.txt
```diff
@@ -137,38 +147,37 @@ You can configure the following constraints for a given property:
...
-       
-       Objects that are missing a required field without a defined
-       default value will fail validation when you attempt to create
-       them and will not persist to the {+realm+}.
+       - IList<T>
+
+   * - Required Properties
+     - Non-nullable properties are always required, unless you specifically 
+       declare them with the optional type (``int?``, for example). To make nullable 
+       types like ``object``, ``string`` and ``byte[]`` required, you mark the
```
nirinchev, line 166: Objects can never be required

```suggestion
       types like ``string`` and ``byte[]`` required, you mark the 
```

source/dotnet/objects.txt
```diff
@@ -137,38 +147,37 @@ You can configure the following constraints for a given property:
...
+
+   * - Required Properties
+     - Non-nullable properties are always required, unless you specifically 
+       declare them with the optional type (``int?``, for example). To make nullable 
+       types like ``object``, ``string`` and ``byte[]`` required, you mark the 
+       properties with the 
+       :dotnet-sdk:`[Required] <reference/Realms.RequiredAttribute.html>` attribute. 
+       Similarly, if you have an ``IList<T>``, which is always required because 
+       it isn't nullable, you can use the ``[Required]`` attribute to ensure the 
+       list is never empty.
```
nirinchev, line 171: I'm sorry - I must have not been clear enough - `[Required]` on a list doesn't force the list to not be empty, but rather ensures that the list cannot contain `null` - it essentially applies the `[Required]` attribute on the elements in the list. E.g.

```csharp
[Required]
public IList<string> Strings { get; }

foo.Strings.Add(null); // will throw
foo.Strings.Clear(); // Empty the list - that's perfectly fine
```

Again, this attribute only makes sense for `IList<string>` and `IList<byte[]>` because the other types are not nullable by default.

examples/dotnet/Examples/Examples.cs
```diff
@@ -317,10 +327,42 @@ public class MyClass : RealmObject
...
+        [Required]
+        public string Name { get; set; }
+
+        public int Age { get; set; }
+        public string Breed { get; set; }
+        public IList<Person> Owners { get; }
+
+        public Dog()
+        {
+            this.Owners = new List<Person>();
```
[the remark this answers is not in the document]
MongoCaleb: That's why I removed it. :) Look again!

source/dotnet/objects.txt
```diff
@@ -32,14 +32,20 @@ class implementation.
...
 
-   .. literalinclude:: /examples/Schemas/DogSchema.cs
+   .. literalinclude:: /examples/generated/code/start/Examples.codeblock.dog_class.cs
       :language: csharp
 
 
+.. note::
+   
+   To define a collection of objects within an object, use an ``IList<T>`` with 
+   only a getter.
```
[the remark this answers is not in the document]
MongoCaleb: @nirinchev - added:

   To define a collection of objects within an object, use an ``IList<T>`` with 
   only a getter. You do not need to initialize it in the constructor, as {+realm+} 
   will generate a collection instance the first time the property is accessed.

examples/dotnet/Examples/Examples.cs
```diff
@@ -317,10 +327,37 @@ public class MyClass : RealmObject
...
+    public class Dog : RealmObject
+    {
+        [Required]
+        public string Name { get; set; }
+
+        public int Age { get; set; }
+        public string Breed { get; set; }
+        public IList<Person> Owners { get; }
+
+        public Dog() { }
```
nirinchev, line 347:
```suggestion
```

examples/dotnet/Examples/Examples.cs
```diff
@@ -317,10 +327,37 @@ public class MyClass : RealmObject
...
+    public class Person : RealmObject
+    {
+        [Required]
+        public string Name { get; set; }
+        //etc...
+    }
+    /*  To add items to the IList<T>:
+     
+        var dog = new Dog();
+        dog.Owners.Add("Caleb");
```
nirinchev, line 359:
```suggestion
        var dog = new Dog();
        var caleb = new Person { Name = "Caleb" };
        dog.Owners.Add(caleb);
```

examples/dotnet/Examples/Examples.cs
```diff
@@ -317,10 +327,42 @@ public class MyClass : RealmObject
...
+        {
+            this.Owners = new List<Person>();
+        }
+
+        public Dog(Person owner)
+        {
+            this.Owners = new List<Person>
+            {
+                owner
+            };
```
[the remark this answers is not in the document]
MongoCaleb: gone

examples/dotnet/Examples/Examples.cs
MongoCaleb: doh!

nirinchev (approved):

PR #572, opened 2020-10-21
Postreleasecleanup: quick-start fixes

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12841

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/postreleasecleanup/dotnet/quick-start.html


examples/dotnet/Examples/Examples.cs
```diff
@@ -131,7 +131,7 @@ public async System.Threading.Tasks.Task ModifiesATask()
             var realm = await Realm.GetInstanceAsync(config);
             // :code-block-start: modify
             var t = realm.All<Task>()
-                .Where(t => t._id == testTaskId)
+                .Where(t => t.Id == testTaskId)
                 .FirstOrDefault();
```
nirinchev, line 135: [nit] These can be combined in a single `FirstOrDefault` call: `realm.All<Task>().FirstOrDefault(t => t.Id == testTaskId)`.

source/dotnet/quick-start.txt
```diff
@@ -98,14 +98,14 @@ how to do this by calling ``LogOutAsync()``:
 .. literalinclude:: /examples/generated/code/start/Examples.codeblock.logout.cs
```
nirinchev, line 98: I can't comment on the right line because it's not in the diff, but the "Import Realm" section states the following:

> In the shared code project of your solution, go to Manage NuGet Packages. In the NuGet Package Manager, switch to the Browse tab and then search for Realm. Select The Realm driver and then click Add Package.

This should be "Select the Realm package".

source/dotnet/quick-start.txt
```diff
@@ -98,14 +98,14 @@ how to do this by calling ``LogOutAsync()``:
...
    :language: csharp
 
-.. seealso:: :ref:`Authenticate a User <android-authenticate>`
```
nirinchev, line 101: Again, can't comment on the correct line, but the section where it says 

> ... and are using the BsonElement attribute so we can use

This should be `MapTo` attribute

examples/dotnet/Examples/Examples.cs
```diff
@@ -302,13 +302,14 @@ public async System.Threading.Tasks.Task TearDown()
...
                 {
-                    realm.RemoveAll<Task>();
+                    realm.Remove(myTask);
                 });
                 // :code-block-end:
-
+                realm.RemoveAll<Task>();
                 var realmUser = await app.LogInAsync(Credentials.Anonymous());
                 // :code-block-start: logout
                 await realmUser.LogOutAsync();
```
nirinchev, line 315: [nit] The first example is `var user = await app.LogInAsync...`, but here we're logging out `realmUser`. We should use the same variable name for consistency sake.

nirinchev (commented):
A few nits, looking good otherwise.

examples/dotnet/Examples/Relationships.cs
```diff
@@ -0,0 +1,93 @@
...
+                Age = 7,
+                Breed = "mutt"
+            });
+            // :code-block-end:
+        }
+        // :code-block-start: one-to-many
+        public class Person : RealmObject
+        {
+            public string Name { get; set; }
+            public DateTime Birthdate { get; set; }
```
nirinchev, line 47:
```suggestion
            public DateTimeOffset Birthdate { get; set; }
```

examples/dotnet/Examples/Relationships.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace Examples
+{
+    public class OneToOneRelationship
+    {
+        // :code-block-start: one-to-one
+        public class Person : RealmObject
+        {
+            public string Name { get; set; }
+            public DateTime Birthdate { get; set; }
```
nirinchev, line 15:
```suggestion
            public DateTimeOffset Birthdate { get; set; }
```

Does this not result in errors during compilation? `DateTime` is not a valid Realm type and I thought we'd error out if we encounter it.

examples/dotnet/Examples/Relationships.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        // :code-block-start: inverse
+        public class User : RealmObject
+        {
+            [PrimaryKey]
+            [MapTo("_id")]
+            public ObjectId Id { get; set; } = ObjectId.GenerateNewId();
+
+            [MapTo("_partition")]
+            public string Partition { get; set; }
```
nirinchev, line 70: I feel we can generally skip showing the `_partition` property - it doesn't convey something useful and it may be a bit confusing for people.

source/dotnet/relationships.txt
```diff
@@ -126,14 +124,8 @@ set the value of an inverse relationship property.
      automatically updates to refer back to any User object that contains the
```
nirinchev, line 124: I can't comment on the correct line, but there's a section that says:

> To define an inverse relationship, define a LinkingObjects property in your object model. The LinkingObjects definition specifies the object type and property name of the relationship that it inverts.

This is a bit vague because it uses `LinkingObjects` whereas the .NET SDK uses the `Backlink` terminology and the sentence doesn't actually tell you how to define an inverse relationship. I suggest something along the lines of:

> To define an inverse relationship, define a getter-only `IQueryable<T>` property in your object model, where `T` is the source type of the forward relationship, and annotate it with the `[Backlink(sourceProperty)]` attribute. In this case `sourceProperty` is the relationship property in the source type.

source/dotnet/realms.txt
```diff
@@ -54,137 +51,38 @@ A {+realm+} is not a collection of schemaless documents.
...
+
+Realm Schema
+~~~~~~~~~~~~
+
+A **{+backend-schema+}** is a list of valid :ref:`object schemas <dotnet-realm-objects>` that
+each define an object type that a {+app+} may persist. All objects in a {+realm+} must
+conform to the {+backend-schema+}.
+
+Client applications provide a {+frontend-schema+} when they open a {+realm+}. If a
+{+realm+} already contains data, then {+client-database+} validates each existing object
```
nirinchev, line 79: "validates each existing object" is not correct. If a Realm already contains data, then it already has a schema. When it is opened, the new schema is validated against the existing schema. But we don't iterate over the data to check for conformity against the new schema. Moreover, even if no data exists in the Realm, but it has already been opened with a particular schema, trying to open it with a different schema will require a migration if it's a local Realm or will be handled based on sync schema rules if it's a synchronized Realm.

source/dotnet/realms.txt
```diff
@@ -54,137 +51,38 @@ A {+realm+} is not a collection of schemaless documents.
...
+to ensure that an object schema was provided for its type and that it meets all of the
+constraints specified in the schema. To determine the schema in .NET applications,
+{+service-short+} automatically checks at runtime for any class that inherits from
```
nirinchev, line 82: `at runtime` is not correct - we generate the schema at compile time and store it in the built application. This is much faster than looking it up at runtime.

In this section, you can mention that, if the developer doesn't specify the `ObjectClasses` property on their RealmConfiguration, all classes that inherit from `RealmObject` or `EmbeddedObject` will be added to the schema. And if they want to use only a subset of these classes for a particular Realm, they can explicitly set the `ObjectClasses` property to an array containing just the types they want.

examples/dotnet/Examples/Relationships.cs
MongoCaleb: I'm not clear on when/if it is ever _required_.

nirinchev: It is fine to omit it in client models. It will always contain the same value (the value you opened the Realm with), so it's not particularly useful in most cases. There are some very niche cases where you open several Realms with different partitions, but somehow want to combine all their data in the same view, so having the partition on the objects is useful, but I don't think it's worth optimizing for those.

nirinchev (approved):

PR #580, opened 2020-11-02
(DOCSP-12222): Link Credentials Node/React Native SDKs Doc(s)

mongomoe:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12222

* Node: https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/docsp-12222-link-identity-javascript/node/link-identities.html
* React Native: https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/docsp-12222-link-identity-javascript/react-native/link-identities.html

### Passing Tests:
Screen Shot 2020-11-02 at 10 41 24 AM

### DIFF
--- a/examples/node/Examples/link-identities.js
+++ b/examples/node/Examples/link-identities.js
@@ -1,0 +1,86 @@
+import Realm from "realm";
+const randomEmail = require('random-email');
+
+let app;
+let anonUser;
+let emailPasswordUser;
+const email = randomEmail();
+const password = "Pa55w0rd";
+
+const credentials = Realm.Credentials.emailPassword(
+    email,
+    password
+);
+
+beforeAll(async () => {
+    app = new Realm.App({ id: "tutsbrawl-qfxxj" });
+
+    async function loginAnonymously(){
+        const anonymousCredentials = Realm.Credentials.anonymous();
+        anonUser = await app.logIn(anonymousCredentials);
+        console.log(`Logged in with the user id: ${anonUser.id}`);
+        return anonUser;
+    }
+    async function registerNewAccount(email,password){
+        await app.emailPasswordAuth.registerUser(email, password).catch((err) => console.log(`An error occurred while registering: ${JSON.stringify(err, 2, null)}`));
+    }
+    // application user tries out the app by logging in anonymously
+    anonUser = await loginAnonymously().catch((err) => console.log(`An error occurred while logging in anonymously: ${JSON.stringify(err, 2, null)}`));
+    // after using the app for awhile the user decides to register:
+    await registerNewAccount(email,password);
+});
+
+afterAll(async () => {
+    async function deleteAnonUser(anonUser){
+        // logging out of an anonymous user will delete the user
+        await anonUser.logOut().catch((err) => console.log(`An error occurred while logging out: ${JSON.stringify(err, 2, null)}`));
+    }
+    // delete the anon user after logging in with an anonymous identity, then
+    // registering as email/pass identity, then linking the two identities
+    if(anonUser){
+        await deleteAnonUser(anonUser);
+    }
+});
+
+
+/* 
+    Steps the app user follows:
+    1. Creates an anonymous account to try out the app
+    2. Decides to create a more permanent account (email/pass) once they decide they enjoy the app
+    3. Links the temporary anonymous account with the permanent
+       email-password account in order to retain their user data
+    4. Deletes the temporary anonymous account
+*/
+describe("Linking Identities Tests", () => {
+    afterEach(async () => {});
+
+    test("links anon identity with email/pass identity", async () => {   
+        expect.assertions(1); 
+        // :code-block-start: link-identities
+        async function linkAccounts(user,email,password){
+            const emailPasswordUserCredentials = Realm.Credentials.emailPassword(
+                email,
+                password
+              );
+            return user.linkCredentials(emailPasswordUserCredentials);
+        }
+
+        async function run(){
+            const linkedAccount = await linkAccounts(anonUser,email, password).catch((err) => console.log(`An error occurred while linking accounts: ${JSON.stringify(err, 2, null)}`));
+            // :hide-start:
+            const emailPasswordUser = await app.logIn(credentials).catch((err) => console.log(`An error occurred while logging in anonymously; ${JSON.stringify(err, 2, null)}`));
+            
+            expect(linkedAccount).toStrictEqual(emailPasswordUser); // test if the linked account is the same user as the emailPass user
+            // :hide-end:
+        }
+        // :hide-start:
+        await run();
+        /*
+        // :replace-with:
+        run();
+        // :hide-end:
+        // :code-block-end:
+        */
+    })
+});
+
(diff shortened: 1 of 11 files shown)


examples/node/Examples/link-identities.js
```diff
@@ -0,0 +1,86 @@
...
+            const emailPasswordUser = await app.logIn(credentials).catch((err) => console.log(`An error occurred while logging in anonymously; ${JSON.stringify(err, 2, null)}`));
+            
+            expect(linkedAccount).toStrictEqual(emailPasswordUser); // test if the linked account is the same user as the emailPass user
+            // :hide-end:
+        }
+        // :hide-start:
+        await run();
+        /*
+        // :replace-with:
+        run();
```
cbush, line 80: - Why is this in a block comment?
- Why is this replaced? We should only need "start" blocks for code examples, so there should be no use of replace-with

source/node/link-identities.txt
```diff
@@ -0,0 +1,43 @@
...
+to a unique user identity, {+service-short+} lets you merge multiple identities
+corresponding to a single user into one user identity.
+
+Example
+-------
+
+For instance, consider an application that enables users to authenticate via
+:ref:`anonymous authentication <anonymous-authentication>` so they may explore
+functionality without linking to an existing SSO provider or generating a new
+password. When that user finally decides to create a full account with an SSO
```
cbush, line 21: Reduce wordiness:
```suggestion
password. When that user decides to create a full account with an SSO
```

source/node/link-identities.txt
```diff
@@ -0,0 +1,43 @@
...
+
+{+service+} provides :ref:`multiple ways <authentication-providers>` for users
+to log into your application. Because each method of authentication corresponds
+to a unique user identity, {+service-short+} lets you merge multiple identities
+corresponding to a single user into one user identity.
+
+Example
+-------
+
+For instance, consider an application that enables users to authenticate via
```
cbush, line 18: "Example", "for instance", and "consider..." are all doing the same thing:
```suggestion
Consider an application that enables users to authenticate via
```

cbush (commented):
Withholding approval for questions

examples/node/Examples/link-identities.js
```diff
@@ -0,0 +1,86 @@
+import Realm from "realm";
+const randomEmail = require('random-email');
```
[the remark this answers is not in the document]
mongomoe: Thanks for pointing this out. I've added formatting on-save now.  I've updated this to be imports only for the JS file, however, for TS I can only do a require statement because the package random-email does not have a @types/random-email.

examples/node/Examples/link-identities.js
```diff
@@ -0,0 +1,86 @@
...
+        const anonymousCredentials = Realm.Credentials.anonymous();
+        anonUser = await app.logIn(anonymousCredentials);
+        console.log(`Logged in with the user id: ${anonUser.id}`);
+        return anonUser;
+    }
+    async function registerNewAccount(email,password){
+        await app.emailPasswordAuth.registerUser(email, password).catch((err) => console.log(`An error occurred while registering: ${JSON.stringify(err, 2, null)}`));
+    }
+    // application user tries out the app by logging in anonymously
+    anonUser = await loginAnonymously().catch((err) => console.log(`An error occurred while logging in anonymously: ${JSON.stringify(err, 2, null)}`));
```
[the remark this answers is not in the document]
mongomoe: Before any attempt to be linked, a user should be logged in anonymously and registered. Where would you put it alternatively?

examples/node/Examples/link-identities.js
```diff
@@ -0,0 +1,86 @@
...
+/* 
+    Steps the app user follows:
+    1. Creates an anonymous account to try out the app
+    2. Decides to create a more permanent account (email/pass) once they decide they enjoy the app
+    3. Links the temporary anonymous account with the permanent
+       email-password account in order to retain their user data
+    4. Deletes the temporary anonymous account
+*/
+describe("Linking Identities Tests", () => {
+    afterEach(async () => {});
```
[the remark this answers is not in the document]
mongomoe: Sorry is what needed? The comment explaining the thought process of the developer? I suppose it's not needed, but at the time I figured it would be a nice additional comment

examples/node/Examples/link-identities.js
mongomoe: To answer your questions:
1. It's in a block comment so that Jest doesn't pick it up and have it interfere with the tests. However, Bluehawk needs to pick it up, in order for it to replace the await statement.
2. It is replaced because Jest needs to `await run()` (see: https://jestjs.io/docs/en/asynchronous.html#asyncawait). But outside of Jest/In a non-unit test/standard node environment, you do not need to await run(). So to the docs reader, the code should be something like:

```
async function run(){ // do stuff } 
run();
```
So in certain cases, we need to do a replace-with. Though I'm happy to alter this if you know of a way to accomplish the same functionality while staying true to our precedents/rules for Bluehawk usage for our code-examples.

examples/node/Examples/link-identities.js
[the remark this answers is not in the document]
mongomoe: OH the empty afterEach method! I've removed that now, good catch :)

examples/node/Examples/link-identities.js
cbush: Maybe you can put the entire example into an async function and use await in the final example code? If you use replace-with, the code is not really tested, which defeats the purpose. Also, the behavior of replace-with on code blocks in Bluehawk is currently ill-defined, so this will probably break at some point.

examples/node/Examples/link-identities.js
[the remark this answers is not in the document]
cbush: This text:

>// application user tries out the app by logging in anonymously

Looks like it belongs in the code example?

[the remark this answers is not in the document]
mongomoe: Any comments in the BeforeAll block (and all parts of the file that are not after a blue hawk start statement) are intended for the docs writer/developer of docs-realm, not the reader of the published doc

examples/node/Examples/link-identities.js
mongomoe: So I essentially went with your suggestion ^ (except instead of using await in the final example code, I used jest's promise.resolves() (resolves with an s) method. This means that we only show the method, but not the calling of the method (which is fine for this example, but I still wish we could've shown it with replace-with)

source/node.txt
```diff
@@ -83,6 +83,8 @@ To learn how to handle schema updates in your client application, see
 To learn how to query for data in MongoDB with the Node.js SDK, see
 :ref:`MongoDB Data Access <node-mongodb-data-access>`.  
 
+foobar
```
mongomoe, line 86:
```suggestion
```

source/node.txt
```diff
@@ -83,6 +83,7 @@ To learn how to handle schema updates in your client application, see
 To learn how to query for data in MongoDB with the Node.js SDK, see
 :ref:`MongoDB Data Access <node-mongodb-data-access>`.  
 
+
```
mongomoe, line 86:
```suggestion
```

cbush (approved):

PR #587, opened 2020-11-03
DOCSP-12863: bulkWrite

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12863

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/12863/mongodb/bulk-write.html


nathan-contino-mongo (approved):
LGTM, though I wonder if we could provide a little more detail about the format and limitations (for comparison/inspiration, you can take a look at the new Node Driver Usage Example for Bulk Writes: https://docs.mongodb.com/drivers/node/usage-examples/bulkWrite). Specifically noting limitations like "only one collection" and the consequences of ordered vs. unordered bulk operations (unordered tries to run everything even if it hits a failure; ordered fails fast as soon as one operation encounters an error; unordered executes operations in parallel) could be helpful for some users. Alternatively, we could always just direct users to the Node Driver documentation to single-source information. It's really down to where you personally draw the line of "too much information".

source/includes/extracts-mongodb-action.yaml
```diff
@@ -862,4 +862,61 @@ content: |
...
+
+          .. code-block:: javascript
+
+             { insertOne: { document: { a: 1 } } }
+
+             { updateOne: { filter: {a:2}, update: {$set: {a:2}}, upsert:true } }
+
+             { updateMany: { filter: {a:2}, update: {$set: {a:2}}, upsert:true } }
+
+             { updateMany: { filter: {}, update: {$set: {"a.$[i].x": 5}}, arrayFilters: [{ "i.x": 5 }]} }
```
arahmanan, line 890: `arrayFilters` is not supported yet, but it hopefully will be once  https://jira.mongodb.org/browse/REALMC-5891 is done.

source/includes/extracts-mongodb-action.yaml
```diff
@@ -862,4 +862,61 @@ content: |
...
+              - Description
+  
+            * - ordered
+              - boolean (default = true)
+              - Execute write operation in ordered or unordered fashion
+
+            * - session
+              - client session
+              - optional session to use for this operation
+
```
arahmanan, line 920: I took a second look and also `bypassDocumentValidation` is supported.

arahmanan (approved):
LGTM! just a couple of comments from my end

PR #591, opened 2020-11-05
fix default domain name - UI

nathan-contino-mongo:


ccho-mongodb (approved):
Just assuming this is correct!

PR #599, opened 2020-11-09
(DOCSP-12926): Update for iOS 10.1.1

cbush:
- Changed partition value to avoid conflict with other tests on other platforms
- 10.1.2 does not install - see ask-realm-sdks channel

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12926


nathan-contino-mongo (approved):
some LGTM value

PR #609, opened 2020-11-16
(DOCSP-13017): JS SDK Open a Realm Synchronously

mongomoe:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13017

Node: https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-13017-JS-SDK-Open-Realm-Synchronously/node/open-a-realm.html
React Native: https://docs-mongodbcom-staging.corp.mongodb.com/realm/mohammad.hunan/DOCSP-13017-JS-SDK-Open-Realm-Synchronously/react-native/open-a-realm.html

### DIFF
--- a/source/node/open-a-realm.txt
+++ b/source/node/open-a-realm.txt
@@ -38,7 +38,7 @@ this:
 
 In the above example, the code shows how to open the {+realm+} *asynchronously*
 by calling ``realm.open()``. You can also open a realm synchronously by passing
-a ``Configuration object`` to a new instance of the :js-sdk:`Realm
+a ``Configuration`` object to a new instance of the :js-sdk:`Realm
 <Realm.html>` object, which is necessary if the device is offline.
 
 .. code-block:: javascript
@@ -54,11 +54,12 @@ a ``Configuration object`` to a new instance of the :js-sdk:`Realm
 
 Open a Local (Non-Synced) Realm
 -------------------------------
-When opening a local (non-synced) {+realm+}, you can pass a
-:js-sdk:`Configuration() <Realm.html#~Configuration>` object, which may include
-the path property defining the local file path to use when storing the data, to
-either :js-sdk:`Realm.open() <Realm.html#.open>` or :js-sdk:`new Realm()
-<Realm.html>`. 
+To open a local (non-synced) {+realm+}, pass a :js-sdk:`Configuration()
+<Realm.html#~Configuration>` object to either :js-sdk:`Realm.open()
+<Realm.html#.open>` or :js-sdk:`new Realm() <Realm.html>`. The following example
+creates a ``Configuration`` object with a path property that defines the local
+file path to store data. 
+
 
 .. code-block:: javascript
 
--- a/source/node/open-a-realm.txt
+++ b/source/node/open-a-realm.txt
@@ -39,7 +39,7 @@ this:
 In the above example, the code shows how to open the {+realm+} *asynchronously*
 by calling ``realm.open()``. You can also open a realm synchronously by passing
 a ``Configuration object`` to a new instance of the :js-sdk:`Realm.App
-<Realm.App.html>` object, which is necessary if the device is offline.
+<Realm.html>` object, which is necessary if the device is offline.
 
 .. code-block:: javascript
 
@@ -77,4 +77,3 @@ It is important to remember to call the ``close()`` method when done with a
 .. code-block:: javascript
 
    realm.close();
-
--- a/source/node/open-a-realm.txt
+++ b/source/node/open-a-realm.txt
@@ -38,7 +38,7 @@ this:
 
 In the above example, the code shows how to open the {+realm+} *asynchronously*
 by calling ``realm.open()``. You can also open a realm synchronously by passing
-a ``Configuration object`` to a new instance of the :js-sdk:`Realm.App
+a ``Configuration object`` to a new instance of the :js-sdk:`Realm
 <Realm.html>` object, which is necessary if the device is offline.
 
 .. code-block:: javascript
--- a/source/node/open-a-realm.txt
+++ b/source/node/open-a-realm.txt
@@ -14,8 +14,8 @@ Open and Close a Realm
 
 Open a Synced Realm
 -------------------
-To open a synced {+realm+}, call :js-sdk:`Realm.open() <Realm.html#.open>`. 
-Pass in a :js-sdk:`Configuration() <Realm.html#~Configuration>`
+To open a synced {+realm+}, call :js-sdk:`Realm.open() <Realm.html#.open>`, 
+pass in a :js-sdk:`Configuration() <Realm.html#~Configuration>`
 object, which must include the ``sync`` property defining a 
 :js-sdk:`SyncConfiguration <Realm.Sync.html#~SyncConfiguration>` object.
 
--- a/source/node/open-a-realm.txt
+++ b/source/node/open-a-realm.txt
@@ -14,8 +14,8 @@ Open and Close a Realm
 
 Open a Synced Realm
 -------------------
-To open a synced {+realm+}, call :js-sdk:`Realm.open() <Realm.html#.open>`, 
-pass in a :js-sdk:`Configuration() <Realm.html#~Configuration>`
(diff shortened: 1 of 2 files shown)


source/node/open-a-realm.txt
```diff
@@ -34,15 +34,31 @@ this:
       }
    };
     
-   realm = await Realm.open(config);
+   const realm = await Realm.open(config);
 
+In the above example, the code shows how to open the {+realm+} *asynchronously*
+by calling ``realm.open()``. You can also open a realm synchronously by passing
+a ``Configuration object`` to a new instance of the :js-sdk:`Realm.App
```
nlarew, line 41: * What is a `Configuration object`? It's a Realm [Configuration](https://docs.mongodb.com/realm-sdks/js/latest/Realm.html#~Configuration) - let's link to the API docs for it here.
* This is about `Realm` not `Realm.App`

source/node/open-a-realm.txt
```diff
@@ -34,15 +34,31 @@ this:
...
+<Realm.App.html>` object, which is necessary if the device is offline.
```
nlarew, line 42: Why is it necessary if the device is offline? From the ticket - `Realm.open` **downloads all the server data that is currently available** but requires an internet connection.

source/node/open-a-realm.txt
```diff
@@ -34,15 +34,31 @@ this:
...
+.. code-block:: javascript
+
+   const realm = new Realm.App(config);
+
+.. note::
+
+   The first time a user logs on to your realm app, you should open the realm 
+   *asynchronously* to sync data from the server to the device. After that initial 
+   connection, you can open a realm *synchronously* to ensure the app works in 
+   an offline state.
```
nlarew, line 53: Based on the ticket, it's not about the first time a particular user logs in but rather about whether or not a user is logged in at the time that you open the realm.

I think the motivation is that if a user is already logged in, we want to load the UI as fast as possible. `Realm.open()` doesn't resolve until all pending changes are fully synced so rather than making the user wait for all of that sync to happen we use the `new Realm()` constructor to open the Realm locally. Then it downloads changes in the background and we can update the UI to match the state of the on-device Realm as synced changes come in.

source/node/open-a-realm.txt
```diff
@@ -34,15 +34,31 @@ this:
...
 -------------------------------
-When opening a local (non-synced) {+realm+}, call :js-sdk:`Realm.open() <Realm.html#.open>`, 
-passing in a :js-sdk:`Configuration() <Realm.html#~Configuration>`
-object, which may include the ``path`` property defining the local file path to 
-use when storing the data. 
+When opening a local (non-synced) {+realm+}, you can pass a
+:js-sdk:`Configuration() <Realm.html#~Configuration>` object, which may include
+the path property defining the local file path to use when storing the data, to
+either :js-sdk:`Realm.open() <Realm.html#.open>` or :js-sdk:`new Realm.App()
+<Realm.App.html>`.
```
nlarew, line 61: nits:
* When opening -> To open
* You can pass -> pass

When should I include the path property? If we're going to mention details of the config file let's give a bit more guidance.

source/node/open-a-realm.txt
```diff
@@ -34,15 +34,31 @@ this:
...
+   const realm = await Realm.open(config);
 
+In the above example, the code shows how to open the {+realm+} *asynchronously*
+by calling ``realm.open()``. You can also open a realm synchronously by passing
+a ``Configuration object`` to a new instance of the :js-sdk:`Realm.App
+<Realm.App.html>` object, which is necessary if the device is offline.
+
+.. code-block:: javascript
+
+   const realm = new Realm.App(config);
```
nlarew, line 46:
```suggestion
   const realm = new Realm(config);
```

nlarew (changes requested):

source/node/open-a-realm.txt
mongomoe: To address your points:
* It is a Realm Configuration! It's linked just a bit above this (line 18) and later linked below in a different section. I didn't want to include two external links to the same thing in the same section.
* Whoops sorry about that, fixed it now!

source/node/open-a-realm.txt
mongomoe, line 42:
```suggestion
<Realm.html>` object, which is necessary if the device is offline.
```

source/node/open-a-realm.txt
mongomoe, line 41:
```suggestion
a ``Configuration object`` to a new instance of the :js-sdk:`Realm
```

source/node/open-a-realm.txt
```diff
@@ -34,15 +34,31 @@ this:
...
 Open a Local (Non-Synced) Realm
 -------------------------------
-When opening a local (non-synced) {+realm+}, call :js-sdk:`Realm.open() <Realm.html#.open>`, 
-passing in a :js-sdk:`Configuration() <Realm.html#~Configuration>`
-object, which may include the ``path`` property defining the local file path to 
-use when storing the data. 
+When opening a local (non-synced) {+realm+}, you can pass a
+:js-sdk:`Configuration() <Realm.html#~Configuration>` object, which may include
+the path property defining the local file path to use when storing the data, to
+either :js-sdk:`Realm.open() <Realm.html#.open>` or :js-sdk:`new Realm.App()
```
mongomoe, line 60:
```suggestion
either :js-sdk:`Realm.open() <Realm.html#.open>` or :js-sdk:`new Realm()
```

source/node/open-a-realm.txt
```diff
@@ -34,15 +34,31 @@ this:
...
 -------------------------------
-When opening a local (non-synced) {+realm+}, call :js-sdk:`Realm.open() <Realm.html#.open>`, 
-passing in a :js-sdk:`Configuration() <Realm.html#~Configuration>`
-object, which may include the ``path`` property defining the local file path to 
-use when storing the data. 
+When opening a local (non-synced) {+realm+}, you can pass a
+:js-sdk:`Configuration() <Realm.html#~Configuration>` object, which may include
+the path property defining the local file path to use when storing the data, to
+either :js-sdk:`Realm.open() <Realm.html#.open>` or :js-sdk:`new Realm()
+<Realm.App.html>`.
```
mongomoe, line 61:
```suggestion
<Realm.html>`. 
```

source/node/open-a-realm.txt
mongomoe: The ticket specifies to open a realm synchronously (via new Realm) if the app user is already logged in. This is not to load the UI as possible, but rather to open a realm even if the phone is offline (it's about internet connectivity not speed). This note is taken from the published [DOTNET page](https://docs.mongodb.com/realm/dotnet/open-a-realm/) about opening a realm, and it functions the same way. 

Let me know if this makes sense or if you still have any questions/concerns about this.

source/node/open-a-realm.txt
mongomoe: Sorry - it should be "This works even if the device is offline." not that it is necessary. Updated it now.

source/node/open-a-realm.txt
```diff
@@ -34,15 +34,31 @@ this:
       }
    };
     
-   realm = await Realm.open(config);
+   const realm = await Realm.open(config);
 
+In the above example, the code shows how to open the {+realm+} *asynchronously*
+by calling ``realm.open()``. You can also open a realm synchronously by passing
+a ``Configuration object`` to a new instance of the :js-sdk:`Realm
+<Realm.html>` object, which is necessary if the device is offline.
```
mongomoe, line 42:
```suggestion
<Realm.html>` object. This works even if the device is offline. 
```

source/node/open-a-realm.txt
mongomoe: * When opening -> To open [Fixed]
* You can pass -> pass [Fixed]

Sure, I've altered the sentence to be more like the .net and ios adjacent sentences: "To open a local (non-synced) realm, pass a Configuration() object to either Realm.open() or new Realm(). The following example creates a Configuration object with a path property that defines the local file path to store data." 

Let me know if that works, or if you have any specific alternatives.

nlarew (approved):

PR #611, opened 2020-11-17
CI on PRs

nathan-contino-mongo:
## Pull Request Info
Adding test runs to our PRs so we can ensure that:
1) Tests are frequently run (as often as we PR, at least)
2) New changes do not break existing tests

See the Actions page: https://github.com/mongodb/docs-realm/actions of this repo to see our tests. Currently iOS, .NET, and Android pass reliably; Node.js passes most tests, but often fails because of async callbacks.

### DIFF
--- a/examples/dotnet/Examples/EmbeddedExamples.cs
+++ b/examples/dotnet/Examples/EmbeddedExamples.cs
@@ -119,6 +119,7 @@ namespace UnitTests
             }
         }
 
+
         [Test]
         public async Task QueryEmbeddedObject()
         {
@@ -146,7 +147,7 @@ namespace UnitTests
                 // :code-block-start:query
                 // Find All Contacts with an Address of "Los Angeles"
                 var losAngelesContacts = realm.All<Contact>()
-                    .Filter("address.city == 'Los Angeles'");
+                    .Filter("address.City == 'Los Angeles'");
 
                 foreach (var contact in losAngelesContacts)
                 {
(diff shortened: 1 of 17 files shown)


examples/dotnet/Examples/EmbeddedExamples.cs
```diff
@@ -156,8 +157,8 @@ public async Task QueryEmbeddedObject()
                 }
                 //:code-block-end:
 
-                /* Test that the query worked and that the Contacts returned 
-                 * actually are from 'Los Angeles'. */
+                // Test that the query worked and that the Contacts returned 
+                // actually are from 'Los Angeles'.
```
MongoCaleb, line 161: Why this change? Doesn't matter, just curious.

examples/dotnet/Examples/EmbeddedExamples.cs
```diff
@@ -187,16 +188,16 @@ public class Address : EmbeddedObject
             [MapTo("_partition")]
             public string Partition { get; set; }
 
-            [MapTo("street")]
+            //[MapTo("street")]
```
MongoCaleb, line 191: These need to be in here and not commented out. They are part of a bluehawk code example and are important.

MongoCaleb: What's the error when you don't comment them out?

MongoCaleb (commented):
One blocking change; otherwise lgtm

examples/dotnet/Examples/EmbeddedExamples.cs
nathan-contino-mongo: It seems the mapped fields don't actually match the field names in the schema -- if we just fix the backend schema (to use lowercase letters) we should be able to throw these changes away.

examples/dotnet/Examples/EmbeddedExamples.cs
nathan-contino-mongo: Originally I had to comment out this whole test, and you can't nest block comments in comments (the outer block comment ends when the inner block comment ends, leaving you with a hanging block comment end somewhere later on. Definitely can change this back too!

MongoCaleb (approved):
LGTM

cbush (approved):

PR #630, opened 2020-11-30
13105

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13105

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/13105/dotnet/async-console


source/dotnet/async-console.txt
```diff
@@ -0,0 +1,30 @@
...
+   :local:
+   :backlinks: none
+   :depth: 3
+   :class: singlecol
+
+Overview
+--------
+
+We recommend calling {+realm+} methods asynchronously. However, .NET console apps 
+do not support background threads. So if you are building a .NET
```
cbush, line 19: Arguably the "so" doesn't need to be stated here:
```suggestion
do not support background threads. If you are building a .NET 
```

examples/dotnet/ConsoleTests/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    await Task.Delay(10); // Simulates some background work
+                    Console.WriteLine(foo.Bar);
+                }
+            });
+        }
+        //:code-block-end:
+    }
+
+    class Foo : RealmObject
```
cbush, line 39: Why not include the Foo class definition in the example for the sake of completeness?

cbush (approved):
LGTM with minor comments

source/dotnet/async-console.txt
```diff
@@ -0,0 +1,30 @@
...
+   :depth: 3
+   :class: singlecol
+
+Overview
+--------
+
+We recommend calling {+realm+} methods asynchronously. However, .NET console apps 
+do not support background threads. So if you are building a .NET 
+console app with the async {+realm+} API, you must include a library that 
+provides support for asynchronous tasks.
```
nirinchev, line 21: I would reword this paragraph to something along the lines of:

> Realm instances and objects are SynchronizationContext-confined (*link to the threading page*). This means that they can only be accessed on the thread they were created on. .NET's [Task Asynchronous Programming model](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/) allows developers to seamlessly write asynchronous code as a sequence of statements preceded by the `await` keyword. On platforms with a UI thread the framework installs a synchronization context on the main thread, so the continuation after an `await` is dispatched to original context, allowing you to intersperse reads and writes to the database with asynchronous calls. On server/console apps, however, there is no UI thread/SynchronizationContext, which means that `await`-ing a Task will dispatch the continuation on a random thread from the thread pool, preventing you from accessing previously opened Realm instances. To be able to efficiently use Realm in between asynchronous calls, we recommend installing a SynchronizationContext - either one you implemented yourself or one provided by a 3rd party library.

That may be a bit heavy, so feel free to reword it however you want, but I hope it clarifies the problem - it's not about the lack of background threads, but instead, because there's no UI thread, a developer would be forced to reopen Realm after every `await`, otherwise they'd get invalid thread exception.

Also, you can weave in a reference to Stephen Cleary's (the author of Nito.AsyncEx) writeup about synchronization contexts - it's excellent, but also very technical: https://docs.microsoft.com/en-us/archive/msdn-magazine/2011/february/msdn-magazine-parallel-computing-it-s-all-about-the-synchronizationcontext

examples/dotnet/ConsoleTests/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        //:code-block-start: async-console
+        public static void Main(string[] args)
+        {
+            AsyncContext.Run(async () => await MainAsync(args));
+        }
+
+        private static async Task MainAsync(string[] args)
+        {
+            AsyncContext.Run(async () =>
```
nirinchev, line 22: MainAsync is already invoked with `AsyncContext.Run`, which installs the synchronization context - there's no need to do it again here.

nirinchev (changes requested):
I have a minor code comment and a suggestion for a rewrite of the opening section.

examples/dotnet/ConsoleTests/Program.cs
MongoCaleb: It's not relevant to this discussion, I think, and is covered in several other places. ¯\_(ツ)_/¯

nirinchev (approved):

PR #631, opened 2020-12-01
(DOCSP-13416): Describe how to return JSON (not EJSON) from a function

nlarew:
### Issue JIRA link:

(DOCSP-13416): Describe how to return JSON (not EJSON) from a function

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/ejson/functions#behavior

### DIFF
--- a/source/functions.txt
+++ b/source/functions.txt
@@ -50,7 +50,7 @@ Realm functions that take specific arguments depending on the service.
 Behavior
 ~~~~~~~~
 
-- Realm automatically serializes values returned from functions as
+- Realm automatically serializes values returned from functions to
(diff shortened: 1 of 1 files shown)


ccho-mongodb (commented):
LGTM, 3 nitpick suggestions:
1) I think "serializes ... to" is more common than "serializes ... as", but it probably doesn't affect the meaning.
2) "call... and return the stringified" -> "call... to return the stringified" seems to make more sense to me.
3) It could be helpful to provide a lead-in to the code block to provide context. E.g. " ... as shown in the code example below".

ccho-mongodb (approved):
Whoops meant to approve instead of comment.

PR #633, opened 2020-12-01
(DOCSP-12990): Add missing parts of query engine page

cbush:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12990

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/query-engine/ios/query-engine

### DIFF
--- a/source/ios/query-engine.txt
+++ b/source/ios/query-engine.txt
@@ -166,7 +166,7 @@ Database includes the object in the results collection.
          .. literalinclude:: /examples/generated/code/start/QueryEngine.codeblock.models.m
              :language: objective-c
 
-   You can set up the {+realm+} for these examples as follows:
+   You can set up the {+realm+} for these examples with the following code:
 
    .. tabs-realm-languages::
 
--- a/source/ios/query-engine.txt
+++ b/source/ios/query-engine.txt
@@ -476,4 +476,5 @@ the object.
 Summary
 -------
 
+- Realm accepts predicates based on NSPredicate to filter results.
 - There are several categories of **operators** available to filter results: comparison, logical, string, aggregate, and set operators.
--- a/source/ios/query-engine.txt
+++ b/source/ios/query-engine.txt
@@ -23,6 +23,7 @@ For an introduction to NSPredicate in general, See Apple’s
 `Predicates Programming Guide
 <https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Predicates/AdditionalChapters/Introduction.html>`__.
 
+
 Predicates
 ----------
 
--- a/source/ios/query-engine.txt
+++ b/source/ios/query-engine.txt
@@ -56,7 +56,7 @@ one of the following:
 Dot Notation
 ~~~~~~~~~~~~
 
-When referring to an object property, you can use the **dot notation** to refer
+When referring to an object property, you can use **dot notation** to refer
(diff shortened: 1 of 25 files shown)


source/ios/query-engine.txt
```diff
@@ -94,15 +208,15 @@ values.
      - Evaluates to ``true`` if the left-hand numerical or date expression is greater than or equal to the right-hand numerical or date expression. For dates, this evaluates to ``true`` if the left-hand date is later than or the same as the right-hand date.
 
    * - ``in``
-     - Evaluates to ``true`` if the left-hand expression is in the right-hand list.
+     - Evaluates to ``true`` if the left-hand expression is in the right-hand list or string.
```
nathan-contino-mongo, line 211: do you mean:
```suggestion
     - Evaluates to ``true`` if the left-hand expression is in the right-hand list of strings.
```
?

source/ios/query-engine.txt
```diff
@@ -56,17 +155,32 @@ An **expression** consists of one of the following:
...
+             :language: swift
 
       .. tab::
-          :tabid: objective-c
+         :tabid: objective-c
+
+         .. literalinclude:: /examples/generated/code/start/QueryEngine.codeblock.models.m
+             :language: objective-c
+
+   You can set up the realm for these examples as follows:
```
nathan-contino-mongo, line 169:
```suggestion
   You can set up the {+realm+} for these examples as follows:
```

nathan-contino-mongo: "as follows" is a little vague. Could we change this to "with the following Realm object models" or something like that?

source/ios/query-engine.txt
```diff
@@ -24,6 +24,111 @@ For an introduction to NSPredicate in general, See Apple’s
...
+one of the following:
+
+- The name of a property of the object currently being evaluated.
+- An operator and up to two argument expression(s).
+- A literal string, number, or date.
+
+Dot Notation
+~~~~~~~~~~~~
+
+When referring to an object property, you can use the **dot notation** to refer
```
nathan-contino-mongo, line 59: I feel like the article isn't necessary here. What do you think?
```suggestion
When referring to an object property, you can use **dot notation** to refer
```

nathan-contino-mongo (approved):
LGTM aside from a couple of small wording suggestions.

nathan-contino-mongo:
Also, you may want to do a little work to ensure that the iOS test suite continues to reliably run in CI :)

source/ios/query-engine.txt
cbush: No. I mean the property can be a list or a string (but not, say, a number). Open to suggestions.

nathan-contino-mongo: No problem, seems like i misunderstood what was going on here.

nathan-contino-mongo (approved):
LGTM

astigsen:
A few minor things missing:

In the list of comparison operators:
*  `=` is missing as alternative to `==`.
* `between` could really use an example. It may not be clear to users how to define a range. It should be written as `BETWEEN {a, b}`.

In the list of string comparison operators:
*  `=` is missing as alternative to `==`.
* missing `!=`

Set operators
* `SOME` is missing as alternative to `ANY`

PR #648, opened 2020-12-08
(DOCSP-13322): Add open realm examples for obj-c

cbush:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13322

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/open-close-objc/ios/open-a-realm


source/ios/open-a-realm.txt
```diff
@@ -20,17 +20,40 @@ passing in the user's
...
+   .. tab::
+      :tabid: objective-c
+
+      .. literalinclude:: /examples/generated/code/start/OpenCloseRealm.codeblock.open-synced-realm.m
+         :language: objective-c
 
 In the above example, the code shows how to open the {+realm+} *asynchronously* by using 
 ``asyncOpen``. You can also open a {+realm+} *synchronously* by using one of the 
 :swift-sdk:`Realm() <Structs/Realm.html>` initializers, but this may 
 lead to temporary data inconsistencies while the remote data is downloaded, and
```
mongomoe, line 40: I realize this paragraph is already in master but does it make sense to have a paragraph about something that is "generally not recommended"? Is it even generally not recommended? Correct me if I'm wrong but wouldn't it might be recommended in the use-case that the developer/reader is building an offline-first application or an application that has many users with a poor quality network connection?  


Would something like the alternate phrasing below (based on the adjacent [.NET SDK example](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/open-close-objc/dotnet/open-a-realm)) be better:

"In the above example, the code shows how to open the realm asynchronously by using asyncOpen. You can also open a realm synchronously by using one of the Realm() initializers. This works if the device is offline, but may lead to temporary data inconsistencies while the remote data is downloaded."

examples/ios/Examples/OpenCloseRealm.m
```diff
@@ -0,0 +1,92 @@
...
+    }];
+}
+
+- (void)testOpenLocalRealm {
+    // :code-block-start: open-local-realm
+    // Open the default realm
+    RLMRealm *defaultRealm = [RLMRealm defaultRealm];
+
+    // Open the realm with a specific file URL, for example a username
+    NSString *username = @"GordonCole";
```
mongomoe, line 79: Haha I like the twin peaks reference

examples/ios/Examples/OpenCloseRealm.m
```diff
@@ -0,0 +1,92 @@
...
+    NSString *partitionValue = @"some partition value";
+
+    RLMRealmConfiguration *configuration = [user configurationWithPartitionValue:partitionValue];
+
+    [RLMRealm asyncOpenWithConfiguration:configuration
+                           callbackQueue:dispatch_get_main_queue()
+                                callback:^(RLMRealm *realm, NSError *error) {
+        
+        if (error != nil) {
+            NSLog(@"Failed to open realm: %@", [error localizedDescription]);
```
mongomoe, line 28:
```suggestion
            NSLog(@"Failed to open realm: %@", [error localizedDescription]);
            // handle error
```
should prob have the same code comment as swift

source/ios/open-a-realm.txt
```diff
@@ -20,17 +20,40 @@ passing in the user's
...
+   .. tab::
+      :tabid: swift
+
+      .. literalinclude:: /examples/generated/code/start/OpenCloseRealm.codeblock.open-synced-realm-synchronously.swift
+         :language: swift
+
+   .. tab::
+      :tabid: objective-c
+
+      .. literalinclude:: /examples/generated/code/start/OpenCloseRealm.codeblock.open-synced-realm-synchronously.m
```
mongomoe, line 54: * Why does the Swift example for synchronously opening of a realm include just the actual opening of the realm + print, but the Objective C example for synchronously opening includes the user, partition and configuration value, and the actual opening of the realm?
* Can the line "RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration" be highlighted, it kind of gets lost in the weeds that this is the actual opening of the realm?

examples/ios/Examples/OpenCloseRealm.m
```diff
@@ -0,0 +1,92 @@
...
+
+    // Open the realm with a specific file URL, for example a username
+    NSString *username = @"GordonCole";
+    RLMRealmConfiguration *configuration = [RLMRealmConfiguration defaultConfiguration];
+    configuration.fileURL = [[[configuration.fileURL URLByDeletingLastPathComponent]
+                             URLByAppendingPathComponent:username]
+                             URLByAppendingPathExtension:@"realm"];
+    NSError *error = nil;
+    RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration
+                                                 error:&error];
```
mongomoe, line 86: Can an `NSLog(@"Opened realm: \(realm)")` be added here to be consistent with the adjacent swift example.

mongomoe (approved):
LGTM w/ some suggestions and questions

examples/ios/Examples/OpenCloseRealm.m
cbush: Removed from the other example.

mongomoe: Sounds good

PR #670, opened 2020-12-15
(DOCSP-13324): Read & Write Data Swift & Obj-C examples

cbush:
## Pull Request Info
Examples only

Please review output in source/examples/generated/

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13324

### DIFF
--- a/examples/ios/Examples/ReadWriteData.m
+++ b/examples/ios/Examples/ReadWriteData.m
@@ -1,0 +1,148 @@
+#import <XCTest/XCTest.h>
+#import <Realm/Realm.h>
+
+// :code-block-start: models
+@interface Example_Dog : RLMObject
+@property NSString *name;
+@property NSInteger age;
+@end
+
+// Enable Example_Dog for use in RLMArray
+RLM_ARRAY_TYPE(Example_Dog)
+
+// A dog owner has a primary key ID, a collection of dogs, and can be a member of multiple clubs.
+@interface Example_DogOwner : RLMObject
+@property NSInteger id;
+
+// To-many relationship - a dog owner can have many dogs
+@property RLMArray<Example_Dog *><Example_Dog> *dogs;
+
+// Inverse relationship - an owner can be a member of many clubs
+@property (readonly) RLMLinkingObjects *clubs;
+@end
+
+RLM_ARRAY_TYPE(Example_DogOwner)
+
+@interface Example_DogClub : RLMObject
+@property NSString *name;
+@property RLMArray<Example_DogOwner *><Example_DogOwner> *members;
+@end
+
+@implementation Example_Dog
+@end
+
+@implementation Example_DogOwner
+// Define the primary key for the class
++ (NSString *)primaryKey {
+    return @"id";
+}
+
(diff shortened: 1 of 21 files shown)


terakilobyte (approved):
lgtm

examples/ios/Examples/ReadWriteData.m
```diff
@@ -0,0 +1,148 @@
+#import <XCTest/XCTest.h>
+#import <Realm/Realm.h>
+
+// :code-block-start: models
+@interface Example_Dog : RLMObject
```
tgoyne, line 5: Not at all a big deal, but underscores in obj-c type names are unusual enough that it looks pretty weird to see.

tgoyne (approved):

examples/ios/Examples/ReadWriteData.m
cbush: I could use help with this. I was trying to find a way to 'namespace' the models so that they would be specific to the given examples. The goal is that we can put the model in the same unit test file. We might use `Dog` in a few places but with slightly different properties, hence `ReadWriteExample_Dog`, `NotificationsExample_Dog`. Obviously it would be better if it were just `Dog` in both places. Thoughts?

tgoyne: Wouldn't it be better to use an entirely different type if you need a different set of properties? It might be less confusing, but I guess it depends on what exactly the differences are.

The idiomatic way to do namespaces in obj-c is that you pick a three letter prefix that you stick in front of all the type names (which for Realm is RLM), so you'd have RWEDog and NFEDog..

cbush: Colocation of the type near the relevant code is another reason. I don't want someone later down the line updating the model thinking it affects only one page of examples when it actually affects other pages.

Dogs, people, tasks, and projects comprise the basic contrived set of example types. We could get creative here. Namespacing (or doing something like Dog1, Dog2 etc.) is a workaround, but it has the benefit that people could paste all of the examples together in the same app without colliding object types.

By the way, what would be the convention for namespacing in Swift? (Since I'm doing the same thing in Swift and the Realm objects must be at the top scope).

tgoyne: Swift namespacing is a big can of worms. The built-in thing is that each module is a namespace, and if you want to use the same name more than once in a module you're doing it wrong and should just not do that. I've recently seen people start talking about how Swift code should probably use obj-c style prefixes because Swift namespacing really doesn't work well in practice, but the mainstream view is still "just don't reuse names" and there's no widely accepted way to do so.

cbush: Good to know.

Another option entirely is to make a separate test target for each set of models, but that seems like overkill just to avoid some ugly names. We could also make it so the actual name in the test suite is SomeLongSpecificName_Dog and we automatically replace it with just Dog, but I would hate to introduce an error this way and defeat the purpose of hosting the code examples in a test suite.

I'm thinking I'll just go for minimal variations on the name for now.

PR #671, opened 2020-12-16
Move service advice into the table, no longer say "deprecated"

nathan-contino-mongo:
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/deprecated-to-removed/android/migrate

### DIFF
--- a/source/android/migrate.txt
+++ b/source/android/migrate.txt
@@ -27,15 +27,6 @@ New Features
 - The {+service-short+} Android SDK now includes {+client-database+}, which includes
   local object storage and :ref:`{+sync+} <android-sync-data>`.
 
-Deprecated Features
--------------------
-
-- The {+service-short+} SDKs do not provide hooks for calling Services (like the 
-  :ref:`Twilio <twilio-service>` and :ref:`AWS <aws-service>` services). We 
-  recommend using the AWS and/or Twilio ``npm`` packages within {+realm+} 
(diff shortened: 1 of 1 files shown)


source/android/migrate.txt
```diff
@@ -72,3 +63,6 @@ Changes
 
    * - The BSON package, which contains MongoDB query filters and the ``Document`` data type for reading/writing to MongoDB Atlas has moved from ``com.mongodb.stitch.core.internal.common`` to ``org.bson``.
      - Refactor all imports of the BSON package to use ``io.realm.mongodb.mongo``.
+     
+   * - The {+service-short+} SDK does not provide an interface for calling services such as :ref:`Twilio <twilio-service>` and :ref:`AWS <aws-service>`.
```
ccho-mongodb, line 67: Question: are these the correct links? I would have expected the Stitch links since those are the "services" instead of the Realm ones.

ccho-mongodb (approved):
LGTM, just one question about links.

source/android/migrate.txt
nathan-contino-mongo: The capabilities of the Stitch SDK services and the functions services are the same (functions are basically their own SDK, actually) so unfortunately there's just this one page. I could link to the Stitch legacy docs but it's essentially the same content without Android/iOS examples so probably easier to just link within the same docs.

PR #675, opened 2020-12-17
DOCSP-13320 revisions

ksadoff:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13320

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-13320-revisions/dotnet/objects


source/dotnet/objects.txt
```diff
@@ -182,25 +180,26 @@ class.
...
 - ``ObjectId``
-- ``Integer`` or ``int``
-- ``Long`` or ``long``
-- ``Short`` or ``short``
-- ``Byte`` or ``byte[]``
+- ``int``
+- ``long``
+- ``short``
+- ``byte``
+- ``char``
```
nirinchev, line 202: [nit] let's order these and move `char` closer to `string`:

```suggestion
- ``byte``
- ``short``
- ``int``
- ``long``
```

source/dotnet/objects.txt
```diff
@@ -215,52 +214,42 @@ objects of the same type in a {+realm+}.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If a field is normally nullable, you can make it required to disallow ``null`` values
-using the ``[Required]`` annotation. In .NET, fields are nullable only if marked with with the 
-C# `? operator <https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types>`__ except
-for the following types:
+If a property is normally nullable, you can make it required to disallow ``null`` values
+using the ``[Required]`` attribute. In .NET, properties are nullable only if marked with 
+`?  <https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types>`__ except
+for ``string`` and ``byte[]``.
```
nirinchev, line 233: Let's reword this and remove the second paragraph which is a bit repetitive:

```suggestion
In C#, value types are implicitly non-nullable, but can be made optional by appending 
`?  <https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types>`.
Reference types, such as ``string`` and ``byte[]``, are implicitly nullable and can be
made required by annotating them with the ``[Required]`` attribute. Properties of types
that inherit from ``RealmObject`` or ``EmbeddedObject`` are special and can always be
null. Applying ``[Required]`` on such a property will result in a compilation error.
```

source/dotnet/objects.txt
```diff
@@ -286,34 +275,35 @@ documents that it must inspect.
...
 certain queries faster. Indexes require space in your {+realm+} file, so
-adding an index to a field will increase disk space consumed by your
+adding an index to a property will increase disk space consumed by your
 {+realm+} file.
 
-You can index fields with the following types:
+You can index properties with the following types:
 
-- ``String``
+- ``string``
```
nirinchev, line 289: Can't add it as a suggestion, but let's order these too :)

```
- ``bool``
- ``byte``
- ``short``
- ``int``
- ``long``
- ``DateTimeOffset``
- ``char``
- ``string``
- ``ObjectId``
```

source/examples/Schemas/Inverse.cs
```diff
@@ -1,6 +1,6 @@
 class Dog : RealmObject
 {
-    // One to many relationship with Person.Dogs
+    // One to one relationship with Person.Dogs
```
nirinchev, line 3: This comment is a bit confusing. It should probably say `To-one relationship from the Dog to its owner`.

source/examples/Schemas/Inverse.cs
```diff
@@ -9,7 +9,7 @@ class Person : RealmObject
     [Backlink(nameof(Dog.Owner))]
```
nirinchev, line 9:
```suggestion
    // An inverse relationship that returns all Dog instances that have Dog.Owner set to
    // the current Person.
    [Backlink(nameof(Dog.Owner))]
```

source/examples/Schemas/Inverse.cs
```diff
@@ -9,7 +9,7 @@ class Person : RealmObject
...
     public IQueryable<Dog> Dogs { get; }
 
-    // Many to many relationship with Hobby.PeopleWithThatHobby
+    // One to many relationship with Hobby.PeopleWithThatHobby
```
nirinchev, line 12: Similarly for this comment - let's reword to something closer to `To-many relationship, containing a collection of all hobbies the current person enjoys`.

And on line 18 (can't comment on it), we can add a comment similar to what we had for the dogs property: 
```
    // An inverse relationship that returns all Person instances that have the current Hobby
    // instance in their Hobbies list.
```

source/dotnet/objects.txt
```diff
@@ -432,7 +423,7 @@ this behavior:
 - To change a class name in .NET without forcing a migration.
```
nirinchev, line 423: I can't comment on the correct line, but on 420, where it says `same simple name in different packages`, it should say `same simple name in different namespaces`.

nirinchev (commented):
Looks good. I have a bunch of nits + a few suggestions about rewording certain sections.

source/dotnet/objects.txt
```diff
@@ -100,57 +100,52 @@ objects in a {+realm+} conform to the
...
+
+.. _dotnet-property-lists:
 
 Lists
 ~~~~~
 
-Realm objects can contain lists of non-{+service-short+}-object data
-types. You can model these collections with the standard .NET interface
+Realm objects can contain lists of all supported data types. 
+You can model these collections with the standard .NET interface
```
nirinchev, line 139:
```suggestion
You can model these collections by defining a getter-only property of type
```

source/dotnet/objects.txt
```diff
@@ -100,57 +100,52 @@ objects in a {+realm+} conform to the
...
 
-Unlike lists of {+service-short+} objects, these lists can contain
-null values. If null values shouldn't be allowed, use the
-:ref:`[Required] <dotnet-required-field>` annotation.
-
-.. literalinclude:: /examples/Schemas/Lists.cs
-  :language: csharp
+These lists can contain null values. 
+If null values should not be allowed, do not use nullable types
+in the list declaration (i.e., use ``IList<double>`` instead of ``IList<double?>``).
```
nirinchev, line 145: Maybe expand a little on this like:

```suggestion
Lists of objects cannot contain null values - i.e. deleting an object from the database
will also remove it from any lists where it existed. Lists of primitive types can contain null
values. If null values should not be allowed, do not use nullable types
in the list declaration (i.e., use ``IList<double>`` instead of ``IList<double?>``) or
add the ``[Required]`` attribute if the list contains reference types, such as ``string``
or ``byte[]``.
```

Also, might want to link to the required/optional properties section from the `[Required]` attribute.

source/dotnet/objects.txt
```diff
@@ -215,52 +211,39 @@ objects of the same type in a {+realm+}.
...
-- ``Date``
-- ``ObjectId``
-- ``Decimal128``
-
-Any value type in C# can be made nullable with the ``?``, although by
-default they are non-nullable without it. 
-Fields with the ``RealmObject`` type are always
-nullable and cannot be required.
+In C#, value types are implicitly non-nullable, but can be made optional by appending 
+`?  <https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types>`.
```
nirinchev, line 228: This doesn't render as a proper link:

Screen Shot 2021-01-08 at 10 37 36

nirinchev (commented):
Two final clarifications for lists from me, otherwise seems good to go!

nlarew (approved):

nlarew:
@nirinchev Thanks for the review! Looks like all of your comments & suggestions are integrated so I'm gonna go ahead and merge this in. Feel free to ping me if there's anything else we should add/change!

PR #677, opened 2020-12-17
how to query inverse

MongoCaleb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13654

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/inverse-query/dotnet/relationships#inverse-relationship

### DIFF
--- a/examples/dotnet/Examples/Relationships.cs
+++ b/examples/dotnet/Examples/Relationships.cs
@@ -124,8 +124,8 @@ namespace Examples
                 realm.Add(user);
             });
 
-            var task1 = new Task() { Text = "oh hai", Assignee = user };
-            var task2 = new Task() { Text = "ok bai", Assignee = user };
+            var task1 = new Task() { Text = "Defribillate the master oscillator", Assignee = user };
+            var task2 = new Task() { Text = "Subvert the paradigm", Assignee = user };
             realm.Write(() =>
             {
                 realm.Add(task1);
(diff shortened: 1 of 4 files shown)


nirinchev (commented):
The code looks fine, but I feel we should come up with a better example for inverse relationships (I know this was not written in this PR, but we might as well fix it while we're at it).

`IQueryable<User>` is a collection so it's a bit inconsistent that we're calling the property `Assignee`. This gets even more confusing when we get to the querying part because the query is `$"Assignee._id == '{katie.Id}'"` which kind of implies that `Assignee` is an instance of user rather than a collection of users. If we were to translate this query to English, it would look something like "Give me all tasks where at least one element in the Assignee collection has _id equal to `katie.Id`". But because of the singular property name, a user might incorrectly assume that it's "Give me all tasks where the Assignee property has _id equal to `katie.Id`".

So I have two suggestions about how we can improve it:
1. Keep the current model and rename `Assignee` to `Assignees` and clarify in the docs that the reason why `Assignees` is a collection is because multiple users may have the same task in their `Tasks` collection. This makes sense in a model where many users could be collaborating on the same task.
2. Modify the model so that the forward relationship is on the `Task` - i.e. we'll define the property as `User Assignee { get; set; }` and move the inverse relationship on `User` - i.e. `[Backlink(nameof(Task.Assignee)] IQueryable<Task> Tasks { get; }`. That way we enforce a single assignee per task and the User gets an automatically populated collection of tasks where they are the assignee.

ksadoff (approved):
LGTM (once Nikola's suggestions are implemented). Based on this approach, I'll go ahead and add the other examples of how to query on relationships to the Relationships page, rather than the Query Engine page (for DOCSP-13325).

examples/dotnet/Examples/Relationships.cs
```diff
@@ -96,13 +99,48 @@ public class Task : RealmObject
...
+            var task2 = new Task() { Text = "ok bai", Assignee = user };
+            realm.Write(() =>
+            {
+                realm.Add(task1);
+                realm.Add(task2);
+            });
+
+            // :code-block-start: inverse-query
+            var katie = realm.All<User>().Where(u => u.Name == "Katie").FirstOrDefault();
+            var katiesTasks = realm.All<Task>().Filter($"Assignee._id == '{katie.Id}'");
```
nirinchev, line 137: With the new model, this is no longer demonstrating querying an inverse relationship. In fact, one can get Katie's tasks by using the inverse property directly: `var katiesTasks = katie.Tasks`.

Instead what we can do is something like:

```
// This returns all users that have at least one task assigned to them
// with the text "office party" ([c] is used for case insensitive lookup)
var partyOrganizers = realm.All<User>().Filter("Tasks.Text CONTAINS[c] 'office party'");
```

nirinchev (changes requested):

PR #682, opened 2021-01-04
DOCSP-13329: Modify an Object Schema (.NET)

ksadoff:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13329

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-13329/dotnet/migrations

### DIFF
--- a/source/dotnet/migrations.txt
+++ b/source/dotnet/migrations.txt
@@ -96,7 +96,7 @@ deletes it.
  
          .. literalinclude:: /examples/Migrations/PersonClassV4/PersonClassV4.cs
            :language: csharp
-           :emphasize-lines: 3
+           :emphasize-lines: 4
 
 
    To migrate the {+realm+} to conform to the updated ``Person`` schema, the
@@ -107,7 +107,7 @@ deletes it.
    
    .. literalinclude:: /examples/Migrations/LocalMigration/LocalMigration.cs
       :language: csharp
-      :emphasize-lines: 3, 15
+      :emphasize-lines: 3, 20, 26
 
 
 .. _dotnet-synced-migration:
--- a/source/dotnet/migrations.txt
+++ b/source/dotnet/migrations.txt
@@ -68,7 +68,7 @@ deletes it.
      .. tab::
          :tabid: c-sharp
  
-         .. literalinclude:: /examples/Migrations/PersonClassV1/PersonClassV2.cs
+         .. literalinclude:: /examples/Migrations/PersonClassV2/PersonClassV2.cs
            :language: csharp
            :emphasize-lines: 4
 
--- a/source/dotnet/migrations.txt
+++ b/source/dotnet/migrations.txt
@@ -61,7 +61,7 @@ deletes it.
            :language: csharp
            :emphasize-lines: 3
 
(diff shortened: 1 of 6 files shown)


source/dotnet/migrations.txt
```diff
@@ -59,31 +59,55 @@ deletes it.
...
+           :language: csharp
+           :emphasize-lines: 4
+
 
-   
    The developer decides that the ``Person`` class should use a combined
-   ``fullName`` field instead of the separate ``firstName`` and
-   ``lastName`` fields:
+   ``FullName`` field instead of the separate ``FirstName`` and
+   ``LastName`` fields:
```
nlarew, line 78: Can we make it clearer that this decision/change happens at a later time? i.e. after the developer has updated the app to use the separate `FirstName` and `LastName` property model

This is important because new version numbers only apply when you actually open the Realm with an updated schema and explicitly specify the new version number, so to get to version 4 we'd presumably need to open with versions 2 & 3 first

nirinchev:
I'll review this tomorrow morning.

source/examples/Migrations/LocalMigration/LocalMigration.cs
```diff
@@ -12,7 +12,19 @@
...
             var newPerson = newPeople.ElementAt(i);
             
-            newPerson.FullName = oldPerson.FirstName + " " + oldPerson.LastName;
+            // Changes from version 1 to 2 (adding LastName) will occur automatically when Realm detects the change
+
+            // Migrate Person from version 2 to 3: replace FirstName and LastName with FullName
+            if (oldSchemaVersion < 3)
+            {
+                newPerson.FullName = oldPerson.FirstName + " " + oldPerson.LastName;
+            }
```
nirinchev, line 21: This code is incorrect - if the old Realm is at version 1, then it won't have the `LastName` property. You'll need to guard against that:

```csharp
if (oldSchemaVersion < 2)
{
    newPerson.FullName = oldPerson.FirstName;
}
else if (oldSchemaVersion < 3)
{
    newPerson.FullName = $"{oldPerson.FirstName} {oldPerson.LastName}";
}
```

source/examples/Migrations/LocalMigration/LocalMigration.cs
```diff
@@ -12,7 +12,19 @@
...
+
+            // Migrate Person from version 3 to 4: replace Age with Birthday
+            if (oldSchemaVersion < 4)
+            {
+                newPerson.Birthday = DateTimeOffset.Now.AddYears(-(int)oldPerson.Age);
```
nirinchev, line 26: This is bound to produce incorrect results - while we can't reasonably determine someone's birthday knowing their age, it's certainly not desirable to assume their birthday was X years ago on today's date. Instead, we can set their birthday to January 1st 00:00:00 (freehand code, so there could certainly be errors):

```csharp
var birthYear = DateTimeOffset.UtcNow.Year - oldPerson.Age;
newPerson.Birthday = new DataTimeOffset(birthYear, 1, 1, 0, 0, 0, TimeSpan.Zero);
```

nirinchev (approved):

PR #690, opened 2021-01-11
(DOCSP-13327): Modify an Object Schema (Java/Kotlin)

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13327

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-13327/android/migrations#local-migration


source/android/migrations.txt
```diff
@@ -49,6 +49,15 @@ you must set to the value of the old property before {+service+} automatically
 deletes it.
 
 .. example::
+
+   .. note::
+
+      Note: assume that each schema change in this example occurs after
```
nlarew, line 55:
```suggestion
      In this example, assume that each schema change in this example occurs after
```

The extra "note" is a bit redundant

source/android/migrations.txt
```diff
@@ -49,6 +49,15 @@ you must set to the value of the old property before {+service+} automatically
 deletes it.
 
 .. example::
+
+   .. note::
```
nlarew, line 53:
```suggestion
   .. note:: Versions Update on Realm Open
```

Adds a title for tl;dr devs - just a suggestion, feel free to omit/change

source/examples/Migrations/PersonClassV1/PersonClassV1.kt
```diff
@@ -1,5 +1,4 @@
 class Person: RealmObject {
```
nlarew, line 1:
```suggestion
// Realm schema version 1
class Person: RealmObject {
```

Thoughts on adding a comment like this to each relevant snippet?

i.e. identify the version at-a-glance

nlarew (approved):
LGTM w/ some small suggestions

source/android/migrations.txt
nathan-contino-mongo: D'oh. Fixed.

source/android/migrations.txt
nathan-contino-mongo: I couldn't come up with a satisfying title earlier, thanks for coming up with one!

source/examples/Migrations/PersonClassV1/PersonClassV1.kt
nathan-contino-mongo: This could definitely make things a little clearer, particularly for users who don't read the full page end-to-end. Added!

cmelchior:
```
MongoDB Realm automatically migrates certain changes, such as new and deleted properties, but does not automatically set values for new properties unless the updated object schema specifies a default value. You can define additional logic in the migration that dynamically calculates a new property's value. Migrations do not allow you to directly rename a property. Instead, it treats a renamed property as a new property that you must set to the value of the old property before MongoDB Realm automatically deletes it.
```

This is not correct for Java. Java doesn't have automatic migrations, but manual ones. This means you need to define new and deleted properties, but renames are also easier as you can directly rename an existing field.

cmelchior (changes requested):
No offense, but this PR feels very rushed. The code wouldn't compile, nor is the logic in the examples accurate. I rewrote the migration code so it is accurate. Perhaps you can use it as a basis for reworking the PR?

```
        val migration = object: RealmMigration {
            override fun migrate(realm: DynamicRealm, oldVersion: Long, newVersion: Long) {
                var version: Long = oldVersion

                // DynamicRealm exposes an editable schema
                val schema: RealmSchema = realm.schema

                // Changes from version 1 to 2: Adding lastName.
                // All properties will be initialized with the default value "".
                if (version == 1L) {
                    schema.get("Person")!!
                            .addField("lastName", String::class.java, FieldAttribute.REQUIRED)
                    version++
                }

                // Changes from version 2 to 3: Combining firstName/lastName into fullName
                if (version == 2L) {
                    schema.get("Person")!!
                            .addField("fullName", String::class.java, FieldAttribute.REQUIRED)
                            .transform { obj: DynamicRealmObject ->
                                val name = "${obj.getString("firstName")} ${obj.getString("lastName")}"
                                obj.setString("fullName", name)
                            }
                            .removeField("firstName")
                            .removeField("lastName")
                    version++
                }

                // Changes from version 3 to 4: Replace age with birthday
                if (version == 3L) {
                    schema.get("Person")!!
                            .addField("birthday", Date::class.java, FieldAttribute.REQUIRED)
                            .transform { obj: DynamicRealmObject ->
                                var birthYear = Date().year - obj.getInt("age")
                                obj.setDate("birthday", Date(birthYear, 1, 1))
                            }
                            .removeField("age")
                    version++
                }
            }
        }
```

The migrations from 2 -> 3 and 3 -> 4, are very similar though, and just setting a fixed birthday feels a bit odd to me. So I would consider coming up with something else to show, like adding primary keys or maybe just consider removing the last step?

nathan-contino-mongo:
On the subject of:

```
The migrations from 2 -> 3 and 3 -> 4, are very similar though, and just setting a fixed birthday feels a bit odd to me. So I would consider coming up with something else to show, like adding primary keys or maybe just consider removing the last step?
```

Happy to change the last example, I was just following the example that we've set for other SDKs for consistency.

source/android/migrations.txt
```diff
@@ -39,18 +39,23 @@ To perform a local migration you must specify a new schema
...
 
-{+service+} automatically migrates certain changes, such as new and deleted
-properties, but does not automatically set values for new properties
-unless the updated object schema specifies a default value. You can
-define additional logic in the migration that dynamically calculates a
-new property's value. Migrations do not allow you to directly rename a
-property. Instead, it treats a renamed property as a new property that
-you must set to the value of the old property before {+service+} automatically
-deletes it.
+Because Android does not offer automatic migrations, you will need to
```
cmelchior, line 42: Maybe this is just me, but this feels written in a pessimistic way. 

There are tradeoffs between automatic and manual migrations, but none is inherently better. "automatic" migrations are easier for simple cases, but are harder to maintain for advanced use cases.

"manual" migrations require more boilerplate, but are write-once-and-forget.

We had long discussions around this in the SDK teams back in the day, but my impression is that either approach seems to work for users. At least we haven't seen the negative side effect of either materialize in a meaningful way.

For more context, you can see https://github.com/realm/realm-java/issues/2530#issuecomment-203845589

source/examples/Migrations/LocalMigration/LocalMigration.java
```diff
@@ -1,22 +1,42 @@
...
+     // All properties will be initialized with the default value "".
+     if (version == 0L) {
+        schema.get("Person")
+            .addField("lastName", String.class, FieldAttribute.REQUIRED);
+        version++;
+     }
+
+     // Changes from version 1 to 2: combine firstName/lastName into fullName
+     if (version == 1L) {
+        schema.get("Person")!!
```
cmelchior, line 19:
```suggestion
        schema.get("Person")
```
`!!` is only supported in Kotlin

source/examples/Migrations/LocalMigration/LocalMigration.java
```diff
@@ -1,22 +1,42 @@
...
+                obj.setString("fullName", name);
+            })
+            .removeField("firstName")
+            .removeField("lastName");
+        version++;
+     }
+
+     // Changes from version 2 to 3: replace age with birthday
+     if (version == 2L) {
+        schema.get("Person")!!
```
cmelchior, line 32:
```suggestion
        schema.get("Person")
```

cmelchior (approved):
Nice 👍

source/android/migrations.txt
nathan-contino-mongo: Updated the verbiage here to hopefully make this more optimistic. Can you take one last look at the updated text @cmelchior before I merge this in?

cmelchior (approved):
Looks great 👍

PR #695, opened 2021-01-11
(DOCSP-14116): Only run CI when that example app is modified

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14116


.github/workflows/dotnet-core.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches: [ master ]
```
nlarew, line 5: Should we remove the `on.push`?

nlarew (approved):
LGTM - Can you throw up a test branch with these updated workflows? e.g. a commit that updates only a file in `/examples/node`

Wanna make sure it works before merge

nathan-contino-mongo:
@nlarew I made a change to the android example app (renamed a top-level folder that was not well named) and I can see it building only the android CI app here as a result. Looks like it's working to me!

PR #701, opened 2021-01-13
Tutorial + S

nathan-contino-mongo:
This keeps driving me crazy.

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -7,7 +7,7 @@ MongoDB Realm
    :hidden:
    
    Get Started </get-started>
-   Tutorial </tutorial>
+   Tutorials </tutorial>
    Realm Database SDKs </sdk>
    MongoDB Realm Cloud </cloud>
    MongoDB Stitch (Legacy) </stitch>


cbush (changes requested):
Why?

nathan-contino-mongo:
@cbush The "tutorial" section contains a tutorial for setting up a backend, a set of mostly-equivalent tutorials for each platform, and a sort-of equivalent tutorial for getting started with Node. Many docs sites use the title "tutorials" (such as https://www.typescriptlang.org/docs/ and https://www.twilio.com/docs/sms for example) and even our other sections, like "Realm SDKs", "Third Party Services", "Values and Secrets", etc. are written as plurals when they encapsulate multiple things on one abstraction level instead of a single overarching concept.

cbush:
Counterpoint: each client is part of a single tutorial app. Referring to multiple tutorials might make it seem more disjointed than it is.

I don't personally feel too strongly either way, but let's get some form of consensus from the team and stakeholders before moving forward with this.

PR #702, opened 2021-01-13
DEPM-584: Redirect triggers overview to new triggers overview

nathan-contino-mongo:

### DIFF
--- a/config/redirects
+++ b/config/redirects
@@ -12,4 +12,4 @@ raw: realm/authentication/linking -> ${base}/authentication
 raw: realm/getting-started/configure-rules-based-access-to-mongodb -> ${base}/mongodb/define-roles-and-permissions
 raw: realm/procedures/create-stitch-app -> ${base}/get-started/create-realm-app
 raw: realm/graphql/graphql-types-and-resolvers/index.html -> ${base}/graphql/types-and-resolvers/
-raw: realm/triggers/overview -> ${base}/triggers
+raw: realm/triggers/overview -> ${base}/triggers/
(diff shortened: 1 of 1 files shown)


config/redirects
```diff
@@ -12,3 +12,4 @@ raw: realm/authentication/linking -> ${base}/authentication
 raw: realm/getting-started/configure-rules-based-access-to-mongodb -> ${base}/mongodb/define-roles-and-permissions
 raw: realm/procedures/create-stitch-app -> ${base}/get-started/create-realm-app
 raw: realm/graphql/graphql-types-and-resolvers/index.html -> ${base}/graphql/types-and-resolvers/
+raw: realm/triggers/overview -> realm/triggers
```
jdestefano-mongo, line 15: Pretty sure this needs the `${base}/` prefix, which includes "realm" so we'll need to drop that too

```suggestion
raw: realm/triggers/overview -> ${base}/triggers
```

jdestefano-mongo: fwiw I don't know if the trailing slash matters or not. I posted the question in the platform channel, check there for an update.

jdestefano-mongo (changes requested):
I believe there is an issue with the syntax and an open question with DOP. Once that's all sorted out, please deploy and test right away. Thanks!

config/redirects
nathan-contino-mongo: d'oh. Thank you for noticing this!

nathan-contino-mongo: As for the trailing slash: my basic testing shows that either works in my browser. So perhaps it does not matter?

config/redirects
```diff
@@ -12,3 +12,4 @@ raw: realm/authentication/linking -> ${base}/authentication
 raw: realm/getting-started/configure-rules-based-access-to-mongodb -> ${base}/mongodb/define-roles-and-permissions
 raw: realm/procedures/create-stitch-app -> ${base}/get-started/create-realm-app
 raw: realm/graphql/graphql-types-and-resolvers/index.html -> ${base}/graphql/types-and-resolvers/
+raw: realm/triggers/overview -> ${base}/triggers
```
nathan-contino-mongo, line 15:
```suggestion
raw: realm/triggers/overview -> ${base}/triggers/
```

nathan-contino-mongo (commented):
adding trailing slash

jdestefano-mongo (approved):
lgtm

terakilobyte (approved):
LGTM

PR #718, opened 2021-01-25
(DOCSP-14132): add note for electron writes from both processes

mongomoe:
### Issue JIRA link:
* https://jira.mongodb.org/browse/DOCSP-14132

* Electron: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/links-to-advanced-electron-guide/node/electron
* Electron with React: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/links-to-advanced-electron-guide/node/electron-cra

### DIFF
--- a/source/node/electron.txt
+++ b/source/node/electron.txt
@@ -99,4 +99,4 @@ renderer.js file:
    - :ref:`Sync Data <node-sync-data>`
    - :ref:`Define Your Schema <node-realm-objects>`
    - :ref:`Authenticate a User <node-authenticate>`
-   - :ref:`Write to a Realm <node-sync-data>`
+   - :ref:`Write to a Realm <node-realm-database-writes>`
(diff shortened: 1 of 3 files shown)


source/node/electron.txt
```diff
@@ -85,13 +85,18 @@ renderer.js file:
      
      // The myPartition realm is now synced to the device. You can
      // access it through the `realm` object returned by `Realm.open()`
+
+     // write to the realm
```
ccho-mongodb, line 89: Question: are there other potential actions that a user can take other than writing to the realm? This comment seems a little out of place because it is under the heading "Open a Realm" which doesn't imply writing to it. You might instead recommend a specific action to take once they have the sync code in place rather than just listing them in the "See Also" section.

ccho-mongodb (approved):
LGTM! Added a question about a new commented line added to the example.

source/node/electron.txt
mongomoe: Using the realm instance, you can either perform a write transaction or watch for changes to the realm. Good idea on recommending a specific action! Will add that in.

PR #719, opened 2021-01-26
Complete "Define a Realm Object Schema" page

cbush:
On "Define a Realm Object Schema" page, please review:
- All Swift examples
- All Objective-C examples
- Copy on both tabs
- Please check links to SDK reference manuals

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia-ios2/sdk/ios/examples/define-a-realm-object-model


source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+      .. literalinclude:: /examples/generated/code/start/ObjectModels.codeblock.define-a-model.swift
+         :language: swift
+
+   .. tab::
+      :tabid: objective-c
+
+      You can define a Realm object by deriving from the
+      :objc-sdk:`RLMObject <Classes/RLMObject.html>` or
+      :objc-sdk:`RLMEmbeddedObject
+      <Classes/RLMEmbeddedObject.html.html>` class. The name of the
```
nathan-contino-mongo, line 42:
```suggestion
      <Classes/RLMEmbeddedObject.html>` class. The name of the
```
Broken link to RLMEmbeddedObject

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+      class becomes the table name in the realm, and properties of the
+      class persist in the database. This makes it as easy to work with
+      persisted objects as it is to work with regular Objective-C
+      objects.
```
nathan-contino-mongo, line 46: In the Objective C code sample, it looks like you actually define an interface and then implement that interface with a class. Is this just the standard way to define classes in ObjC? I don't have much experience in that language, but I've never used a language that did things quite that way before. If that *isn't* the standard way of defining classes in ObjC, would it make sense to either:
- call that out as a best practice explicitly
- adjust the example to define a class in a more standard way (that most users will use)

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
 
-      Because Swift supports optional values, whether a property is
-      **optional** or **required** is built into the type of that
-      property. You can use ``RealmOptional`` to make optional integral
+      You can use the following types to define your object model
       properties:
 
-      .. literalinclude:: /examples/generated/code/start/Models.codeblock.person-model.swift
-         :language: swift
+      - Boolean ``Bool``
```
nathan-contino-mongo, line 65: nit: why is "Boolean" before ``Bool`` here, when for other types below (decimal128, double, etc.) we just use the type itself? I'm guessing it's for parity with the ObjC version of this copy, but for users who read through only the swift side of this, it could seem out of place.

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+      <Extensions/Object.html#/c:@CM@RealmSwift@@objc(cs)RealmSwiftObject(cm)primaryKey>`
+      to set the model’s primary key.
+
+      .. literalinclude:: /examples/generated/code/start/ObjectModels.codeblock.specify-a-primary-key.swift
+         :language: swift
+
+   .. tab::
+      :tabid: objective-c
+
+      Override :objc-sdk:`+[RLMObject primaryKey]
```
nathan-contino-mongo, line 214: Should this line include the + symbol? Not familiar enough with ObjC to say for sure.

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+      <Extensions/Object.html#/c:@CM@RealmSwift@@objc(cs)RealmSwiftObject(cm)indexedProperties>`
+      and return a list of indexed property names.
+
+      .. literalinclude:: /examples/generated/code/start/ObjectModels.codeblock.index-a-property.swift
+         :language: swift
+
+   .. tab::
+      :tabid: objective-c
+
+      To index a property, override :objc-sdk:`+[RLMObject
```
nathan-contino-mongo, line 251: same as above

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+
 .. _ios-ignore-a-property:
 
 Ignore a Property
 ~~~~~~~~~~~~~~~~~
 
-.. _ios-remap-a-property:
+Ignored properties behave exactly like normal properties. They can’t be
+used in queries and won’t trigger Realm notifications. They can still be
+observed using KVO.
```
nathan-contino-mongo, line 266: What is KVO? Should we place a link here?

nathan-contino-mongo: nit: passive voice

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+      and return a list of ignored property names.
+
+      .. literalinclude:: /examples/generated/code/start/ObjectModels.codeblock.ignore-a-property.swift
+         :language: swift
+
+   .. tab::
+      :tabid: objective-c
+
+      If you don’t want to save a field in your model to its realm,
+      override :swift-sdk:`+[RLMObject ignoredProperties]
```
nathan-contino-mongo, line 289: Broken link.

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+
+      .. literalinclude:: /examples/generated/code/start/EmbeddedObjects.codeblock.models.swift
+         :language: swift
+
+   .. tab::
+      :tabid: objective-c
+
+      You can define an embedded object by deriving from the
+      :objc-sdk:`RLMEmbeddedObject
+      <Classes/RLMEmbeddedObject.html.html>` class. You can use your
```
nathan-contino-mongo, line 436: broken link

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+      .. literalinclude:: /examples/generated/code/start/ObjectModels.codeblock.specify-a-primary-key.m
+         :language: objectivec
+
+
 .. _ios-index-a-property:
 
 Index a Property
 ~~~~~~~~~~~~~~~~
 
+Indexes make queries using equality and IN operators faster in exchange
```
nathan-contino-mongo, line 227: "IN" took me a second to comprehend here. Could we perhaps link to the queries page?

source/sdk/ios/examples/define-a-realm-object-model.txt
```diff
@@ -12,91 +12,429 @@ Define a Realm Object Schema
...
+for slightly slower writes. Indexes take up more space in the realm
+file. It’s best to only add indexes when optimizing the read performance
+for specific situations.
+
+Realm supports indexing for string, integer, boolean, Date, and ObjectId
```
nathan-contino-mongo, line 232: I feel like these properties should be monospaced (since they are referring to specific classes), except for maybe integer and string as "categories" of types. If the types aren't common between objc and swift, maybe we could move this sentence into the tabs and list types explicitly.

nathan-contino-mongo (approved):
LGTM with a few link fixes and nitpicks.

source/sdk/ios/examples/define-a-realm-object-model.txt
cbush: That is just how it's done - think header file vs. implementation file in C++

source/sdk/ios/examples/define-a-realm-object-model.txt
cbush: Yes, this shows that it is a "class" method (think `static` member function in C++)

source/sdk/ios/examples/define-a-realm-object-model.txt
cbush: We don't currently have a spot for the query page on this feature branch. Let's re-add later.

leemaguire (approved):
LGTM

PR #705, opened 2021-01-15
New SDK IA (Node) - Usage Examples

nlarew:
## Pull Request Info

All implemented/portable Usage Examples for the Node SDK

Still TODO:
- Fundamentals
- Advanced Guides
- Data Types (non-blocking)

### Issue JIRA link:

(DOCSP-13773): Update TOC & Pages to New IA (Node)

### Docs staging link (requires sign-in on MongoDB Corp SSO):

- [ 👍 ] [Open & Close a Local Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia/sdk/node/examples/open-and-close-a-local-realm)
- [👍 ] [Read & Write Data](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia/sdk/node/examples/read-and-write-data)
- [ 👍 ] [React to Changes](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia/sdk/node/examples/react-to-changes)
- [ 👍 ] [Authenticate Users](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia/sdk/node/examples/authenticate-users)
- [👍  ] [Manage Email/Password Users](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia/sdk/node/examples/manage-email-password-users)
- [ 👍 ] [Create & Manage User API Keys](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia/sdk/node/examples/manage-user-api-keys)
- [ 👍 ] [Access a Linked MongoDB Cluster](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia/sdk/node/examples/access-a-linked-mongodb-cluster)
- [ 👍 ] [Call a Function](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia/sdk/node/examples/call-a-function)

PR #723, opened 2021-01-27
(DOCSP-13321) Define a Realm Object Schema

mongomoe:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13321

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-13321-define-a-realm-object-schema/sdk/node/examples/define-a-realm-object-model


source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -51,18 +199,71 @@ Property Annotations
...
 
 Remap a Property
 ~~~~~~~~~~~~~~~~
+To use a different property name in your code than is stored in Realm DB, use
+the ``mapTo`` property. 
+
+.. example::
+   
+   A developer opens a realm client using a ``DogSchema`` with an alias of
+   ``firstName`` for it's ``first_name`` field.
```
cbush, line 260:
```suggestion
   ``firstName`` for its ``first_name`` field.
```

source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -51,18 +199,71 @@ Property Annotations
...
+.. _node-ignore-a-property:
 
 Remap a Property
 ~~~~~~~~~~~~~~~~
+To use a different property name in your code than is stored in Realm DB, use
+the ``mapTo`` property. 
+
+.. example::
+   
+   A developer opens a realm client using a ``DogSchema`` with an alias of
```
cbush, line 259: Do you mean `realm`?
```suggestion
   A developer opens a realm using a ``DogSchema`` with an alias of
```

source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -51,18 +199,71 @@ Property Annotations
...
+   
+         .. literalinclude:: /examples/Schemas/Indexed.js
+            :language: javascript
+            :emphasize-lines: 9
+
+.. _node-ignore-a-property:
 
 Remap a Property
 ~~~~~~~~~~~~~~~~
+To use a different property name in your code than is stored in Realm DB, use
```
cbush, line 254: We have been asked to avoid "Realm DB":
```suggestion
To use a different property name in your code than is stored in Realm Database, use
```

source/examples/Schemas/OptionalProperty.js
```diff
@@ -0,0 +1,7 @@
+const DogSchema = {
+    name: "Dog",
+    properties: {
+      breed: "string?"
+    }
+  };
+
```
cbush, line 7: Are these unit tested/bluehawked?

source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -17,30 +17,178 @@ Define a Realm Object Schema
 Object Properties
 -----------------
 
+Every {+service-short+} object has an *object type* that refers to the object's
+class. Objects of the same type share an :ref:`object schema
+<node-object-schema>` that defines the properties and relationships of those
+objects. Your {+service-short+} object should contain a ``name`` field and a
```
cbush, line 23: Is this a requirement? Should this be "must"? Otherwise, what happens if we don't include name and properties?

source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -51,18 +199,71 @@ Property Annotations
...
+Primary keys enforce uniqueness among objects in a realm. 
+
+You can efficiently find and update objects by primary key.
+
+As long as a realm manages an object, that object's primary key value is
+immutable.
+
+.. example::
+   
+   A developer creates a ``TaskSchema`` and defines a ``_id`` property as it's primary key.
```
cbush, line 211:
```suggestion
   A developer creates a ``TaskSchema`` and defines a ``_id`` property as its primary key.
```

cbush (approved):
Looks good with suggestions and questions. Can we get the JS team to look at this? Thanks!

source/examples/Schemas/OptionalProperty.js
mongomoe: No to both. JS has a lot of code that needs to be unit tested and bluehawked, including the code from the [new-ia branch](https://github.com/mongodb/docs-realm/pull/705/files). I have another [ticket](https://jira.mongodb.org/browse/DOCSP-13431) that will involve unit testing and bluehawking node content such as the code samples on this page.

source/sdk/node/examples/define-a-realm-object-model.txt
mongomoe: ``name`` and ``properties`` are the only required fields of the js-sdk's [object schema](https://docs.mongodb.com/realm-sdks/js/latest/Realm.html#~ObjectSchema). An error would be thrown if name isn't, but no error is thrown if "properties" aren't included (though I think this is a bug since it is a required field).

cbush: Then it sounds like it should be:

>You object must contain...

source/examples/Schemas/OptionalProperty.js
cbush: Aren't these net new files? Wouldn't it be easier to add them to the unit test now? Aren't we making more technical debt for ourselves later?

source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -17,30 +17,178 @@ Define a Realm Object Schema
...
+ 
+         .. literalinclude:: /examples/Schemas/DefineAProperty.js
+           :language: javascript
+           :emphasize-lines: 4, 5
 
 .. _node-specify-an-optional-required-property:
 
 Specify an Optional/Required Property
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+To mark a property as optional, append a question mark `?` to its value.
```
kneth, line 50: Is it to complicate things to say that you can also say `optional: true`?

source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -17,30 +17,178 @@ Define a Realm Object Schema
...
+     .. tab::
+         :tabid: typescript
+          
+         .. literalinclude:: /examples/Relationships/Inverse/Inverse.ts
+           :language: javascript
+           :emphasize-lines: 5, 14, 23, 32-36
+
 Define an Embedded Object Property
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+To specify that a Realm object model define an embedded object, set ``embedded``
```
kneth, line 172: I think you need to explain in more detail what an embedded object is. Luckily it is not a remote concept for a JS developer: an embedded object can be seen as a nested object:

```
let contact = {
  name: "Arthur Dent",
  address: {
    street: "155 Country Lane",
     city: "Cottington"
  }
};
```

source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -51,18 +199,71 @@ Property Annotations
...
+   
+         .. literalinclude:: /examples/Schemas/PrimaryKey.js
+            :language: javascript
+            :emphasize-lines: 9
 
 .. _node-index-a-property:
 
 Index a Property
 ~~~~~~~~~~~~~~~~
+Indexes make queries using equality and IN operators faster in exchange
```
kneth, line 226: Maybe you can mention that primary keys are always indexed.

kneth (approved):
I have only a few minor comments.

source/sdk/node/examples/define-a-realm-object-model.txt
mongomoe: 👍

source/examples/Schemas/OptionalProperty.js
mongomoe: Most aren't truly new files, there code snippets that are taken from other sections of the node docs. Yes doing this would be more technical debt (although arguably not much more considering these are schema files) for us, but given that our current priority is finishing the content rather than testing right away, this is preferable.

source/sdk/node/examples/define-a-realm-object-model.txt
mongomoe: 👍

source/sdk/node/examples/define-a-realm-object-model.txt
mongomoe: Added a brief explanation that it is a nested object without going into depth. This page is meant to be additional quick code examples that the reader can use after they have already read the fundamentals sections, so they should already know what an embedded object is.

source/sdk/node/examples/define-a-realm-object-model.txt
mongomoe: I think it's preferable to show the single example on this quick code examples page.

PR #724, opened 2021-01-27
Complete "open and close a local realm" page

cbush:
Note: this is based on https://github.com/mongodb/docs-realm/pull/719 - please ignore the first two commits until that is merged

Please review the code examples, links and text on the [Open & Close a Local Realm
](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia-ios3/sdk/ios/examples/open-and-close-a-local-realm) page.

PR #725, opened 2021-01-27
Complete "React to Changes" page

cbush:
This is based on https://github.com/mongodb/docs-realm/pull/724

Please review the [React to Changes](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia-ios4/sdk/ios/examples/react-to-changes) page.


source/examples/generated/code/start/Notifications.codeblock.register-a-collection-change-listener.m
```diff
@@ -0,0 +1,46 @@
...
+            [tableView reloadRowsAtIndexPaths:[changes modificationsInSection:0]
+                             withRowAnimation:UITableViewRowAnimationAutomatic];
+        } completion:^(BOOL finished) {
+            // ...
+        }];
+    }];
+}
+
+- (void)dealloc {
+    [self->_notificationToken invalidate];
```
leemaguire, line 44:
```suggestion
    [_notificationToken invalidate];
```
A comment about why you're doing this would be good. Maybe also a bit about where it should be placed in a view controllers lifecycle. For example, dealloc won't necessarily be called if they push to another view controller.

source/sdk/ios/examples/react-to-changes.txt
```diff
@@ -113,13 +111,40 @@ You can write to a {+realm+} *without* sending a notification to a specific
...
+      .. literalinclude:: /examples/generated/code/start/Notifications.codeblock.silent-write.m
+        :language: objectivec
+
+
+.. _stop-watching-for-changes:
+
+Stop Watching for Changes
+-------------------------
 
+Observation stops when the token returned by an ``observe`` call becomes
```
tgoyne, line 134: It might be worth reiterating that notifications will stop if the token is in a local variable that goes out of scope. It's stated above that you have to hold a strong reference to it, but this is something that seems to regularly trip up new users so repeating it in different ways could be good.

source/examples/generated/code/start/OpenCloseRealm.codeblock.configure-object-types.m
```diff
@@ -1,8 +1,12 @@
 RLMRealmConfiguration *config = [RLMRealmConfiguration defaultConfiguration];
 
 // Given a RLMObject subclass called `Task`
+<<<<<<< HEAD
```
tgoyne: Merge conflict markers

examples/ios/Examples/Notifications.m
```diff
@@ -0,0 +1,173 @@
...
+                             withRowAnimation:UITableViewRowAnimationAutomatic];
+        } completion:^(BOOL finished) {
+            // ...
+        }];
+    }];
+}
+
+- (void)dealloc {
+    [self->_notificationToken invalidate];
+}
```
tgoyne: The token doesn't have to be explicitly invalidated in dealloc.

source/sdk/ios/examples/react-to-changes.txt
cbush: I have updated to really drive the point home. Please check it out.

tgoyne (approved):

source/examples/generated/code/start/Notifications.codeblock.register-a-collection-change-listener.m
cbush: Good point! Per @tgoyne: "adding the notification in viewWillAppear: and removing it in viewWillDisappear might make sense" ... "I'm not totally confident that it's correct so I'd be inclined to just leave it alone"

Therefore let's think about addressing this later.

PR #729, opened 2021-01-28
(DOCSP-14171) Add best practices and service limitations for Realm Sync

dacharyc:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14171

https://docs-mongodborg-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14171/

Specifically, I've updated these pages:
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14171/sync
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14171/mongodb/service-limitations

Link to build log confirming I have not introduced any new errors: https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=6012fc03687cdd451a27c519

### DIFF
--- a/source/mongodb/service-limitations.txt
+++ b/source/mongodb/service-limitations.txt
@@ -139,6 +139,37 @@ following table lists the limitations for each cluster size:
    - Avoid data models that require you to Watch or Sync an unbounded
      number of collections.
 
+.. _mongodb-service-limitations-atlas-support:
+
+Atlas Support 
+-------------
+
+{+sync+} requires {+atlas+} clusters to run
+MDB 4.4.0 or greater. 
+
+We do not currently support sharded {+atlas+} clusters.
+
+When data is bulk/batch loaded into {+atlas+}, you may see a delay in
+data appearing on devices while {+realm+} processes changes.
+
+.. _mongodb-service-limitations-concurrent-workloads:
+
+Concurrent Workloads 
+--------------------
+
+{+realm+} supports up to 30
+users concurrently working with data from an underlying MongoDB
+document. With more than 30 concurrent users, you may see delays in
+syncing or conflict resolution.
+
+.. _mongodb-service-limitations-rules:
+
+Rules 
+-----
+
+{+sync+} supports :doc:`{+sync-short+} rules</sync/rules>`.
+{+sync+} does not support :doc:`roles/field-level
+rules</mongodb/define-roles-and-permissions>` at this time.
 
 Static Hosting
 --------------
--- a/source/mongodb/service-limitations.txt
+++ b/source/mongodb/service-limitations.txt
@@ -157,10 +157,21 @@ data appearing on devices while {+realm+} processes changes.
 Concurrent Workloads 
 --------------------
 
-{+service+} supports up to 30
-users concurrently working with data from an underlying MongoDB
-data source. With more than 30 concurrent users, you may see delays in
-syncing or conflict resolution.
+Writing:
+  {+service+} supports up to 30 users concurrently writing data to an
+  underlying MongoDB document. With more than 30 concurrent writers, you
+  may see delays in syncing or conflict resolution.
+
+Reading:
+  Frontend clients reading from a global {+realm+} can scale past tens
+  of thousands of concurrent users.
+
+.. _scalability:
+
+.. tip:: Scalability
+   
+   When you intend to have many concurrent writers, split projects
+   across {+realms+}. 
 
 .. _mongodb-service-limitations-rules:
 
--- a/source/mongodb/service-limitations.txt
+++ b/source/mongodb/service-limitations.txt
@@ -147,7 +147,7 @@ Atlas Support
 {+sync+} requires {+atlas+} clusters to run
 MDB 4.4.0 or greater. 
 
-We do not currently support sharded {+atlas+} clusters.
+{+service+} does not currently support sharded {+atlas+} clusters.
 
 When data is bulk/batch loaded into {+atlas+}, you may see a delay in
 data appearing on devices while {+realm+} processes changes.
--- a/source/mongodb/service-limitations.txt
+++ b/source/mongodb/service-limitations.txt
@@ -157,7 +157,7 @@ data appearing on devices while {+realm+} processes changes.
 Concurrent Workloads 
 --------------------
 
-{+realm+} supports up to 30
+{+service+} supports up to 30
 users concurrently working with data from an underlying MongoDB
 data source. With more than 30 concurrent users, you may see delays in
 syncing or conflict resolution.
--- a/source/mongodb/service-limitations.txt
+++ b/source/mongodb/service-limitations.txt
@@ -159,7 +159,7 @@ Concurrent Workloads
 
 {+realm+} supports up to 30
 users concurrently working with data from an underlying MongoDB
-document. With more than 30 concurrent users, you may see delays in
+data source. With more than 30 concurrent users, you may see delays in
 syncing or conflict resolution.
 
 .. _mongodb-service-limitations-rules:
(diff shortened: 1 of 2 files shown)


source/mongodb/service-limitations.txt
```diff
@@ -139,6 +139,37 @@ following table lists the limitations for each cluster size:
...
+
+.. _mongodb-service-limitations-concurrent-workloads:
+
+Concurrent Workloads 
+--------------------
+
+{+service+} supports up to 30
+users concurrently working with data from an underlying MongoDB
+data source. With more than 30 concurrent users, you may see delays in
+syncing or conflict resolution.
```
nathan-contino-mongo, line 163: Looks like the ticket suggests the move to this wording from the "30 concurrent writers" that we've currently got on the sync page. However, I think "30 concurrent writers" is a lot clearer to users -- this wording makes it sound like you shouldn't have more than 30 users on a realm, reader *or* writer! I would personally reach out to drew to see if the 30 concurrent writers recommendation has changed, because if it hasn't, I think we should use the wording "30 concurrent writers" instead (since it is much clearer).

nathan-contino-mongo (commented):
A few suggestions, but looking really good so far.

cbush (commented):
Commenting for now - will continue in a bit

source/sync.txt
```diff
@@ -199,6 +199,27 @@ To get the best performance from {+sync+}, design your
 intend to have many concurrent writers, split projects
 across {+realms+}.
 
+.. _realm-sync-best-practices:
```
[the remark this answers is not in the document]
dacharyc: I wasn't sure what to do with that point after removing the "30 concurrent writers" line. Without that line there, I'm not sure that paragraph makes sense. Do you think it still works if I remove the "As such" from the trailing line? Then the paragraph would be:

>Multiple writer, multiple reader:
  The real-time collaboration app use case.
  Each 'project' on which users can collaborate works best in its own {+realm+}.

If we remove the entire "Multiple writer, multiple reader" point, then I feel like we're losing the pattern of the "Use Case Profiles and Scalability" section. But without the 30 concurrent writers element, the two lines feel disconnected to me. Maybe the second line belongs in the new best practices section? But then that point becomes only: 

>Multiple writer, multiple reader:
  The real-time collaboration app use case.

Y'all are more familiar with the product, so I'd love a suggestion about what might be relevant to include here.

source/mongodb/service-limitations.txt
dacharyc: Is the preferred way to clarify a detail like this by leaving a comment on the Jira ticket, or asking in Slack? I know some places prefer to have the entire discussion in Jira so the details are captured in the ticket, but others prefer the faster communication that can happen via Slack.

dacharyc:
Alright, @nathan-contino-mongo and @cbush - I removed the section we discussed, which also had a related `ref` re: scalability - so I moved that `ref` to the new `Concurrent Workload` section on the service-limitations page. I've added a `tip` directive there regarding scalability. I'm torn on whether that covers _all_ of the relevant info from the section I removed.

For easy reference, here are the staging links for the two pages in question:
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14171/sync
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14171/mongodb/service-limitations

The section I removed is `Use Case Profiles and Scalability` - here's a link to the relevant page in the existing docs for your convenience: https://docs.mongodb.com/realm/sync

The thing that I think is really missing from the new `Concurrent Workload` content is this section:
>Single writer, single reader:
>Each user has their own private realm to store, e.g., user account information, preferences, or the individual user's shopping cart. As long as only the user or a backend application interacts with the realm data, the performance is ideal and scales linearly with number of users.

How important is this information? Should I find a way to incorporate that into the new content?

Here's an updated build log - I don't _think_ moving the `ref` has broken anything, but maybe one of you would spot something in the errors that I'm missing: https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=60133345ab23346c20569aa9

source/mongodb/service-limitations.txt
```diff
@@ -139,6 +139,48 @@ following table lists the limitations for each cluster size:
...
+Atlas Support 
+-------------
+
+{+sync+} requires {+atlas+} clusters to run
+MDB 4.4.0 or greater. 
+
+{+service+} does not currently support sharded {+atlas+} clusters.
+
+When data is bulk/batch loaded into {+atlas+}, you may see a delay in
+data appearing on devices while {+realm+} processes changes.
```
cbush, line 153: This should be {+service+} here. "realm" (lowercase) refers to a specific instance of Realm Database.
```suggestion
data appearing on devices while {+service+} processes changes.
```

source/mongodb/service-limitations.txt
```diff
@@ -139,6 +139,48 @@ following table lists the limitations for each cluster size:
    - Avoid data models that require you to Watch or Sync an unbounded
      number of collections.
 
+.. _mongodb-service-limitations-atlas-support:
+
+Atlas Support 
+-------------
+
+{+sync+} requires {+atlas+} clusters to run
+MDB 4.4.0 or greater.
```
cbush, line 148: Looking at this page overall, I wonder if we should mention this in the overview section as well? This page was originally about "Remote MongoDB Access" but now has useful Sync information. Just to make the whole page cohesive, maybe we should remove this line from the overview:

>This page provides a high-level overview of MongoDB feature availability in Realm.

cbush, line 148: Prefer full name:
```suggestion
MongoDB 4.4.0 or greater. 
```

source/mongodb/service-limitations.txt
```diff
@@ -139,6 +139,48 @@ following table lists the limitations for each cluster size:
...
+Reading:
+  Frontend clients reading from a global {+realm+} can scale past tens
+  of thousands of concurrent users.
+
+.. _scalability:
+
+.. tip:: Scalability
+   
+   When you intend to have many concurrent writers, split projects
+   across {+realms+}.
```
cbush, line 174: I know this is reworked from the original "use case profiles" section, but it doesn't really seem clear what "projects" we're referring to anymore or how to follow this advice. I might say something like:

>...concurrent writers, have each writer work on a separate document.

cbush (approved):
Looks great to me with a few comments. I would not worry about the "Single writer, single reader" case anymore. I was going for a rule of three originally, and it doesn't apply in the new page arrangement.

(Note: I usually "approve" instead of "request changes" when there are no major blocking issues, but I do appreciate you at least considering my comments before merging. Thanks!)

source/mongodb/service-limitations.txt
dacharyc: Good note. I took your suggestion and removed this line from the overview. I'm uncertain about mentioning this version requirement in the overview section; it seems redundant with this info now contained in a dedicated Atlas heading. However, I see that we do refer to some other version-specific information in the Overview.

I wonder if the current overview section should actually be renamed to something more specific to the CRUD and Aggregation functionality it details, and the page should get a new overview that more accurately captures the broader scope of the page content?

cbush: Agreed, might be redundant. I worry that it's jarring to read "3.x" here and then see "4.4" later if someone is scanning. I think your idea of updating the title of the "Overview" section would work well, but we're getting a bit out of scope for this ticket.

dacharyc: Good point, @cbush . I've made another ticket for me to go back and revisit this later: https://jira.mongodb.org/browse/DOCSP-14228

PR #739, opened 2021-02-01
Complete Migrations page

cbush:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13328

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/migrations/sdk/ios/examples/modify-an-object-schema

### DIFF
--- a/source/examples/generated/code/start/Migrations.codeblock.model-v1.m
+++ b/source/examples/generated/code/start/Migrations.codeblock.model-v1.m
@@ -1,0 +1,12 @@
+// Version 1 had separate fields for first name and last name
+@interface Person : RLMObject
+@property NSString *firstName;
+@property NSString *lastName;
+@property int age;
+@end
+
+@implementation Person
++ (NSArray<NSString *> *)requiredProperties {
+    return @[@"firstName", @"lastName"];
+}
+@end
--- a/source/examples/generated/code/start/Migrations.codeblock.model-v1.m
+++ b/source/examples/generated/code/start/Migrations.codeblock.model-v1.m
@@ -1,4 +1,5 @@
-// Version 1 had separate fields for first name and last name
+// In the first version of the app, the Person model
+// has separate fields for first and last names.
 @interface Person : RLMObject
 @property NSString *firstName;
 @property NSString *lastName;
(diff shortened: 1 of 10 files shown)


source/examples/generated/code/start/Migrations.codeblock.model-v1.m
```diff
@@ -0,0 +1,12 @@
+// Version 1 had separate fields for first name and last name
```
terakilobyte, line 1: nitpick: temporal framing

It initially struck me as odd that the choice of word was "had" here. Considering this is the first mention of Version, I think present tense is appropriate and won't take away from the message. Thoughts?

source/sdk/ios/examples/modify-an-object-schema.txt
```diff
@@ -12,22 +12,80 @@ Modify an Object Schema
...
+         :tabid: objective-c
+
+         .. literalinclude:: /examples/generated/code/start/Migrations.codeblock.model-v2.m
+            :language: objectivec
+
+
+   To migrate the {+realm+} to conform to the updated ``Person`` schema,
+   the developer sets the {+realm+}'s schema version to ``2`` and defines a
+   migration function to set the value of ``fullName`` based on the
+   existing ``firstName`` and ``lastName`` properties:
```
terakilobyte, line 76: suggestion: reorder

I think if you reorder this, it may be more clear.

Version 1
What the developer wants
How to achieve it
Code for end state
Code to get there

> The developer decides that the Person class should use a combined fullName field instead of the separate firstName and lastName fields. To migrate the realm to conform to the updated Person schema, the developer sets the realm's schema version to 2 and defines a migration function to set the value of fullName based on the existing firstName and lastName properties.

> .. updated person schema
> .. migration code

Thoughts?

terakilobyte (approved):
LGTM. I left some comments/suggestions but they aren't blockers.

source/examples/generated/code/start/Migrations.codeblock.model-v1.m
cbush: For sure. Updated!

source/sdk/ios/examples/modify-an-object-schema.txt
cbush: Updated. Let me know what you think.

terakilobyte:
To me it's easier to follow along with now. Much clearer end-state.

PR #744, opened 2021-02-02
(DOCSP-14228): Improve structure and organization of service-limitations page

dacharyc:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14228

https://docs-mongodborg-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14228/

Couple of notes for this one, @cbush and @nathan-contino-mongo :
- I wasn't sure if I needed to add a redirect for moving `service-limitations` from `mongodb` to `reference`.  The docs say to use redirects when renaming files, and this isn't a rename, but it's a different path so I assume it also needs a redirect. I did attempt to add one directly to `config` -> `redirects`.
- Reference as a directory seems like an appropriate place for this content now, since its scope has changed, but I found the only `Reference` in the ToC was under "MongoDB Realm Application Services." I don't know if that's the right place for this info - open to thoughts since you're both more familiar with the structure of these docs. I did also include it in the top-level intro doc under the "Billing & Reference" heading: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14228/


source/reference/service-limitations.txt
```diff
@@ -17,6 +17,17 @@ Service Limitations
 Overview
 --------
 
+{+service+} has several guidelines to keep in mind when architecting how
+your tools and clients interact with MongoDB through {+service-short+}.
+Keep this guidance in mind when deciding how to structure queries,
+selecting which CRUD and aggregation operations to use, and determining
+how to handle concurrent workloads.
```
cbush, line 24: Can we call the section "Concurrent Workloads with Sync"? On the combined page, it's not 100% clear that the limitation applies to Sync.

cbush (approved):
Looks good. Application Services is the right place for it. Our reference section is growing too much though. We will be revisiting the Application Services IA soon.

cbush:
A redirect would be appropriate.

nathan-contino-mongo:
🥳 🍾

PR #746, opened 2021-02-02
(DOCSP-14270): Fix broken literalinclude on Electron React QuickStart

mongomoe:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14270

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14270-Electron-cra-broken-link/node/electron-cra

### DIFF
--- a/source/includes/steps-realm-with-electron-using-create-react-app.yaml
+++ b/source/includes/steps-realm-with-electron-using-create-react-app.yaml
@@ -61,6 +61,7 @@ content: |
   application. Add the following to this file:
 
   .. literalinclude:: /examples/QuickStarts/electron/quickstart-setup-react-cracoconfig.js
+     :language: javascript
 
   The `target <https://webpack.js.org/configuration/target/>`_ is set to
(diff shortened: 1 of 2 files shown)


source/includes/steps-realm-with-electron-using-create-react-app.yaml
```diff
@@ -106,8 +106,9 @@ content: |
 
   .. literalinclude:: /examples/QuickStarts/electron/quickstart-setup-react-electron.js
```
cbush, line 107: I also notice the syntax highlighting is missing here. Should we fix that?

cbush (approved):
Looks fixed.

source/includes/steps-realm-with-electron-using-create-react-app.yaml
mongomoe: Good catch, added highlighting :)

PR #749, opened 2021-02-04
Migrate install and quickstarts

cbush:
Migrating existing content. Nothing really to review, but please check that the TOC under iOS now looks good.

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia-ios-toc/sdk/ios

### DIFF
--- a/source/sdk/ios.txt
+++ b/source/sdk/ios.txt
@@ -13,9 +13,8 @@ both Swift and Objective-C apps.
 .. toctree::
    :titlesonly:
    
-   Install Realm for iOS (TODO) </sdk/ios/install>
-   Quick Start (TODO) </sdk/ios/quick-start>
+   Install Realm for iOS, macOS, tvOS, and watchOS </sdk/ios/install>
+   Quick Start </sdk/ios/quick-start>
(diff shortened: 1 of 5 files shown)


source/sdk/ios.txt
```diff
@@ -13,9 +13,8 @@ both Swift and Objective-C apps.
 .. toctree::
    :titlesonly:
    
-   Install Realm for iOS (TODO) </sdk/ios/install>
-   Quick Start (TODO) </sdk/ios/quick-start>
+   Install Realm for iOS, macOS, tvOS, and watchOS </sdk/ios/install>
```
nlarew, line 16:
```suggestion
   Install the Realm SDK (Swift/Objective-C) </sdk/ios/install>
```

just a suggestion given that this is a super long page name and that it seems the official name is changing to Swift/Obj-C SDK instead of iOS

nlarew (approved):
LGTM w/ one suggestion, feel free to disregard

source/sdk/ios.txt
cbush: Fair point, but that hasn't happened yet. We can make that change in one go.

PR #758, opened 2021-02-08
(DOCSP-13993): Add User Id Type for Realm Custom User Data

nlarew:
### Issue JIRA link:

(DOCSP-13993): Add User Id Type for Realm Custom User Data

- [Enable Custom User Data > Procedure (Realm UI & CLI)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/custom-user-id/users/enable-custom-user-data/#procedure)
  ^ Each changed page maps to a tabbed procedure

### DIFF
--- a/source/includes/steps-define-custom-user-data-import-export.yaml
+++ b/source/includes/steps-define-custom-user-data-import-export.yaml
@@ -69,6 +69,10 @@ content: |
        - The name of a field that contains an application user's ID
          value. All user documents in the collection must have this
          field and no two documents may contain the same user ID.
+
+         The name of the field that maps each document to a specific application
+         user. The field must have the same name in each document and contain
+         the user's ID as a string.
 ---
 title: Deploy the Updated Application
 ref: deploy-the-updated-application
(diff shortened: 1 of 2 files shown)


source/includes/steps-define-custom-user-data-import-export.yaml
```diff
@@ -66,9 +66,14 @@ content: |
          data.
      
      * - ``user_id_field``
-       - The name of a field that contains an application user's ID
-         value. All user documents in the collection must have this
-         field and no two documents may contain the same user ID.
+       - The name of the field that maps each document to a specific application
+         user. The field must have the same name in each document and contain
```
ccho-mongodb, line 70: Not sure if my suggestion is better, but can a field be considered the same if the name is different? Same comment for both instances.

```suggestion
         user. The field must be present in every document and contain
```

ccho-mongodb (approved):
LGTM! Just one optional suggestion.

source/includes/steps-define-custom-user-data-import-export.yaml
nlarew: It technically doesn't have to be in every document, just in documents that you actually want to map to users. Updated for clarity.

PR #763, opened 2021-02-09
iOS fundamentals: write transactions

nathan-contino-mongo:
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/ios-write-transactions/sdk/ios/fundamentals/write-transactions/
Compare to node equivalent: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/ios-write-transactions/sdk/node/fundamentals/write-transactions/


source/sdk/ios/fundamentals/write-transactions.txt
```diff
@@ -11,3 +11,98 @@ Write Transactions
...
+{+client-database+} represents each transaction as a callback function
+that contains zero or more read and write operations. To run
+a transaction, define a transaction callback and pass it to
+the {+realm+}'s ``write`` method. Within this callback, you are
+free to create, read, update, and delete on the {+realm+}. If
+the code in the callback throws an exception when {+client-database+} runs
+it, {+client-database+} cancels the transaction. Otherwise, {+client-database+} commits
+the transaction immediately after the callback.
+
+.. important:: Concurrency Concerns
```
cbush, line 78: Let's delete or seriously trim down this note, per a recent discussion with the Cocoa team. The SwiftUI integration basically expects you to write on the main thread and it doesn't ultimately make a difference where the user writes. It's going to be out of the user's control if Sync locks everyone else out for a long time. Instead, it's something to be improved in the Sync core side.

source/sdk/ios/fundamentals/write-transactions.txt
```diff
@@ -11,3 +11,98 @@ Write Transactions
...
+
+When you are done with your transaction, {+client-database+} either
+**commits** it or **cancels** it:
+
+- When {+client-database+} **commits** a transaction, {+client-database+} writes
+  all changes to disk. For :term:`synced {+realms+} <{+sync+}>`, {+client-database+} queues the change
+  for synchronization with :term:`{+service+}`.
+- When {+client-database+} **cancels** a write transaction or an operation in
+  the transaction causes an error, all changes are discarded
+  (or "rolled back").
```
cbush, line 46: One thing I'd like to see near the top of these pages, possibly after this line, is something like a "tip" that says "To learn how to read and write data..." then links to the relevant Usage Examples page. I don't want people to land on this page and have to read too far if they were looking for examples.

cbush (approved):
Looks good with a couple comments.

source/sdk/ios/fundamentals/write-transactions.txt
nathan-contino-mongo: Instead I'm following the pattern I've already established in the Android SDK docs, which is a "see also" link after the relevant section (in this case, "run a transaction"). Let me know if you disagree, or we can just modify all of them later.

source/sdk/ios/fundamentals/write-transactions.txt
nathan-contino-mongo: Removed for now. My gut instinct is to leave the note as long as the warning is relevant (it sounds like it still is, but SwiftUI makes you ignore it anyway) but if it works well enough in SwiftUI I think we can live without the note.

cbush: It turns out that it has actually not really been super relevant to iOS, but more for Android which is more aggressive when it comes to non-responsive apps.

PR #773, opened 2021-02-11
Modify object schema dotnet

nathan-contino-mongo:
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/modify-object-schema-dotnet/sdk/dotnet/examples/modify-an-object-schema/


source/sdk/dotnet/examples/modify-an-object-schema.txt
```diff
@@ -14,71 +14,82 @@ Modify an Object Schema - .NET SDK
...
 Modify a Property Type or Rename a Property
 -------------------------------------------
 
+The following example modifies the ``Age`` property in the
+:ref:`original Person schema <dotnet-og-schema>` by
+renaming it to ``Birthday`` and changing the type to ``DateTimeOffset``:
+
+.. literalinclude:: /examples/generated/dotnet/Migrations.codeblock.ro4.cs
+   :language: csharp
+   :emphasize-lines: 8
```
cbush, line 76: Do we need the emphasize-lines anymore? I'm concerned these will drift fast.

cbush (approved):

source/sdk/dotnet/examples/modify-an-object-schema.txt
nathan-contino-mongo: I am OK with leaving them for the time being because I do think some of the larger examples (migration functions in particular) benefit from them. But perhaps the solution is actually to create a bluehawk command that lets you embed comments to generate a codeblock to literalinclude, such as:

```
Foo.kt
// :codeblock-start:
public Foo() : String {
     var foo: String
     foo = "test"
     return foo // :emphasize-line:
}
// :codeblock-end:

```
which would generate:
```
.. codeblock:: kotlin
   :emphasize-lines: 4
   
   public Foo() : String {
      var foo: String
      foo = "test"
      return foo // :emphasize-line:
   }

```
That way we could avoid emphasize-lines drifting from the source material.

nathan-contino-mongo: @cbush Should I make a github issue for this on Bluehawk?

PR #774, opened 2021-02-11
More iOS for new IA

cbush:
Various things per the [final checklist](https://github.com/mongodb/docs-realm/pull/720).

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia-ios-final/sdk/ios/


source/sdk/ios/examples/authenticate-users.txt
```diff
@@ -12,11 +12,13 @@ Authenticate Users
    :depth: 2
    :class: singlecol
 
+Log In
+^^^^^^
```
nathan-contino-mongo, line 16: Why not follow our standard of "---" for first-level headers, "~~~" for second level?

nathan-contino-mongo (approved):
LGTM

source/sdk/ios/examples/authenticate-users.txt
cbush: Laziness - fewer things to change this way. But updated.

PR #778, opened 2021-02-12
Remaining iOS Checklist

cbush:
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-ia-ios-final/sdk/ios/

- Add Data Types section
- Call a Function in Obj-C
- Link to the jazzy docs for Manage User API Keys

### DIFF
--- a/examples/ios/Examples/Functions.m
+++ b/examples/ios/Examples/Functions.m
@@ -1,0 +1,38 @@
+#import "AnonymouslyLoggedInTestCase.h"
+#import "MyRealmApp.h"
+
+@interface FunctionsObjc : AnonymouslyLoggedInTestCase
+@end
+
+@implementation FunctionsObjc
+- (void)testCallFunction {
+    XCTestExpectation *expectation = [self expectationWithDescription:@"it completes"];
+    // :code-block-start: call-a-function
+    RLMApp *app = [RLMApp appWithId:YOUR_REALM_APP_ID];
+    
+    // ... log in ...
+    
+    RLMUser *user = [app currentUser];
+
+    // Call sum function
+    [user callFunctionNamed:@"sum"
+                  arguments:@[@1, @2]
+            completionBlock:^(id<RLMBSON> result, NSError *error) {
+        if (error) {
+            NSLog(@"Function call failed: %@", [error localizedDescription]);
+            return;
+        }
+        NSLog(@"Called function 'sum' and got result: %@", result);
+        assert([result isEqual:@3]);
+        // :remove-start:
+        [expectation fulfill];
+        // :remove-end:
+    }];
+
+    // :code-block-end:
+    [self waitForExpectationsWithTimeout:10 handler:^(NSError *error) {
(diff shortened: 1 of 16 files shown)


source/sdk/ios/data-types.txt
```diff
@@ -1,3 +1,10 @@
...
 Realm Data Types - iOS SDK
-==============================
+==========================
+
+.. toctree::
+   :titlesonly:
+   
+   Supported Property Types </sdk/ios/data-types/supported-property-types>
+   Collections </sdk/ios/data-types/collections>
+   Embedded Objects </sdk/ios/data-types/embedded-objects>
```
nathan-contino-mongo, line 10: Should we add a bulleted list of doc links here so this page isn't empty?

source/sdk/ios/examples/manage-user-api-keys.txt
```diff
@@ -1,8 +1,8 @@
 .. _manage-user-api-keys:
 
-====================
-Manage User API Keys
-====================
+==============================
+Manage User API Keys (iOS SDK)
```
nathan-contino-mongo, line 4: Not to be nitpicky, but the android and dotnet SDKs use the format "Page - Android SDK". Since we already have two SDKs following that format, could we update these SDKs to use that format as well? I know it is an annoying change to make, so if you feel like this is a better format we can always update the Android/.NET SDKs to use this format instead.

examples/ios/Examples/Functions.m
```diff
@@ -0,0 +1,38 @@
...
+- (void)testCallFunction {
+    XCTestExpectation *expectation = [self expectationWithDescription:@"it completes"];
+    // :code-block-start: call-a-function
+    RLMApp *app = [RLMApp appWithId:YOUR_REALM_APP_ID];
+    
+    // ... log in ...
+    
+    RLMUser *user = [app currentUser];
+
+    // Call sum function
```
nathan-contino-mongo, line 17: The Swift code sample for this includes a pretty lengthy comment explaining the parameters passed to call this function. Should we add a similar comment here? Or should we maybe consider trimming down the Swift comment if the code is self-explanatory, because over time the code may change and we could forget to update the comment?

nathan-contino-mongo (approved):
LGTM with a couple of small comments and one big suggestion that might warrant further discussion.

source/sdk/ios/examples/manage-user-api-keys.txt
cbush: Oops, not sure why this slipped through. Will update. Thanks!

examples/ios/Examples/Functions.m
cbush: The big difference is because the Swift one uses some mind-bending dynamic lookup stuff that actually needs to be explained while this just takes the function name as a string.

nathan-contino-mongo: Good enough for me.

PR #781, opened 2021-02-16
[New Section] Realm Application Configuration (Feature Branch)

nlarew:
## Pull Request Info

Scaffolds the section + adds a first batch of pages

Details about what changed are in each ticket + https://docs.google.com/document/d/1J61TRBSNt87sUNsfNFiWopK_PC2vmvFCgadnkog7Q28/edit

### Issue JIRA link:

- (DOCSP-14493): [Config Files] Realm App
- (DOCSP-14494): [Config Files] User & Authentication Provider
- (DOCSP-14503): [Config Files] Static Hosting
- (DOCSP-14501): [Config Files] Value
- (DOCSP-14504): [Config Files] Application (Legacy)

- [Application Configuration Files (Legacy)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/app-structure/config/legacy) (current page for existing users - adds a legacy note to the top)
- [Realm Application Configuration (Overview)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/app-structure/config/) (new page, general outline of the section + key terms)
- [Realm App Configuration Files](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/app-structure/config/app) (same as previous config.json, but w/o a couple fields and renamed to `realm_config.json`)
- [User & Authentication Provider Configuration Files](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/app-structure/config/auth) (provider configs didn't change, but now they're all defined in a single JSON object instead of individual files)
- [Static Hosting Configuration Files](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/app-structure/config/hosting) (same config but now in its own file instead of `realm_config.json`)
- [Value Configuration Files](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/app-structure/config/values) (no changes from current implementation)


source/config/auth.txt
```diff
@@ -0,0 +1,195 @@
...
+.. code-block:: javascript
+   :caption: /auth/providers.json
+   
+   {
+     "<Provider Name>": {
+       "name": "<Provider Name>",
+       "type": "<Provider Type>",
+       "disabled": <Boolean>,
+       "config": {
+         "<Configuration Option>": <Configuration Value>
```
nathan-contino-mongo, line 47: Syntax highlighting falls apart after the <Configuration Value> field value -- consider putting it in quotes to keep syntax highlighting functioning:
```suggestion
         "<Configuration Option>": "<Configuration Value>"
```

source/config/hosting.txt
```diff
@@ -0,0 +1,118 @@
...
+              | String
+            - The name of the metadata attribute. This should be one of
+              the :doc:`file metadata attributes
+              </hosting/file-metadata-attributes>` that Realm supports.
+
+          * - | ``value``
+              | String
+            - The value of the metadata attribute.
+
+.. include:: /hosting/includes/note-infer-content-type.rst
```
nathan-contino-mongo, line 118: Should this note exist on multiple pages in this section? Seems like a generally useful statement, though I'm not sure how if it has an actionable consequence for most users.

nathan-contino-mongo (approved):
LGTM, with one caveat: I'm not sure that "Legacy Configuration Files" will make much sense to users, even those who are using earlier versions of realm-cli at the moment to manage application configurations. I think we should probably consult with engineering to see if there's a clearer label we can use than the very generic word "legacy". Otherwise, looks good.

source/config/hosting.txt
nlarew: It's specific to static hosting file metadata so not sure it's appropriate for any other config file pages

nathan-contino-mongo: OK. It only refers to "files", so maybe we should make the note more specific? Or maybe I just missed something.

source/config/auth.txt
nlarew: Well that's pretty annoying, good catch!

Technically the value is a product type like `string | number | object` so the quotes aren't 100% correct but I'd rather have syntax highlighting than pedantic correctness in this case :)

nlarew:
Added a note to our next product/docs sync to discuss legacy config files

PR #806, opened 2021-02-22
(DOCSP-14353): Node Client Resets Documentation

mongomoe:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14353

* Sync Changes Between Devices > "Perform a Client Reset" Subsection: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14353-Node-Client-Resets/sdk/node/examples/sync-changes-between-devices/#perform-a-client-reset
* Node Client Reset Doc: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14353-Node-Client-Resets/sdk/node/advanced/client-reset/#std-label-node-client-resets

### DIFF
--- a/examples/node/Examples/sync-changes-between-devices.js
+++ b/examples/node/Examples/sync-changes-between-devices.js
@@ -17,7 +17,7 @@ describe("Sync Changes Between Devices", () => {
     const credentials = Realm.Credentials.anonymous();
     await app.logIn(credentials);
     // :code-block-start: sync-changes-between-devices-perform-a-client-reset
-    let realm;
+    let realm = await Realm.open(config);
     function errorSync(_session, error) {
       if (realm) {
         if (error.name === "ClientReset") {
@@ -26,7 +26,7 @@ describe("Sync Changes Between Devices", () => {
           realm.close();
 
           console.log(`Error ${error.message}, need to reset ${realmPath}…`);
-          Realm.App.Sync.initiateClientReset(app, realmPath);
+          Realm.App.Sync.initiateClientReset(app, realmPath); // pass your realm app instance, and realm path to initiateClientReset()
           console.log(`Creating backup from ${error.config.path}…`);
           // Move backup file to a known location for a restore
           fs.renameSync(error.config.path, realmPath + "~");
@@ -37,7 +37,7 @@ describe("Sync Changes Between Devices", () => {
         }
       }
     }
-    const config = {
+    var config = {
       schema: [DogSchema], // predefined schema
       sync: {
         user: app.currentUser,
@@ -45,7 +45,6 @@ describe("Sync Changes Between Devices", () => {
         error: errorSync,
       },
     };
-    realm = await Realm.open(config);
     // :code-block-end:
     expect(1).toBe(2);
     realm.close();
--- a/examples/node/Examples/sync-changes-between-devices.js
+++ b/examples/node/Examples/sync-changes-between-devices.js
@@ -1,0 +1,53 @@
+import Realm from "realm";
+import fs from "fs";
+
+const DogSchema = {
+  name: "Dog",
+  properties: {
+    name: "string",
+    age: "int?",
+  },
+};
+
+describe("Sync Changes Between Devices", () => {
+  // this test is skipped because we currently are unable to test Synced Realms
+  // via jest, to track the progress of this issue see: https://jira.mongodb.org/browse/RJS-1008
+  test.skip("should perform a client reset", async () => {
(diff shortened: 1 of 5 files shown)


examples/node/Examples/sync-changes-between-devices.js
```diff
@@ -0,0 +1,53 @@
...
+    name: "string",
+    age: "int?",
+  },
+};
+
+describe("Sync Changes Between Devices", () => {
+  // this test is skipped because we currently are unable to test Synced Realms
+  // via jest, to track the progress of this issue see: https://jira.mongodb.org/browse/RJS-1008
+  test.skip("should perform a client reset", async () => {
+    const app = new Realm.App({ id: "<Your App ID>" });
```
cbush, line 16: It would be really nice if we could actually include this app in the code block. Otherwise we don't really know what "app" is.

examples/node/Examples/sync-changes-between-devices.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+    const config = {
+      schema: [DogSchema], // predefined schema
+      sync: {
+        user: app.currentUser,
+        partitionValue: "MyPartitionValue",
+        error: errorSync,
+      },
+    };
+    realm = await Realm.open(config);
```
cbush, line 48: It feels a bit confusing to have realm initialized down here. Why not initialize above when you declare the variable?

source/sdk/node/advanced/client-reset.txt
```diff
@@ -0,0 +1,61 @@
...
+
+Overview
+--------
+
+When using :doc:`{+sync+} </sync>`, a **client
+reset** is a serious error recovery task that your client
+app must perform in the following situation:
+
+- The given synced {+realm+} on the server was restored from a backup.
+  For example, due to a {+service-short+} server crash.
```
cbush, line 24: I know this might be inherited, but "For example, due to a crash" is not a complete sentence. I also don't know how much this applies since people can't just run their own ROS's anymore. Can we just delete the example line?
```suggestion
```

cbush (approved):
I really think the `app` instance declaration needs to be in the code example(s). Otherwise, looks great!

examples/node/Examples/sync-changes-between-devices.js
mongomoe: That would kind of bulk up this code sample (we would also have to include the lines under app instantiation), and then we would be back at the "using emphasize-lines vs concise code samples" debate. I did, however, add a comment clarifying what "app" was. Let me know if that doesn't work for you.

source/sdk/node/advanced/client-reset.txt
mongomoe: fair point

examples/node/Examples/sync-changes-between-devices.js
mongomoe: It is a bit confusing in general (largely due to the nature of client resets), unfortunately we would get the error: `ReferenceError: Cannot access 'config' before initialization` if we tried to initialize realm when the variable was declared (because const variables are not hoisted)

mongomoe: Consider the following oversimplified example:

```
class Realm {
  static open(config){
    return {openedRealm: true, myConfig: config}
  }
}
let realm = Realm.open(config);
function errorSync(){
  if(realm){
    console.log('this is an error handling function that uses a realm instance in some manner', realm);
    return `this is an error handling function that uses a realm instance in some manner ${realm}`;
  }else{
    return `an alternative error`;
  }

}

const config = {
  id: 1,
  sync: {
    user: "sample user",
    error: errorSync()
  }
}
```

here you would have the error: `ReferenceError: Cannot access 'config' before initialization`

mongomoe: Alternatively, the following oversimplified code example works but uses `var config`:
```
class Realm {
  static open(config){
    return {openedRealm: true, myConfig: config}
  }
}
let realm = Realm.open(config);
function errorSync(){
  if(realm){
    console.log('this is an error handling function that uses a realm instance in some manner', realm);
    return `this is an error handling function that uses a realm instance in some manner ${realm}`;
  }else{
    return `an alternative error`;
  }

}

var config = {
  id: 1,
  sync: {
    user: "sample user",
    error: errorSync()
  }
}

console.log('realm', realm);
```

I think this might be the best example, and I'll adapt our real example to be similar to this.

PR #813, opened 2021-02-24
(DOCSP-14219): Add new API endpoint to openAPI docs, add description to Authenticate Client Requests page

dacharyc:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14219

New endpoint documentation:
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14219/admin/api/v3/#post-/groups/%7Bgroupid%7D/apps/%7Bappid%7D/users/verify_token
Updated "Authenticate HTTP Client Requests" page (added the "Verify a Client Access Token" section): https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14219/reference/authenticate-http-client-requests/

### Docs staging build link:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=6037b63775e817aef7fe3c7e


source/openapi-admin-v3.yaml
```diff
@@ -838,6 +838,37 @@ paths:
...
+                token: {type: string}
+              required: [token]
+      responses:
+        '200':
+          description: Success returns this object.
+          content:
+            application/json:
+              schema: 
+                properties:
+                  sub: {type: string, description: Contains user ID and custom user data if the app uses it.}
```
nathan-contino-mongo, line 862: I don't have any background on this field, so feel free to ignore this, but I suspect it might be more accurate to say "if they exist for the user". Thoughts?

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,47 @@ following URL:
...
+Verify a Client Access Token
+----------------------------
+
+If you want to verify that a client access token is valid, you can do so using
+the verify access token endpoint in the :ref:`administration API <admin-api>`.
+
+The Admin API verify access token endpoint accepts a ``POST`` request that
+includes these details:
+
+- Use the client access token you got using one of the above methods for the payload
```
nathan-contino-mongo, line 165:
```suggestion
- the client access token you got using one of the :ref:`above methods <get-client-api-access-token>` for the payload
```

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,47 @@ following URL:
...
+- You'll need an :ref:`Admin API access token <realm-api-authentication>` ``Authorization`` header
```
nathan-contino-mongo, line 166:
```suggestion
- an :ref:`Admin API access token <realm-api-authentication>` ``Authorization`` header
```

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,47 @@ following URL:
...
+- The {groupId} and {appId} form the URL. View :ref:`the admin API documentation <realm-api-project-and-application-ids>` for details on how to find these
```
nathan-contino-mongo, line 167:
```suggestion
- the :ref:`{groupId} and {appId} <realm-api-project-and-application-ids>` for your backend {+app+}
```

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,47 @@ following URL:
...
+
+With this information, call this API endpoint to verify that the client access 
+token is valid:
+
+.. code-block:: text
+   
+   https://realm.mongodb.com/api/admin/v3.0/​groups/​{groupId}/​apps/​{appId}/​users/​verify_token
+
+.. important::
+   Because this endpoint requires admin API access, implementing this is best
```
nathan-contino-mongo, line 177: We tend to add an extra newline after note directives like this, so I would suggest adding that for consistency (though it is not required, as your staging output shows). I would also suggest adding a title, maybe something like "Do Not Access the Admin API Directly from Client Applications"

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,47 @@ following URL:
...
+With this information, call this API endpoint to verify that the client access 
+token is valid:
+
+.. code-block:: text
+   
+   https://realm.mongodb.com/api/admin/v3.0/​groups/​{groupId}/​apps/​{appId}/​users/​verify_token
+
+.. important::
+   Because this endpoint requires admin API access, implementing this is best 
+   done as a backend service behind an authorization layer - not directly
```
nathan-contino-mongo, line 178: For some reason DevEd seldom uses dashes in our documentation. But I'm inclined to say that this point is important enough to draw attention to in this way.

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,47 @@ following URL:
...
+   https://realm.mongodb.com/api/admin/v3.0/​groups/​{groupId}/​apps/​{appId}/​users/​verify_token
+
+.. important::
+   Because this endpoint requires admin API access, implementing this is best 
+   done as a backend service behind an authorization layer - not directly 
+   in the client app.
+
+.. example::
+
+   The following request demonstrates how to use a the Admin API endpoint to
```
nathan-contino-mongo, line 183:
```suggestion
   The following request demonstrates how to use the Admin API endpoint to 
```

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,47 @@ following URL:
...
+   done as a backend service behind an authorization layer - not directly 
+   in the client app.
+
+.. example::
+
+   The following request demonstrates how to use a the Admin API endpoint to 
+   verify that a user's client access token is valid. Replace ``<API Access Token>`` 
+   with the ``access_token`` value you get after authenticating with the
+   Admin API, and ``<Client Access Token>`` with the client access token you
+   got using one of the above methods.
```
nathan-contino-mongo, line 187: I would also mention that users need to replace the groupId and appId placeholders here with the values obtained from their Realm app, probably with another link to the admin api docs on those concepts.

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,47 @@ following URL:
...
+   got using one of the above methods. 
+   
+   .. code-block:: shell
+      :emphasize-lines: 3
+   
+      curl --request POST \
+        --url https://realm.mongodb.com/api/admin/v3.0/groups/{groupId}/apps/{appId}/users/verify_token \
+        --header 'Authorization: Bearer <API Access Token>' \
+        --header 'Content-Type: application/json' \
+        --data '{ "token": "<Client Access Token>" }'
```
nathan-contino-mongo, line 196: Finally, I would add a sentence after this explaining that if you get a 200 OK, the token is valid, otherwise it is invalid. And maybe mention the type of response you get if your admin auth token is expired/invalid, since that is a pretty common thing to hit.

source/openapi-admin-v3.yaml
```diff
@@ -838,6 +838,37 @@ paths:
...
+                  exp: {type: integer}
+                  iat: {type: integer}
+                  iss: {type: string}
+                  domain_id: {type: string}
+                  device_id: {type: string}
```
nathan-contino-mongo, line 867: I would love to see short explanations of these fields if they are at all relevant to users (maybe we can at least get that for exp?). But if they aren't relevant, this is fine.

nathan-contino-mongo (approved):
LGTM with a few style suggestions and minor nitpicks. Looking good!

source/openapi-admin-v3.yaml
dacharyc: I asked the engineer who created this endpoint about these fields, and she said that "sub" was the notable one, and gave me the info about user ID and custom data. I gathered that the rest aren't noteworthy, but I can ask for a brief description of all of them if it makes sense for documentation consistency. Thoughts?

nathan-contino-mongo: +1, sounds like a good solution to me!

source/reference/authenticate-http-client-requests.txt
dacharyc: Amusingly, I had tried adding a title before, but it showed up in the body of the directive as regular text, so I'm guessing that was related to not having the newline. I'll add the newline and put in a title again 👍

source/reference/authenticate-http-client-requests.txt
dacharyc: I'm glad you mentioned this! Turns out, you get a 200 OK if the token is valid, OR if the token is expired, but the response body is different in those cases. I'll cover that here, and maybe add a note in the openAPI doc about it, too.

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,95 @@ following URL:
...
+
+.. _verify-client-access-token-responses:
+
+Responses
+~~~~~~~~~
+
+**Valid access token**: If you get a ``200 OK`` with this object, the <Client Access Token> is valid:
```
nathan-contino-mongo, line 205:
```suggestion
**Valid access token**: If you get a ``200 OK`` with a response that contains the following fields, the <Client Access Token> is valid:
```

When I first read this, I translated it to "if you get a 200 OK, it's valid" in my head, which the second bullet point contradicts. I figure if that can happen to me, it can happen to any reader, so we should be a tiny bit more specific here than "this object".

nathan-contino-mongo (approved):
LGTM! Merge when the feature is available.

source/openapi-admin-v3.yaml
```diff
@@ -839,6 +839,38 @@ paths:
...
+                token: {type: string}
+              required: [token]
+      responses:
+        '200':
+          description: "Success returns this object. A 200 :ref:`may also return token expired<verify-client-access-token-responses>`."
+          content:
+            application/json:
+              schema: 
+                properties:
+                  sub: {type: string, description: "Contains user ID and :ref:`custom user data <custom-user-data>` if it exists for the user."}
```
caiyingshen, line 863: this is largely my fault since I phrased it weirdly 😅  but `custom_user_data` is actually an additional property. 
`sub` will just be the user ID and `custom_user_data` will include the user data if it is enabled on Realm. 

if it helps, this is a [direct link to the code ](https://github.com/caiyingshen/baas/blob/a9c6bc02fbd00e5bc91e9f2ec3f2eab3d1cc9cd2/api/models/v3/auth/access_token.go#L210-L218)that includes these fields. 
notably, `custom_user_data` and `aud` are also possible fields but they may be omitted if empty.

caiyingshen (commented):
this looks really good! 🎉  i just have one comment about some of the missing properties in the response. that's totally my bad for not linking you to those earlier!

source/openapi-admin-v3.yaml
dacharyc: Thanks, @caiyingshen ! I just updated the fields for this object; let me know if you spot anything else. 😃

source/reference/authenticate-http-client-requests.txt
```diff
@@ -150,3 +150,95 @@ following URL:
...
+.. _verify-client-access-token:
+
+Verify a Client Access Token
+----------------------------
+
+If you want to verify that a client access token is valid, you can do so using
+the `verify access token endpoint <{+base-url+}{+admin-api-page+}post-/groups/{groupId}/apps/{appId}/users/verify-token>`__ 
+in the administration API.
+
+This endpoint accepts ``POST`` request that includes these details:
```
cbush, line 163:
```suggestion
This endpoint accepts a ``POST`` request that includes these details:
```

cbush (approved):
Looks great!

source/openapi-admin-v3.yaml
mburdette: one clarification on `custom_user_data`, it will only contain that data if this access token was create AFTER custom user data was enabled and configured, if you were to get the access token via login and then enable custom user data we will not fetch this data separately, custom user data is stored in the access token

mburdette (approved):
LGTM I think @caiyingshen covered everything.

source/openapi-admin-v3.yaml
dacharyc: Thanks, @mburdette ! Just added a note summarizing that.

caiyingshen (approved):
LGTM 🎉  thanks Dachary!

dacharyc:
I've confirmed this endpoint is live in production now, so I'll go ahead and merge. Thanks for the feedback, all!

PR #829, opened 2021-02-25
(DOCSP-14676) iOS Local-Only Quickstart 

dacharyc:
## Pull Request Info

Ready for review!

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14676

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/ios-local-quickstart/sdk/ios/quick-start-local/

### Docs staging build link:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=603d088a75e817aef7ecde1f

### Review Guidelines


examples/ios/Examples/LocalOnlyCompleteQuickStart.swift
```diff
@@ -0,0 +1,97 @@
...
+        self.init()
+        self.name = name
+    }
+}
+
+// Entrypoint. Call this to run the example.
+func runLocalOnlyExample() {
+    // Open the local-only default realm
+    let localRealm = try! Realm()
+        // Get all tasks in the realm
```
cbush, line 26: This line and everything below it to the end of the function must be dedented by 1.

examples/ios/Examples/LocalOnlyCompleteQuickStart.swift
```diff
@@ -0,0 +1,97 @@
...
+        let taskToDelete = tasks[0]
+        try! localRealm.write {
+            // Delete the LocalOnlyQsTask.
+            localRealm.delete(taskToDelete)
+        }
+
+        print("A list of all tasks after deleting one: \(tasks)")
+
+        // :hide-start:
+        gExpectation!.fulfill()
```
cbush, line 83: It seems we should be able to get rid of gExpectation completely.

cbush (commented):
One formatting issue. Otherwise, code looks great.

source/sdk/ios.txt
```diff
@@ -14,6 +14,7 @@ both Swift and Objective-C apps.
    :titlesonly:
    
    Install Realm for iOS, macOS, tvOS, and watchOS </sdk/ios/install>
+   Quick Start Local </sdk/ios/quick-start-local>
```
cbush, line 17: From the user's perspective, "local" should be default and they won't know what we mean by local-only. Sync is the add-on, so we need not say local. (If possible, please update the file name too)
```suggestion
   Quick Start </sdk/ios/quick-start-local>
```

source/sdk/ios/quick-start-local.txt
```diff
@@ -0,0 +1,170 @@
+.. _ios-client-quick-start-local:
+
+==================================
+Quick Start (Local-only) - iOS SDK
```
cbush, line 4: I would also remove "Local-only" here.

source/sdk/ios/quick-start-local.txt
```diff
@@ -0,0 +1,170 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+This page contains information to quickly get started with a local-only
```
cbush, line 18: I know this is inherited, but maybe we should replace this "Overview" with a "Prerequisites" section that jumps straight into the bullet list. The preamble here feels redundant.

source/sdk/ios/quick-start-local.txt
```diff
@@ -0,0 +1,170 @@
...
+{+client-database+}. Before you begin, ensure you have:
+
+- :ref:`Installed the iOS SDK <ios-install>`
+
+.. note:: This Quickstart is Local-only
+   
+   This page contains only the essential information to get started with
+   a local-only {+client-database+}. For a quickstart featuring communicating
```
cbush, line 26: I would remove this first line. I feel we're really looking at this as if people are assuming Realm Sync is the default, but they won't be.

source/sdk/ios/quick-start-local.txt
```diff
@@ -0,0 +1,170 @@
...
+
+   import RealmSwift
+
+Define Your Object Model
+------------------------
+
+For a local-only {+client-database+} you can define your :ref:`object model
+<ios-realm-objects>` directly in code.
+
+.. code-block:: swift
```
cbush, line 47: The Sync quick start was written well before Bluehawk 0.1 which supports nested code snippets. So you can now go through and add code-block tags around each of these things you want to include.

cbush (changes requested):
Great first start! Let's get all of those code blocks into Bluehawk. Thanks!

source/sdk/ios/quick-start-local.txt
```diff
@@ -0,0 +1,170 @@
+.. _ios-client-quick-start-local:
```
dacharyc, line 1: So question, @cbush - does removing the "local-only" references you identified in your comments preclude us from using "local" in the :ref: label here? The existing quick start, which is now "Quick Start with Sync" uses the plain quick start label: `.. _ios-client-quick-start:`. 

Should I: 
A) Change that to `.. _ios-client-quick-start-with-sync:` (and update associated :ref: links to it) and then use `.. _ios-client-quick-start:` here, or;
B) Leave this `local` in the :ref: label since that's really just for internal use?

cbush: I would do A, because refs do show up as anchor links (# in URL). That said, it's not a big deal to have -local in there too.

source/sdk/ios.txt
dacharyc: I'll make these changes, but I'm wondering if leaving the "local" would be helpful for users who are aware there is both a local-only Realm and a Realm w/Sync option. Having "local" in the title and key places in the doc might make it easier for folks who are specifically looking for local-only documentation to identify this as the right page for them. Thoughts?

source/sdk/ios/quick-start-local.txt
dacharyc: So in updating the associated :ref: links, I see there's a link in the "Check Out the Quick Start" note admonition on the Swift Task Tracker tutorial (ios-swift.txt): 

>If you prefer to explore on your own rather than follow a guided tutorial,
   check out the :ref:`iOS Quick Start <ios-client-quick-start>`. It includes
   copyable code examples and the essential information that you need to set up
   a MongoDB Realm application.

Since the tutorial features sync, I assume this should point to the "with Sync" quick start?

cbush: Yes, I would link to the "with Sync" version there.

cbush (approved):
Looks good! Could you ask the Realm Swift team to take a look?

PR #835, opened 2021-03-01
Add Dictionaries and Sets, and update to Collections

MongoCaleb:
# DO NOT DEPLOY UNTIL SDK IS PUBLISHED

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14562
https://jira.mongodb.org/browse/DOCSP-14570

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/dicts_and_sets/sdk/dotnet/data-types/dictionaries/
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/dicts_and_sets/sdk/dotnet/data-types/sets/
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/dicts_and_sets/sdk/dotnet/data-types/collections/

**NEW**
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/dicts_and_sets/sdk/dotnet/data-types/data-type-queries/

### Docs staging build link:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=603d307c75e817aef7f40cc4

### Review Guidelines


source/sdk/dotnet/data-types/dictionaries.txt
```diff
@@ -0,0 +1,48 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+A {+service-short+} dictionary is an implementation of   
+`IDictionary <https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2?view=net-5.0>`_ 
+that has keys of type ``String`` and supports values of any 
+:ref:`{+service-short+} type <dotnet-objects>` except :ref:`dotnet-client-collections`.
```
nirinchev, line 20: The `dotnet-client-collections` ref renders as "values of any Realm type except Collections - .NET SDK." which reads pretty weird.

source/sdk/dotnet/data-types/dictionaries.txt
```diff
@@ -0,0 +1,48 @@
...
+A {+service-short+} dictionary is an implementation of   
+`IDictionary <https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2?view=net-5.0>`_ 
+that has keys of type ``String`` and supports values of any 
+:ref:`{+service-short+} type <dotnet-objects>` except :ref:`dotnet-client-collections`. 
+To define a dictionary, use 
+a getter-only ``IDictionary<string, TValue>`` property, where ``TValue`` is any 
+of the supported types.
+
+Deleting an object from the database will remove it from any dictionaries 
+in which it existed. Therefore, a dictionary of objects will never contain null objects.
```
nirinchev, line 26: This is not true - it will set the value to `null`, but the key will remain.

source/sdk/dotnet/data-types/dictionaries.txt
```diff
@@ -0,0 +1,48 @@
...
+However, dictionaries of primitive types can contain null values. If you do not 
+want to allow null values in a dictionary, then either use non-nullable types in 
+the dictionary declaration (for example, use ``IDictionary<double>`` instead of
```
nirinchev, line 29: These definitions need to be `IDictionary<string, double>`, etc.

examples/dotnet/Examples/DataTypes.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            var foo = new SetStuff();
+            foo.DoubleSet.Add(22.22);
+
+        }
+        //:code-block-start:sets
+        public class SetStuff : RealmObject
+        {
+            // A Set can contain any Realm-supported type, including
+            // objects that inherit from RealmObject
```
nirinchev, line 35: or EmbeddedObject

source/sdk/dotnet/data-types/dictionaries.txt
```diff
@@ -0,0 +1,48 @@
...
+contains nullable reference types, such as ``string`` or ``byte[]``.
+
+The following code shows examples of dictionary types:
+
+.. literalinclude:: /examples/generated/dotnet/DataTypes.codeblock.dictionaries.cs
+   :language: csharp
+
+.. note::
+   
+   You cannot query against a {+service-short+} Dictionary.
```
nirinchev, line 41: This is true for LINQ, but not for string queries. String queries are fully supported and have the following syntax (which is common for all SDKs, so probably a good idea to extract it to a shared page):

```csharp
// Find all Inventory items that have "Petunia" as a key in their PlantDict.
var matches = realm.All<Inventory>().Filter("PlantDict.@keys == 'Petunia'");

// Find all Inventory items that have at least one value in their IntDict that is larger than 5
var matches = realm.All<Inventory>().Filter("IntDict.@values > 5");

// Find all Inventory items where StringDict has the key "Foo" and its value
// contains the phrase "bar" regardless of what the case is
var matches = realm.All<Inventory>().Filter("StringDict['Foo'] CONTAINS[c] 'bar'");
```

nirinchev (commented):
A few quick corrections about dictionaries. Will review the rest later today.

source/sdk/dotnet/data-types/dictionaries.txt
MongoCaleb: @nirinchev on the last example, since keys are unique, finding a key of "Foo" will always result in 0 or 1 results, right? So the rest of that query (the contains) isn't necessary. Yes?

nirinchev: This is filtering for inventory objects, each of which contains a dictionary. It may or may not contain the Foo key and many dictionaries can contain it. The [c] part is the query syntax for case insensitive.

MongoCaleb: Also, I'll need examples for Sets, too, as I can't get them working correctly.

cbush (changes requested):
Looks great. Blocking until available.

PR #836, opened 2021-03-01
Subsection organization for Android SDK landing page

nathan-contino-mongo:
## Pull Request Info
- staging and evaluating updates based on chats with Realm stakeholders

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/android-landing-page-subsections/sdk/android/

### Docs staging build link:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=603d365075e817aef7f51bcf

### Review Guidelines

### DIFF
--- a/source/sdk/android.txt
+++ b/source/sdk/android.txt
@@ -214,7 +214,7 @@ For practical code samples of common tasks in {+client-database+} and
    * - Get Started
      - Fundamentals
      - Examples
-     - Sync Examples
+     - MongoDB Realm App Examples
      
    * - :doc:`Install Realm for Android </sdk/android/install>`
    
--- a/source/sdk/android.txt
+++ b/source/sdk/android.txt
@@ -154,8 +154,8 @@ The {+service+} Android SDK optionally contains the ability to access
 backend {+backend+} apps running in the cloud. In addition to local
 {+client-database+} functionality, {+app+}s provide:
 
-Automatic Sync
-~~~~~~~~~~~~~~
+Realm Sync
+~~~~~~~~~~
 
 Automatically sync data between {+realm+}s on client devices and your
 MongoDB Atlas data store with :ref:`Sync <android-sync-changes-between-devices>`:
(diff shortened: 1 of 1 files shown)


source/sdk/android.txt
```diff
@@ -40,91 +40,106 @@ Local Realm Database
...
+      .. literalinclude:: /examples/generated/android/local/LandingPageTest.codeblock.query.java
+         :language: java
 
-- :ref:`update <android-modify-an-object>` stored objects by simply
-  updating field values on an instance of the object:
+Update Live Objects
+~~~~~~~~~~~~~~~~~~~
 
-  .. tabs-realm-languages::
+:ref:`Update <android-modify-an-object>` stored objects by simply
```
dacharyc, line 86: Nit: I don't know if we have a policy around this in MongoDB docs, but using "simply" in technical documentation is a habit I've been trying to break. Not everyone has the same level of comfort with things, so saying something is "simple" may be true for some people, but not for others. Folks who struggle with something that is purported to be "simple" may be less likely to try to proceed, and may instead look for another option.

It's also a bit marketing-esque, and can put some people off when encountering assertions like that in documentation. In this case, probably not, but it can be easier to avoid using terms like this than to weigh individually every instance of using it.

dacharyc (approved):
I like it! I think the headers between each code block make the page easier to parse, and I like that it now appears in the "On this page." 

I also like the table vs. the lists of bullets, but the borders in the header row (specifically the lack of borders above the top of the headers) make me wonder if people would think there's a bug or a rendering issue and find that off-putting. Overall, though, I think it's enough of an improvement that we should probably go with the table vs. the lists of links, since we don't have a better visual way of presenting that.

Also, the `MongoDB Realm App Examples` in the header row seems quite long compared to the other headers. Maybe we can lose the `Examples` there?

source/sdk/android.txt
nathan-contino-mongo: Yeah i totally agree here. I was trying to be a bit "marketing-y" and used "simply" to try to make it clearer that writing to the object instance == writing to the database but I think this general wisdom still applies. Will simply rewrite it instead 😉

PR #852, opened 2021-03-04
Update tutorials for new bluehawk markup

cbush:
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/update-tutorials/tutorial/

PR #854, opened 2021-03-04
Automate push to artifact repos for tutorials

cbush:
Yes, this PR cannot be on my fork or else GitHub secrets won't be applied


.github/workflows/push-to-artifact-repos.yml
```diff
@@ -0,0 +1,80 @@
+name: Push to Artifact Repos
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "tutorial/**"
```
nathan-contino-mongo, line 8: Interesting choice -- so we'll publish to all of the tutorial repos every time we push one of them?

tutorial/kotlin-android/app/src/main/java/com/mongodb/tasktracker/LoginActivity.kt
```diff
@@ -69,7 +69,7 @@ class LoginActivity : AppCompatActivity() {
         if (createUser) {
             // register a user using the Realm App we created in the TaskTracker class
             // :code-block-start: create-user
-            // :hide-start:
+            // :state-start: final
```
nathan-contino-mongo, line 72: Thanks for porting this over!

nathan-contino-mongo (approved):
LGTM, though unless I missed it I think we're still missing the README for kotlin in here. Should we add that so we can just update it here?

.github/workflows/push-to-artifact-repos.yml
cbush: Yes, it will run, but if there are no changes then it won't actually commit/push anything.

cbush:
@nathan-contino-mongo I will re-add the README etc. over here: https://github.com/mongodb/docs-realm/pull/852

PR #865, opened 2021-03-05
[Feature] App Structure V2

nlarew:
## Pull Request Info

- Adds a new **Realm App Configuration** section
- Adds a page for each domain/subdirectory in an exported app (e.g. functions, http_endpoints, etc.)

### Issue JIRA link:

**Epic:** https://jira.mongodb.org/browse/DOCSP-14491

- [Realm App Configuration](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/app-structure/config/) **New Section**

### Review Guidelines


source/config/functions.txt
```diff
@@ -0,0 +1,119 @@
...
+
+Every :ref:`function <functions>` in your app has a corresponding metadata entry
+in the function manifest file: ``/functions/config.json``.
+
+.. tip::
+   
+   {+service-short+} automatically adds functions to the manifest on import if
+   they don't already have a configuration defined. If you're okay with the
+   default settings, you can skip defining the configuration for a function and
+   let {+service-short+} do it for you.
```
andrewzlchen, line 29: suggestion: It's true that we do this on the backend side, but it may be worth calling out that it does not modify the manifest file on the user's local file system

source/config/environments.txt
```diff
@@ -0,0 +1,53 @@
+.. _appconfig-environment:
+
+========================================
+Environment Variable Configuration Files
```
andrewzlchen, line 4: question: Are we officially calling these environment variables?

In the web UI, I think we refer to them as environment values to market this as a variant of the two existing "values". 

This may be a product-level decision, but whatever name we choose, I'd like for us to refer to them in a consistent manner.

source/config/functions.txt
```diff
@@ -0,0 +1,119 @@
...
+--------------------
+
+You define a function's source code in a ``.js`` file within the ``/functions``
+directory that uses the function name as its file name. Each file must export
+the main function that runs whenever a request calls the function.
+
+.. important::
+   
+   You can organize source code files by defining arbitrary subdirectories
+   within ``/functions``. However, {+service-short+} ignores file paths so every
```
andrewzlchen, line 108: nit: I wouldn't say that we ignore the file paths since we store them to be used during export.

Instead, I think we should just say that the functions must have a unique filename, and leave it at that

Maybe we can say something along the lines of:

```
Important:

You can organize source code files by defining arbitrary subdirectories within ``/functions``. 

However, Realm uses filename as the function resource identifier, so
 functions must have unique filenames, even if they are in different directories.
```

andrewzlchen (approved):
Hey, great work on this! This is seriously well written. Everything's really clear to me.

I did have a few small nits/questions, but other than that, this looks good to me!

I'll approve so as not to block you, but wanted to call out that there's a discrepancy between the UI where we call Environment variables "Environment Values", but the docs here calls them "Environment Variables".

source/config/environments.txt
```diff
@@ -0,0 +1,59 @@
...
+       ├── no-environment.json
+       ├── development.json
+       ├── testing.json
+       ├── qa.json
+       └── production.json
+
+Environment Configuration
+-------------------------
+
+You can define variable values for each :ref:`environment <environment-values>`
```
andrewzlchen, line 32: I tried clicking this in the staging link, but this link doesn't seem to go anywhere

nlarew: Yup it doesn't point anywhere until we merge https://github.com/mongodb/docs-realm/pull/874

andrewzlchen: Ah okay. Makes sense. My bad then

source/config.txt
```diff
@@ -0,0 +1,80 @@
...
+   app/
+   ├── realm_config.json
+   ├── auth/
+   ├── data_sources/
+   ├── environments/
+   ├── functions/
+   ├── graphql/
+   ├── hosting/
+   ├── http_endpoints/
+   ├── sync/
```
andrewzlchen, line 63: Also, just noticed this, but the services directory that holds the legacy services are still in this version of the app structure.

andrewzlchen: It may go away after the service deprecation project, but for the time being, it still exists, so i think it's still worth documenting

nlarew: Ah okay I can add them back in!

PR #873, opened 2021-03-10
DOP-1883:Update links to Android SDK to /latest

schmalliso:
## Pull Request Info

Turns out the TOC links are also causing redirection woes 😁 

### Issue JIRA link:
https://jira.mongodb.org/browse/DOP-1883

### Staging (built on our QA servers using the build we'll be releasing tomorrow)
https://docs-mongodborg-staging.corp.mongodb.com/realm/docsworker-xlarge/DOP-1883-TOC/

Alternative solution would be to add `/kotlin-extensions/` after `10.3.1` for Kotlin and a trailing slash to the Javadoc link, but this seems like it would save some maintenance burden? Let me know what y'all would prefer.

### DIFF
--- a/source/sdk/android.txt
+++ b/source/sdk/android.txt
@@ -15,8 +15,8 @@ MongoDB Realm Android SDK
    Data Types </sdk/android/data-types>
    Usage Examples </sdk/android/examples>
    Advanced Guides </sdk/android/advanced-guides>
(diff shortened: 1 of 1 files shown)


schmalliso:
Note: because I built using our QA environment, your Autobuilder error check is looking for the log in the wrong DB, which is why it failed.

nlarew (approved):
LGTM

PR #888, opened 2021-03-16
(DOCSP-14644) Add Property Cheatsheet for both Swift and Objective-C

dacharyc:
## Pull Request Info

Couple of notes:
- The property cheat sheet from the Realm legacy doc site lists Float, which is not on our current page listing supported property types, but there are tests for it in the realm-cocoa repo.
- The property cheat sheet page from the Realm legacy docs does _not_ list Decimal128, which _is_ on our current page listing supported property types, so I've attempted to add appropriate declarations, but would love confirmation that I've done it correctly.

The Jira ticket doesn't request the Objective-C property cheatsheet, but since both tabs exist in our current documentation set, I've ported over the Objective-C properties, too.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14644

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14644/sdk/ios/data-types/supported-property-types/

### Review Guidelines

### DIFF
--- a/source/sdk/ios/data-types/supported-property-types.txt
+++ b/source/sdk/ios/data-types/supported-property-types.txt
@@ -113,11 +113,11 @@ Supported Property Types - iOS SDK
            - ``@property RLMObjectId *objectId;``
            - ``@property RLMObjectId *objectId;``
          * - :objc-sdk:`RLMArray <Classes/RLMArray.html>`
-           - ``@property RLMArray<MyArray *><MyArray> *arrayItems;``
-           - ``@property RLMArray<MyArray *><MyArray> *arrayItems;``
+           - ``@property RLMArray<MyClass *><MyClass> *items;``
+           - ``@property RLMArray<MyClass *><MyClass> *items;``
          * - User-defined :objc-sdk:`RLMObject<Classes/RLMObject.html>`
            - N/A
-           - ``@interface MyClass : RLMObject``
+           - ``@property MyClass *value;``
 
       Additionally:
 
--- a/source/sdk/ios/data-types/supported-property-types.txt
+++ b/source/sdk/ios/data-types/supported-property-types.txt
@@ -58,13 +58,12 @@ Supported Property Types - iOS SDK
            - ``let value = List<Type>()``
            - 
          * - User-defined :swift-sdk:`Object <Extensions/Object.html>`
-           - N/A: Must be optional
-           - ``@objc dynamic var value: Class?``
+           - N/A
+           - ``@objc dynamic var value: MyClass?``
       
       Additionally:
 
-      - User-defined :swift-sdk:`Object <Extensions/Object.html>`- and 
-        :swift-sdk:`EmbeddedObject <Extensions/EmbeddedObject.html>`-derived types
+      - :swift-sdk:`EmbeddedObject <Extensions/EmbeddedObject.html>`-derived types
       - :swift-sdk:`Enum <Protocols.html#/s:10RealmSwift0A4EnumP>`
 
       You can use :swift-sdk:`RealmOptional <Classes/RealmOptional.html>` to
@@ -107,15 +106,23 @@ Supported Property Types - iOS SDK
          * - Date
            - ``@property NSDate *value;``
            - ``@property NSDate *value;``
+         * - Decimal128
+           - ``@property RLMDecimal128 *value;``
+           - ``@property RLMDecimal128 *value;``
+         * - :objc-sdk:`RLMObjectId <Classes/RLMObjectId.html>`
+           - ``@property RLMObjectId *objectId;``
+           - ``@property RLMObjectId *objectId;``
+         * - :objc-sdk:`RLMArray <Classes/RLMArray.html>`
+           - ``@property RLMArray<MyArray *><MyArray> *arrayItems;``
+           - ``@property RLMArray<MyArray *><MyArray> *arrayItems;``
+         * - User-defined :objc-sdk:`RLMObject<Classes/RLMObject.html>`
+           - N/A
+           - ``@interface MyClass : RLMObject``
 
       Additionally:
 
       - Integral types ``int``, ``NSInteger``, ``long``, ``long long``
-      - ``RLMDecimal128``
-      - :objc-sdk:`RLMObjectId <Classes/RLMObjectId.html>`
-      - User-defined :objc-sdk:`RLMObject<Classes/RLMObject.html>`- and 
-        :objc-sdk:`RLMEmbeddedObject<Classes/RLMEmbeddedObject.html>`-derived types
-      - :objc-sdk:`RLMArray <Classes/RLMArray.html>`
+      - :objc-sdk:`RLMEmbeddedObject<Classes/RLMEmbeddedObject.html>`-derived types
 
       ``CGFloat`` properties are discouraged, as the type is not
       platform independent.
--- a/source/sdk/ios/data-types/supported-property-types.txt
+++ b/source/sdk/ios/data-types/supported-property-types.txt
@@ -56,11 +56,13 @@ Supported Property Types - iOS SDK
            - ``@objc dynamic var objectIdOpt: ObjectId?``
          * - :swift-sdk:`List <Classes/List.html>`
            - ``let value = List<Type>()``
+           - 
          * - User-defined :swift-sdk:`Object <Extensions/Object.html>`
            - N/A: Must be optional
            - ``@objc dynamic var value: Class?``
       
       Additionally:
+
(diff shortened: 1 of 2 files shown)


source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -20,23 +20,54 @@ Supported Property Types - iOS SDK
...
+         * - :swift-sdk:`List <Classes/List.html>`
+           - ``let value = List<Type>()``
+           - 
+         * - User-defined :swift-sdk:`Object <Extensions/Object.html>`
+           - N/A: Must be optional
+           - ``@objc dynamic var value: Class?``
+      
+      Additionally:
+
       - User-defined :swift-sdk:`Object <Extensions/Object.html>`- and
```
nathan-contino-mongo, line 66: For consistency, I would stick to either including or excluding "user defined Object" in both tables, since it's mentioned below each of them.

nathan-contino-mongo (approved):
LGTM with some small nitpicks. Looks great!

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -48,26 +79,44 @@ Supported Property Types - iOS SDK
...
+      Additionally:
+
       - Integral types ``int``, ``NSInteger``, ``long``, ``long long``
       - ``RLMDecimal128``
-      - ``double``
-      - ``NSString``
-      - ``NSDate``
-      - ``NSData``
       - :objc-sdk:`RLMObjectId <Classes/RLMObjectId.html>`
       - User-defined :objc-sdk:`RLMObject<Classes/RLMObject.html>`- and
```
nathan-contino-mongo, line 116: Boy, could we use monospace-formatted links...

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -48,26 +79,44 @@ Supported Property Types - iOS SDK
...
+      .. list-table::
+         :header-rows: 1
+         :stub-columns: 1
+      
+         * - Type
+           - Required
+           - Optional
+         * - Bool
+           - ``@property BOOL value;``
+           - ``@property NSNumber<RLMBool> *value;``
```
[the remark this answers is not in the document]
cbush: NSNumber is not a supported type in Swift, I guess because it can't be specialized for a specific numeric type like in Obj-C (i.e. `NSNumber<RLMInt>`).

Realm property types must be representable in Obj-C (hence the `@objc dynamic ...` annotation). Realm depends on this to be able to dynamically see which properties a class has so it can be represented in the database. `Int?` (i.e. optional Int) cannot be represented in Obj-C, so you need another way to represent a number as an Objective-C class instance.

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: YES!

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -48,25 +78,51 @@ Supported Property Types - iOS SDK
...
+           - ``@property NSDate *value;``
+           - ``@property NSDate *value;``
+         * - Decimal128
+           - ``@property RLMDecimal128 *value;``
+           - ``@property RLMDecimal128 *value;``
+         * - :objc-sdk:`RLMObjectId <Classes/RLMObjectId.html>`
+           - ``@property RLMObjectId *objectId;``
+           - ``@property RLMObjectId *objectId;``
+         * - :objc-sdk:`RLMArray <Classes/RLMArray.html>`
+           - ``@property RLMArray<MyArray *><MyArray> *arrayItems;``
```
cbush, line 116: Not MyArray - MyClass
```suggestion
           - ``@property RLMArray<MyClass *><MyClass> *arrayItems;``
```

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -48,25 +78,51 @@ Supported Property Types - iOS SDK
...
+           - ``@property RLMArray<MyArray *><MyArray> *arrayItems;``
```
cbush, line 117:
```suggestion
           - ``@property RLMArray<MyClass *><MyClass> *items;``
```

source/sdk/ios/data-types/supported-property-types.txt
cbush, line 116: I think it would be more conventional to drop "array" (even for example):
```suggestion
           - ``@property RLMArray<MyArray *><MyArray> *items;``
```

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -48,25 +78,51 @@ Supported Property Types - iOS SDK
...
+         * - User-defined :objc-sdk:`RLMObject<Classes/RLMObject.html>`
+           - N/A
+           - ``@interface MyClass : RLMObject``
```
cbush, line 120:
```suggestion
           - ``@property MyClass *value;``
```

cbush (changes requested):
See suggestions.

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -48,25 +78,51 @@ Supported Property Types - iOS SDK
...
 
-      - Boolean types ``BOOL``, ``bool``
+      .. list-table::
+         :header-rows: 1
+         :stub-columns: 1
+      
+         * - Type
+           - Required
+           - Optional
+         * - Bool
```
cbush, line 88: This "type" column actually seems redundant on the Obj-C side. Swift used type inference so it's useful to have the type explicitly listed, but the type is always after "@property" in Obj-C. In any case, these are generally using the Swift type names so we might want to use the proper Obj-C name (i.e. `BOOL`, `int`, etc.) or a more generic name (boolean, integer, etc.).

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -20,23 +20,53 @@ Supported Property Types - iOS SDK
...
+           - ``@objc dynamic var value: String? = nil``
+         * - Data
+           - ``@objc dynamic var value = Data()``
+           - ``@objc dynamic var value: Data? = nil``
+         * - Date
+           - ``@objc dynamic var value = Date()``
+           - ``@objc dynamic var value: Date? = nil``
+         * - Decimal128
+           - ``@objc dynamic var decimal: Decimal128 = 0``
+           - ``@objc dynamic var decimalOpt: Decimal128?``
```
cbush, line 53: Minor: I would avoid baking the optionality into the name.
```suggestion
           - ``@objc dynamic var decimal: Decimal128?``
```

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -20,23 +20,53 @@ Supported Property Types - iOS SDK
...
+         * - :swift-sdk:`ObjectId <Classes/ObjectId.html>`
+           - ``@objc dynamic var objectId = ObjectId()``
+           - ``@objc dynamic var objectIdOpt: ObjectId?``
```
cbush, line 56:
```suggestion
           - ``@objc dynamic var objectId: ObjectId?``
```

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -20,23 +20,53 @@ Supported Property Types - iOS SDK
...
+           - ``@objc dynamic var value = Data()``
+           - ``@objc dynamic var value: Data? = nil``
+         * - Date
+           - ``@objc dynamic var value = Date()``
+           - ``@objc dynamic var value: Date? = nil``
+         * - Decimal128
+           - ``@objc dynamic var decimal: Decimal128 = 0``
+           - ``@objc dynamic var decimalOpt: Decimal128?``
+         * - :swift-sdk:`ObjectId <Classes/ObjectId.html>`
+           - ``@objc dynamic var objectId = ObjectId()``
```
cbush, line 55: Perhaps:
```suggestion
           - ``@objc dynamic var objectId = ObjectId.generate()``
```

cbush (approved):
Few more minor comments. Looks good!

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: One nice thing about the "type" column in the Obj-C table is that it gives me a place to link out to the SDK docs for things like `RLMObjectId`, `RLMArray`, and `User-defined RLMObject`. I think I'd prefer to keep it, but your point is valid re: the Swift type names. I'll swap Bool and Int to more generic names.

PR #894, opened 2021-03-17
(DOCSP-14600): Realm lifecycle - android sdk

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14600

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14600/sdk/android/fundamentals/realms/#the-realm-lifecycle (linked section only)

### Review Guidelines

### DIFF
--- a/source/sdk/android/fundamentals/realms.txt
+++ b/source/sdk/android/fundamentals/realms.txt
@@ -99,6 +99,37 @@ all configured schema, migration, and schema versions.
 
    :ref:`Usage Examples: Dynamic Realms <android-dynamic-realms>`
 
+.. _android-realm-lifecycle:
+
+The Realm Lifecycle
+-------------------
+
+Every {+realm+} instance consumes a significant amount of resources.
+Opening and closing a {+realm+} are both expensive operations, but
+keeping a {+realm+} open also incurs significant resource overhead. To
+maximize the performance of your application, you should minimize the
+number of open {+realm+}s at any given time and limit the number of
+open and close operations used.
+
+One way you can follow this advice involves tying the {+realm+} instance
+lifecycle to the lifecycles of the views that observe the {+realm+}. For
+instance, if you embed a `RecyclerView` in a `Fragment`,
+you could open a single {+realm+} that contains the data for that view
+in the ``Fragment.onCreateView()`` lifecycle method of the `Fragment`.
+You could close that same {+realm+} in the ``Fragment.onDestroyView()``
+lifecycle method. And if multiple `Fragment` instances require access to
+the same dataset, you could manage a single {+realm+} in the enclosing
+`Activity` by opening the {+realm+} in ``Activity.onCreate()`` and
+closing the {+realm+} in ``Activity.onDestroy()``.
+
+.. note::
+
+   If your {+realm+} is especially large, fetching a {+realm+} instance
+   in ``Fragment.onCreateView()`` may briefly block rendering. If
+   opening your {+realm+} in ``onCreateView()`` causes performance
+   issues, consider managing the {+realm+} from ``Fragment.onStart()``
+   and ``Fragment.onStop()`` instead.
+
 Comparison with Other Databases
 -------------------------------
 
@@ -166,8 +197,8 @@ in the schema.
 
 .. _android-realm-path:
 
-Realms are Partitions of Atlas Data
------------------------------------
+Synced Realms are Partitions of Atlas Data
+------------------------------------------
 
 Each {+realm+} corresponds to a subset of the data in your
 :term:`{+app+}'s <{+app+}>`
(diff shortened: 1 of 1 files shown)


source/sdk/android/fundamentals/realms.txt
```diff
@@ -99,6 +99,37 @@ all configured schema, migration, and schema versions.
 
    :ref:`Usage Examples: Dynamic Realms <android-dynamic-realms>`
 
+.. _android-realm-lifecycle:
+
+The Realm Lifecycle
+-------------------
+
+Every {+realm+} instance consumes a significant amount of resources.
```
dacharyc, line 107: While this is helpful information, it also gives me pause presented in this way. I wonder if developers who are considering using Realm might rethink that when they see "consumes a significant amount of resources," "expensive operations," and "significant resource overhead." I wonder if there's a way we can present this that is technically accurate and helpful, but also less off-putting?

Maybe something like: "To optimize your application's performance, minimize the number of open {+realm+}s at any given time, and limit the number of open and close operations used." Just sidestep the computationally expensive element entirely?

source/sdk/android/fundamentals/realms.txt
```diff
@@ -99,6 +99,37 @@ all configured schema, migration, and schema versions.
...
+Opening and closing a {+realm+} are both expensive operations, but
+keeping a {+realm+} open also incurs significant resource overhead. To
+maximize the performance of your application, you should minimize the
+number of open {+realm+}s at any given time and limit the number of
+open and close operations used.
+
+One way you can follow this advice involves tying the {+realm+} instance
```
dacharyc, line 114: "Follow this advice" is a bit colloquial/idiomatic. Maybe just "One way to do this is to tie the {+realm+} instance..."

Or even something like: "As a best practice, we recommend..." if we're comfortable making this as a formal recommendation. Lots of hedge words otherwise.

source/sdk/android/fundamentals/realms.txt
```diff
@@ -99,6 +99,37 @@ all configured schema, migration, and schema versions.
...
+lifecycle to the lifecycles of the views that observe the {+realm+}. For
+instance, if you embed a ``RecyclerView`` in a ``Fragment``,
```
dacharyc, line 116: This chunk of text is a bit dense. I wonder if it would work as bullet points for easier scannability? Maybe something like:

>For example:
>- Embed a ``RecyclerView`` in a ``Fragment``
>- Open the {+realm+} that contains the data for that view in the ``Fragment.onCreateView()`` lifecycle method of the ``Fragment``
>- Close the {+realm+} in the ``Fragment.onDestroyView()`` lifecycle method.

Then the "If multiple ``Fragment`` instances..." isn't quite so dense to parse because you've already gotten the context from the easily-scannable bullet points. 

Just a thought!

dacharyc (approved):
I love the lifecycle recommendations - this seems really helpful for developers! I had a couple of suggestions about ways to possibly make it easier to parse, but they're just thoughts - feel free to ignore.

source/sdk/android/fundamentals/realms.txt
nathan-contino-mongo: Taking that second suggestion -- folks often ask us for best practice recommendations, and I think this a reasonable one to give.

source/sdk/android/fundamentals/realms.txt
cmelchior: This information is actually not entirely accurate, or at least not the way it is represented here.

It is true that the Realm database itself is heavyweight, but there are 3 scenarios when opening and closing Realms:

1) Opening the Realm for the first time in the entire application. This is probably as heavyweight as described in this section.

2) Calling `Realm.getInstance()` on the same thread as a Realm that was already opened is very lightweight, equivalent to a HashMap lookup. Calling close is also just as cheap.

3) Calling `Realm.getInstance()` on a thread, with a Realm already open on another thread, is slightly more expensive than 2, but less expensive than 1. This is because we already have validated schemas.

How we manage the resources are done the following way:

When you open the Realm for the first time, we do the memory-mapping required to read/write data from the file. This is why it is expensive to open the first time.

When you "open" a Realm the 2nd time on the same thread, we just increment a thread specific reference count, and decrease it when closing it. When the reference count reaches 0 on that thread, we release the resources for that thread.

When the reference count reaches 0 across all threads, we fully close the Realm and release all resources.

This pattern is why that we recommend calling open/close in all activities/fragments as each is relatively cheap and the Android lifecycle ensure the reference count reaches 0 when no views no longer use Realm.





So minimizing the number of open/close operations is not actually a goal, in most cases, it is actually the desired pattern as it makes it a lot i

nathan-contino-mongo: Addressing this in a separate PR, thank you!

PR #896, opened 2021-03-17
(DOCSP-14624): Intents, Android SDK

nathan-contino-mongo:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14624

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14624/sdk/android/advanced-guides/threading/#intents (just the intents subsection)

### Review Guidelines

### DIFF
--- a/source/sdk/android/advanced-guides/threading.txt
+++ b/source/sdk/android/advanced-guides/threading.txt
@@ -126,7 +126,7 @@ for that identifier.
 
 .. seealso::
 
-   You can find working examples in the :guilabel:`Object Passing`
+   You can find working examples in the `Passing Objects <https://github.com/realm/realm-java/blob/master/examples/threadExample/src/main/java/io/realm/examples/threads/PassingObjectsFragment.java>`__
    portion of the `Android SDK Threading Example
    <https://github.com/realm/realm-java/tree/master/examples/threadExample>`__.
    The example shows you how to pass IDs and retrieve a RealmObject in
--- a/source/sdk/android/advanced-guides/threading.txt
+++ b/source/sdk/android/advanced-guides/threading.txt
@@ -97,7 +97,7 @@ more space-efficient, and more performant as a result.
 When you need to communicate across threads, you have
 several options depending on your use case:
 
-- To work modify the data on two threads, :ref:`query <android-client-query-engine>`
+- To modify the data on two threads, :ref:`query <android-client-query-engine>`
   for the object on both threads.
 
 - To react to changes made on any thread, use {+client-database+}'s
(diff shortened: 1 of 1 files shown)


source/sdk/android/advanced-guides/threading.txt
```diff
@@ -112,6 +112,22 @@ several options depending on your use case:
...
+~~~~~~~
+
+``RealmObject`` instances are not thread-safe or ``Parcelable``, so you
+cannot pass them between activities or threads via an ``Intent``. Instead,
+you can pass an object identifier, like a :ref:`primary key
+<android-primary-key-fundamentals>`, in the ``Intent`` extras bundle,
+and then open a new {+realm+} instance in the separate thread to query
+for that identifier.
+
+.. seealso::
```
dacharyc, line 127: Nit: I am a little bummed that this `seealso` doesn't give me context for what I'm looking at when I click the link. I appreciate that the legacy page had this text to orient the user to what they could expect to find in that repo:

>You can find working examples in the `Object Passing` portion of the threading example. The example shows you how to pass id’s and retrieve the `RealmObject` in common Android use cases.

A `seealso` can't contain text, can it? I wonder if there's another way we can present this link so users don't have the cognitive load of figuring out what they should expect to see when they arrive at a repo with no details.

dacharyc (approved):
Minor nit about the link, but otherwise LGTM!

source/sdk/android/advanced-guides/threading.txt
nathan-contino-mongo: It can! I'll incorporate this text.

source/sdk/android/advanced-guides/threading.txt
```diff
@@ -112,6 +112,26 @@ several options depending on your use case:
 - To keep and share many read-only views of the object in your app, copy
   the object from the {+realm+}.
 
+.. _android-intents:
+
+Intents
+~~~~~~~
+
+``RealmObject`` instances are not thread-safe or ``Parcelable``, so you
```
cmelchior, line 120: Generally, this is true, and I'm not sure if it is worth polluting this section, but there is a difference between managed and unmanaged objects.

Managed objects are the ones that read data from the database and generally what we try to describe in the docs, and it is true these are not parcelable.

Unmanaged objects, i.e. objects you create before adding them to Realm or the ones you get back when calling `Realm.copyFromRealm(...)`. These are parcelable. Of course the problem is that both types use the same class definition, so using these correctly would be entirely up to the user.

nathan-contino-mongo: Addressing this in a separate PR, thanks!

PR #898, opened 2021-03-18
(DOCSP-14645): Incorporate Swift SDK Troubleshooting info from legacy docs

dacharyc:
## Pull Request Info

The Jira ticket links to this Swift SDK troubleshooting section from legacy docs: https://docs.mongodb.com/realm-legacy/docs/swift/latest/index.html#troubleshooting

After doing some digging, I think the **Xcode 12/Arm64** instructions detailed there are no longer relevant, and I don't see any questions about **Realm Core binary fails to download** in several years. Aside from some of the general things I've included here, I think the rest of the content is not relevant to port over.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14645

New troubleshooting page; so far, the only content is about resolving build issues, but I think we can expand it as we work on the debugging guide content: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14645/sdk/ios/troubleshooting/

New Tip at the top of this page: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14645/sdk/ios/examples/open-and-close-a-local-realm/#std-label-ios-provide-a-subset-of-classes-to-a-realm

### Review Guidelines


source/sdk/ios/troubleshooting.txt
```diff
@@ -0,0 +1,49 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. _ios-resolve-build-issues:
```
nlarew, line 13: Should we link to this from `/sdk/ios/install`?

source/sdk/ios/examples/open-and-close-a-local-realm.txt
```diff
@@ -4,6 +4,13 @@
 Open & Close a Local Realm - iOS SDK
 ====================================
 
+.. tip:: Operating with Low Memory Constraints
+   
+   Are you developing watchOS apps or App Extensions? Avoid a costly 
+   call to ``objc_copyClassList()``. Instead, use a ``Realm.Configuration``
+   object to :ref:`provide a subset of classes to the Realm 
+   <ios-provide-a-subset-of-classes-to-a-realm>`.
```
nlarew, line 12: nit: Perhaps this makes sense to a practicing iOS developer, but I'm not really sure _why_ I'd use `objc_copyClassList()` in the first place - we don't mention it anywhere else on the page.

nlarew (approved):
LGTM w/ some comments

source/sdk/ios/troubleshooting.txt
dacharyc: Yes, good point! Added. Thanks!

source/sdk/ios/examples/open-and-close-a-local-realm.txt
dacharyc: That is a good question. I inferred from the context on the original page: https://docs.mongodb.com/realm-legacy/docs/swift/latest/index.html#troubleshooting

That this is the behavior that occurs when you don't explicitly pass a subset of classes when opening a Realm. Maybe it would be cleaner to rewrite it in a way that doesn't presuppose that knowledge, and why it's bad? Something like:

> Are you developing watchOS apps or App Extensions? Use a `Realm.Configuration` object to optimize for low-memory environments. Explicitly specify a subset of classes when opening a Realm.

source/includes/steps-troubleshoot-carthage-install.yaml
```diff
@@ -0,0 +1,29 @@
+title: Reset Carthage-managed Dependency State
+ref: remove-carthage-managed-dependencies
+content: |
+
+   Run these commands in the terminal, in the root of your project:
+   
+   .. code-block:: bash
+      
+      rm -rf Carthage
+      rm -rf ~/Library/Developer/Xcode/DerivedData
```
jsflax, line 10: This directory is the default DerivedData directory, but may or may not be true for everyone if they've set it to a custom location. Not sure how you want to handle this case, but I'm just pointing it out.

source/includes/steps-troubleshoot-cocoapods-install.yaml
```diff
@@ -0,0 +1,26 @@
...
+content: |
+
+   Run these commands in the terminal, in the root of your project:
+   
+   .. code-block:: bash
+      
+      pod cache clean Realm
+      pod cache clean RealmSwift
+      pod deintegrate || rm -rf Pods
+      pod install --verbose
```
jsflax, line 12: Would want to consider also adding `--repo-update`

dacharyc: Thanks, @jsflax - added!

source/includes/steps-troubleshoot-carthage-install.yaml
dacharyc: Added a caveat about this. Thanks!

source/sdk/ios/examples/open-and-close-a-local-realm.txt
```diff
@@ -6,10 +6,9 @@ Open & Close a Local Realm - iOS SDK
 
 .. tip:: Operating with Low Memory Constraints
    
-   Are you developing watchOS apps or App Extensions? Avoid a costly 
-   call to ``objc_copyClassList()``. Instead, use a ``Realm.Configuration``
-   object to :ref:`provide a subset of classes to the Realm 
-   <ios-provide-a-subset-of-classes-to-a-realm>`.
+   Are you developing watchOS apps or App Extensions? Use a 
+   ``Realm.Configuration`` object to optimize for low-memory environments.
+   :ref:`Provide a subset of classes to the Realm <ios-provide-a-subset-of-classes-to-a-realm>`.
```
nlarew, line 11: nit: technically one _always_ uses a `Realm.Configuration` when they open a realm, so just saying "Use it in low-memory environments" feels a bit redundant. It's more about _how_ you use it, i.e. by providing a subset of classes."

Thoughts on something like the following?
```rst
.. tip:: Memory-Constrained Apps
   
   Some applications, such as watchOS apps and iOS app extensions, have tight
   constraints on their memory footprint. To optimize your data model for
   low-memory environments like these, you can :ref:`provide a subset of classes
   <ios-provide-a-subset-of-classes-to-a-realm>` when you open the realm.

```

PR #913, opened 2021-03-23
(DOCSP-14982): New Swift SDK landing page

dacharyc:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14982

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14982/sdk/ios/

Build log is complaining about errors resolved in #914 , so please disregard build log failure. Log for reference: https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=605a10ea2723a9e448e1ffcd

### Review Guidelines

### DIFF
--- a/examples/ios/Examples/LandingPageCodeExamples.swift
+++ b/examples/ios/Examples/LandingPageCodeExamples.swift
@@ -1,0 +1,192 @@
+// :replace-start: {
+//   "terms": {
+//     "LandingPageExamples_": "",
+//     "landingPageExamples_": ""
+//   }
+// }
+
+import XCTest
+import RealmSwift
+
+// MARK: Define an Object Schema
+
+// :code-block-start: coffee-drink-object
+class LandingPageExamples_CoffeeDrink: Object {
+    @objc dynamic var name = ""
+    @objc dynamic var hotOrCold: String?
+    @objc dynamic var rating = 0
+}
+// :code-block-end:
+
+class LandingPageExamples_CoffeeShop: Object {
+    // Required string property
+    @objc dynamic var name = ""
+    let drinks = List<LandingPageExamples_CoffeeDrink>()
+}
+
+// MARK: Query Realm Database
+
+class LandingPageExamples: XCTestCase {
+
+    override func tearDown() {
+        let realm = try! Realm()
+        try! realm.write {
+            realm.deleteAll()
+        }
+    }
+
+    func testQueryCoffeeRatings() {
+
+        // :code-block-start: query
+        let realm = try! Realm()
+        // :hide-start:
+        try! realm.write {
+            // Add coffee shop and drink info here.
+            let shop = LandingPageExamples_CoffeeShop()
+            shop.name = "Better Coffee"
+            let drink = LandingPageExamples_CoffeeDrink()
+            drink.name = "Maple Latte"
+            drink.rating = 7
+            shop.drinks.append(drink)
+            realm.add(shop)
+            // ...
+        }
+        // :hide-end:
+
+        let drinks = realm.objects(LandingPageExamples_CoffeeDrink.self)
+
+        let wellRatedDrinks = drinks.filter("rating > 6")
+        print("Well-rated drinks: \(wellRatedDrinks.count)")
+
+        let mapleOrCaramelLattes = drinks.filter("name IN {'Maple Latte', 'Caramel Latte'}")
+        print("Number of maple or caramel lattes: \(mapleOrCaramelLattes.count)")
+
+        let drinkTempNotSpecified = drinks.filter("hotOrCold == nil")
+        print("No info about drink temp: \(drinkTempNotSpecified.count)")
+
+        // :code-block-end:
+    }
+
+    // MARK: Update Live Objects
+
+    func testUpdateMapleLatte() {
+        // :code-block-start: update-live-objects
+        let realm = try! Realm()
+
+        // :hide-start:
+        try! realm.write {
+            // Add coffee shop and drink info here.
+            let shop = LandingPageExamples_CoffeeShop()
+            shop.name = "Better Coffee"
+            let drink = LandingPageExamples_CoffeeDrink()
+            drink.name = "Maple Latte"
+            drink.rating = 7
+            shop.drinks.append(drink)
+            realm.add(shop)
+            // ...
+        }
+        // :hide-end:
+
+        // Get a maple latte
+        let mapleLatte = realm.objects(LandingPageExamples_CoffeeDrink.self).filter("name == 'Maple Latte'").first!
+
+        // Open a thread-safe transaction
+        try! realm.write {
+            // Change the name of the maple latte.
+            mapleLatte.name = "Maple Delight"
+            // Specify that the maple latte is a hot drink.
+            mapleLatte.hotOrCold = "Hot"
+        } // When the transaction completes, the drink details are updated in the database.
+        // :code-block-end:
+    }
+
+    // MARK: Watch for Object Updates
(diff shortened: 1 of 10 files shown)


examples/ios/Examples/LandingPageCodeExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+    // MARK: Watch for Object Updates
+    // Note: This example was adapted from "Notifications.swift"
+    // and is not really tested, but makes a convenient full-program
+    // example.
+
+    // :code-block-start: watch-for-object-updates
+
+    var drinkNotificationToken: NotificationToken?
+
+    func landingPageExamples_objectNotificationExample() {
```
cbush, line 112: It's not really required to namespace this function ("landingPageExamples_...") since it's already [scoped](https://docs.swift.org/swift-book/ReferenceManual/Declarations.html) by the class. We only use the prefix to prevent global namespace pollution as in the case of Realm object classes, which much be declared at the module (top) level and therefore can be seen by any Swift file in this module.

examples/ios/Examples/LandingPageCodeExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    // MARK: Update Live Objects
+
+    func testUpdateMapleLatte() {
+        // :code-block-start: update-live-objects
+        let realm = try! Realm()
+
+        // :hide-start:
+        try! realm.write {
```
cbush, line 77: Why not include this in the snippet? Makes it a bit longer, but then someone can easily copy and paste. Or add it to a "write" example earlier on.

examples/ios/Examples/LandingPageCodeExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+        XCTAssert(drinkA.name == drinkB.name)
+        // Update drink B's rating
+        try! realm.write {
+            drinkB.rating = 4
+        }
+        // See that drink A instance updates with the new rating
+        XCTAssert(drinkB.rating == drinkA.rating)
+
+        // :hide-start:
+        expectation.fulfill()
```
cbush, line 187: I don't think you need to use expectations here at all, since all of the code in this function is synchronous, and you're not waiting for anything.

source/sdk/ios.txt
```diff
@@ -25,3 +19,190 @@ both Swift and Objective-C apps.
...
    Release Notes <https://github.com/realm/realm-cocoa/releases>
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+The {+service+} Swift SDK enables mobile applications to access data
+stored in local {+realms+} and interact with {+backend+} services like
```
cbush, line 30: It might be worth finding some way to emphasize that all MongoDB Realm/Atlas/App Services stuff is optional.
```suggestion
stored in local {+realms+} and optionally interact with {+backend+} services like
```

source/sdk/ios.txt
```diff
@@ -25,3 +19,190 @@ both Swift and Objective-C apps.
...
+
+Define an Object Schema
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Use Swift to idiomatically :ref:`define an object schema <ios-define-a-realm-object-schema>`.
+
+.. literalinclude:: /examples/generated/code/start/LandingPageCodeExamples.codeblock.coffee-drink-object.swift
+   :language: swift
+
+Query Realm Database
```
cbush, line 47: This page seems to be missing the important "write an object".

examples/ios/Examples/LandingPageCodeExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+            shop.name = "Better Coffee"
+            let drink = LandingPageExamples_CoffeeDrink()
+            drink.name = "Maple Latte"
+            drink.rating = 7
+            shop.drinks.append(drink)
+            realm.add(shop)
+            // ...
+        }
+        // :hide-end:
+
```
cbush, line 89: Deleting this (blank) line prevents double blank lines in the bluehawked output. (Unless you keep the above snippet).
```suggestion
```

examples/ios/Examples/LandingPageCodeExamples.swift
```diff
@@ -0,0 +1,192 @@
...
+        let drinks = realm.objects(LandingPageExamples_CoffeeDrink.self)
+
+        let wellRatedDrinks = drinks.filter("rating > 6")
+        print("Well-rated drinks: \(wellRatedDrinks.count)")
```
cbush, line 59: Perhaps "Highly rated"?

cbush (approved):
Looks good! Just a few comments for your consideration.

examples/ios/Examples/LandingPageCodeExamples.swift
dacharyc: That's helpful - thank you!

source/sdk/ios.txt
dacharyc: Huh, good point! Added it - thanks!

source/sdk/ios.txt
dacharyc: I rewrote this intro section a little bit to emphasize that. Thanks!

examples/ios/Examples/LandingPageCodeExamples.swift
dacharyc: I wanted to keep the snippets short and easily scannable for folks who just wanted a quick overview. But I did add this as a write example earlier on, and refactored the tests a little bit. Would you please take another look and make sure everything looks good, @cbush ?

cbush (approved):
Looks great!

examples/ios/Examples/LandingPageCodeExamples.swift
```diff
@@ -0,0 +1,176 @@
...
+            shop.drinks.append(drink)
+            realm.add(shop)
+        }
+        // :code-block-end:
+    }
+
+    func testQueryCoffeeRatings() {
+
+        // :code-block-start: query
+        setUp()
```
cbush, line 72: Wait, this changed.

setUp() is automatically run before each test. So you shouldn't be calling it twice. Also, I think this might be stretching the Bluehawk replace functionality since this inserts a big chunk of code rather than renaming something.

cbush (changes requested):

examples/ios/Examples/LandingPageCodeExamples.swift
```diff
@@ -0,0 +1,176 @@
...
+class LandingPageExamples: XCTestCase {
+
+    override func tearDown() {
+        let realm = try! Realm()
+        try! realm.write {
+            realm.deleteAll()
+        }
+    }
+
+    let realm = try! Realm()
```
cbush, line 38: I would just put this in every test case where you want to use it. It's ok to open it again and again in the examples. (It will automatically close when it goes out of scope, i.e. at the end of each test function.)

examples/ios/Examples/LandingPageCodeExamples.swift
dacharyc: Alrighty, thanks! I've removed the manual call to setUp() and have removed the Bluehawk replace for that.

examples/ios/Examples/LandingPageCodeExamples.swift
dacharyc: Got it, thanks! Added it where I'd like to illustrate that, and it didn't appear to cause any issues - tests are still passing - so I think these should be all set now.

cbush (approved):

PR #918, opened 2021-03-23
(DOCSP-14635): Incorporate info on encrypted Realms and multiple processes

dacharyc:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14635

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14635/sdk/ios/advanced-guides/encrypt-a-realm/

### Review Guidelines

### DIFF
--- a/source/sdk/ios/advanced-guides/encrypt-a-realm.txt
+++ b/source/sdk/ios/advanced-guides/encrypt-a-realm.txt
@@ -58,8 +58,8 @@ unencrypted {+realm+}. Encrypted {+realms+} cannot be accessed by
 multiple processes at the same time. 
 
 Store data that you want to encrypt as ``NSData`` properties on {+realm+} 
-objects. Use the Security and CommonCrypto system frameworks to encrypt 
-and decrypt the data.
+objects. Use the :apple:`Security and Common Crypto system frameworks
+<security>` to encrypt and decrypt the data.
 
 Example
 -------
--- a/source/sdk/ios/advanced-guides/encrypt-a-realm.txt
+++ b/source/sdk/ios/advanced-guides/encrypt-a-realm.txt
@@ -63,7 +63,7 @@ decrypt individual fields.
 
 One possible tool to encrypt and decrypt fields is :apple:`Apple's 
 CryptoKit framework <documentation/cryptokit>`. You can use :github:`Swift 
-Crypto <apple/swift-crypto>`, to simplify app 
+Crypto <apple/swift-crypto>` to simplify app 
 development with CryptoKit.
 
 Example
--- a/source/sdk/ios/advanced-guides/encrypt-a-realm.txt
+++ b/source/sdk/ios/advanced-guides/encrypt-a-realm.txt
@@ -63,7 +63,7 @@ decrypt individual fields.
 
 One possible tool to encrypt and decrypt fields is :apple:`Apple's 
 CryptoKit framework <documentation/cryptokit>`. You can use `Swift 
-Crypto<https://github.com/apple/swift-crypto>`__ to simplify app 
+Crypto<https://github.com/apple/swift-crypto>`__, to simplify app 
 development with CryptoKit.
 
 Example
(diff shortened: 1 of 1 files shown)


source/sdk/ios/advanced-guides/encrypt-a-realm.txt
```diff
@@ -50,6 +50,17 @@ You can encrypt a :ref:`synced realm
 on the device and stores the data unencrypted in your {+atlas+}
 :term:`data source`.
 
+Multiple Processes
+~~~~~~~~~~~~~~~~~~
+
+If multiple processes need to access a {+realm+} simultaneously, use an 
+unencrypted {+realm+}. Encrypted {+realms+} cannot be accessed by
```
mongomoe, line 57: Suggestion:

Let's change that second sentence to avoid the passive voice. An alternative would be: "Multiple processes cannot access encrypted realms at the same time."

source/sdk/ios/advanced-guides/encrypt-a-realm.txt
```diff
@@ -50,6 +50,17 @@ You can encrypt a :ref:`synced realm
...
 
+Multiple Processes
+~~~~~~~~~~~~~~~~~~
+
+If multiple processes need to access a {+realm+} simultaneously, use an 
+unencrypted {+realm+}. Encrypted {+realms+} cannot be accessed by 
+multiple processes at the same time. 
+
+Store data that you want to encrypt as ``NSData`` properties on {+realm+} 
+objects. Use the Security and CommonCrypto system frameworks to encrypt
```
mongomoe, line 61: Question: Are there external links to the frameworks 'Security' and 'CommonCrypto'? 

Suggestion: If so, can we link to those, as a non-frequent swift developer I wasn't aware these existed.

mongomoe (approved):
LGTM 👍  with some suggestions

source/sdk/ios/advanced-guides/encrypt-a-realm.txt
dacharyc: Thanks for spotting that, @mongomoe ! In general, I'm in favor of avoiding passive voice. However, in this instance, I've intentionally structured this sentence to lead with "Encrypted {+realms+}..." because that seems like the important key to communicate vs. the prior sentence. I think it's worth breaking the rules here to communicate the key point more effectively to the readers.

source/sdk/ios/advanced-guides/encrypt-a-realm.txt
dacharyc: Fair point! I added a link to the Security page, which gives an overview of the Security framework and mentions the CommonCrypto library. Unfortunately, that seems to be the only place where CommonCrypto is mentioned, and a different Apple documentation page mentions "preferring CryptoKit over lower-level interfaces": https://developer.apple.com/documentation/cryptokit

It looks to me like CommonCrypto is an older library that may not be preferred these days. Perhaps a technical reviewer can suggest whether we'd still recommend it here?

ericjordanmossman: I would go with [SwiftCrypto](https://swift.org/blog/crypto/) and cryptoKit. SwiftCrypto is the swift package of CryptoKit.

I would also make it clear that these libraries are a _possible_ way to encrypt the necessary fields and not the _only_ way (maybe a development team has a third party library they prefer).
The important part is that since the database as a whole can't be encrypted for a multiprocess case, the fields will have to be.

ericjordanmossman:
The two main technical points check out:
1) Realm-cocoa currently doesn't support multiprocess on encrypted realms.
2) Encrypting necessary fields is best alternative.

See comment on library link.

source/sdk/ios/advanced-guides/encrypt-a-realm.txt
dacharyc: Awesome, thanks! Adding those to this section.

PR #921, opened 2021-03-24
(DOCSP-14654): Port interface-driven write text

cbush:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14654

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/carb3/sdk/ios/fundamentals/write-transactions/#interface-driven-writes

### Review Guidelines

### DIFF
--- a/source/sdk/ios/fundamentals/write-transactions.txt
+++ b/source/sdk/ios/fundamentals/write-transactions.txt
@@ -110,3 +110,45 @@ the transaction immediately after the callback.
 .. seealso::
 
    :ref:`Usage Example: Read and Write Data <ios-read-and-write-data>`
+
+.. _ios-interface-driven-writes:
+.. _ios-silent-writes:
+
+Interface-Driven Writes
+-----------------------
+
+{+service-short+} always delivers notifications asynchronously, so they
+never block the UI thread. However, there are situations when the UI
+must reflect changes instantly. If you update the UI directly at the
+same time as the write, the eventual notification could double that
+update. This could lead to your app crashing due to inconsistent state
+between the UI and the backing data store.
+
+Fortunately, you can write without sending a notification to a specific
+handler. We call this type of transaction an **interface-driven write**.
+
+.. example::
+
+   Say we decide to manage a table view's data source manually, because
+   our app design requires an instantaneous response to UI-driven table
+   updates. As soon as a user adds an item to the table view, we insert
+   it to our data source, which writes to the {+realm+} but also
+   immediately kicks off the animation. However, when {+service-short+}
+   delivers the change notification for this insertion a little later,
+   it indicates that an object has been added. But we already updated
+   the table view for it! Do we need to write some code to distinguish
+   UI-driven table updates from backend-driven table updates? No, we can
+   use interface-driven writes to prevent a specific notification
+   handler from firing for that specific write.
+
+Interface-driven writes, also known as silent writes, are especially
+useful when using fine-grained collection notifications with a
+:ref:`synchronized {+realm+} <sync>`. Many of the workarounds to account
+for interface-driven writes relied on controlling the full state of when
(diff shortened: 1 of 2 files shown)


source/sdk/ios/fundamentals/write-transactions.txt
```diff
@@ -110,3 +110,45 @@ the transaction immediately after the callback.
...
+
+{+service-short+} always delivers notifications asynchronously, so they
+never block the UI thread. However, there are situations when the UI
+must reflect changes instantly. If you update the UI directly at the
+same time as the write, the eventual notification could double that
+update. This could lead to your app crashing due to inconsistent state
+between the UI and the backing data store.
+
+Fortunately, you can write without sending a notification to a specific
+handler. We call this type of transaction an **interface-driven write**.
```
nlarew, line 128: nit: IMO this paragraph lacks some "oomph" in that it may not be immediately obvious why this is a "fortunate" feature to have.

Maybe we could add a sentence? Goal is to make it just a bit clearer that this means "you can update your UI optimistically without dealing with the double update".

source/sdk/ios/fundamentals/write-transactions.txt
```diff
@@ -110,3 +110,45 @@ the transaction immediately after the callback.
...
+   our app design requires an instantaneous response to UI-driven table
+   updates. As soon as a user adds an item to the table view, we insert
+   it to our data source, which writes to the {+realm+} but also
+   immediately kicks off the animation. However, when {+service-short+}
+   delivers the change notification for this insertion a little later,
+   it indicates that an object has been added. But we already updated
+   the table view for it! Do we need to write some code to distinguish
+   UI-driven table updates from backend-driven table updates? No, we can
+   use interface-driven writes to prevent a specific notification
+   handler from firing for that specific write.
```
nlarew, line 142: nit: This example could maybe do with some small code samples / paragraphs to break up the text. It feels a bit dense and maybe too socratic IMO. Thoughts?

source/sdk/ios/fundamentals/write-transactions.txt
```diff
@@ -110,3 +110,45 @@ the transaction immediately after the callback.
...
+
+Interface-driven writes, also known as silent writes, are especially
+useful when using fine-grained collection notifications with a
+:ref:`synchronized {+realm+} <sync>`. Many of the workarounds to account
+for interface-driven writes relied on controlling the full state of when
+the app can perform changes. With synchronized {+realms+}, changes apply
+when they’re synchronized, which can happen at any point in the app’s
+lifetime.
```
nlarew, line 150: A bit confused by this paragraph. Which workarounds?

Also I'm uncertain why interface-driven writes would be more useful for sync. If I understand correctly, the goal is to prevent *my* local updates (that I already sent to the UI) from firing notifications. I'd use them whether I was using sync or not.

Wouldn't synced writes from other devices always need to update my UI through a notification because they're not *my* local updates?

nlarew (approved):
Looks good overall, a few nits and questions

source/sdk/ios/fundamentals/write-transactions.txt
cbush: How about:

>While you use interface-driven writes for the current user's updates and update the UI immediately, the sync process can use standard notifications to update the UI.

source/sdk/ios/fundamentals/write-transactions.txt
cbush: Reworked a bit and removed question(able) tone.

source/sdk/ios/fundamentals/write-transactions.txt
cbush: Reworked a bit, let me know what you think!

PR #923, opened 2021-03-25
(DOCSP-14653): Add key-value coding to docs

dacharyc:
## Pull Request Info

Porting the KVC example from Realm.io legacy docs.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14653

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14653/sdk/ios/examples/read-and-write-data/#update-properties-with-key-value-coding

### Review Guidelines

### DIFF
--- a/examples/ios/Examples/ReadWriteData.swift
+++ b/examples/ios/Examples/ReadWriteData.swift
@@ -435,7 +435,7 @@ class ReadWriteData: XCTestCase {
 
         try! realm.write {
             allDogs.first?.setValue("Sparky", forKey: "name")
-            // An international dognapping ring moves all the dogs to Toronto
+            // Move the dogs to Toronto for vacation
             allDogs.setValue("Toronto", forKey: "currentCity")
         }
         // :code-block-end:
--- a/examples/ios/Examples/ReadWriteData.swift
+++ b/examples/ios/Examples/ReadWriteData.swift
@@ -14,6 +14,7 @@ class ReadWriteDataExamples_DogToy: Object {
 class ReadWriteDataExamples_Dog: Object {
     @objc dynamic var name = ""
     @objc dynamic var age = 0
+    @objc dynamic var currentCity = ""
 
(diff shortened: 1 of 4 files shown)


cbush (approved):
Looks good. My only question is whether it would be better as two higher-level headings ("Modify an Object", "Update Properties with Key-Value Coding") so they appear in the "On This Page" box. Up to you.

examples/ios/Examples/ReadWriteData.swift
```diff
@@ -398,6 +398,19 @@ class ReadWriteData: XCTestCase {
...
+
+    func testKeyValueCoding() {
+        // :code-block-start: key-value-coding
+        let realm = try! Realm()
+
+        let allDogs = realm.objects(ReadWriteDataExamples_Dog.self)
+
+        try! realm.write {
+            allDogs.first?.setValue("Sparky", forKey: "name")
+            allDogs.first?.setValue(3, forKey: "age")
```
ericjordanmossman, line 410: Not required, just a thought:
Maybe consider dropping the `first?` from one the lines to demonstrate the "_Avoid the overhead of iterating over a collection while creating accessors for every item._" capability.

```
try! realm.write {
            allDogs.first?.setValue("Sparky", forKey: "name")
            // set each dogs' planet to Earth
            allDogs.setValue("Earth", forKey: "planet")
}
```

The only problem is I couldn't think of any Dog example that all dogs would have in common - other than maybe planet (which I just saw in the old docs).

ericjordanmossman (approved):
LGTM, I added one comment but not required change.

examples/ios/Examples/ReadWriteData.swift
dacharyc: Good note. I've updated the example to add a `currentCity`, and then update it for all dogs. Thanks for the feedback!

PR #926, opened 2021-03-25
(DOCSP-14649): Port "Initializing Swift properties using Realm APIs"

cbush:
- Not sure where this should go. Added "Considerations" to Realms fundamentals page.
- Theoretical Swift example doesn't need to be Bluehawked?

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14649

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/carb4/sdk/ios/examples/configure-and-open-a-realm/

### Review Guidelines

### DIFF
--- a/source/sdk/ios/examples/configure-and-open-a-realm.txt
+++ b/source/sdk/ios/examples/configure-and-open-a-realm.txt
@@ -171,8 +171,52 @@ instance of ``SomeSwiftType`` before
 ``applicationDidFinishLaunching()``, you might be accessing your
 {+realm+} before it has been correctly configured.
 
-To avoid such issues, consider doing one of the following options:
+To avoid such issues, consider doing one of the following:
 
 - Defer instantiation of any type that eagerly initializes properties using {+client-database+} APIs until after your app has completed setting up its {+realm+} configurations. 
 - Define your properties using Swift’s ``lazy`` keyword. This allows you to safely instantiate such types at any time during your application’s lifecycle, as long as you do not attempt to access your ``lazy`` properties until after your app has set up its {+realm+} configurations.
 - Only initialize your properties using Realm APIs that explicitly take in user-defined configurations. You can be sure that the configuration values you are using have been set up properly before they are used to open {+realms+}.
+
+.. _use-realm-when-the-device-is-locked:
+
+Use Realm When the Device is Locked
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, iOS 8 and above encrypts app files using
+``NSFileProtection`` whenever the device is locked. If your app attempts
+to do any work on a {+realm+} while the device is locked, you might see
+the following error:
+
+.. code-block:: text
+
+   open() failed: Operation not permitted
+
+To handle this, you can downgrade the file protection of the folder[1]_
+containing :ref:`the Realm file itself and its auxiliary files
+<ios-realm-files>` to a less strict protection level that allows file
+access even when the device is locked, such as
+:apple:`NSFileProtectionCompleteUntilFirstUserAuthentication
+</documentation/foundation/nsfileprotectioncompleteuntilfirstuserauthentication>`.
+
+.. [1] It is recommended that you apply the file protection attribute to
+   the folder containing the Realm files rather than to the specific
+   files themselves. {+service-short+} may create and deletes auxiliary
+   files at any time. By applying the file protection attribute to the
+   parent folder, you can ensure the file protection applies to all of
+   the relevant files, regardless of when they were created.
+
+.. tip::
+
+   If you choose to opt out of complete iOS file encryption in this way,
+   we encourage you to use Realm's own :ref:`built-in encryption
+   <ios-encrypt-a-realm>` to secure your data.
+
+The following example shows how to apply a less strict protection level
+to the parent directory of the default {+realm+}:
+
+.. literalinclude:: /examples/generated/code/start/Encrypt.codeblock.background-app-refresh-encryption.swift
+   :language: swift
+
+.. seealso::
+   
+   :ref:`sync-changes-in-the-background`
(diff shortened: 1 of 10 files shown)


source/sdk/ios/fundamentals/realms.txt
```diff
@@ -202,3 +202,37 @@ receieved out of order.
    
    For code examples that show how to work with a synced realm in the
    iOS SDK, see :ref:`ios-sync-changes-between-devices`.
+
+.. _ios-realm-considerations:
+
+Considerations
```
dacharyc, line 208: I'm not thrilled with "Considerations" as a header or a section. That sounds like it wants to be an admonition type. It also sounds like a laundry list of caveats, which I'd prefer to pull out into a more meaningful form.

FWIW, I'm working on #929 which contains a section about Property Attributes. Maybe there should just be a section for Properties, and both of these things would go under that - maybe under Fundamentals?

dacharyc (approved):
I think it's fine not to Bluehawk this theoretical example, and the content itself is fine, but I'm not sure this is the best place to present it. I suggested an alternative - curious about your thoughts.

source/sdk/ios/fundamentals/realms.txt
cbush: I agree with your feedback. I don't think this belongs with Properties though. Even though this is about "properties", it's not about Realm properties per se. It's more about timing of Configuration. And I wouldn't necessarily want to create a new page for property attributes when we already have a page for object models.

dacharyc: Fair. Since it's about the timing of Configuration, maybe we should have a section about that. The other places we talk about Configuration are in the "Open and Close a Local Realm" page and the "Sync Changes Between Devices" - in both places, we link away to SDK docs for more details about Configuration. Maybe we should just add a section for it to our docs somewhere, and this would really belong to that?

cbush: That sounds good. Would a Configuration page be a Usage Example or a Fundamental?

cbush: Or maybe Configuration as a section under Fundamentals > Realms?

dacharyc: I wonder if we should actually have a page about Usage Examples -> Working with Realms, where Configuration and/or Opening a Realm would be one (or more) section(s). I'm working on a PR about Compacting Realms, which might also be relevant as a section on that page. I notice there's also a section about Deleting Realm files in the legacy docs that would also be relevant there, although I don't know if that ticket has come up yet.

It's tricky because these are both fundamentals _and_ contain usage examples, so that divide is proving to be problematic.

source/sdk/ios/examples/configure-and-open-a-realm.txt
```diff
@@ -142,3 +146,77 @@ Provide a Subset of Classes to a Realm
...
+
+.. code-block:: text
+
+   open() failed: Operation not permitted
+
+To handle this, you can downgrade the file protection of the folder
+containing :ref:`the Realm file itself and its auxiliary files
+<ios-realm-file>` to a less strict protection level that allows file
+access even when the device is locked, such as
+:apple:`NSFileProtectionCompleteUntilFirstUserAuthentication
```
dacharyc, line 198: I'm getting a `Page Not Found` for this link.

source/sdk/ios/examples/configure-and-open-a-realm.txt
```diff
@@ -142,3 +146,77 @@ Provide a Subset of Classes to a Realm
...
+   we encourage you to use Realm's own :ref:`built-in encryption
+   <ios-encrypt-a-realm>` to secure your data.
+
+The following example shows how to apply a less strict protection level
+to the parent directory of the default {+realm+}. 
+
+.. literalinclude:: /examples/generated/code/start/Encrypt.codeblock.background-app-refresh-encryption.swift
+   :language: swift
+
+You should apply the file protection attribute to the folder containing
```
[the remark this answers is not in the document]
dacharyc: This section was also a bit dense. Breaking it down more makes it more readable; something like:
> {+service-short+} may create and delete auxiliary files at any time. Instead of applying the file protection attribute directly to files, apply it to the parent folder. This way, you ensure the file protection applies to all relevant files.

source/sdk/ios/examples/sync-changes-between-devices.txt
```diff
@@ -86,6 +86,10 @@ to see the most recent data; without Background App Refresh, {+service+}
 updates the data when the user launches the app, potentially resulting
 in noticeable lag.
 
+In order to use the {+realm+} while the device is locked, you must
```
dacharyc, line 89:
```suggestion
To use the {+realm+} while the device is locked, you must
```

source/sdk/ios/examples/configure-and-open-a-realm.txt
```diff
@@ -142,3 +146,77 @@ Provide a Subset of Classes to a Realm
...
+To handle this, you can downgrade the file protection of the folder
+containing :ref:`the Realm file itself and its auxiliary files
+<ios-realm-file>` to a less strict protection level that allows file
+access even when the device is locked, such as
+:apple:`NSFileProtectionCompleteUntilFirstUserAuthentication
+</documentation/foundation/nsfileprotectioncompleteuntilfirstuserauthentication>`.
+
+.. tip::
+
+   If you choose to opt out of complete iOS file encryption in this way,
```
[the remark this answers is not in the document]
cbush: Reworked.

source/sdk/ios/examples/configure-and-open-a-realm.txt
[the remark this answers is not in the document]
cbush: Excellent. Reworked.

PR #936, opened 2021-03-26
(DOCSP-15345): Add Realm legacy links

cbush:
https://jira.mongodb.org/browse/DOCSP-15345

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/realm-legacy/

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -11,6 +11,7 @@ MongoDB Realm
    Realm Database SDKs </sdk>
    MongoDB Realm </cloud>
    MongoDB Stitch (Legacy) </stitch>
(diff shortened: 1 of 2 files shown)


dacharyc (approved):
Nice to have this here! I wonder if the header might be easier to quickly parse without the `pre-v10` in the header? Maybe something like "Realm Database Documentation (Legacy)" without pre-v10, to match the legacy Stitch documentation header?

I also wonder if it would make sense for the "Realm Legacy" documentation to appear higher on the page than the Stitch legacy documentation, since I generally see more discussion of Realm/Realm than Realm/Stitch?

PR #929, opened 2021-03-25
(DOCSP-14639): Port info about default realms, file size, and property attributes

dacharyc:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14639

### Docs staging link (requires sign-in on MongoDB Corp SSO):
New sections on the [Fundamentals -> Realms page](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14639/sdk/ios/fundamentals/realms): 
- File Size
- Default Realm

New section on [Fundamentals -> Object Models & Schemas page](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14639/sdk/ios/fundamentals/object-models-and-schemas/#property-attributes):
- Property Attributes

### Review Guidelines
[REVIEWING.md](https://github.com/mongodb/docs-realm/blob/master/REVIEWING.md)

PR #946, opened 2021-03-29
(DOCSP-14632): Port content on compacting from legacy Swift docs

dacharyc:
## Pull Request Info

The scope of this ticket is to port the content from the [compacting realms section of the legacy Swift docs](https://docs.mongodb.com/realm-legacy/docs/swift/latest/index.html#compacting-realms).

Specifically, this documentation covers `shouldCompactOnLaunch`.

However, PR #929 adds a section about File Size that lists another method to manage compacting: `writeCopy(toFile:encryptionKey:)`. That isn't covered in the legacy documentation.

I do plan to link between the two places, so it seems a bit odd to me that the File Sizes page lists two methods to manage compacting a Realm, but the actual page on compacting makes no mention of the second method. However, that seems out of scope for this ticket on porting legacy content.

Thoughts on how to proceed, reviewer?

### Jira

- https://jira.mongodb.org/browse/DOCSP-14632

- [Compact a Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14632/sdk/ios/advanced-guides/compacting/)
- [Adding page to Advanced Guides](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14632/sdk/ios/advanced-guides/)

### Review Guidelines

### DIFF
--- a/source/sdk/ios/advanced-guides/compacting.txt
+++ b/source/sdk/ios/advanced-guides/compacting.txt
@@ -1,8 +1,8 @@
-.. _ios-client-compacting:
+.. _ios-client-compact-a-realm:
 
-====================
-Compacting - iOS SDK
-====================
+=========================
+Compact a Realm - iOS SDK
+=========================
 
 .. default-domain:: mongodb
 
@@ -17,26 +17,32 @@ Overview
 
 The size of a {+client-database+} file is always larger than the total 
 size of the objects stored within it. This architecture enables some of 
-{+client-database+}s great performance, concurrency, and safety benefits. 
+{+realm+}'s great performance, concurrency, and safety benefits. 
 
 To avoid making expensive system calls, {+realm+} files are rarely 
 shrunk at runtime. Instead, they grow by specific size increments. 
-{+realm+} writes new data within unused space tracked inside the file. 
+{+backend-short+} writes new data within unused space tracked inside the file. 
 In some situations, unused space may comprise a significant portion of a 
 {+realm+} file. 
 
-.. _ios-compacting-realms:
+.. _ios-how-to-compact-a-realm:
 
-Compacting Realms
-~~~~~~~~~~~~~~~~~
+Compact a Realm
+~~~~~~~~~~~~~~~
 
-Use ``shouldCompactOnLaunch()`` on a Realm's configuration object to 
-manage compacting. This method is a boolean that determines whether to 
+Use ``shouldCompactOnLaunch()`` on a {+realm+}'s configuration object to 
+compact a {+realm+}. This method is a boolean that determines whether to 
 compact a realm when it first opens. Specify conditions to execute this 
 method, such as:
 
 - The size of the file on disk
 - How much free space the file contains
+
+.. important:: Compacting may not occur
+
+   If another process is accessing the {+realm+}, compacting does not occur. 
+   It doesn't matter if the configuration block's conditions apply. 
+   Compacting cannot be safely performed while a {+realm+} is being accessed.
 
 .. tabs-realm-languages::
 
@@ -49,23 +55,17 @@ method, such as:
    .. tab::
       :tabid: objective-c
 
-      .. literalinclude:: /examples/generated/code/start/Compacting.codeblock.compacting.swift
-        :language: swift
-
-.. important:: Compacting may not occur
-
-   If another process is accessing the {+realm+}, compacting does not occur. 
-   It doesn't matter if the configuration block's conditions apply. 
-   Compacting cannot be safely performed while a Realm is being accessed.
+      .. literalinclude:: /examples/generated/code/start/Compacting.codeblock.compacting.m
+        :language: objectivec
 
 .. _ios-how-compacting-works:
 
 How Compacting Works
 ~~~~~~~~~~~~~~~~~~~~
 
-Realm compacting works by:
+{+backend-short+} compacting works by:
 
-1. Reading the entire contents of the {+client-database+} file
+1. Reading the entire contents of the {+realm+} file
 2. Writing the contents to a new file at a different location
 3. Replacing the original file
 
@@ -75,18 +75,18 @@ Best Practices
 ~~~~~~~~~~~~~~
 
 Experiment with conditions to find the right balance of how often to 
-compact a {+client-database+} file. Because the operation can be 
+compact a {+realm+} file. Because the operation can be 
 expensive, don't compact every time you open a {+realm+}. Do compact 
 often enough to prevent the file size from growing too large.
 
 Summary
 -------
 
-- {+client-database+}s architecture enables benefits, but can result in 
+- {+client-database+}'s architecture enables benefits, but can result in 
   file size growth.
 
 - Use compacting to manage file size growth.
 
-- Define conditions for ``shouldCompactOnLaunch()`` to manage compacting.
+- Define conditions for ``shouldCompactOnLaunch()`` to compact a {+realm+}.
 
 - Compacting cannot occur if another process is accessing the {+realm+}.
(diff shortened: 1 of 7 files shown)


source/sdk/ios/advanced-guides/compacting.txt
```diff
@@ -0,0 +1,95 @@
...
+{+backend-short+} writes new data within unused space tracked inside the file. 
+In some situations, unused space may comprise a significant portion of a 
+{+realm+} file. 
+
+.. _ios-how-to-compact-a-realm:
+
+Compact a Realm
+~~~~~~~~~~~~~~~
+
+Use ``shouldCompactOnLaunch`` on a {+realm+}'s configuration object
```
MongoCaleb, line 33: Maybe this is just the nature of supporting 2 languages on one page, but I always prefer to see the first occurence of a method as the link, like you have below in "For more info, see: :swift-sdk:`shouldCompactOnLaunch()<Structs/Realm/Configuration.html>`" (that latter of which feels weird to me). I wonder if it wouldn't be possible to do something like:

> To compact a Realm, use :swift-sdk:`shouldCompactOnLaunch()<Structs/Realm/Configuration.html>` (Swift) or :objc-sdk:`shouldCompactOnLaunch <Classes/RLMRealmConfiguration.html>` (Obj-C) on a blah blah blah...

source/sdk/ios/advanced-guides/compacting.txt
```diff
@@ -0,0 +1,95 @@
...
+   If another process is accessing the {+realm+}, compacting does not occur. 
+   It doesn't matter if the configuration block's conditions apply. 
+   Compacting cannot be safely performed while a {+realm+} is being accessed.
+
+.. tabs-realm-languages::
+
+   .. tab::
+      :tabid: swift
+
+      For more info, see: :swift-sdk:`shouldCompactOnLaunch()<Structs/Realm/Configuration.html>`
```
MongoCaleb, line 51: see above about moving these links to the intro para.

source/sdk/ios/advanced-guides/compacting.txt
```diff
@@ -0,0 +1,95 @@
...
+Use ``shouldCompactOnLaunch`` on a {+realm+}'s configuration object 
+to compact a {+realm+}. Specify conditions to execute this method, such 
+as:
+
+- The size of the file on disk
+- How much free space the file contains
+
+.. important:: Compacting may not occur
+
+   If another process is accessing the {+realm+}, compacting does not occur.
```
MongoCaleb, line 42: The first and last sentences feel redundant to me. Can it be simplified? Perhaps:

> Compacting cannot occur while a realm is being accessed, regardless of any configuration settings.

source/sdk/ios/advanced-guides/compacting.txt
```diff
@@ -0,0 +1,95 @@
...
+How Compacting Works
+~~~~~~~~~~~~~~~~~~~~
+
+{+backend-short+} compacting works by:
+
+1. Reading the entire contents of the {+realm+} file
+2. Writing the contents to a new file at a different location
+3. Replacing the original file
+
+If the file contains a lot of data, this can be an expensive operation.
```
MongoCaleb, line 75: Can / should we say more? Like what's "a lot of data" or what is expensive -- time, storage space, memory usage?

source/sdk/ios/advanced-guides/compacting.txt
```diff
@@ -0,0 +1,95 @@
...
+
+Best Practices
+~~~~~~~~~~~~~~
+
+Experiment with conditions to find the right balance of how often to
```
MongoCaleb, line 80: nit: Perhaps worth starting with "Because compacting is expensive and depends on your app requirements, we recommend experimenting with... " Or maybe adding something about logging information so you have data about where to find the balance?

source/sdk/ios/advanced-guides/compacting.txt
```diff
@@ -0,0 +1,95 @@
...
+
+Experiment with conditions to find the right balance of how often to 
+compact a {+realm+} file. Because the operation can be 
+expensive, don't compact every time you open a {+realm+}. Do compact 
+often enough to prevent the file size from growing too large.
+
+Summary
+-------
+
+- {+client-database+}'s architecture enables benefits, but can result in
```
MongoCaleb, line 88: what does "enable benefits" mean? Fluffy; suggest something more specific.

source/sdk/ios/advanced-guides/compacting.txt
dacharyc: Yeah, I waffled on this. Moving it, thanks!

source/sdk/ios/advanced-guides/compacting.txt
dacharyc: I left out a phrase like:

> "Because compacting is expensive and depends on your app requirements, we recommend experimenting with... "

Partially because the last line before this section talks about how it can be expensive, so saying it here felt redundant - and partially because I think I read in onboarding docs somewhere that we want to avoid saying things like "we recommend." I think the recommendation is implicit in the "Best Practices" heading, though.

I'm uncertain what to add about logging. I didn't see anything in the Realm -> Logs documentation that jumped out at me as applicable here (although it's entirely possible I'm missing something in a quick scan). If we're suggesting that devs use other logging tools to evaluate performance, that feels more like general app development stuff that is out of scope of this ticket. Maybe I'm misunderstanding what you mean here - open to suggestion if you've got one, @MongoCaleb !

source/sdk/ios/advanced-guides/compacting.txt
dacharyc: Fair. The original page linked to Threading, where there was a discussion of those benefits. It looks like we cover it pretty well on our threading page, too, so I've added a `seealso` where I mention it in the `Overview`, and just added a link here.

source/sdk/ios/advanced-guides/compacting.txt
dacharyc: Good question. Perhaps a technical reviewer can suggest what we might safely say here. I did see [a Slack discussion](https://mongodb.slack.com/archives/CL9JSLTGE/p1617103754085000) around this just this morning.

MongoCaleb (approved):
Love it!

source/sdk/ios/advanced-guides/compacting.txt
```diff
@@ -0,0 +1,95 @@
...
+The size of a {+client-database+} file is always larger than the total 
+size of the objects stored within it. This architecture enables some of 
+{+realm+}'s great performance, concurrency, and safety benefits. 
+
+.. seealso::
+
+   :ref:`ios-client-threading`
+
+To avoid making expensive system calls, {+realm+} files are rarely 
+shrunk at runtime. Instead, they grow by specific size increments.
```
tgoyne, line 27: We currently never shrink the file except when compacting. The second sentence here was referring to a pre-5.0 behavior where we sometimes rounded up the file size to simplify the implementation, which we no longer do.

source/sdk/ios/advanced-guides/compacting.txt
tgoyne: I don't think we have a good understanding of when compaction performance becomes an issue. We've seen people run into issues because they just set the block to `return true` and from people failing to compact the Realm before it got too large to open, but seemingly everything in between is either fine or people haven't bothered to complain to us about it not being fine.

source/sdk/ios/advanced-guides/compacting.txt
dacharyc: Thanks, @tgoyne ! I reworked this paragraph a little, removing that sentence and trying to reflect the convos here. Does this seem like a reasonable balance, or are there additional recommendations we could make or info we could/should provide?

PR #954, opened 2021-03-31
(DOCSP-12905): Port tvOS considerations page

cbush:
- Move to integrations section

## Jira

https://jira.mongodb.org/browse/DOCSP-12905

## Staged

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/tvos-carb/sdk/ios/integrations/tvos/


source/sdk/ios/integrations/tvos.txt
```diff
@@ -0,0 +1,53 @@
...
+This page details considerations when using {+client-database+} on tvOS.
+
+.. seealso::
+
+   :ref:`ios-install`
+
+Avoid Storing Important User Data
+---------------------------------
+
+Because tvOS restricts writes to the "Documents" directory, the default
```
dacharyc, line 27: You could break this up a little bit for easier readability. Something like:
> The default Realm location is NSCachesDirectory. This is because tvOS restricts writes to the "Documents" directory. 
> 
> Avoid storing important user data in a realm. Instead, it's best to treat Realm as a rebuildable cache. This is because tvOS may purge the "Caches" directory at any point.

Although honestly, I'm uncertain how important the second line is. Do readers of Realm documentation for tvOS care to learn from us that tvOS restricts writes to the "Documents" directory?

dacharyc (approved):
LGTM! The first paragraph read a bit awkwardly to me, so I made a suggestion about how you could break it down more, but it's just a nit.

source/sdk/ios/integrations/tvos.txt
tgoyne: It was something we added to the old docs because we got questions about why we put the default Realm in a different place on tvOS. People using us on tvOS and watchOS are generally people who were already using Realm on iOS and are now adding a tvOS version of their app, and often have never touched tvOS before.

dacharyc: Ahh, that makes sense, thanks for the context, @tgoyne !

PR #960, opened 2021-04-01
(DOCSP-14656): iOS: Port content on key-value observing

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-14656

- [React to Changes: Key Value Observation section](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14656/sdk/ios/examples/react-to-changes/#key-value-observation)

### Review Guidelines

### DIFF
--- a/source/sdk/ios/examples/react-to-changes.txt
+++ b/source/sdk/ios/examples/react-to-changes.txt
@@ -221,13 +221,13 @@ Observing changes made to {+service-short+} ``List`` properties is simpler than
 
 - You don't have to mark ``List`` properties as dynamic to observe them.
 - You can call modification methods on ``List`` directly. Anyone observing 
-the property that stores it gets a notification.
+  the property that stores it gets a notification.
 
 You don't need to use ``mutableArrayValueForKey(_:)``, although {+realm+} 
 does support this for code compatibility. 
 
 .. seealso::
 
-   Example of using Realm with :github:`ReactiveCocoa from 
+   Examples of using Realm with :github:`ReactiveCocoa from 
    Objective‑C<realm/realm-cocoa/tree/master/examples/ios/objc/RACTableView>`
    , and :github:`ReactKit from Swift<realm/realm-cocoa/tree/v2.3.0/examples/ios/swift-2.2/ReactKit>`.
--- a/source/sdk/ios/examples/react-to-changes.txt
+++ b/source/sdk/ios/examples/react-to-changes.txt
@@ -195,7 +195,7 @@ Observing the properties of managed objects works differently. With
   automatically on a runloop thread
 - You begin a write transaction after changes on another thread
 
-{+backend-short+} applies changes made in the write transaction(s) on other threads 
+{+service-short+} applies changes made in the write transaction(s) on other threads 
 at once. Observers see Key-value observation notifications at once. 
 Intermediate steps do not trigger KVO notifications.
 
--- a/source/sdk/ios/examples/react-to-changes.txt
+++ b/source/sdk/ios/examples/react-to-changes.txt
@@ -191,7 +191,7 @@ Observing the properties of managed objects works differently. With
 {+realm+}-managed objects, the value of a property may change when:
 
 - You assign to it
-- The realm is refreshed, either manually with ``realm.refresh()``, or 
+- The realm is refreshed, either manually with ``realm.refresh()`` or 
   automatically on a runloop thread
 - You begin a write transaction after changes on another thread
 
--- a/source/sdk/ios/examples/react-to-changes.txt
+++ b/source/sdk/ios/examples/react-to-changes.txt
@@ -191,11 +191,9 @@ Observing the properties of managed objects works differently. With
 {+realm+}-managed objects, the value of a property may change when:
 
 - You assign to it
-- You call ``realm.refresh()``
-- The {+realm+} is automatically refreshed after a write transaction 
-  occurs on a different thread
-- You call ``realm.beginWrite()`` after changes to a different thread which 
-  have not been picked up by a refresh on the current thread
+- The realm is refreshed, either manually with ``realm.refresh()``, or 
+  automatically on a runloop thread
+- You begin a write transaction after changes on another thread
 
 {+backend-short+} applies changes made in the write transaction(s) on other threads 
 at once. Observers see Key-value observation notifications at once. 
--- a/source/sdk/ios/examples/react-to-changes.txt
+++ b/source/sdk/ios/examples/react-to-changes.txt
@@ -192,7 +192,7 @@ Observing the properties of managed objects works differently. With
 
 - You assign to it
 - You call ``realm.refresh()``
-- The {+realm+} is automatically refreshed when a write transaction 
+- The {+realm+} is automatically refreshed after a write transaction 
   occurs on a different thread
 - You call ``realm.beginWrite()`` after changes to a different thread which 
   have not been picked up by a refresh on the current thread
--- a/source/sdk/ios/examples/react-to-changes.txt
+++ b/source/sdk/ios/examples/react-to-changes.txt
@@ -158,3 +158,76 @@ invalid. You can explicitly invalidate a token by calling its
 
       .. literalinclude:: /examples/generated/code/start/Notifications.codeblock.stop-watching.m
         :language: objectivec
+
+.. _ios-key-value-observation:
+
+Key-value Observation
+---------------------
+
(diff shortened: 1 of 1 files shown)


source/sdk/ios/examples/react-to-changes.txt
```diff
@@ -158,3 +158,76 @@ invalid. You can explicitly invalidate a token by calling its
...
+.. _ios-key-value-observation:
+
+Key-value Observation
+---------------------
+
+Key-value Observation Compliance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+{+client-database+} objects are :apple:`key-value observing 
+compliant<documentation/swift/cocoa_design_patterns/using_key-value_observing_in_swift>`
```
cbush, line 171: We say "KVO" below. Should we add a "(KVO)" upon first usage here?

source/sdk/ios/examples/react-to-changes.txt
```diff
@@ -158,3 +158,76 @@ invalid. You can explicitly invalidate a token by calling its
...
+
+.. important::
+
+   You cannot add an object to a {+realm+} (with ``realm.add(obj)`` or similar 
+   methods) while it has any registered observers.
+
+Managed vs. Unmanaged KVO Considerations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Working with unmanaged instances of ``Object`` subclasses? Observing their
```
cbush, line 187: Heh. I kind of like this but I feel the tone doesn't match our other docs. I might just go for the boring version.

source/sdk/ios/examples/react-to-changes.txt
```diff
@@ -158,3 +158,76 @@ invalid. You can explicitly invalidate a token by calling its
...
+
+Working with unmanaged instances of ``Object`` subclasses? Observing their 
+properties works like any other dynamic property.
+
+Observing the properties of managed objects works differently. With 
+{+realm+}-managed objects, the value of a property may change when:
+
+- You assign to it
+- You call ``realm.refresh()``
+- The {+realm+} is automatically refreshed when a write transaction
```
cbush, line 195: I know this comes from the original but I feel like this is a bit misleading. This seems to imply a realm will notice write transactions on another thread. My understanding was that changes from other threads are only picked up upon refresh(). It's just that a realm on a runloop calls refresh() automatically for you. @tgoyne, could you clarify?

source/sdk/ios/examples/react-to-changes.txt
```diff
@@ -158,3 +158,76 @@ invalid. You can explicitly invalidate a token by calling its
...
+properties works like any other dynamic property.
+
+Observing the properties of managed objects works differently. With 
+{+realm+}-managed objects, the value of a property may change when:
+
+- You assign to it
+- You call ``realm.refresh()``
+- The {+realm+} is automatically refreshed when a write transaction 
+  occurs on a different thread
+- You call ``realm.beginWrite()`` after changes to a different thread which
```
cbush, line 197: Should this be more general? ``realm.write`` also starts a transaction which should also trigger the updates.

source/sdk/ios/examples/react-to-changes.txt
```diff
@@ -158,3 +158,76 @@ invalid. You can explicitly invalidate a token by calling its
...
+Intermediate steps do not trigger KVO notifications.
+
+.. example::
+
+   Say your app performs a write transaction that increments a property 
+   from 1 to 10. On the main thread, you get a single notification of a 
+   change directly from 1 to 10. You won't get notifications for every 
+   incremental change between 1 and 10.
+
+We don't recommend modifying managed Realm objects from within
```
cbush, line 211: "We don't recommend" -> "Avoid"?

cbush (commented):
It looks good, but I have some questions.

source/sdk/ios/examples/react-to-changes.txt
dacharyc: I waffled on using "KVO" at all. I wasn't going to use it, but then the header below got really long. I'll add the "(KVO)" on first usage here, but I'm equally happy to revise the header in 184 and the usage in 202 to omit it. Do you have thoughts, @cbush ?

cbush: I think "KVO" is fairly standard among the Apple developer community. I just think it's worth mentioning the first time.

source/sdk/ios/examples/react-to-changes.txt
dacharyc: I had a difficult time parsing and disentangling this paragraph from the original documentation:

> Observing properties of managed objects (those previously added to a Realm) works a little differently. With managed objects, there are three times when the value of a property may change: when you directly assign to it; when you call realm.refresh() or the Realm is automatically refreshed after a write transaction is committed on a different thread; and when you call realm.beginWrite() after changes on a different thread which have not been picked up by a refresh on the current thread.

I translated it to the bullet points from 193 to 198, which you're asking about here and in the next comment, @cbush , but I'd love some technical clarification around this paragraph. I had a hard time communicating this accurately and effectively.

source/sdk/ios/examples/react-to-changes.txt
dacharyc: Alright, I added it here at first mention. Thanks!

source/sdk/ios/examples/react-to-changes.txt
tgoyne: The wording here is correct. Starting a write transaction acquires the write lock and then implicitly calls `refresh()`.

tgoyne: Or nevermind, I misread it. "is automatically refreshed when a write transaction" should be "is automatically refreshed after a write transaction"

dacharyc: Updated the wording to correct that. Thanks @tgoyne !

source/sdk/ios/examples/react-to-changes.txt
```diff
@@ -158,3 +158,76 @@ invalid. You can explicitly invalidate a token by calling its
...
+{+realm+}-managed objects, the value of a property may change when:
+
+- You assign to it
+- You call ``realm.refresh()``
+- The {+realm+} is automatically refreshed after a write transaction 
+  occurs on a different thread
+- You call ``realm.beginWrite()`` after changes to a different thread which 
+  have not been picked up by a refresh on the current thread
+
+{+backend-short+} applies changes made in the write transaction(s) on other threads
```
cbush, line 200: Don't use backend-short here. That could become "Application Services":

```suggestion
{+service-short+} applies changes made in the write transaction(s) on other threads 
```

source/sdk/ios/examples/react-to-changes.txt
cbush: Ah, I see. I feel like we're missing a "at some point", as in "When the realm is automatically refreshed at some point after a write on another thread". It still reads as if the write on the other thread _causes_ the automatic refresh on this thread. (Correct me if I'm wrong, but I don't believe it could do that and still benefit from MVCC.) Anyway, it could be combined with above point, making it about a realm being refreshed (whether manually or automatically).

source/sdk/ios/examples/react-to-changes.txt
dacharyc: Yeah, I've tried rewording these bullets like three different times, but haven't been happy with anything I've been able to come up with. I'd love your thoughts on a better way to present this section next week, @cbush !

source/sdk/ios/examples/react-to-changes.txt
```diff
@@ -158,3 +158,74 @@ invalid. You can explicitly invalidate a token by calling its
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Observing the properties of unmanaged instances of ``Object`` subclasses 
+works like any other dynamic property.
+
+Observing the properties of managed objects works differently. With 
+{+realm+}-managed objects, the value of a property may change when:
+
+- You assign to it
+- The realm is refreshed, either manually with ``realm.refresh()``, or
```
cbush, line 194: [No comma](https://www.grammarly.com/blog/comma-between-correlative-conjunction-sets/) in "Either ... or"
```suggestion
- The realm is refreshed, either manually with ``realm.refresh()`` or 
```

cbush (approved):

PR #966, opened 2021-04-02
Fix Autobuilder CI

cbush:
- Update autobuilder checker for new backend log style
- Reduce time waiting for ongoing builds if none exists


nathan-contino-mongo (approved):
LGTM!

PR #970, opened 2021-04-05
(DOCSP-14636): Update threading with thread-safe reference docs

cbush:
### Jira

https://jira.mongodb.org/browse/DOCSP-14636

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/threading/sdk/ios/advanced-guides/threading/#pass-instances-across-threads

### Review Guidelines

### DIFF
--- a/examples/ios/Examples/Threading.swift
+++ b/examples/ios/Examples/Threading.swift
@@ -1,7 +1,29 @@
+// :replace-start: {
+//   "terms": {
+//     "ThreadingExamples_": ""
+//   }
+// }
 import XCTest
 import RealmSwift
 
+class ThreadingExamples_Person : Object {
+    @objc dynamic var name = ""
+    
+    convenience init(name: String) {
+        self.init()
+        self.name = name
+    }
+}
+
 class Threading: XCTestCase {
+    override func setUp() {
+        Realm.Configuration.defaultConfiguration = Realm.Configuration(inMemoryIdentifier: self.name)
+    }
+
+    override func tearDown() {
+        Realm.Configuration.defaultConfiguration = Realm.Configuration()
+    }
+
     func testFreeze() {
         // :code-block-start: freeze
         let realm = try! Realm()
@@ -40,4 +62,32 @@ class Threading: XCTestCase {
         assert(frozenTask.realm!.isFrozen)
         // :code-block-end:
     }
+    
+    func testPassAcrossThreads() {
+        // :code-block-start: pass-across-threads
+        let person = ThreadingExamples_Person(name: "Jane")
+        let realm = try! Realm()
+
+        try! realm.write {
+            realm.add(person)
+        }
+        
+        // Create thread-safe reference to person
+        let personRef = ThreadSafeReference(to: person)
+        
+        // Pass the reference to a background thread
+        DispatchQueue(label: "background").async {
+            autoreleasepool {
+                let realm = try! Realm()
+                guard let person = realm.resolve(personRef) else {
+                    return // person was deleted
+                }
+                try! realm.write {
+                    person.name = "Jane Doe"
+                }
+            }
+        }
+        // :code-block-end:
+    }
 }
+// :replace-end:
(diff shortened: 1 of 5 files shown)


examples/ios/Examples/Threading.swift
```diff
@@ -40,4 +89,54 @@ class Threading: XCTestCase {
...
+
+        // Create thread-safe reference to person
+        let personRef = ThreadSafeReference(to: person)
+
+        // Pass the reference to a background thread
+        DispatchQueue(label: "background").async {
+            autoreleasepool {
+                let realm = try! Realm()
+                guard let person = realm.resolve(personRef) else {
+                    return // person was deleted
```
ericjordanmossman, line 111: I think resolving a reference to an object which has been deleted would pass this guard. Similar to what's described [here](https://github.com/realm/realm-cocoa/issues/5420#issuecomment-338811085). ex)
<details>
Screen Shot 2021-04-05 at 5 15 05 PM
</details>

I believe it should be:
```
DispatchQueue(label: "background").async {
            autoreleasepool {
                let realm = try! Realm()
                realm.refresh()
                guard let person = realm.resolve(personRef) else {
// ...
```
@jsflax , does this change look right? ^ @cbush is using the [person example](https://docs.mongodb.com/realm-legacy/docs/swift/latest/index.html#threading)  from the old docs. So I want to confirm there isn't a reason the old example doesn't refresh. Otherwise it means the old example may have been wrong.

cbush: Does `try! Realm()` not implicitly refresh? Also, what happens if the delete on the other thread happened between refresh() and resolve()? Isn't this always going to be an issue?

tgoyne: `try! Realm()` doesn't refresh implicitly, but here it's going to be a fresh instance and not a cached one so it'll always be the latest version. If it was a cached instance older than when `personRef` was created then `resolve()` would refresh the Realm. If a write happens between the call to resolve() and the call to `write()`, `write()` will refresh the Realm and it will crash inside the write transaction if the object was deleted; any sort of checking to see if the objects you're working with are still valid needs to happen inside write transactions rather than before it as those checks will be stale after starting the write.

cbush: I see. I have moved the checks into the transactions. Could you verify that it's now correct? Thanks!

source/sdk/ios/advanced-guides/threading.txt
```diff
@@ -120,6 +122,66 @@ several options depending on your use case:
...
+
+   You must resolve a ``ThreadSafeReference`` exactly once. Otherwise,
+   the source {+realm+} will remain pinned until the reference gets
+   deallocated. For this reason, ``ThreadSafeReference`` should be
+   short-lived.
+
+.. literalinclude:: /examples/generated/code/start/Threading.codeblock.pass-across-threads.swift
+   :language: swift
+
+You can use ``ThreadSafeReference`` with Realm objects whether they have
```
nathan-contino-mongo, line 150:
```suggestion
You can use ``ThreadSafeReference`` with {+service-short+} objects whether they have
```

source/sdk/ios/advanced-guides/threading.txt
```diff
@@ -120,6 +122,66 @@ several options depending on your use case:
...
+a primary key or not. You can also use it with lists and results. The
```
nathan-contino-mongo, line 151: While I naturally think of primary keys as a means of implementing this, is it potentially confusing to mention them here? Would it be better to say "with any realm object"?

nathan-contino-mongo (approved):
LGTM with a couple of thoughts.

source/sdk/ios/advanced-guides/threading.txt
cbush: The point is that the other option (just query again on the other thread) gets complicated when the object doesn't have a primary key. I'll try to make the connection clearer.

PR #975, opened 2021-04-06
(DOCSP-14651): Initialize an object by value

cbush:
### Jira

https://jira.mongodb.org/browse/DOCSP-14651

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/carb-nested/sdk/ios/examples/read-and-write-data/#initialize-an-object-with-a-value

### Review Guidelines

### DIFF
--- a/source/sdk/ios/examples/read-and-write-data.txt
+++ b/source/sdk/ios/examples/read-and-write-data.txt
@@ -334,14 +334,14 @@ Create a New Object
 
 .. _ios-initialize-an-object-with-a-value:
 
-Initialize an Object with a Value
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Initialize Objects with a Value
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You can initialize an object directly by passing an initializer value to
+You can initialize an object by passing an initializer value to
 :swift-sdk:`Object.init(value:)
 <Extensions/Object.html#/s:So16RealmSwiftObjectC0aB0E5valueAByp_tcfc>`.
-The initializer value can be a :apple:`key-value
-coding<library/archive/documentation/Cocoa/Conceptual/KeyValueCoding/>`
+The initializer value can be a :apple:`key-value coding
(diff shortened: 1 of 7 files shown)


source/sdk/ios/examples/read-and-write-data.txt
```diff
@@ -332,6 +332,43 @@ Create a New Object
       .. literalinclude:: /examples/generated/code/start/ReadWriteData.codeblock.create.m
          :language: objectivec
 
+.. _ios-initialize-an-object-with-a-value:
+
+Initialize an Object with a Value
```
dacharyc, line 337: I wonder if it makes sense to communicate this only as an "Initialize an Object with a Value" example? Couple of thoughts:
- "Initialize with a Value" as a header doesn't necessarily jump out at me as something that will also cover nested objects. So if I'm just on this page scanning the "On This Page" to find out about dealing with nested objects, I don't see anything obvious to me.
- Can you also update an object with a nested object? If I'm on the page looking for info about updating an object, I might miss the note nesting/embedding objects here in initializing.

dacharyc (approved):
Looks good. I did have a couple of thoughts about presenting this solely as an "Initialize" example, but I don't think it necessarily requires changes if you don't agree - just musing.

Related: I do think the Read/Write page is getting a bit long and unwieldy at this point. When even the "On This Page" requires non-trivial scrolling, my eyes glaze over and it's probably time to figure out a good way to break this up more. Out of scope for this ticket, though.

source/sdk/ios/examples/read-and-write-data.txt
cbush: I felt the nested object support was a bit overblown in the ticket and original doc, so I shifted focus onto value initialization in general. I will add a link from the KVC update section below.

cbush:
I agree about the length. The obvious split would be "Read Data" and "Write Data". Let's discuss later.

source/sdk/ios/examples/read-and-write-data.txt
dacharyc: That's fair. I do feel the value initialization... adds value. 👍 for the changes.

PR #981, opened 2021-04-07
(DOCSP-14638): Part 1: updates to object models - primary keys

cbush:
### Jira

https://jira.mongodb.org/browse/DOCSP-14638

- [Object Models & Schemas](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/carb-objects/sdk/ios/fundamentals/object-models-and-schemas/#primary-keys)
- [Upsert](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/carb-objects/sdk/ios/examples/read-and-write-data/#upsert-an-object)

### Review Guidelines

### DIFF
--- a/source/sdk/ios/examples/read-and-write-data.txt
+++ b/source/sdk/ios/examples/read-and-write-data.txt
@@ -484,7 +484,7 @@ Objective-C object.
    dot-notation or bracket-notation as if it were in a regular, nested
    object.
 
-.. _ios-upsert-an-object:
+.. _ios-update-with-key-value-coding:
 
 Update Properties with Key-value Coding
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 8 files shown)


source/sdk/ios/examples/read-and-write-data.txt
```diff
@@ -530,13 +533,27 @@ Upsert an Object
...
+
+      .. literalinclude:: /examples/generated/code/start/ReadWriteData.codeblock.partial-update.swift
+         :language: swift
+
+   .. tab::
+      :tabid: objective-c
+
+      .. literalinclude:: /examples/generated/code/start/ReadWriteData.codeblock.partial-update.m
+         :language: objectivec
+
```
nathan-contino-mongo, line 556: Thought: Should we add a new section, perhaps titled "Replace an Object", that discusses the `.all` update policy? I think the ticketed sections are the only sections of the realm.io docs that discuss the concept, so either we should make the choice to omit mention of `.all` from the docs or maybe just create a backlog ticket for it.

Additional thought: is `.all` the default update policy? Should we mention that here?

nathan-contino-mongo (approved):
LGTM with one rambly thought

source/sdk/ios/examples/read-and-write-data.txt
cbush: This is documented in the SDK reference docs. I think we're getting into API specifics here best covered by the reference docs.

PR #1008, opened 2021-04-23
(DOCSP-15555): Add "thaw" to frozen object documentation

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-15555

- [Threading: Frozen Objects](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-15555/sdk/ios/advanced-guides/threading/#frozen-objects)

### Review Guidelines

### DIFF
--- a/examples/ios/Examples/Threading.swift
+++ b/examples/ios/Examples/Threading.swift
@@ -90,17 +90,17 @@ class Threading: XCTestCase {
         assert(frozenTask.realm!.isFrozen)
 
         // When you want to modify a frozen object, you can thaw it
-        let thawedTask = task.thaw()
+        let thawedTask = frozenTask.thaw()
 
         assert(thawedTask!.isFrozen == false)
 
         // You can thaw collections
-        let thawedTasks = tasks.thaw()
+        let thawedTasks = frozenTasks.thaw()
 
         assert(thawedTasks!.isFrozen == false)
 
         // And you can thaw a frozen Realm
-        let thawedRealm = realm.thaw()
+        let thawedRealm = frozenRealm.thaw()
 
         assert(thawedRealm.isFrozen == false)
         // :code-block-end:
--- a/examples/ios/Examples/Threading.swift
+++ b/examples/ios/Examples/Threading.swift
@@ -88,21 +88,44 @@ class Threading: XCTestCase {
         assert(frozenTask.isFrozen)
         // Frozen objects have a reference to a frozen realm
         assert(frozenTask.realm!.isFrozen)
+        // :code-block-end:
+    }
 
-        // When you want to modify a frozen object, you can thaw it
+    func testThaw() {
+        let realm = try! Realm()
+
+        try! realm.write {
+            realm.add(Task())
+        }
+
+        let frozenRealm = realm.freeze()
+
+        // :code-block-start: thaw
+        // Read from a frozen realm
+        let frozenTasks = frozenRealm.objects(Task.self)
+
+        // The collection that we pull from the frozen realm is also frozen
+        assert(frozenTasks.isFrozen)
+
+        // Get an individual task from the collection
+        let frozenTask = frozenTasks.first!
+
+        // To modify the task, you must first thaw it
+        // You can also thaw collections and realms
         let thawedTask = frozenTask.thaw()
 
         assert(thawedTask!.isFrozen == false)
 
-        // You can thaw collections
-        let thawedTasks = frozenTasks.thaw()
+        // Thawing the task also thaws the frozen realm it references
+        assert(thawedTask!.realm!.isFrozen == false)
 
-        assert(thawedTasks!.isFrozen == false)
+        // Let's make the code easier to follow by naming the thawed realm
+        let thawedRealm = thawedTask!.realm!
 
-        // And you can thaw a frozen Realm
-        let thawedRealm = frozenRealm.thaw()
-
-        assert(thawedRealm.isFrozen == false)
+        // Now, you can modify the task
+        try! thawedRealm.write {
+           thawedTask!.status = "Done"
+        }
         // :code-block-end:
     }
 
--- a/examples/ios/Examples/Threading.swift
+++ b/examples/ios/Examples/Threading.swift
@@ -114,7 +114,10 @@ class Threading: XCTestCase {
         // You can also thaw collections and realms
         let thawedTask = frozenTask.thaw()
 
-        assert(thawedTask!.isFrozen == false)
+        // Check to make sure this task is valid. An object is
+        // invalidated when it is deleted from its managing realm,
+        // or when its managing realm has invalidate() called on it.
+        assert(thawedTask?.isInvalidated == false)
 
         // Thawing the task also thaws the frozen realm it references
         assert(thawedTask!.realm!.isFrozen == false)
--- a/examples/ios/Examples/Threading.swift
+++ b/examples/ios/Examples/Threading.swift
@@ -88,6 +88,21 @@ class Threading: XCTestCase {
         assert(frozenTask.isFrozen)
         // Frozen objects have a reference to a frozen realm
         assert(frozenTask.realm!.isFrozen)
+
+        // When you want to modify a frozen object, you can thaw it
+        let thawedTask = task.thaw()
+
+        assert(thawedTask!.isFrozen == false)
+
+        // You can thaw collections
+        let thawedTasks = tasks.thaw()
+
+        assert(thawedTasks!.isFrozen == false)
+
+        // And you can thaw a frozen Realm
+        let thawedRealm = realm.thaw()
+
+        assert(thawedRealm.isFrozen == false)
         // :code-block-end:
     }
 
--- a/source/examples/generated/code/start/Threading.codeblock.freeze.swift
+++ b/source/examples/generated/code/start/Threading.codeblock.freeze.swift
@@ -27,18 +27,3 @@ let frozenTask = task.freeze()
 assert(frozenTask.isFrozen)
 // Frozen objects have a reference to a frozen realm
 assert(frozenTask.realm!.isFrozen)
-
-// When you want to modify a frozen object, you can thaw it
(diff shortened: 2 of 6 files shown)


examples/ios/Examples/Threading.swift
```diff
@@ -88,6 +88,21 @@ class Threading: XCTestCase {
...
+
+        assert(thawedTask!.isFrozen == false)
+
+        // You can thaw collections
+        let thawedTasks = tasks.thaw()
+
+        assert(thawedTasks!.isFrozen == false)
+
+        // And you can thaw a frozen Realm
+        let thawedRealm = realm.thaw()
```
nathan-contino-mongo, line 103: Interesting that you thaw the objects *before* you thaw the realm -- I wonder if we should explain how a live object can exist when it's tied to a frozen realm?

source/sdk/ios/advanced-guides/threading.txt
```diff
@@ -294,8 +292,9 @@ thread-safe.
              // ...
          }
 
-To modify a frozen object, query for it on an unfrozen
-{+realm+}, then modify it.
+To modify a frozen object, you thaw the object. Alternately, you can
```
nathan-contino-mongo, line 295: Ditto on the "but what if the realm is frozen" point -- I'm immediately curious how this works, and I feel like there could be some gotchas that I don't know about here.

nathan-contino-mongo (approved):
LGTM with one big ol' question

source/sdk/ios/advanced-guides/threading.txt
ericjordanmossman: > I'm immediately curious how this works

Thaw effectively just creates a ThreadSafeReference, and resolves it into a live copy.
To thaw a frozen object, it's realm is [thawed](https://github.com/realm/realm-cocoa/blob/master/Realm/RLMResults.mm#L519). To [thaw a realm](https://github.com/realm/realm-cocoa/blob/master/Realm/RLMRealm.mm#L959), the frozen realm's configuration is used to open a realm reference.

The thaw API is more or less the same concept that Chris Bush originally had [here](https://github.com/cbush/docs-realm/blob/fd557b604ce72f97fee4131c268d2d8cf576decb/examples/ios/QuickStartSwiftUI/QuickStart.swift#L286). This has since been replaced by [ObservedRealmObject/Results](https://github.com/cbush/docs-realm/blob/4130d84173727e5998418c8230c4d464936e7449/examples/ios/QuickStartSwiftUI/QuickStart.swift#L237), which uses [thaw](https://github.com/realm/realm-cocoa/blob/master/RealmSwift/SwiftUI.swift#L27) for various operations.

source/sdk/ios/advanced-guides/threading.txt
```diff
@@ -294,8 +292,9 @@ thread-safe.
              // ...
          }
 
-To modify a frozen object, query for it on an unfrozen
-{+realm+}, then modify it.
+To modify a frozen object, you thaw the object. Alternately, you can 
+query for it on an unfrozen {+realm+}, then modify it. Calling ``thaw`` 
+on a live collection returns itself.
```
ericjordanmossman, line 297: Small thing: an `Object` and `Realm` will also return self if they're already live. Not just collections.

source/examples/generated/code/start/Threading.codeblock.freeze.swift
```diff
@@ -27,3 +27,18 @@ let frozenTask = task.freeze()
 assert(frozenTask.isFrozen)
 // Frozen objects have a reference to a frozen realm
 assert(frozenTask.realm!.isFrozen)
+
+// When you want to modify a frozen object, you can thaw it
```
ericjordanmossman, line 31: Suggestion: "you can thaw it" -> "you must thaw it first".
Thawing alone won't modify anything.

ericjordanmossman (approved):
LGTM.
I left some optional comments.

source/examples/generated/code/start/Threading.codeblock.freeze.swift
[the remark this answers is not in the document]
dacharyc: Updated - thanks!

source/sdk/ios/advanced-guides/threading.txt
dacharyc: Based on the conversation here, I think my code example was too arbitrary and didn't accurately demonstrate a real use case for `thaw`. 

What do we think of this new code snippet, @nathan-contino-mongo ? It's after the "Important: On caching frozen objects" callout: 

Staging: [Frozen Objects](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-15555/sdk/ios/advanced-guides/threading/#frozen-objects)

[Threading.codeblock.thaw.swift](https://github.com/mongodb/docs-realm/pull/1008/files#diff-c523092d15f99f94445a27099fa9eb08d893411768d26fe4d20f18126dea81e6)

source/sdk/ios/advanced-guides/threading.txt
```diff
@@ -234,20 +234,19 @@ immutable copies around to many threads for processing
 before ultimately ending up on the UI thread. Making a deep
 copy every time would be expensive, and {+client-database+} does not allow
 live instances to be shared across threads. In this case,
-you can **freeze** objects, collections, and {+realms+}.
+you can **freeze** and **thaw** objects, collections, and {+realms+}.
 
 Freezing creates an immutable view of a specific object,
 collection, or {+realm+} that still exists on disk and does not
 need to be deeply copied when passed around to other
 threads. You can freely share a frozen object across threads
```
nathan-contino-mongo, line 242: I wonder if, instead of "does not need to be deeply copied when passed around to other threads", we could simply change "immutable view" to "thread-safe, immutable copy". Personally that makes more sense to me. Especially since the next sentence repeats the idea that you can freely pass frozen objects across threads.

nathan-contino-mongo: What does "specific object, collection , or realm *that still exists on disk*" refer to? I think I might be missing the point of the sentence, in which case I suspect we should rewrite the sentence to better convey that intended meaning.

source/examples/generated/code/start/Threading.codeblock.thaw.swift
```diff
@@ -0,0 +1,25 @@
...
+// Thawing the task also thaws the frozen realm it references
+assert(thawedTask!.realm!.isFrozen == false)
+
+// Let's make the code easier to follow by naming the thawed realm
+let thawedRealm = thawedTask!.realm!
+
+// Now, you can modify the task
+try! thawedRealm.write {
+   thawedTask!.status = "Done"
+}
```
nathan-contino-mongo, line 25: I'm confused -- the staging link shows me a code snippet that includes `frozenTasks2`, which doesn't reflect the code snippet I see here. Maybe there was an error staging this commit?

source/examples/generated/code/start/Threading.codeblock.thaw.swift
```diff
@@ -0,0 +1,25 @@
...
+let frozenTask = frozenTasks.first!
+
+// To modify the task, you must first thaw it
+// You can also thaw collections and realms
+let thawedTask = frozenTask.thaw()
+
+assert(thawedTask!.isFrozen == false)
+
+// Thawing the task also thaws the frozen realm it references
+assert(thawedTask!.realm!.isFrozen == false)
```
nathan-contino-mongo, line 17: Thought: if a task has been deleted since freezing, this line (and the previous assert) would presumably throw an exception. Maybe we should add a line above the previous assert that does something like this:
```
// check to make sure the thawed object is valid and has not been deleted since freezing
assert(thawedTask.isValid)
```

nathan-contino-mongo (approved):
LGTM

source/examples/generated/code/start/Threading.codeblock.thaw.swift
dacharyc: I've created a second code snippet that deals with thawing, a bit further down the page. It's after the "Important: On caching frozen objects" callout. `frozenTasks2` is part of Bush's original freeze example.

source/examples/generated/code/start/Threading.codeblock.thaw.swift
dacharyc: I'm a little envious, @nathan-contino-mongo - looks like the Swifty way of doing this is a little less elegant:

`      // Check to make sure this task is valid and has not been deleted since it was frozen.
        assert(thawedTask!.isInvalidated == false)`

Would you then proceed directly from this to line 14 above? Or would you wrap this and everything following it in a `guard`? I guess I'm wondering what our preferred balance is in handling exceptions versus adding complexity to the code examples.

nathan-contino-mongo: i would probably either proceed directly from this or just perform an if check -- just to show folks a best practice of checking to make sure the thawed object actually exists.

dacharyc: Alright, I've added that, along with a heading and a ref so we can get to this part more directly: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-15555/sdk/ios/advanced-guides/threading/#modify-a-frozen-object

Thanks for the feedback!

source/sdk/ios/advanced-guides/threading.txt
ericjordanmossman: > or realm that still exists on disk" refer to?
I think it's describing that frozen realms are on disk. ie. when you freeze something, it's not being copied.
It's fair to split the clause into a new sentence if you think it makes sense for docs. It makes sense technically:
"Freezing creates an immutable view of a specific object,
collection, or {+realm+}. Frozen objects/collections/realm still exist on disk and do not
need to be deeply copied when passed around to other
threads."

ericjordanmossman (approved):

PR #1009, opened 2021-04-23
(DOCSP-15702): Crypto is not fully supported

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-15702

- [Built-In Module Support > Partially Supported Modules > Crypto](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-15702-crypto-not-fully-supported/functions/built-in-module-support/#crypto)

### Review Guidelines

### DIFF
--- a/source/functions/built-in-module-support.txt
+++ b/source/functions/built-in-module-support.txt
@@ -123,7 +123,9 @@ dgram
 
 crypto
 ~~~~~~
-{+service-short+} supports the :nodejs:`crypto <api/crypto.html>` module with the following **exceptions**:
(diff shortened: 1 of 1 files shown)


ccho-mongodb (approved):
LGTM. Just a suggestion, but I think it could be less verbose if in the form:
"Realm supports all the methods in the crypto library except the following:
- method name 1
- method name 2
- method name 3
"

When you describe the following items as "exceptions", it seems redundant to restate the methods as items that Realm does not support.

PR #1027, opened 2021-05-03
(DOCSP-14568): [iOS] Add mixed type to Supported Property Types

dacharyc:
## Pull Request Info

The links to Jazzy docs for `RealmProperty` and `AnyRealmValue` are broken currently, but I believe they'll be valid once the auto-generated docs are updated.

As part of this work, I updated the docs where `RealmOptional` was specified. In most instances where, I replaced `RealmOptional` with `RealmProperty`, but on the Supported Properties page, I also added directives about the version in which this change occurred. I don't know the version number yet; I'll update that before I publish the docs.

I also updated the code examples that used `RealmOptional` to use `RealmProperty` and generated updated snippets with Bluehawk.

### Jira

- https://jira.mongodb.org/browse/DOCSP-14568

- [Supported Property Types](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14568/sdk/ios/data-types/supported-property-types/#mixed-data-type): Add Mixed Data Type section
- [Define a Realm Object Model](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14568/sdk/ios/examples/define-a-realm-object-model/#specify-an-optional-required-property): Update "Specify an Optional/Required Property" section to use `RealmProperty`
- [Object models & Schemas](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14568/sdk/ios/fundamentals/object-models-and-schemas/): Update "Property Attributes" section to use `RealmProperty`

### Review Guidelines

### DIFF
--- a/source/sdk/ios/data-types/supported-property-types.txt
+++ b/source/sdk/ios/data-types/supported-property-types.txt
@@ -183,7 +183,7 @@ permitted value, you can't declare an ``AnyRealmValue`` as optional.
 .. literalinclude:: /examples/generated/code/start/AnyRealmValue.codeblock.mixed-data-type.swift
    :language: swift
 
-You can compare these mixed value types:
+You can :ref:`compare <ios-filter-data/#operators>` these mixed value types:
 
 - Numeric: int, bool, float, double, decimal
 - Byte-based: string, binary
@@ -191,8 +191,11 @@ You can compare these mixed value types:
 
 When using the ``AnyRealmValue`` mixed data type, keep these things in mind:
 
-- Equals queries match on the value and the type
-- Not equals queries match objects with either different values or 
+- ``equals`` queries match on value and type
+- ``not equals`` queries match objects with either different values or 
   different types
-- Realm converts comparable numeric properties where possible
-- String properties do not match numeric queries.
+- {+realm+} converts comparable numeric properties where possible. For example,
+  in a mixed type field, 1 matches all of 1.0, 1, and true.
+- String properties do not match numeric queries. For example, in a mixed
+  type field, 1 does not match "1". "1" does not match 1, 1.0, or true.
+  
--- a/source/sdk/ios/data-types/supported-property-types.txt
+++ b/source/sdk/ios/data-types/supported-property-types.txt
@@ -66,6 +66,9 @@ Supported Property Types - iOS SDK
          * - :swift-sdk:`MutableSet <Classes/MutableSet.html>`
            - ``let value = MutableSet<Type>()``
            - 
+         * - :swift-sdk:`Mixed Data Type <Enums/AnyRealmValue.html>`
+           - ``let value = RealmProperty<AnyRealmValue>()``
+           - N/A
          * - User-defined :swift-sdk:`Object <Extensions/Object.html>`
            - N/A
            - ``@objc dynamic var value: MyClass?``
@@ -146,3 +149,44 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
 globally-unique value. You can :ref:`index <ios-index-a-property>` both types,
 and use either as a :ref:`primary key <ios-specify-a-primary-key>`.
 
+.. _ios-mixed-data-type:
+
+Mixed Data Type
+---------------
+
+.. versionadded:: 10.?
+   The ``AnyRealmValue`` type is new in this version.
+
+``AnyRealmValue`` is a {+service-short+} property type that can hold different 
+data types. Supported ``AnyRealmValue`` data types include:
+
+- Int
+- Float
+- Double
+- Decimal128
+- ObjectID
+- UUID
+- Bool
+- Date
+- Data
+- String
+- Object
+
+This mixed data type is indexable, but you can't use it as a primary key. 
+Because ``null`` is a permitted value, you can't declare an ``AnyRealmValue`` 
+as optional.
+
+You can compare these mixed value types:
+- Numeric: int, bool, float, double, decimal
+- Byte-based: string, binary
+- Time-based: timestamp, objectId
+
+When using the ``AnyRealmValue`` mixed data type, keep these things in mind:
+
+- Equals queries match on the value and the type
+- Not equals queries match objects with either different values or different types
+- Realm converts comparable numeric properties where possible
+- String properties do not match numeric queries.
+
+.. literalinclude:: /examples/generated/code/start/AnyRealmValue.codeblock.mixed-data-type.swift
+   :language: swift
--- a/source/sdk/ios/data-types/supported-property-types.txt
+++ b/source/sdk/ios/data-types/supported-property-types.txt
@@ -183,7 +183,7 @@ permitted value, you can't declare an ``AnyRealmValue`` as optional.
 .. literalinclude:: /examples/generated/code/start/AnyRealmValue.codeblock.mixed-data-type.swift
    :language: swift
 
-You can :ref:`compare <ios-filter-data/#operators>` these mixed value types:
+You can :ref:`compare <ios-filter-data-operators>` these mixed value types:
 
 - Numeric: int, bool, float, double, decimal
 - Byte-based: string, binary
--- a/source/sdk/ios/data-types/supported-property-types.txt
+++ b/source/sdk/ios/data-types/supported-property-types.txt
@@ -12,6 +12,9 @@ Supported Property Types - iOS SDK
    :depth: 2
    :class: singlecol
 
+Property Cheat Sheet
+--------------------
+
 .. tabs-realm-languages::
 
    .. tab::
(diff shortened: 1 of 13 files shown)


source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
...
+- Bool
+- Date
+- Data
+- String
+- Object
+
+This :swift-sdk:`mixed data type <Enums/AnyRealmValue.html>` is 
+:ref:`indexable <ios-index-a-property>`, but you can't use it as a 
+:ref:`primary key <ios-specify-a-primary-key>`. Because ``null`` is a 
+permitted value, you can't declare an ``AnyRealmValue`` as optional.
```
nathan-contino-mongo, line 181: I do not think that the second part of this sentence follows from the first -- maybe we should explain that because AnyRealmValue actually *refers* to an enclosed object (or something like that), an AnyRealmValue itself cannot be null -- but the object it refers to can be null. Sorta. Or however we want to word that. Basically they're pointers but they're not.

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
...
+
+.. literalinclude:: /examples/generated/code/start/AnyRealmValue.codeblock.mixed-data-type.swift
+   :language: swift
+
+You can compare these mixed value types:
```
nathan-contino-mongo, line 186: on first read, it was confusing to me what you meant by "compare". Perhaps we can link here to the query guide?

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
...
+
+- Numeric: int, bool, float, double, decimal
+- Byte-based: string, binary
+- Time-based: timestamp, objectId
+
+When using the ``AnyRealmValue`` mixed data type, keep these things in mind:
```
nathan-contino-mongo, line 192: "things" just read like most MongoDB documentation to me, but I can't quite put my finger on why. Maybe we should reword?

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
...
+
+- Equals queries match on the value and the type
```
nathan-contino-mongo, line 194:
```suggestion
- ``equals`` queries match on value and type
```
I think we don't need the articles here, and we *could* monospace equals to make it clearer that we're talking about the query language term.

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
...
+- Not equals queries match objects with either different values or
```
nathan-contino-mongo, line 195: Same as above regarding monospacing `not equals`

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
...
+- Numeric: int, bool, float, double, decimal
+- Byte-based: string, binary
+- Time-based: timestamp, objectId
+
+When using the ``AnyRealmValue`` mixed data type, keep these things in mind:
+
+- Equals queries match on the value and the type
+- Not equals queries match objects with either different values or 
+  different types
+- Realm converts comparable numeric properties where possible
```
nathan-contino-mongo, line 197: If you take my advice above to use literal query keywords above, we'll probably want to lowercase the beginning of the other bullet points as well. Besides, it (could) help avoid confusion since Realm != realm
```suggestion
- {+realm+} converts comparable numeric properties where possible
```

Nitpick: "converts comparable numeric properties where possible" is so vague that it's basically meaningless. I wonder if we can specify something a little clearer. For instance, what happens if I try to compare "equals 2.0" to an integer value of 2? What about the other direction? Are there any serious limitations converting between doubles and integers or vice versa? How exact are double comparisons (for instance, is 2.0000001 == 2.0? is it == 2?)

nathan-contino-mongo (approved):
LGTM with some nitpicks and suggestions and mostly-not-necessary thoughts from me.

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
 globally-unique value. You can :ref:`index <ios-index-a-property>` both types,
 and use either as a :ref:`primary key <ios-specify-a-primary-key>`.
 
+.. _ios-mixed-data-type:
+
+Mixed Data Type
+---------------
+
+.. versionadded:: 10.?
+   The ``AnyRealmValue`` type is new in this version.
```
nathan-contino-mongo, line 161: Not sure if this second line really adds much here -- maybe we just stick to `..versionadded:: 10.?` on its own?

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
...
+- Float
+- Double
+- Decimal128
+- ObjectID
+- UUID
+- Bool
+- Date
+- Data
+- String
+- Object
```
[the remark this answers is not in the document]
dacharyc: The list is tricky. It's a good practice to list what a thing supports vs. what a thing does not support. Listing "can't use" requires the reader to do an additional cognitive step when deciding whether they can implement their thing. i.e. "Oh, this mixed type - can it be a bool? Look at list... there's bool." vs. "Can it be a bool? Look at list, ok, the list is what you can't use, but bool isn't there, so I guess you can use it?" It's like saying something in a positive way vs. using a double negative to say something in a positive way. Both are positive, but the latter requires the reader to do extra mental math to get to the same result.

If that is a tradeoff we want to make in order to reduce maintenance overhead, we can do that. But my default state is to err on the side of "make things easier for the reader." Does this context affect your recommendation, or would you still prefer to see this as a shorter list of "can't use" @nathan-contino-mongo ?

Re: "Object" here - in Swift, "Object" is the Realm class you use to subclass "Object" to create Realm model objects: https://github.com/realm/realm-cocoa/blob/v10.7.4/RealmSwift/Object.swift 

I think this is the equivalent of Java's "RealmObject"?

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -143,3 +152,47 @@ ObjectId is a MongoDB-specific 12-byte unique value. UUID is a 16-byte
...
+
+This :swift-sdk:`mixed data type <Enums/AnyRealmValue.html>` is 
+:ref:`indexable <ios-index-a-property>`, but you can't use it as a
```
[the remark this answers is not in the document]
dacharyc: Good point - added, thanks!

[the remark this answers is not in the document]
dacharyc: This looks like a double post of the above comment, so I'm resolving. I did add this to our list of indexable types - thanks!

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: I think this refers more to how optionals are handled in Swift, and less to do with how the mixed data type is stored in Realm. I could be wrong, though. @ericjordanmossman - I got this concept from your PR description for the mixed data type; can you clarify for us whether this line can stand in a Swifty way, or whether we should do more clarifying similar to what @nathan-contino-mongo has suggested here?

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: Good suggestion, thanks! Added a link.

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: Fair point. Eric had some examples in his PR description; I've added those here.

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: I think that if we don't say "what" was added in version, it is ambiguous. Is it the whole type? Is it some aspect of the type?

PR #1038, opened 2021-05-05
update info about when a token is auto-refreshed

MongoCaleb:
### Jira

- https://jira.mongodb.org/browse/DOCSP-15924

- https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/cht-15924/sdk/ios/advanced-guides/custom-user-data/

### Review Guidelines

### DIFF
--- a/source/includes/custom-data-may-be-stale.rst
+++ b/source/includes/custom-data-may-be-stale.rst
@@ -5,6 +5,7 @@
    of the data whenever a user refreshes their access token, such as
    when they log in. This may mean that the custom data won't
    immediately reflect changes, e.g. updates from an authentication
-   Trigger. The client SDKs automatically refresh a logged-in
-   user's access token periodically, so the user's custom data should
-   never remain stale for more than 30 minutes.
+   Trigger. If the token is not refreshed, {+service-short+} waits 
+   30 minutes and then refreshes it on the next call to the backend, so custom user 
+   data could be stale for up to 30 minutes plus the time until the next SDK 
+   call to the backend occurs.
--- a/source/sdk/ios/advanced-guides/custom-user-data.txt
+++ b/source/sdk/ios/advanced-guides/custom-user-data.txt
@@ -45,9 +45,10 @@ data, see :ref:`Update Custom User Data
    Instead, {+backend+} fetches the most recent version of custom user
    data whenever a user refreshes their :ref:`access token
    <user-sessions>`, which is used by most SDK operations that contact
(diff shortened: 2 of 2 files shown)


source/includes/custom-data-may-be-stale.rst
```diff
@@ -5,6 +5,7 @@
    of the data whenever a user refreshes their access token, such as
    when they log in. This may mean that the custom data won't
    immediately reflect changes, e.g. updates from an authentication
-   Trigger. The client SDKs automatically refresh a logged-in
-   user's access token periodically, so the user's custom data should
-   never remain stale for more than 30 minutes.
+   Trigger. If the token is not refreshed, {+service-short+} waits 
+   30 minutes and then refreshes it on the next call to the backend, so custom user
```
nlarew, line 9: Can we make it clear that the SDK isn't arbitrarily waiting 30 mins? That's just the amount of time between the token's issued_at and expired_at values. 

For each request the SDK checks if the access token needs to be refreshed and handles it if it does. I think "needs to be refreshed" means either the token expired or a request failed - would need to confirm with an SDK engineer though.

source/sdk/ios/advanced-guides/custom-user-data.txt
```diff
@@ -45,9 +45,10 @@ data, see :ref:`Update Custom User Data
    Instead, {+backend+} fetches the most recent version of custom user
    data whenever a user refreshes their :ref:`access token
    <user-sessions>`, which is used by most SDK operations that contact
-   the {+backend+} back end. {+service-short+} refreshes access tokens
-   every 30 minutes, so custom user data can be stale for no more than
-   30 minutes.
+   the {+backend+} back end. If the token is not refreshed before its default 
+   30 minute expiration time, {+service-short+} refreshes the token on the next 
+   call to the backend, so custom user data could be stale for up to 30 minutes 
+   plus the time until the next SDK call to the backend occurs.
```
nlarew, line 51: A bit past the scope of this work, but could we unify & single source this note across all the SDKs? Right now some SDKs use the include version (Node & RN) while the others have individual versions that are almost identical.

The main thing that some of the others have is SDK-specific links to the `resetCustomUserData()` method. Not sure if that's worth keeping, especially given the approach we use here for iOS where the specific method is in a separate note.

nlarew (approved):
LGTM + a potential improvement around includes

jsflax (commented):
Nice.

PR #1053, opened 2021-05-07
(DOCSP-12426): Build out Realm Studio documentation

dacharyc:
## Pull Request Info

A few notes for a reviewer:
- This PR is inconsistent about using step files for procedures. Some of the procedures with more info for each step use step files, but some of the things that are pretty simple "click this, then that" I've formatted as bullet lists. The "On this page" was getting long for some of those pages ([Modify a Realm File](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/modify-objects/) is an example of this). Also, the step files also do some formatting that is pretty... bold, and seemed a bit more than necessary for a simple procedure. I could be convinced otherwise, though. (And perhaps if we do keep the simple list in places, they should actually be `ol`s instead of `ul`s?)
- Some of the pages include a bit of "benefit" type language, but I haven't applied that consistently. I notice that different folks on our team feel differently about "marketing" type language in our docs. Since Realm Studio doesn't have a "marketing" landing page, I feel this is our main opportunity to present those benefits. I'm not super tied to it if people hate it, though.
- The [Query Objects in the Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/explore-realm-file/#query-objects-in-the-realm) section has a couple of `?` after two of the operators. I wasn't able to get those operators to work in Realm Studio, and don't know if they're not supported or if my queries were malformed. I need to do some more digging on that.

### Jira

Note: this Jira ticket makes a smaller ask, but as a team, we have discussed building out a more complete version of the Realm Studio documentation. The scope of this PR is larger than what is requested by this Jira ticket.

- https://jira.mongodb.org/browse/DOCSP-12426

- [Realm Studio Landing Page](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/)
- [Install Realm Studio](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/install/)
- [Open a Realm File](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/open-realm-file/)
- [Explore a Realm File](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/explore-realm-file/)
- [Modify a Realm File](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/modify-objects/)
- [Modify Schema](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/modify-schema/)

### Review Guidelines


source/studio/modify-objects.txt
```diff
@@ -0,0 +1,152 @@
...
+
+Change a Field Value
+~~~~~~~~~~~~~~~~~~~~
+
+You can manually change an object's field values:
+
+- Click into the field of an object
+- Enter a different value
+
+When you change a field value in this way, you can attempt to enter any
```
MongoCaleb, line 65: "you can attempt to enter any value into the field." feels a bit weird to me. 
Perhaps consider re-arranging this paragraph slightly? Like maybe:
> Realm Studio provides simple type validation, such 
as providing a list of acceptable boolean options: :guilabel:`Use "true", 
"false", "0", or "1"`, or automatically converting a date into a valid 
datetime. However, this validation may lead to unexpected behavior. For 
example, typing "123abc" into an ``int`` field simply discards the 
non-numeric characters without providing user feedback or errors. Therefore, it is up to you to ensure the data
you enter into a field value is of the correct type.

source/studio/modify-objects.txt
```diff
@@ -0,0 +1,152 @@
...
+
+When you change a field value in this way, you can attempt to enter any 
+value into the field. Realm Studio provides simple type validation, such 
+as providing a list of acceptable boolean options: :guilabel:`Use "true", 
+"false", "0", or "1"`, or automatically converting a date into a valid 
+datetime. However, this validation may lead to unexpected behavior. For 
+example, typing "123abc" into an ``int`` field simply discards the 
+non-numeric characters without providing user feedback or errors. 
+
+Keep this in mind when editing field values in Realm Studio.
```
MongoCaleb, line 73: Not sure we need the additional "warning" here.

source/studio/modify-objects.txt
```diff
@@ -0,0 +1,152 @@
...
+
+Change an Object Relationship
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can change an object's relationship with other objects using Realm
```
MongoCaleb, line 78: At first, I thought this section was about re-defining schema relationships, like "Task now contains a list of Projects, rather than Project containing a list of Tasks", but then it seems like this is about adding a relationship to a single entity / object? Yes?

source/studio/modify-objects.txt
```diff
@@ -0,0 +1,152 @@
...
+- Right-click (or control-click) the object you want to delete, and 
+  select :guilabel:`Delete selected object`.
+- You'll see a dialog asking you to confirm the action. Press the 
+  :guilabel:`Delete selected object` button to confirm.
+
+.. tip::
+   
+   If you don't see the :guilabel:`Delete selected object` option, you
+   may be viewing it in a linking relationship list view. Go directly to
+   the class where the objects are stored.
```
MongoCaleb, line 118: > where the object is stored.
Since we're talking about a specific object that needs deletion.

source/studio/modify-objects.txt
```diff
@@ -0,0 +1,152 @@
...
+- You'll see a dialog asking you to confirm the action. Press the 
+  :guilabel:`Delete selected object` button to confirm.
+
+.. tip::
+   
+   If you don't see the :guilabel:`Delete selected object` option, you
+   may be viewing it in a linking relationship list view. Go directly to
+   the class where the objects are stored. 
+
+   If you're in the class where the objects are stored, and still don't
```
MongoCaleb, line 120: same change
> class where the object is stored

source/studio/modify-objects.txt
```diff
@@ -0,0 +1,152 @@
...
+
+.. _realm-studio-save-data:
+
+Save the Data
+-------------
+
+Because {+realm+} objects are live objects, you can make changes to a 
+{+realm+} file and see them reflected in a client app in realtime. However,
+if you want to persist your changes after you close the {+realm+} file, 
+you must save the data from Realm Studio.
```
MongoCaleb, line 134: This is surprising, and I wonder if it deserves some kind of admonition?

source/studio/modify-objects.txt
```diff
@@ -0,0 +1,152 @@
...
+To save data from Realm Studio:
+
+- Go to :guilabel:`File` > :guilabel:`Save data` and select :guilabel:`JSON`
+  or :guilabel:`Local Realm`.
+- Select the filename and destination where you want to save the file.
+- Press the :guilabel:`Save` button.
+
+Once you have saved the data, you can import it back into your app. 
+
+If you save it as JSON, you can use your preferred method to import the
```
MongoCaleb, line 145: This sounds scary. Do we want to provide any guidance (even generic) on what the process would look like? I'm tempted to avoid that potential train wreck, but it would serve our users well, I think.

source/studio/modify-schema.txt
```diff
@@ -0,0 +1,140 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Realm Studio gives you tools to make additive changes to your schema. You
```
MongoCaleb, line 18: We should highlight the word additive here, I think, as everything will go to you-know-where if the user makes a non-additive change. Might even be worth an admonition.

source/studio/modify-schema.txt
```diff
@@ -0,0 +1,140 @@
...
+
+When you've completed your changes, you can :ref:`export your model 
+definitions <realm-studio-export-model-definitions>` from Realm Studio. 
+This gives you new class definitions in your language of choice, which 
+you can drop into your code.
+
+You can also :ref:`save an updated version of the {+realm+} file 
+<realm-studio-save-data>` to use in your app. When you save a {+realm+} 
+file with an updated schema, you must :ref:`perform a migration 
+<realm-studio-perform-migration>` to use it in your app.
```
MongoCaleb, line 39: Maybe my comment above is wrong? If we're doing a schema migration, then I would think this supports non-additive changes...but then, my understanding of schema changes is still rather foggy.

source/studio/open-realm-file.txt
```diff
@@ -0,0 +1,109 @@
...
+If you don't have a local {+realm+} file, you can create one from a CSV
+in Realm Studio.
+
+When you create a {+realm+} from CSV, Realm Studio infers these details:
+
+- The name of your CSV becomes the class name in the {+realm+} file
+- The first row of your CSV becomes the class property names in the {+realm+} 
+- The values of each column shape the class property types. A column of 
+  'true' and 'false' values becomes a boolean property type. A column of
+  whole numbers becomes an integer property type, and so on.
```
MongoCaleb, line 93: This is cool. I assume the default is to make the property type `string`? We should add that.

MongoCaleb (commented):
Lots of suggestions about wording; feel free to push back on any of them!

source/includes/steps-realm-studio-create-object.yaml
```diff
@@ -0,0 +1,52 @@
+title: Select a class for the new object
+ref: realm-studio-create-object-select-class
+content: |
+  The left sidebar of Realm Studio displays the classes that constitute 
+  the {+realm+}'s schema. Select the class in the model whose properties 
+  define the object you want to create.
+---
+title: Consider the object relationships in your schema
```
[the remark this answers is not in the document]
dacharyc: Good call, that makes a lot more sense. Great suggestion - thanks!

source/includes/steps-realm-studio-create-object.yaml
```diff
@@ -0,0 +1,52 @@
...
+ref: realm-studio-create-object-consider-object-relationships
+content: |
+  Consider the relationships of the objects in your schema. In a to-do
```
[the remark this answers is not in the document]
dacharyc: Hmm. That type of construct (it is important/helpful to understand...) seems unnecessarily wordy. I really just want people to think about the relationships when they are deciding where to create the object. So I tried just saying that - thoughts?

> When you create a new object, think about the relationships between the objects. For example...

source/includes/steps-realm-studio-create-object.yaml
```diff
@@ -0,0 +1,52 @@
...
+  the {+realm+}'s schema. Select the class in the model whose properties 
+  define the object you want to create.
+---
+title: Consider the object relationships in your schema
+ref: realm-studio-create-object-consider-object-relationships
+content: |
+  Consider the relationships of the objects in your schema. In a to-do 
+  app, you might have categories and tasks. The categories have a to-many
+  relationship to the tasks; one category may have many tasks. The tasks
+  have an inverse relationship to the category.
```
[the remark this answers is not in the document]
dacharyc: I'm not sure "Therefore" really fits here. I'm going on what I understand from the iOS SDK's explanation of these relationships:

> An inverse relationship links an object back to any other objects that refer to it in a defined to-one or to-many relationship. Relationship definitions are unidirectional, so you must explicitly define a property in the object's model as an inverse relationship.

If this is accurate, then "therefore" does not belong here. The tasks don't have an inverse relationship with the parent category "as a consequence of the fact" that the parent category has a to-many relationship to the tasks (which is what "therefore" would mean if we used it here); instead, these are each two uni-directional relationships that we've defined in the model. That's why I've stated them as two independent facts. If we want a linking word here, maybe "Additionally," would work better given the relationships?

I could be misunderstanding how these relationships work, though - let me know if that's the case!

source/includes/steps-realm-studio-create-object.yaml
```diff
@@ -0,0 +1,52 @@
...
+title: Consider the object relationships in your schema
+ref: realm-studio-create-object-consider-object-relationships
+content: |
+  Consider the relationships of the objects in your schema. In a to-do 
+  app, you might have categories and tasks. The categories have a to-many
+  relationship to the tasks; one category may have many tasks. The tasks
+  have an inverse relationship to the category. 
+  
+  In {+client-database+}, this type of relationship is represented as a 
+  ``list`` property linking the two objects. For our to-do app, the
```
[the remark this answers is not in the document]
dacharyc: Nope! I was just using a to-do app as a concrete example of how these relationships might work in Realm. Does rewording this help - something like:

> In the example of a to-do app, the ``Category`` class...

source/includes/steps-realm-studio-modify-object.yaml
```diff
@@ -0,0 +1,52 @@
+title: Select a class for the new object
```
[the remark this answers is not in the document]
dacharyc: Oops! Nope, this is a legacy artifact from when I was going to do something different with this section. I'm deleting it - it's not used here. Sorry to make you look at it twice!

source/studio/explore-realm-file.txt
```diff
@@ -0,0 +1,81 @@
...
+- Query objects in the {+realm+}
+- Export SDK model definitions
+
+This makes it a valuable tool for developing and debugging applications 
+backed by {+client-database+}. 
+
+View Live Objects in Realtime
+-----------------------------
+
+Realm Studio gives you the ability to view live objects while you're
```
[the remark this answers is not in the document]
dacharyc: Yeah, these pages are in each SDK. This is one of the sections I was thinking about (this and the relationship description above) when I mentioned at standup today that we don't have some of these fundamental concepts described at a higher level of abstraction, independent of each SDK. I think we could use something at a higher level that covers some of these things, but I'd have to have a think about what things it would be useful to have outside of an SDK context. (Maybe a "Fundamentals" section in "Get Started"?)

Looks like we touch on that a little bit in the Get Started section for mobile developers: https://docs.mongodb.com/realm/get-started/introduction-mobile/

And there's a glossary term for it: https://docs.mongodb.com/realm/get-started/glossary/#std-term-live-objects

I could link to either of those places, if you think it would be helpful. But if we're talking about creating fundamental content that exists outside of the SDKs, I think that may be a bigger project to identify what would be helpful and where it would live. Would we still capture that in a Jira ticket?

source/studio/explore-realm-file.txt
```diff
@@ -0,0 +1,81 @@
...
+
+To view live objects in Realm Studio, you'll need to :ref:`find your {+realm+} 
+file <realm-studio-find-a-realm-file>`. Once you know your {+realm+} file 
+path, you can open it in Realm Studio while your app is running.
+
+.. note::
+
+   Unfortunately, Android emulators do not support viewing live objects 
+   in Realm Studio. If you're developing for Android, you must download 
+   the file from the emulator before you can access it.
```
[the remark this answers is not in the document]
dacharyc: Yeah. It's really "View live objects in realtime unless you're developing for Android." But this was a feature that Alexander Stigsen specifically requested to have documented in a comment on the Jira ticket, and it had a section in the legacy Realm Studio documentation.

Do you think the Android note should come at the top, instead? And/or should it be a different type of directive, like "Important?" I'm very open to suggestions for portraying this honestly while also calling out how awesome it is for folks that can use it.

source/studio/explore-realm-file.txt
```diff
@@ -0,0 +1,81 @@
...
+
+Query Objects in the Realm
+--------------------------
+
+When you view a {+realm+} file in Realm Studio, you can query the objects
```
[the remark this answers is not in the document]
dacharyc: Much less awkward - thanks for the suggestion!

source/studio/explore-realm-file.txt
```diff
@@ -0,0 +1,81 @@
...
+accepts the :ref:`{+client-database+} JavaScript query syntax<node-client-query-engine>`. 
+You can query using:
+
+- Comparison operators
+- Logical operators
+- String operators
+- Aggregate operators?
+- Set operators?
+
+You can also access a query reference document from within Realm Studio
```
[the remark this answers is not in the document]
dacharyc: That was awkwardly worded. I just switched it to:

> You can also view a reference document that details the query options from within Realm Studio. Press the :guilabel:`?` button next to the query bar.

Fun fact - this link in the product actually points at a deprecated legacy document! I'll add that to my list of bugs to file from documenting Realm Studio.

source/studio/modify-objects.txt
```diff
@@ -0,0 +1,152 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+Realm Studio gives you the ability to not just view, but modify objects
```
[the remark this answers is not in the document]
dacharyc: Good call! I was writing this overview section just after the "explore" overview section, and was definitely feeling some "but wait, there's more!" energy. But this is better.

source/studio/modify-objects.txt
dacharyc: I like it - thank you!

source/studio/modify-objects.txt
dacharyc: Yes.

I updated the section head to "Change an Object's Relationships" - do you think that is clearer? Or would you recommend changing the section head entirely or clarifying in another way?

In my head, it was clear, because everything on this page is about modifying an object, while there's a different page for modifying the schema. But that context is not necessarily immediately apparent to a reader! Open to suggestions if you think there's a better way to communicate about this.

source/studio/modify-schema.txt
dacharyc: The Realm Studio UI does not give you the ability to make non-additive changes to the schema. (I.e. there is no option to delete properties, change their type, delete classes, etc.) So I don't think it's really a worry that a user will make a non-additive change. But I'm not opposed to applying special formatting here if you think it's valuable to make it more apparent. With that context, what do you think?

source/studio/modify-schema.txt
dacharyc: So here's what happens when I make changes to a schema in Realm Studio...

If I make changes to the objects, my client app chugs along quite happily and will stop and start again and again with no issue.

If I make changes to the schema, such as adding a property, and then try to use that realm file with my app, Xcode throws this error: 
> Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=io.realm Code=1 "Provided schema version 0 is less than last set version 2."

The iOS SDK docs say: 
> Schema versions are integers that you may include in the realm configuration when you open a realm. If a client application does not specify a version number when it opens a realm then the realm defaults to version 0.

And then points me to this page about performing a schema migration: https://docs.mongodb.com/realm/sdk/ios/examples/modify-an-object-schema/#std-label-ios-modify-an-object-schema

I could be wrong about a migration being required; maybe just specifying a new version when you open the realm file is sufficient? But I don't think we provide an example of that anywhere in the docs. I do see a `schemaVersion` configuration property in the RealmSwift reference docs - I'll try playing with that a bit more on Monday AM.

source/studio/open-realm-file.txt
dacharyc: Added!

source/studio/modify-objects.txt
dacharyc: I will poke this more on Monday AM and come back to it. There is a transaction option in a drop-down menu, and I wonder if I make changes while in a transaction, it will persist the data even without saving the realm. My example realm file is borked now from making schema changes, so I'll create a new one on Monday AM and come back to this.

source/includes/steps-realm-studio-create-object.yaml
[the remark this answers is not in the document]
MongoCaleb: solid!

source/includes/steps-realm-studio-create-object.yaml
```diff
@@ -0,0 +1,52 @@
...
+  ``Category`` class has a ``list`` property whose value is all of the 
+  related ``Task`` objects.
+
+  In the Realm Studio UI, you would select the ``Category`` class. Click 
+  into the ``[list-of-task]`` property for the category where you want to 
+  add the task - for example, "Work". You'll see all the task objects that 
+  have an inverse relationship with the Work category. 
+
```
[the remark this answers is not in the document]
MongoCaleb: Screenshots are a good thing! I mean, they're a pain to keep fresh, but I think that risk.additional work is well worth it.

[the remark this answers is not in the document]
dacharyc: Just took some screenshots of this flow and will add them to the page.

source/includes/steps-realm-studio-create-object.yaml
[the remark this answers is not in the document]
dacharyc: I follow the logic, but I believe in the iOS SDK, anyway, you have to explicitly define the inverse relationship using `LinkingObjects` - it isn't implicitly created just because the Category has a to-many reference to it. 

My resistance to using a word like "thus" or "therefore" is that it implies the inverse relationship is automagically created by the category's to-many relationship with the task. In reality, the reader needs to take an additional step to define the inverse relationship. Maybe explicitly stating that in the next paragraph after I describe the ``list`` property that represents the to-many relationship would make this clearer?

source/studio/modify-objects.txt
dacharyc: Updated the "Save" section. What do we think now?

source/studio/modify-schema.txt
dacharyc: @MongoCaleb Here's where I am:

- I've commented further on the inverse relationship convo. I think it can stand as it is, but am not opposed to adding some more text in the next paragraph to clarify things for a reader if you think that's valuable.
- I've added screenshots for the [object relationship section](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/modify-objects/#consider-object-relationships) - do we think this is clearer?
- I updated the [Save the Data](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/modify-objects/#save-the-data) section to reflect my additional spelunking, and will put in a ticket to research further the "transaction" option that's in the UI. But I think this section can now stand as-is for publishing, with later refinement.
- I added an admonition about migrating the Realm file after [modifying the schema](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-12426/studio/modify-schema/), because that was kind of a pain! But I did make an additive change, verify that I couldn't use the {+realm+} and had to do a migration, and successfully completed a migration that enabled me to use the {+realm+} again

I think this is ready to go, pending any further comments you might have.

MongoCaleb (approved):
Looks great!

PR #1065, opened 2021-05-11
Node and React Native Data Types

mongomoe:
## Note to the reviewer
- This PR only needs a sanity check review, as the pages of this PR have already been copy and technical reviewed in other PRs.

### Jira

- https://jira.mongodb.org/browse/DOCSP-16176

- [Node.js SDK > Data Types Section](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-data-types-node-rn/sdk/node/data-types)
- [React Native SDK > Data Types Section](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new-data-types-node-rn/sdk/react-native/data-types)

### Review Guidelines


source/sdk/node/data-types/field-types.txt
```diff
@@ -0,0 +1,29 @@
...
+- ``int`` maps to the JavaScript :mdn:`Number <Web/JavaScript/Reference/Global_Objects/Number>` type. Internally, {+client-database+} stores ``int`` with 64 bits.
+- ``float`` maps to the JavaScript :mdn:`Number <Web/JavaScript/Reference/Global_Objects/Number>` type. Internally, {+client-database+} stores ``float`` with 32 bits.
+- ``double`` maps to the JavaScript :mdn:`Number <Web/JavaScript/Reference/Global_Objects/Number>` type. Internally, {+client-database+} stores ``double`` with 64 bits.
+- ``string`` maps to the JavaScript :mdn:`String <Web/JavaScript/Reference/Global_Objects/Boolean>` type.
+- ``decimal128`` for high precision numbers.
+- ``objectId`` maps to BSON :manual:`ObjectId </reference/method/ObjectId/>` type.
+- ``data`` maps to the JavaScript :mdn:`ArrayBuffer <Web/JavaScript/Reference/Global_Objects/ArrayBuffer>` type.
+- ``date`` maps to the JavaScript :mdn:`Date <Web/JavaScript/Reference/Global_Objects/Date>` type.
+- ``list`` maps to the JavaScript :mdn:`Array <Web/JavaScript/Reference/Global_Objects/Array>` type. You can also specify that a field contains a list of a primitive value type by appending ``[]`` to the type name.
```
nathan-contino-mongo, line 24:
```suggestion
- ``list`` maps to the JavaScript :mdn:`Array <Web/JavaScript/Reference/Global_Objects/Array>` type. You can also specify that a field contains a list of a primitive value types by appending ``[]`` to the type name.
```

source/sdk/node/data-types/sets.txt
```diff
@@ -0,0 +1,112 @@
...
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. versionadded:: 10.5.0-beta.1
+
+Overview
+--------
+A ``{+service-short+} Set`` is a special object that allows you to store a
```
nathan-contino-mongo, line 19: It looks like the SDK actually uses the term ``Realm.Set`` (with a period in the middle) to refer to this type. Since "Realm Set" (with a space, no period) isn't actually a type in the SDK, I would suggest either:
- removing the monospace formatting from the term (reserved for literal type names)
- switching the space to a period so we're using the literal name of the type in the SDK.

If the SDK uses the name "Realm Set" with a space in the middle, then you can safely ignore this. But since most languages don't support type names containing spaces, I think we ought to change this.

source/sdk/node/data-types/mixed.txt
```diff
@@ -0,0 +1,65 @@
...
+- bool
+- int 
+- float 
+- double 
+- string
+- Date 
+- Data
+- UUID
+- Set
+- null
```
nathan-contino-mongo, line 30: Since these are all literal type names, can we monospace format these?

source/sdk/node/data-types/mixed.txt
```diff
@@ -0,0 +1,65 @@
...
+
+   The mixed data type is indexable, but you can't use it as a primary key.
+   Because null is a permitted value, you can't declare a Mixed property as
+   optional.
+
+Realm Object Models
+-------------------
+To :ref:`set a property of your object model
+<node-define-a-realm-object-schema>` as mixed, set the property's type to
+``"mixed"``.
```
nathan-contino-mongo, line 42: nitpick: can we move the quotes here outside of the monospace formatting, since it's not part of the literal type name?

source/sdk/node/data-types/mixed.txt
```diff
@@ -0,0 +1,65 @@
...
+- Data
+- UUID
+- Set
+- null
+
+.. note::
+
+   The mixed data type is indexable, but you can't use it as a primary key.
+   Because null is a permitted value, you can't declare a Mixed property as
+   optional.
```
nathan-contino-mongo, line 36: This sentence is a bit confusing to me -- the second bit doesn't really follow from the first. If a Mixed property *can* have the value null, wouldn't it essentially always be optional? Perhaps a second clarifying sentence to explain this could be helpful.

source/sdk/node/data-types/uuid.txt
```diff
@@ -0,0 +1,40 @@
...
+
+``UUID`` (Universal Unique Identifier) is a 16-byte :wikipedia:`unique value
+<Universally_unique_identifier>`. You can use ``UUID`` as an identifier for
+objects. ``UUID`` is :ref:`indexable <node-indexes>` and you can use it as a
+:ref:`primary key <node-primary-keys>`.
+
+.. note:: Using UUID Instead of ObjectId
+
+   In general, you can use ``UUID`` in any field that you need a unique
+   identifier. Using ``UUID`` might be particularly useful if you are migrating
```
nathan-contino-mongo, line 27:
```suggestion
   In general, you can use ``UUID`` for any fields that function as a unique
   identifier. Using ``UUID`` might be particularly useful if you are migrating
```
"any field that you need a unique identifier" doesn't sound like correct English to me, but I can't quite put my finger on why.

source/sdk/node/data-types/uuid.txt
```diff
@@ -0,0 +1,40 @@
...
+   data not stored in MongoDB since it is likely that your object's unique
+   identifiers are already of a ``UUID`` type.
```
nathan-contino-mongo, line 29: No callout here to the fact that you should use ``ObjectId`` for collections of data that already exist in MongoDB? That feels like a significant omission since without that statement we basically never talk about when you would actually use ObjectId.

source/sdk/node/data-types/mixed.txt
```diff
@@ -0,0 +1,65 @@
...
+
+Realm Object Models
+-------------------
+To :ref:`set a property of your object model
+<node-define-a-realm-object-schema>` as mixed, set the property's type to
```
nathan-contino-mongo, line 41:
```suggestion
<node-define-a-realm-object-schema>` as ``Mixed``, set the property's type to
```
The Set and UUID pages use monospace in their equivalents of this sentence, so I feel we should do the same here for consistency.

source/sdk/node/data-types/uuid.txt
```diff
@@ -0,0 +1,40 @@
...
+
+Usage
+----- 
+To define a property as a ``UUID``, set its type to the string ``"uuid"`` in
+your :ref:`object model <node-object-schemas>`. Create a {+service-short+}
+object within a write transaction. To set any unique identifier properties of
```
nathan-contino-mongo, line 35:
```suggestion
your :ref:`object model <node-object-schemas>`. To set any unique identifier properties of
```
I'm not sure if the command/instruction to "create a Realm object within a write transaction" actually fits here. Readers already know that you're talking about a theoretical object model that contains a UUID field, so personally I'd say that the instructions to set properties with ``new UUID()`` should suffice to explain how to generate new UUID instances and put them in such an object.

nathan-contino-mongo (approved):
LGTM with some nitpicks and suggestions on the Node docs. All of my comments apply twofold: once to the Node page I commented on, but also to the React Native equivalent of the page that just has different anchor link names.

source/sdk/node/data-types/field-types.txt
mongomoe: Fixed locally 👍 ,  ignoring the commit suggestion. Changed it to "...contains a list of primitive value types by appending ``[]``"

source/sdk/node/data-types/sets.txt
mongomoe: The SDK term is Realm.Set, but I've defined it as "Realm Set" to indicate that it the term is a proper noun and to avoid confusion with the JavaScript set term.

nathan-contino-mongo: OK, but you shouldn't monospace the term if it isn't a literal code term. You can bold it if you want to highlight it as a key term, but monospace is reserved for code references.

source/sdk/node/data-types/mixed.txt
mongomoe: fixed 👍

source/sdk/node/data-types/mixed.txt
mongomoe: Optional means that it can or cannot contain a value, and null is technically a value. Going to avoid changing this for now, since this sentence has already been copy + tech reviewed, but will add the suggestion to a list of iterative updates.

source/sdk/node/data-types/uuid.txt
mongomoe: Fair point, I've added the sentence "Alternatively, using ``ObjectId`` might be useful for a collection of data that already exists in MongoDB."

source/sdk/node/data-types/uuid.txt
mongomoe: Decided to ignore this one, as it doesn't hurt to re-explain that it needs to be within a write transaction. Personally, I've made the mistake of not writing in a write transaction, and then realized once I got the error, so I think a short reminder sentence might be helpful.

nathan-contino-mongo: Fair enough, but just note that we don't tell users to modify fields in write transactions in the very same paragraph, even though you need to use a write transaction for all modifications.

Perhaps you've missed the point of my comment, though, which was that instructing users to create an instance of the object is probably redundant since they'll know that they need an instance to modify a property.

source/sdk/node/data-types/sets.txt
mongomoe: I see, didn't realize monospace was reserved for code references. Changing it to bold now.

source/sdk/node/data-types/uuid.txt
mongomoe: 'in the very same paragraph' as in the last sentence of that paragraph?: 'Alternatively, set a string to new UUID() to set the unique identifier property to a specific value.'? That sentence is meant to show a specific usage of the new UUID constructor, rather than how to use UUIDs as a whole within the realm and within a write object (which the previous sentences teach). 


Even if it's a bit redundant, I think that sentence still works and could be helpful to readers who may not know or who forget that you need a write transaction and an instance of an object to set or modify a uuid property. 

Let me know if I'm misunderstanding your concern/suggestion though.

PR #1079, opened 2021-05-13
(DOCSP-14573): set data type

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-14573

- [Node.js > Data Types > Sets](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14573-Set-Data-Type/sdk/node/data-types/sets/)

### Review Guidelines

### DIFF
--- a/examples/node/Examples/data-types.js
+++ b/examples/node/Examples/data-types.js
@@ -367,7 +367,7 @@ describe("Node.js Data Types", () => {
 
     // :code-block-start: remove-all-items-from-set
     realm.write(() => {
-      // clear all data from the items slot of the hunter by calling `set.clear()` in a write transaction
+      // clear all data from the inventory slot of the hunter by calling `set.clear()` in a write transaction
       hunter.inventory.clear();
     });
     // :code-block-end:
--- a/examples/node/Examples/data-types.js
+++ b/examples/node/Examples/data-types.js
@@ -356,6 +356,15 @@ describe("Node.js Data Types", () => {
     // :code-block-end:
     expect(hunterHasCompletedLevelThree).toBe(false);
 
+    // :code-block-start: remove-specific-item-from-set
+    realm.write(() => {
+      // remove the compass from link's inventory by calling `set.delete()` within a write transaction
+      link.inventory.delete("compass");
+    });
+
+    // :code-block-end:
+    expect(link.inventory.has("compass")).toBe(false);
+
     // :code-block-start: remove-all-items-from-set
     realm.write(() => {
       // clear all data from the items slot of the hunter by calling `set.clear()` in a write transaction
--- a/examples/node/Examples/data-types.js
+++ b/examples/node/Examples/data-types.js
@@ -299,4 +299,83 @@ describe("Node.js Data Types", () => {
     // close the realm
     realm.close();
   });
+  test("should work with the Set data type", async () => {
+    // :code-block-start: define-set-objects
+    const characterSchema = {
+      name: "Character",
+      primaryKey: "_id",
+      properties: {
+        _id: "objectId",
+        name: "string",
+        levelsCompleted: "int<>",
+        inventory: "string<>",
+      },
+    };
+    // :code-block-end:
+    const realm = await Realm.open({
+      schema: [characterSchema],
+    });
+
+    // :code-block-start: create-set-objects
+    let link, hunter;
+    realm.write(() => {
+      link = realm.create("Character", {
+        _id: new BSON.ObjectId(),
+        name: "Link",
+        inventory: ["elixir", "compass", "glowing shield"],
+        levelsCompleted: [4, 9],
+      });
+      hunter = realm.create("Character", {
+        _id: new BSON.ObjectId(),
+        name: "Hunter",
+        inventory: ["estus flask", "gloves", "rune"],
+        levelsCompleted: [1, 2, 5, 24],
+      });
+    });
+    // :code-block-end:
+
+    expect(link.inventory.has("elixir")).toBe(true);
+    expect(hunter.inventory.has("gloves")).toBe(true);
+
+    // :code-block-start: add-items-to-set
+    realm.write(() => {
+      link.inventory.add("hammer");
+      link.levelsCompleted.add(32);
+    });
+    // :code-block-end:
+
+    expect(link.inventory.size).toBe(4);
+    expect(link.levelsCompleted.size).toBe(3);
+
+    // :code-block-start: check-if-set-has-items
+    // check if the hunter has completed level 3 by calling the `set.has()` method
+    const hunterHasCompletedLevelThree = hunter.levelsCompleted.has(3);
+    console.log(
+      `Is level three completed by the hunter: ${hunterHasCompletedLevelThree}`
+    );
+    // :code-block-end:
+    expect(hunterHasCompletedLevelThree).toBe(false);
+
+    // :code-block-start: remove-all-items-from-set
+    realm.write(() => {
+      // clear all data from the items slot of the hunter by calling `set.clear()` in a write transaction
+      hunter.inventory.clear();
+    });
+    // :code-block-end:
+
+    // :code-block-start: check-set-size
+    // check how many items the hunter has in his inventory through the `set.size` property
+    const hunterInventorySize = hunter.inventory.size;
+    console.log(`The hunter has ${hunterInventorySize} inventory items`);
+    // :code-block-end:
+    expect(hunter.inventory.size).toBe(0);
+
+    // delete the object specifically created in this test to keep tests idempotent
+    realm.write(() => {
+      realm.delete(link);
+      realm.delete(hunter);
+    });
+    // close the realm
+    realm.close();
+  });
 });
--- a/examples/node/Examples/data-types.js
+++ b/examples/node/Examples/data-types.js
@@ -317,72 +317,72 @@ describe("Node.js Data Types", () => {
     });
 
     // :code-block-start: create-set-objects
-    let link, hunter;
+    let playerOne, playerTwo;
     realm.write(() => {
-      link = realm.create("Character", {
+      playerOne = realm.create("Character", {
         _id: new BSON.ObjectId(),
-        name: "Link",
+        name: "PlayerOne",
         inventory: ["elixir", "compass", "glowing shield"],
         levelsCompleted: [4, 9],
       });
-      hunter = realm.create("Character", {
+      playerTwo = realm.create("Character", {
         _id: new BSON.ObjectId(),
-        name: "Hunter",
+        name: "PlayerTwo",
         inventory: ["estus flask", "gloves", "rune"],
         levelsCompleted: [1, 2, 5, 24],
       });
     });
     // :code-block-end:
 
-    expect(link.inventory.has("elixir")).toBe(true);
-    expect(hunter.inventory.has("gloves")).toBe(true);
+    expect(playerOne.inventory.has("elixir")).toBe(true);
+    expect(playerTwo.inventory.has("gloves")).toBe(true);
 
     // :code-block-start: add-items-to-set
     realm.write(() => {
-      link.inventory.add("hammer");
-      link.levelsCompleted.add(32);
+      playerOne.inventory.add("hammer");
+      playerOne.levelsCompleted.add(32);
     });
     // :code-block-end:
 
-    expect(link.inventory.size).toBe(4);
-    expect(link.levelsCompleted.size).toBe(3);
+    expect(playerOne.inventory.size).toBe(4);
+    expect(playerOne.levelsCompleted.size).toBe(3);
 
     // :code-block-start: check-if-set-has-items
-    // check if the hunter has completed level 3 by calling the `set.has()` method
-    const hunterHasCompletedLevelThree = hunter.levelsCompleted.has(3);
+    // check if the playerTwo has completed level 3 by calling the `set.has()` method
+    const playerTwoHasCompletedLevelThree = playerTwo.levelsCompleted.has(3);
     console.log(
-      `Is level three completed by the hunter: ${hunterHasCompletedLevelThree}`
+      `Is level three completed by the playerTwo: ${playerTwoHasCompletedLevelThree}`
     );
     // :code-block-end:
-    expect(hunterHasCompletedLevelThree).toBe(false);
+    expect(playerTwoHasCompletedLevelThree).toBe(false);
 
     // :code-block-start: remove-specific-item-from-set
     realm.write(() => {
-      // remove the compass from link's inventory by calling `set.delete()` within a write transaction
-      link.inventory.delete("compass");
+      // remove the compass from playerOne's inventory by calling `set.delete()` within a write transaction
+      playerOne.inventory.delete("compass");
     });
 
     // :code-block-end:
-    expect(link.inventory.has("compass")).toBe(false);
+    expect(playerOne.inventory.has("compass")).toBe(false);
 
     // :code-block-start: remove-all-items-from-set
     realm.write(() => {
-      // clear all data from the inventory slot of the hunter by calling `set.clear()` in a write transaction
-      hunter.inventory.clear();
+      // clear all data from the inventory slot of the playerTwo by calling `set.clear()` in a write transaction
+      playerTwo.inventory.clear();
     });
     // :code-block-end:
 
     // :code-block-start: check-set-size
-    // check how many items the hunter has in his inventory through the `set.size` property
-    const hunterInventorySize = hunter.inventory.size;
-    console.log(`The hunter has ${hunterInventorySize} inventory items`);
+    // check how many items the playerTwo has in his inventory through the `set.size` property
+    const playerTwoInventorySize = playerTwo.inventory.size;
+    console.log(`The playerTwo has ${playerTwoInventorySize} inventory items`);
     // :code-block-end:
-    expect(hunter.inventory.size).toBe(0);
+    expect(playerTwo.inventory.size).toBe(0);
 
     // delete the object specifically created in this test to keep tests idempotent
     realm.write(() => {
-      realm.delete(link);
-      realm.delete(hunter);
+      realm.delete(playerOne);
+      realm.delete(playerTwo);
     });
     // close the realm
     realm.close();
--- a/source/examples/generated/node/data-types.codeblock.remove-all-items-from-set.js
+++ b/source/examples/generated/node/data-types.codeblock.remove-all-items-from-set.js
@@ -1,4 +1,4 @@
 realm.write(() => {
-  // clear all data from the inventory slot of the characterTwo by calling `set.clear()` in a write transaction
+  // clear all data from the inventory slot of the characterTwo by calling `Realm.Set.clear()` in a write transaction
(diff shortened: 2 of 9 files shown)


source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
 .. versionadded:: 10.5.0-beta.1
 
 Overview
---------
\ No newline at end of file
+--------
+A {+service-short+} set is a special object that allows you to store a
```
ccho-mongodb, line 19: I wonder if we should make it conspicuous that "Realm Set" is a term, either through capitalization similar to Javascript "Set" or using a bold format. 
After reading the first couple of paragraphs, I'm also curious as to whether you can use Javascript Sets instead, or if I can only use Realm Sets since the title of the page is Sets.

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+collection of unique values. {+service-short+} sets work similarly to JavaScript
+:mdn:`Sets <Web/JavaScript/Reference/Global_Objects/Set>`, except they must be a single
+type and can only be modified within a write-transaction.
```
ccho-mongodb, line 22: 1) I think Sets contain types since Set is a type itself, so I would suggest "they must contain a single type".
2) I think it's "write transaction" rather than "write-transaction", unless the latter is terminology specific to Realm. You also use "write transaction" on line 49 of this PR.

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+
+.. note:: {+service-short+} Sets Do Not Guarantee Traversal Order
+   
+   When using a ``forEach()`` loop or other :mdn:`iteration methods
+   <Web/JavaScript/Reference/Global_Objects/Set#iteration_methods>` to iterate
+   through a loop, {+service-short+} sets may be in a different order than
```
ccho-mongodb, line 28: Nitpick:
You probably don't iterate through a loop, so maybe you meant to say something like "to access the contents of the set in a loop" or "to traverse the Set in a loop".
Also, I think it's the contents of the sets that are in a different order rather than "sets may be in a different order".

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+   originally written to. If you require an ordered version of your set, you
+   must implement that order yourself. You can do this by creating an array of
```
ccho-mongodb, line 30: Nitpick:
I don't think you can "implement that order", so maybe you meant "implement that ordering".

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+   the set's values in the insertion order.
```
ccho-mongodb, line 31: I think it could be helpful to show a short example if this is something users need to do often enough.

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+   
+.. _node-define-set-objects:
+
+Realm Object Models
+-------------------
+To define a property's values as a {+service-short+} set, specify the data type
+followed by a less than and equal sign. For instance, for a set made of integer
```
ccho-mongodb, line 38: "less than and equal sign" doesn't match your example. I expected to see "<=", but in the example, I see "<>"

source/sdk/node/data-types/sets.txt
MongoCaleb, line 19: "special object" sounds weird and doesn't really tel us anything. Since everyone is familiar with Arrays, perhaps a start like this?

> A {+service-short+} set is an object that stores a collection of values. A Set is similar to an Array, except:
 - Each value in a Set must be unique,
 - Arrays are ordered by index, while sets are iterable and unordered (this isn't quite right...)

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
 .. versionadded:: 10.5.0-beta.1
 
 Overview
---------
\ No newline at end of file
+--------
+A {+service-short+} set is a special object that allows you to store a
+collection of unique values. {+service-short+} sets work similarly to JavaScript
```
MongoCaleb, line 20: > sets work similarly to JavaScript
to 
> sets are based on JavaScript Sets, but are limited to a single value type and can only be modified...

source/sdk/node/data-types/sets.txt
MongoCaleb, line 28: > through a loop,
to 
> through a set,

source/sdk/node/data-types/sets.txt
MongoCaleb, line 31: The last sentence here is confusing. Are you saying the user should create an Array that gets updated every time the set is updated, and insert into the array in the order I want?

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+
+.. literalinclude:: /examples/generated/node/data-types.codeblock.create-set-objects.js
+    :language: javascript
+
+
+.. _node-add-items-to-set:
+
+Add Items to a Set
+------------------
+To add items to a set, pass the new value to the ``set.add()`` method within a write transaction.
```
ccho-mongodb, line 61: Nitpick:
The plural "items" doesn't match with singular "new value". Instead, consider "To add an item to a set, pass the new value..."

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+   through a loop, {+service-short+} sets may be in a different order than
+   originally written to. If you require an ordered version of your set, you
+   must implement that order yourself. You can do this by creating an array of
+   the set's values in the insertion order.
+   
+.. _node-define-set-objects:
+
+Realm Object Models
+-------------------
+To define a property's values as a {+service-short+} set, specify the data type
```
MongoCaleb, line 37: Can this be shortened to:

> To define a property type as a Set,  specify the data type you want in the set, followed by ``<>``

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+To add items to a set, pass the new value to the ``set.add()`` method within a write transaction.
+
+.. literalinclude:: /examples/generated/node/data-types.codeblock.add-items-to-set.js
+    :language: javascript
+
+.. _node-check-if-set-has-items:
+
+Check if a Set has Specific Items
+---------------------------------
+To find out if a set has a particular value, pass the value to the ``set.has()`` method. The
```
ccho-mongodb, line 70: Nitpick: could be more specific "has" -> "contains"

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+values, specify ``"int<>"``. 
+
+.. literalinclude:: /examples/generated/node/data-types.codeblock.define-set-objects.js
+    :language: javascript
+
+.. _node-create-set-objects:
+
+Create an Object With a Set
+---------------------------
+To create an object with a property that value is of the set data type, create a
```
MongoCaleb, line 48: The wording here is confusing me. 
Does this work?
> To create an object that has one or more set properties, you must create the object within a write transaction. You then initialize the set-type properties as you would an array, by passing the initial values  in brackets (``[]``)

source/sdk/node/data-types/sets.txt
MongoCaleb, line 70: consider changing "To find out" to "To determine"

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+
+.. _node-check-if-set-has-items:
+
+Check if a Set has Specific Items
+---------------------------------
+To find out if a set has a particular value, pass the value to the ``set.has()`` method. The
+``set.has()`` method will return true if the set contains the value specified.
+
+.. literalinclude:: /examples/generated/node/data-types.codeblock.check-if-set-has-items.js
+    :language: javascript
```
ccho-mongodb, line 74: Is the indentation correct? I think the colon should match with the "l" in "literalinclude" above. I know snooty won't error, but occasionally it's updated to be more strict so this could potentially  cause some issues in the future. Applies to all instances.

MongoCaleb (changes requested):
Overall comment: I have no idea WHY I would use a set over an array. Perhaps JS developers already know that? Also, there is no section on how to retrieve a value from a set, and I assume that's because you can't -- you have to iterate? Is it worth any additional info about that?

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,88 @@ Sets - Node.js SDK
...
+
+.. literalinclude:: /examples/generated/node/data-types.codeblock.add-items-to-set.js
+    :language: javascript
+
+.. _node-check-if-set-has-items:
+
+Check if a Set has Specific Items
+---------------------------------
+To find out if a set has a particular value, pass the value to the ``set.has()`` method. The
+``set.has()`` method will return true if the set contains the value specified.
```
ccho-mongodb, line 71: Nitpick: personal preference, but the notation "set.has()" is a little confusing as to whether "set" is the class or a variable. Maybe "has()" or "mySet.has()" would be more clear. Applies to all instances.

examples/node/Examples/data-types.js
```diff
@@ -299,4 +299,92 @@ describe("Node.js Data Types", () => {
...
+        inventory: "string<>",
+      },
+    };
+    // :code-block-end:
+    const realm = await Realm.open({
+      schema: [characterSchema],
+    });
+
+    // :code-block-start: create-set-objects
+    let link, hunter;
```
ccho-mongodb, line 320: Nitpick: this variable name could potentially be difficult for some to follow, especially if they are not familiar with Zelda and/or English is not their first language, since it is similar to the technical term "link".

ccho-mongodb (approved):
There are some changes I think you should make and some nitpicks, but I'll approve since Caleb is also reviewing this.

source/sdk/node/data-types/sets.txt
kneth: I agree: A Realm Set is not a 1:1 mapping of a JavaScript Set (but heavily inspired by it).

examples/node/Examples/data-types.js
mongomoe: Fair point, about link the name vs the technical term. I've altered to 'playerOne' and 'playerTwo' (for consistency)

source/sdk/node/data-types/sets.txt
mongomoe: Fair point, I've altered it to be "``Realm Set``" consistently throughout the doc to indicate it is a term. 

As for indicating that it is not a 1:1 mapping, I've included the line: "Realm sets work similarly to JavaScript Sets, except they must be a single type and can only be modified within a write-transaction"

source/sdk/node/data-types/sets.txt
mongomoe: Thanks, fixed both 👍

source/sdk/node/data-types/sets.txt
mongomoe: Fixed 👍

source/sdk/node/data-types/sets.txt
mongomoe: That's a good idea, I've added it to a jira ticket that contains a list of proposed updates to this doc

source/sdk/node/data-types/sets.txt
mongomoe: Fixed 👍

source/sdk/node/data-types/sets.txt
mongomoe: Fixed 👍

source/sdk/node/data-types/sets.txt
mongomoe: Good point, I've altered it and similar methods to "<Set>.methodName()"

source/sdk/node/data-types/sets.txt
mongomoe: Good point, fixed 👍

source/sdk/node/data-types/sets.txt
mongomoe: Good catch, that's a typo that is supposed to be "less than and greater than sign". Fixed now 👍

source/sdk/node/data-types/sets.txt
mongomoe: Thanks 👍 , I've fixed it to 'to traverse the loop' and 'the contents of the Realm Set  may be in a different order'

source/sdk/node/data-types/sets.txt
mongomoe: 👍

source/sdk/node/data-types/sets.txt
mongomoe: That works, fixed now 👍

source/sdk/node/data-types/sets.txt
mongomoe: Essentially, that's it. I've altered it to be a bit clearer:

"You can do this by creating an array from the set, using Array.from(mySet), and updating that array when the set has been modified through a change listener."

source/sdk/node/data-types/sets.txt
mongomoe: Changed the sentence to: "When using a forEach() loop or other iteration method to traverse the set in a loop, the content of the Realm Set may be in a different order than originally written to."

source/sdk/node/data-types/sets.txt
mongomoe: took this suggestion almost verbatim

source/sdk/node/data-types/sets.txt
mongomoe: I'm down for being a bit clearer on the advantages of Set. I do want to keep it a bit more subtle and not necessarily bring up array for the following reasons:
- JavaScript developers should probably know the differences between set and array. Explaining the differences between the two would be explaining JS, not explaining how to use Realm with JS. For those that aren't familiar with the differences, I've linked out to the mdn docs on set.

As for showing when to use set, I've added some info in the intro, the new intro reads:
```
"A Realm Set is a special object that allows you to store a collection of unique values. Realm Sets are based on JavaScript Sets, but can only contain values of a single type and can only be modified within a write transaction. Sets allow you to perform math operations such as finding the union, intersection, or difference between two sets. To learn more about performing these operations, see the MDN docs for Implementing basic set operations."
```

source/sdk/node/data-types/sets.txt
mongomoe: How about a combo of the two phrasings:

"To create an object with a ``{+service-short+} Set`` property, you must create the object within a write transaction. When defining your Realm object, initialize the set by passing an empty array or an array with your initial values."

mongomoe:
@MongoCaleb per your top level comment: "Overall comment: I have no idea WHY I would use a set over an array. Perhaps JS developers already know that? Also, there is no section on how to retrieve a value from a set, and I assume that's because you can't -- you have to iterate? Is it worth any additional info about that?"

- Added some info about the benefits of using a Set, but I'm opposed to directly mentioning Arrays in this doc since I think it may confuse readers. Fundamentally they are two different data types with unique use cases, see more about my reasoning for not directly mentioning Arrays here: https://github.com/mongodb/docs-realm/pull/1079#discussion_r634546906
- About retrieving a value from a set, you would check if a set has a value (shown in [Node.js > Sets # Check if a Set has Specific Items](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-14573-Set-Data-Type/sdk/node/data-types/sets/#check-if-a-set-has-specific-items) since sets are not retrievable via any index like arrays. Alternatively, you can iterate through the values in a set with a forEach loop, but I don't think we should show a snippet of that since iteration of a set using a forEach loop might confuse readers and imply that a set is similar to an array and since we don't guarantee traversal order.

source/examples/generated/node/data-types.codeblock.remove-all-items-from-set.js
```diff
@@ -0,0 +1,4 @@
+realm.write(() => {
+  // clear all data from the inventory slot of the characterTwo by calling `set.clear()` in a write transaction
```
kneth, line 2:
```suggestion
  // clear all data from the inventory slot of the characterTwo by calling `Realm.Set.clear()` in a write transaction
```

source/examples/generated/node/data-types.codeblock.remove-specific-item-from-set.js
```diff
@@ -0,0 +1,5 @@
+realm.write(() => {
+  // remove the compass from characterOne's inventory by calling `set.delete()` within a write transaction
```
kneth, line 2:
```suggestion
  // remove the compass from characterOne's inventory by calling `Realm.Set.delete()` within a write transaction
```

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,96 @@ Sets - Node.js SDK
...
+these operations, see the MDN docs for :mdn:`Implementing basic set operations
+<Web/JavaScript/Reference/Global_Objects/Set#implementing_basic_set_operations>`
+
+.. note:: ``{+service-short+} Sets`` Do Not Guarantee Traversal Order
+   
+   When using a ``forEach()`` loop or other :mdn:`iteration method
+   <Web/JavaScript/Reference/Global_Objects/Set#iteration_methods>` to traverse
+   the set in a loop, the content of the ``{+service-short+} Set`` may be in a
+   different order than originally written to. If you require an ordered version
+   of your set, you must implement that ordering yourself.If you require an
```
kneth, line 34:
```suggestion
   of your set, you must implement that ordering yourself. If you require an
```

source/sdk/node/data-types/sets.txt
```diff
@@ -15,4 +15,96 @@ Sets - Node.js SDK
...
+   ordered version of your set, you must implement that order yourself. You can
+   do this by creating an array from the set, using :mdn:`Array.from(mySet)
```
kneth, line 36: The spread operator (`...`) will also work.

kneth (approved):

source/sdk/node/data-types/sets.txt
mongomoe: Fixed 👍

source/sdk/node/data-types/sets.txt
fronck: The JS Set does not currently support the math operations.  Its implementation has been kept as close as possible to JS's own set type.

fronck: Maybe it's worth specifying here that Realm Set, unlike JS's own set, does not guarantee iteration in insertion order.

PR #1082, opened 2021-05-14
updating thread info

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-14585

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/14585/sdk/dotnet/advanced-guides/threading/

### Review Guidelines

### DIFF
--- a/source/sdk/dotnet/advanced-guides/threading.txt
+++ b/source/sdk/dotnet/advanced-guides/threading.txt
@@ -29,6 +29,13 @@ maintainable multithreaded code that avoids issues like
 deadlocking and race conditions. {+service+} aims to
 simplify this for you.
 
+.. important:: SynchronizationContext Threads
+
+   Throughout this page, we refer to the "main thread" (or "UI thread") and 
+   "background threads". To be more accurate, any mention of the main or UI thread 
+   refers to **any thread with a ``SynchronizationContext``**, while those threads 
+   without a SynchronizationContext are considered background threads.
+
 .. _dotnet-threading-three-rules:
 
 Three Rules to Keep in Mind
@@ -150,11 +157,13 @@ the current background thread.
 .. note::
 
    In general, since {+realm+} writes are very fast, we recommend using the 
-   ``Write()`` method for interface-driven changes where only a few values 
-   change at a time. The ``WriteAsync()`` method is useful in scenarios where 
-   there are hundreds or thousands of changes (e.g. fetching objects from the 
-   network, or batch-updating multiple objects) and performing the changes 
-   synchronously might lead to interface jitters.
+   ``Write()`` method when only a few values change at a time (like responding 
+   to user interactions.) Conversely, the ``WriteAsync()`` method is useful in 
+   scenarios where there are hundreds or thousands of changes. For example, 
+   fetching objects from the network, or batch-updating multiple objects, can be 
+   time-consuming, and performing the changes synchronously might lead to 
+   interface jitters and a poor user experience. Using ``WriteAsync()`` 
+   prevents this.
 
 .. _dotnet-frozen-objects:
 


source/sdk/dotnet/advanced-guides/threading.txt
```diff
@@ -118,14 +110,51 @@ several options depending on your use case:
...
+   by that version, leading to larger file sizes. 
+
+.. _dotnet-async-writes:
+
+Asynchronous Writes
+~~~~~~~~~~~~~~~~~~~
+
+The :dotnet-sdk:`WriteAsync() <reference/Realms.Realm.html#Realms_Realm_WriteAsync_System_Action_Realms_Realm__>`
+method provides a simple way to offload the UI thread by performing writes in a 
+background thread. When run on the UI thread, {+service-short+} creates a
```
dacharyc, line 143: Nit: this "When run on the UI thread" sentence is long with a lot of handoffs. I wonder if we can make this easier to parse by breaking it up into bullets? A flow diagram would be best, but that's a lot more work. Maybe something like:

> When you run `WriteAsync()` on a UI thread:
> - {+service-short+} creates a temporary {+realm+} instance, which it passes to the Lambda function
> - The Lambda function performs the write, disposes of the temporary instance, and calls `Realm.Refresh()`
> - `Realm.Refresh()` updates the {+realm+} on the UI thread

But that really just makes me want a way to represent that visually...

source/sdk/dotnet/advanced-guides/threading.txt
```diff
@@ -118,14 +110,51 @@ several options depending on your use case:
...
+write, disposes of the temporary instance, and then calls ``Realm.Refresh()`` on 
+the UI thread to update the {+realm+}. When ``WriteAsync()`` is 
+run in a background thread, {+service-short+} forwards the call to ``Write()`` on 
+the current background thread. 
+
+.. note::
+
+   In general, since {+realm+} writes are very fast, we recommend using the 
+   ``Write()`` method for interface-driven changes where only a few values 
+   change at a time. The ``WriteAsync()`` method is useful in scenarios where
```
dacharyc, line 154: Nit: this is a long sentence, and the (e.g. ...) in the middle of it makes it harder to follow. Maybe we could break it up a little. Maybe something like:

> Use the `WriteAsync` method when a large volume of changes could lead to UI performance issues. For example, say your app fetches hundreds or thousands of objects from the network, or batch-updates many objects. Performing these changes synchronously could lead to interface jitters, which `WriteAsync` avoids.

That's still a little awkward, but maybe slightly easier to follow.

dacharyc (approved):
Nice improvement! Just a couple of nits where some sentences feel a little overly complicated.

source/sdk/dotnet/advanced-guides/threading.txt
```diff
@@ -118,14 +110,51 @@ several options depending on your use case:
...
+the current background thread. 
+
+.. note::
+
+   In general, since {+realm+} writes are very fast, we recommend using the 
+   ``Write()`` method for interface-driven changes where only a few values 
+   change at a time. The ``WriteAsync()`` method is useful in scenarios where 
+   there are hundreds or thousands of changes (e.g. fetching objects from the 
+   network, or batch-updating multiple objects) and performing the changes 
+   synchronously might lead to interface jitters.
```
papafe, line 157: What you wrote in this section is true and probably will be enough for most of the people. The thing is that it's not 100% accurate when you say the difference of write async between UI threads and other threads. That is, the difference is actually between threads that have a `SynchronizationContext` (like the UI thread) and threads that don't have it. 
I suppose (but I could be wrong) that for the greatest majority of the developers using realm this does not make a difference, as the UI thread will be the one with a `SynchronizationContext`, and so the explanation is correct.

papafe:
Looks fine to me!

source/sdk/dotnet/advanced-guides/threading.txt
nirinchev: I think it's valuable to mention the synchronizationcontext part at one point or another. We can then clarify that throughout the text "main or UI thread" means a thread with a synchronizationcontext installed while "background thread" is one without. This will be true for the vast majority of use cases, but it's important to document that there's a way to make a background thread behave like a UI one (which is what we do in the https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/14585/sdk/dotnet/async-console/ guide).

nirinchev (approved):
The changes look good to me, though I agree with Ferdinando's point. On an unrelated note, I  know you didn't change that now, but I noticed in the Frozen Objects section on this page, we have `Once frozen, it is not possible to unfreeze an object. You can use isFrozen()...` on line 186. In the .NET SDK, that is a property called `IsFrozen` rather than a method, so might want to drive-by fix that as well.

PR #1084, opened 2021-05-15
(DOCSP-16021): Asynchronous API local-only where appropriate, coroutines & extensions

nathan-contino-mongo:
## Pull Request Info
- please just ignore the .idea files, those will be deleted when we merge in the new datatypes PR

### Jira

- https://jira.mongodb.org/browse/DOCSP-16021

- [Asynchronous API -- entire page](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16021/sdk/android/fundamentals/async-api/)

### Review Guidelines


ianpward:
@nathan-contino-mongo One thing I'd probably note in the Coroutines example is that the toFlow() extension method uses frozen realm objects passed across thread boundaries with a link to the frozen objects doc

examples/android/local/app/src/androidTest/java/com/mongodb/realm/examples/java/AsyncTest.java
```diff
@@ -0,0 +1,130 @@
...
+                @Override
+                public void onSuccess(@NotNull Realm realm) {
+                    Log.v("EXAMPLE", "Successfully fetched realm instance.");
+                }
+                public void onError(Exception e) {
+                    Log.e("EXAMPLE", "Failed to get realm instance: " + e);
+                }
+            });
+            // :code-block-end:
+            expectation.fulfill();
```
cmelchior, line 38: Not exactly sure what this does, but it looks like the test completes before either callback triggers, but that might not matter here?

source/examples/generated/android/local/AsyncTest.codeblock.coroutines.kt
```diff
@@ -0,0 +1,21 @@
+// open a realm asynchronously
+Realm.getInstanceAsync(config, object : Realm.Callback() {
+    override fun onSuccess(realm: Realm) {
+        Log.v("EXAMPLE", "Successfully fetched realm instance")
+
+        CoroutineScope(Dispatchers.IO).launch {
+            // asynchronous transaction
+            realm.executeTransactionAwait { transactionRealm: Realm ->
```
cmelchior, line 8: This is not going to work. The Realm instance is being created on the UI thread (I assume), and then accessed inside an IO thread, so it will throw. Instead. you can do (and replace

```
coroutineScope(Dispatchers.MAIN).launch {
    realm.executeTransactionAwait(Dispatchers.IO).launch {
        ....
    }
}

```

source/examples/generated/android/local/AsyncTest.codeblock.realm-async-task.kt
```diff
@@ -0,0 +1,18 @@
+// using class instances for transaction, success, error
+realm.executeTransactionAsync(Realm.Transaction { transactionRealm ->
+        val item: Item = transactionRealm.createObject(Item::class.java)
```
cmelchior, line 3:
```suggestion
        val item: Item = transactionRealm.createObject<Item>()
```

There is a Kotlin extension function for this.

source/sdk/android/fundamentals/async-api.txt
```diff
@@ -22,45 +22,187 @@ In the callback, you can check if the request executed
...
+~~~~~~~~~~~~
+
+When you query {+service+} App Services like :ref:`Functions
+<android-call-a-function>` and :ref:`user authentication
+<android-authenticate-users>`, asynchronous calls accept a final
+parameter of type  :java-sdk:`App.Callback
+<io/realm/mongodb/App.Callback.html>`.
+You can handle this callback with a lambda function that accepts a
+single parameter of type :java-sdk:`App.Result
+<io/realm/mongodb/App.Result.html`. To retrieve a returned values from
```
cmelchior, line 73: Is a `>`  missing somewhere?

cmelchior (approved):

examples/android/local/app/src/androidTest/java/com/mongodb/realm/examples/java/AsyncTest.java
nathan-contino-mongo: yep, since we're just setting onSuccess/onError the `expectation` bit just ensures that the test doesn't complete before configuring the triggers. We could make this more robust by actually triggering a success/error event, but that seemed like overkill in this case.

source/examples/generated/android/local/AsyncTest.codeblock.coroutines.kt
nathan-contino-mongo: ah, good point. I'm not sure how this hasn't exploded when I ran it, but thanks for pointing this out -- i'll fix it right away.

dacharyc (approved):
I think the content itself is great. But the way it's written now, it kind of reads like a laundry list of Realm-specific method calls and terms, and you have to read each section to find the one that applies to what you want to do. I suggest changing the section headers to reflect when a developer might need to use it. So, for example:

- `Realm.Callback` becomes `Open a Realm`
- `App.Callback` becomes `Query MongoDB Realm App Services`
- `RealmAsyncTask` becomes `Execute Transactions`
- `RealmResults` becomes `Query Realm`
- `Coroutines` becomes `Async Flows in Kotlin`

Or whatever is the most appropriate descriptor for what the user is trying to do with each of these async API elements.

nathan-contino-mongo:
@dacharyc the intent right now is for this page to serve as a catch-all for "how to i deal with this kind of async behavior in the SDK", so when folks query for the async part of those things this page will (hopefully) show up and help them. So I'm a bit torn on renaming the sections since the sections don't actually tell you how to do those things, they just talk about the async *version* of those things. So I'll sit on it!

PR #1089, opened 2021-05-18
(DOCSP-16146): Add run a password reset function to Android

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-16146

- [Manage Email/Password Users -> Run a Password Reset Function](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16146/sdk/android/examples/email-password-users/#run-a-password-reset-function):  new section
- [Manage Email/Password Users -> Reset a User's Password](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16146/sdk/android/examples/email-password-users/#reset-a-user-s-password): new verbiage before email reset section

### Review Guidelines

### DIFF
--- a/examples/android/sync/app/src/androidTest/java/com/mongodb/realm/examples/kotlin/ManageEmailPasswordTest.kt
+++ b/examples/android/sync/app/src/androidTest/java/com/mongodb/realm/examples/kotlin/ManageEmailPasswordTest.kt
@@ -4,6 +4,8 @@ import android.util.Log
 import com.mongodb.realm.examples.Expectation
 import com.mongodb.realm.examples.RealmTest
 import com.mongodb.realm.examples.YOUR_APP_ID
+import com.mongodb.realm.examples.getRandomPartition
+
 import io.realm.mongodb.App
 import io.realm.mongodb.AppConfiguration
 import org.junit.Test
@@ -109,17 +111,16 @@ class ManageEmailPasswordTest : RealmTest() {
 
     @Test
     fun runAPasswordResetFunc() {
-        val random = Random()
-        val email = "test" + random.nextInt(100000)
-        val password = "testtest"
+        val email = getRandomPartition()
         val expectation = Expectation()
         activity!!.runOnUiThread {
             val appID: String = YOUR_APP_ID // replace this with your App ID
             val app = App(AppConfiguration.Builder(appID).build())
+
+            // :code-block-start: run-password-reset-func
             val newPassword = "newFakePassword"
             val args = arrayOf("security answer 1", "security answer 2")
 
-            // :code-block-start: run-password-reset-func
             app.emailPassword.callResetPasswordFunctionAsync(email, newPassword, args) {
                 if (it.isSuccess) {
                     Log.i("EXAMPLE", "Successfully reset the password for $email")
--- a/examples/android/sync/app/src/androidTest/java/com/mongodb/realm/examples/kotlin/ManageEmailPasswordTest.kt
+++ b/examples/android/sync/app/src/androidTest/java/com/mongodb/realm/examples/kotlin/ManageEmailPasswordTest.kt
@@ -106,4 +106,32 @@ class ManageEmailPasswordTest : RealmTest() {
         }
         expectation.await()
     }
+
+    @Test
+    fun runAPasswordResetFunc() {
+        val random = Random()
+        val email = "test" + random.nextInt(100000)
+        val password = "testtest"
+        val expectation = Expectation()
+        activity!!.runOnUiThread {
+            val appID: String = YOUR_APP_ID // replace this with your App ID
+            val app = App(AppConfiguration.Builder(appID).build())
+            val newPassword = "newFakePassword"
+            val args = arrayOf("security answer 1", "security answer 2")
+
+            // :code-block-start: run-password-reset-func
+            app.emailPassword.callResetPasswordFunctionAsync(email, newPassword, args) {
+                if (it.isSuccess) {
+                    Log.i("EXAMPLE", "Successfully reset the password for $email")
+                } else {
+                    Log.e("EXAMPLE", "Failed to reset the password for $email: $it.error")
+                    // :hide-start:
+                    expectation.fulfill()
+                    // :hide-end:
+                }
+            }
+            // :code-block-end:
+        }
+        expectation.await()
+    }
 }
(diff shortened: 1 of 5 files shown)


examples/android/sync/app/src/androidTest/java/com/mongodb/realm/examples/kotlin/ManageEmailPasswordTest.kt
```diff
@@ -106,4 +106,32 @@ class ManageEmailPasswordTest : RealmTest() {
         }
         expectation.await()
     }
+
+    @Test
+    fun runAPasswordResetFunc() {
+        val random = Random()
+        val email = "test" + random.nextInt(100000)
```
nathan-contino-mongo, line 113: there's actually a function in RealmTest to [get a random partition string](https://github.com/mongodb/docs-realm/blob/master/examples/android/sync/app/src/androidTest/java/com/mongodb/realm/examples/RealmTest.kt#L99) -- not required at all, but it could be a nice way to make this test less brittle in the future!

examples/android/sync/app/src/androidTest/java/com/mongodb/realm/examples/kotlin/ManageEmailPasswordTest.kt
```diff
@@ -106,4 +106,32 @@ class ManageEmailPasswordTest : RealmTest() {
...
+        val password = "testtest"
+        val expectation = Expectation()
+        activity!!.runOnUiThread {
+            val appID: String = YOUR_APP_ID // replace this with your App ID
+            val app = App(AppConfiguration.Builder(appID).build())
+            val newPassword = "newFakePassword"
+            val args = arrayOf("security answer 1", "security answer 2")
```
nathan-contino-mongo, line 120: I would suggest pulling the definition of `args` into the code block, because some users might get confused that they have to pass in an array even if they're only passing one arg in.

nathan-contino-mongo: (dittor for the Java version of this)

source/sdk/android/examples/email-password-users.txt
```diff
@@ -128,3 +143,41 @@ methods:
    reset email, you can use a :ref:`custom password reset email subject
    <auth-send-a-password-reset-email>` containing a `deep link
    <https://developer.android.com/training/app-links/deep-linking>`__.
+
+Run a Password Reset Function
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you configure your app to :ref:`run a password reset function 
+<auth-run-a-password-reset-function>`, you'll define the function that 
+should run when you call :java-sdk:`callResetPasswordFunction() <io/realm/mongodb/auth/EmailPasswordAuth.html#callResetPasswordFunction-java.lang.String-java.lang.String-java.lang.Object...->`
```
nathan-contino-mongo, line 152:
```suggestion
runs when you call :java-sdk:`callResetPasswordFunction() <io/realm/mongodb/auth/EmailPasswordAuth.html#callResetPasswordFunction-java.lang.String-java.lang.String-java.lang.Object...->` 
```
One would hope that it's certain that the function runs when you call this method. Of course, maybe it's better to temper user expectations 😉

source/sdk/android/examples/email-password-users.txt
```diff
@@ -128,3 +143,41 @@ methods:
...
+
+Run a Password Reset Function
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you configure your app to :ref:`run a password reset function 
+<auth-run-a-password-reset-function>`, you'll define the function that 
+should run when you call :java-sdk:`callResetPasswordFunction() <io/realm/mongodb/auth/EmailPasswordAuth.html#callResetPasswordFunction-java.lang.String-java.lang.String-java.lang.Object...->` 
+or :java-sdk:`callResetPasswordFunctionAsync() <io/realm/mongodb/auth/EmailPasswordAuth.html#callResetPasswordFunctionAsync-java.lang.String-java.lang.String-java.lang.Object:A-io.realm.mongodb.App.Callback->`
+from the SDK. This function can take a username, a password, and any number 
+of additional arguments. You can use these arguments to specify details
```
nathan-contino-mongo, line 155: Since we pass in the "any number of additional arguments" as a single array, I wonder if it would be best to say "an array of additional arguments" instead to make that third parameter's type clearer.

nathan-contino-mongo (approved):
LGTM with minor nits only. Fantastic job! (as long as the tests pass in CI!)

source/sdk/android/examples/email-password-users.txt
dacharyc: Hmm. The point I wanted to make with this line was that the resetFunc is a custom function that the user defines in the Realm Sync backend. When the user selects the option to run a password reset func in the backend UI, they'll select the function that they want to map it to (from a drop-down menu of available functions) when they make the `callResetPasswordFunction()` call in the SDK. That's different than the password reset email option which basically runs stuff automagically in the backend when you call that method.

I suppose we could say that all backend functions are custom, user-defined functions, anyway, and we don't actually need this part. But I was thinking of communicating how it differs from the send email option. What do you think - should I drop this entirely, @nathan-contino-mongo ?

source/sdk/android/examples/email-password-users.txt
dacharyc: The SDK reference docs call this param an [Object](https://docs.mongodb.com/realm-sdks/java/latest/io/realm/mongodb/auth/EmailPasswordAuth.html#callResetPasswordFunction-java.lang.String-java.lang.String-java.lang.Object...-) (or [Object[]](https://docs.mongodb.com/realm-sdks/java/latest/io/realm/mongodb/auth/EmailPasswordAuth.html#callResetPasswordFunctionAsync-java.lang.String-java.lang.String-java.lang.Object:A-io.realm.mongodb.App.Callback-)  in the Async call)

How would you represent that in this example?

nathan-contino-mongo: Ah, it looks like the sync version uses [variadic arguments](https://www.geeksforgeeks.org/variable-arguments-varargs-in-java/) (basically passing in an array as separate objects that are thrown into an array when they're passed into the function) but the async version uses an array of objects. So I think this is fine and we can just lay the burden on the API reference to explain that (admittedly slightly confusing) distinction.

PR #1092, opened 2021-05-19
(DOCSP-15481): sync errors common errors update

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-15481

- [Sync Errors # Common Sync Protocol Errors ](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-15481-Sync-Errors-Common-Errors-Update/sync/errors/#common-sync-protocol-errors)

### Review Guidelines


source/sync/errors.txt
```diff
@@ -16,13 +16,47 @@ Overview
...
+
+   * - ErrorDivergingHistories
+     - This error indicates that you need to :ref:`perform a client reset <sync-errors-client-resets>`. It may occur if a user connected to a specific identity goes offline, and {+sync-short+} is :doc:`terminated and reenabled </reference/terminating-and-reenabling-realm-sync>` while the user is offline, invalidating that identity. 
+
+   * - ErrorIllegalRealmPath
+     - You may see the message "attempted to bind on illegal realm partition: expected partition to have type objectId but found string". Ensure that the partition value passed to the client matches the partition value in your {+sync-short+} configuration.
+   
+   * - ErrorOtherError
+     - This error indicates an internal communication failure between {+sync+} and {+atlas+} and occurs when you have encountered the limit on the storage quota of a Free tier cluster. If you are encountering this error, you can :doc:`upgrade your cluster tier </reference/upgrade-shared-cluster/>`.
```
ccho-mongodb, line 49: Would the second sentence conform better to the other items if it didn't use "If you are encountering this error" (present continuous) instead of "If you encounter this error" (present)?

ccho-mongodb (approved):
Looks good. I think there are a few items that might make it easier to understand and more uniform.

source/sync/errors.txt
```diff
@@ -16,13 +16,47 @@ Overview
 --------
 
 While you develop applications using {+sync+}, you may run into errors. This
-section details some common errors. 
+section details some common errors and how to log and handle them. 
 
+.. _sync-errors-common-protocol-errors:
+
+Common Sync Protocol Errors
```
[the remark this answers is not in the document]
mongomoe: Fair point, fixed now 👍

source/sync/errors.txt
```diff
@@ -16,13 +16,47 @@ Overview
...
-section details some common errors. 
+section details some common errors and how to log and handle them. 
 
+.. _sync-errors-common-protocol-errors:
+
+Common Sync Protocol Errors 
+---------------------------
+The table below lists some common :doc:`{+sync-short+} protocol </sync/protocol>` errors
+and how to handle them. To see if {+service+} has encountered any of these
+errors, check your :doc:`{+sync-short+} logs </logs>`.
```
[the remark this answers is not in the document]
mongomoe: Yeah that does work better, altered it to your suggestion now 👍

source/sync/errors.txt
```diff
@@ -16,13 +16,47 @@ Overview
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 70
+
+   * - Error
```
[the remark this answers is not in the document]
mongomoe: Changing it to error name now 👍

source/sync/errors.txt
```diff
@@ -16,13 +16,47 @@ Overview
...
+     - Description and How to Handle the Error
```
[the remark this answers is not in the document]
mongomoe: Personally, I think "How to Fix it" is more vague then "How to Handle the Error" (which implies that it can either be dealt with  in code or through the UI, which is the case)

source/sync/errors.txt
```diff
@@ -16,13 +16,47 @@ Overview
...
+
+   * - ErrorBadClientFileIdent
+     - This error indicates that you need to :ref:`perform a client reset <sync-errors-client-resets>`. It occurs when the client is using a {+realm+} file that the server cannot access after :doc:`terminating and re-enabling {+sync-short+} </reference/terminating-and-reenabling-realm-sync>`.
```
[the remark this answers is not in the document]
mongomoe: Fair point, I've swapped the order 👍 . The new phrasing reads:

"This error occurs when the client is using a {+realm+} file that the server cannot access after :doc:`terminating and re-enabling {+sync-short+} </reference/terminating-and-reenabling-realm-sync>`. 

To solve this error, :ref:`perform a client reset <sync-errors-client-resets>`."

source/sync/errors.txt
```diff
@@ -16,13 +16,47 @@ Overview
...
+     - Description and How to Handle the Error
+
+   * - ErrorBadClientFileIdent
+     - This error indicates that you need to :ref:`perform a client reset <sync-errors-client-resets>`. It occurs when the client is using a {+realm+} file that the server cannot access after :doc:`terminating and re-enabling {+sync-short+} </reference/terminating-and-reenabling-realm-sync>`. 
+
+   * - ErrorClientFileUserMismatch
+     - This error occurs when a user tries to synchronize data to a {+realm+} that has already terminated {+sync+} or when you try to synchronize data to a {+realm+} using a different user than the one used to open the synced realm. If you encounter this error, delete the local realm file on the client and re-initialize {+sync+}.
+
+   * - ErrorDivergingHistories
+     - This error indicates that you need to :ref:`perform a client reset <sync-errors-client-resets>`. It may occur if a user connected to a specific identity goes offline, and {+sync-short+} is :doc:`terminated and reenabled </reference/terminating-and-reenabling-realm-sync>` while the user is offline, invalidating that identity.
```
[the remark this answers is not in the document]
mongomoe: Fixed 👍

source/sync/errors.txt
```diff
@@ -16,13 +16,47 @@ Overview
...
+
+   * - ErrorClientFileUserMismatch
+     - This error occurs when a user tries to synchronize data to a {+realm+} that has already terminated {+sync+} or when you try to synchronize data to a {+realm+} using a different user than the one used to open the synced realm. If you encounter this error, delete the local realm file on the client and re-initialize {+sync+}.
+
+   * - ErrorDivergingHistories
+     - This error indicates that you need to :ref:`perform a client reset <sync-errors-client-resets>`. It may occur if a user connected to a specific identity goes offline, and {+sync-short+} is :doc:`terminated and reenabled </reference/terminating-and-reenabling-realm-sync>` while the user is offline, invalidating that identity. 
+
+   * - ErrorIllegalRealmPath
+     - You may see the message "attempted to bind on illegal realm partition: expected partition to have type objectId but found string". Ensure that the partition value passed to the client matches the partition value in your {+sync-short+} configuration.
```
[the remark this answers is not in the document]
mongomoe: Clarified the wording a bit now

source/sync/errors.txt
mongomoe: It would 👍

source/sync/errors.txt
```diff
@@ -16,13 +16,46 @@ Overview
...
+   :widths: 30 70
+
+   * - Error Name
+     - Description and How to Handle the Error
+
+   * - ErrorBadClientFileIdent
+     - This error occurs when the client is using a {+realm+} file that the server cannot access after :doc:`terminating and re-enabling {+sync-short+} </reference/terminating-and-reenabling-realm-sync>`. To solve this error, :ref:`perform a client reset <sync-errors-client-resets>`.
+
+   * - ErrorClientFileUserMismatch
+     - This error occurs when a user tries to synchronize data to a {+realm+} that has already terminated {+sync+} or when you try to synchronize data to a {+realm+} using a different user than the one used to open the synced realm. If you encounter this error, delete the local realm file on the client and re-initialize {+sync+}.
```
mpobrien, line 39: nit: I think a reader could take "re-initialize sync" to mean either "restart sync on the device" or "reinitialize sync in your app (server-side)" so I would suggest making it more explicit about which is being referred to here.

source/sync/errors.txt
```diff
@@ -16,13 +16,46 @@ Overview
...
+
+   * - ErrorDivergingHistories
+     - This error may occur if a user connected to a specific identity goes offline, and {+sync-short+} is :doc:`terminated and reenabled </reference/terminating-and-reenabling-realm-sync>` while the user is offline, invalidating that identity. To solve this error, that you need to :ref:`perform a client reset <sync-errors-client-resets>`.
+
+   * - ErrorIllegalRealmPath
+     - The message "attempted to bind on illegal realm partition: expected partition to have type objectId but found string" is shown in the {+sync+} logs when an ErrorIllegalRealmPath is thrown. To solve this error, ensure that the partition value passed to the client matches the partition value in your {+sync-short+} configuration.
+   
+   * - ErrorOtherError
+     - This error indicates an internal communication failure between {+sync+} and {+atlas+} and occurs when you have encountered the limit on the storage quota of a Free tier cluster. If you encounter this error, you can :doc:`upgrade your cluster tier </reference/upgrade-shared-cluster/>`.
```
mpobrien, line 48: This isn't quite accurate because `ErrorOtherError` is kind of a catch-all error code which could get triggered for a bunch of different reasons, and the exact details would be specified in the full error message - it might have nothing to do with storage quotas or cluster size. 
For this error code, we should collect the most common error messages that are associated with this error code, and give some detail on each individual one.

source/sync/errors.txt
```diff
@@ -16,13 +16,46 @@ Overview
...
+
+   * - ErrorPermissionDenied
+     - This error occurs when a user requests data they don't have access to. Check your :doc:`{+sync-short+} permissions </sync/permissions/>` logic to ensure users have access to the correct data to prevent this error.
```
mpobrien, line 51: I would make this one more precise by explicitly mentioning the sync permissions config. 
In other words, this error means that the user has tried to open a realm (for read access) but their sync permissions for the `read` expression evaluated to `false`.
Or, it could mean that the user opened the realm and was assigned read-only permissions (`read` evaluated to `true`, but `write` evaluated to `false`), but then attempted to modify data in the partition.

mpobrien (commented):
Some notes.
Also, the following two are not protocol errors per se, but they are relatively common and will appear in the UI logs so we should be sure to add a section about them, somewhere:

- `MongoEncodingError` - this occurs when a mongo (i.e. non-sync) writer modifies a document in mongodb such that it no longer conforms to the app's schema. This puts the document into an "unsyncable" state such that subsequent writes to that object won't be propagated to devices until the document is corrected to match the schema again. If the data set reaches the system limit of 100k "unsyncable" docs, sync will terminate with a fatal error. The customer should avoid this by making sure that all other workloads writing to mongoDB are conforming to the schema. Individual documents can be restored back to syncable by issuing another write to make it match schema again. When this error occurs, the message in the logs will include detail about which document had the error and which field didn't match schema. Example log output:

```
Logs:
[
  "Change Event Type: update",
  "Namespace: foo.Locations",
  "Partition: 123456",
  "Document ID: ObjectID(\"609a6eb979c03a221efe3249\")",
  "Detailed Error: could not convert MongoDB value to Realm payload for { table: Locations, path: a }, value=43 : cannot transform int value for non-int property of type string",
]
```
- `TranslatorCorrectiveErasure` - this occurs when sync is attempting to propagate a change from a device back to the mongo collection, but mongo rejected the write. This usually occurs due to a duplicate key exception, meaning that two objects in two different partitions were created with the _same_ primary key. Customers should avoid this by using `ObjectID` or `UUID` as their partition key, or designing their app such that primary keys are unique even across partitions.

source/sync/errors.txt
nlarew: Not actually sure which interpretation is correct here - do you happen to know if it's "try again on the client" or "terminate and re-enable sync on the server"?

source/sync/errors.txt
nlarew: Changed this to be more generic for now but still mention storage limits as an example. We can add more cases for this as they come up / in a future project.

nlarew:
@mpobrien Added a new [MongoDB Translator Errors](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-15481-Sync-Errors-Common-Errors-Update/sync/errors/#mongodb-translator-errors) section with those two errors + made some minor tweaks in other places.

source/sync/errors.txt
mpobrien: > Pretty sure it's "If you encounter this error, delete the local realm file on the client and then re-open the realm to download the correct data.

Yeah, that's correct. Wiping the data on the client side will fix it, resetting on the server end isn't necessary.

source/sync/errors.txt
```diff
@@ -16,89 +16,163 @@ Overview
...
-
-.. _sync-error-logging:
-
-Error Logging
--------------
-When actively troubleshooting an issue, set the logging level to ``debug`` or
-``trace`` to diagnose errors while developing your application. When running in a
-production environment, you will want to decrease the logging level for improved
-performance.
+section details some common errors and how to log and handle them.
```
mpobrien, line 19: "how to log" seems unusual here?

source/sync/errors.txt
```diff
@@ -16,89 +16,163 @@ Overview
...
+     - Description and How to Handle the Error
+
+   * - ErrorBadClientFileIdent
+     - This error occurs when the client is using a {+realm+} file that the
+       server cannot access after :doc:`terminating and re-enabling {+sync-short+}
+       </reference/terminating-and-reenabling-realm-sync>`. To solve this error,
+       :ref:`perform a client reset <sync-errors-client-resets>`.
+
+   * - ErrorClientFileUserMismatch
+     - This error occurs when a user tries to synchronize data to a {+realm+}
```
mpobrien, line 49: nit: I think the description under `ErrorDivergingHistories` actually does a really good job of explaining the same situation that the first part of this paragraph is getting at; I'd suggest just reusing part of that here?

source/sync/errors.txt
```diff
@@ -16,89 +16,163 @@ Overview
...
+
+   * - ErrorDivergingHistories
+     - This error may occur if a user connected to a specific identity goes
+       offline, and {+sync-short+} is :doc:`terminated and reenabled
+       </reference/terminating-and-reenabling-realm-sync>` while the user is
+       offline, invalidating that identity. To solve this error, that you need to
+       :ref:`perform a client reset <sync-errors-client-resets>`.
+
+   * - ErrorIllegalRealmPath
+     - The message "attempted to bind on illegal realm partition: expected
```
mpobrien, line 63: I think this one should be reworded to match the pattern that the others use, like "This error occurs when..."
And maybe specify that `expected partition to have type objectId but found string` is just an _example_ of what the details look like when the error is encountered (the actual types in the message could be number, string, objectid, uuid, etc.)

mpobrien (approved):
Couple minor notes, but looks solid 👍

PR #1110, opened 2021-05-27
Add code to warn if user == null but don't require it

MongoCaleb:
## Pull Request Info

-https://jira.mongodb.org/browse/DOCSP-16259

### DIFF
--- a/tutorial/dotnet/realm-tutorial-dotnet/ProjectPage.xaml.cs
+++ b/tutorial/dotnet/realm-tutorial-dotnet/ProjectPage.xaml.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.ObjectModel;
 using System.Linq;
+using System.Threading;
 using RealmDotnetTutorial.Models;
 using Realms;
 using Realms.Sync;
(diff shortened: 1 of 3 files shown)


tutorial/dotnet/realm-tutorial-dotnet/ProjectPage.xaml.cs
```diff
@@ -68,7 +70,29 @@ private async AsyncTask LoadProjects()
...
                 // :state-uncomment-end:
                 // :code-block-end:
-                if (user != null) SetUpProjectList();
+
+                if (user == null)
+                {
+                    // Either the trigger hasn't completed yet, has failed,
+                    // or was never created on the backend
+                    // So let's wait a few seconds and check again...
+                    Thread.Sleep(5000);
```
cbush, line 79: So on a new user, you always wait 5 seconds?

cbush (approved):

tutorial/dotnet/realm-tutorial-dotnet/ProjectPage.xaml.cs
MongoCaleb: changed as discussed.

PR #1113, opened 2021-05-29
remove flipper

ianpward:
## Pull Request Info

why are we making our docs users go through an extra step to run the app? the final app should be as close to clone and go as possible - it seems odd to include this since we do not support flpper 

### Jira

### Review Guidelines


mongomoe (approved):
LGTM 👍 

@ianpward about your description: "why are we making our docs users go through an extra step to run the app? the final app should be as close to clone and go as possible - it seems odd to include this since we do not support flpper"

Flipper is just included with 'react-native init', we can remove it from this tutorial but even without it users will still have to 'pod install', so this PR doesn't decrease any steps from the user's perspective.

PR #1119, opened 2021-06-02
(DOCSP-15923): Call out "delta sync"

cbush:
### Jira

https://jira.mongodb.org/browse/DOCSP-15923

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/delta/sync/protocol/#changeset

### Review Guidelines

### DIFF
--- a/source/sync/protocol.txt
+++ b/source/sync/protocol.txt
@@ -53,9 +53,12 @@ and resolve conflicting changesets before sending them to connected clients.
 
 .. note:: Delta Sync
 
-   To maximize bandwidth in mobile networks, the {+sync+} service only
-   transmits the changes (deltas) at the field level. The service
-   compresses the deltas on the fly with :wikipedia:`zlib <Zlib>` compression.
+   When you make a change to a synced object, {+service-short+} does not
+   re-upload the entire object. Instead, {+service-short+} sends only
+   the difference ("delta") between before and after. The service
+   compresses the deltas with :wikipedia:`zlib <Zlib>` compression. This
+   reduces network load, which is especially useful in mobile network
+   conditions.
 
 .. _operational-transformations:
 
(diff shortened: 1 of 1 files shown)


source/sync/protocol.txt
```diff
@@ -49,6 +49,14 @@ client (including changes in a synced MongoDB cluster) at any time and uses an
 operational transformation algorithm to serialize changes into a linear order
 and resolve conflicting changesets before sending them to connected clients.
 
+.. _delta-sync:
+
+.. note:: Delta Sync
+
+   To maximize bandwidth in mobile networks, the {+sync+} service only
+   transmits the changes (deltas) at the field level. The service
+   compresses the deltas on the fly with zlib compression.
```
mongomoe, line 58:
```suggestion
   compresses the deltas on the fly with :wikipedia:`zlib <Zlib>` compression.
```

source/sync/protocol.txt
```diff
@@ -49,6 +49,14 @@ client (including changes in a synced MongoDB cluster) at any time and uses an
 operational transformation algorithm to serialize changes into a linear order
 and resolve conflicting changesets before sending them to connected clients.
 
+.. _delta-sync:
+
+.. note:: Delta Sync
+
+   To maximize bandwidth in mobile networks, the {+sync+} service only
```
mongomoe, line 56: It took me a few re-reads to understand what the first sentence was saying. I think it would be a bit simpler to understand if we avoided overloading the paragraph with technical phrases like "maximize bandwidth" and "deltas". I think we need to have the phrase "delta sync" at least once for SEO/search? Here's an alternative that attempts to accomplish that in simpler terms that still, hopefully, satisfies the keyword/search requirement:

"Realm Sync uses a process known as ``delta sync`` that transmits only the changes to the altered fields in order to synchronize data faster. The service compresses the changes on the fly with zlib compression."

mongomoe (approved):
LGTM but 1 link suggestion and 1 wording suggestion. Technically this PR satisfies the ticket but given that it took me a couple re-reads to understand, I can see others needing to re-read this as well.

source/sync/protocol.txt
cbush: Thanks for the feedback. I have completely reworked it. Let me know what you think!

mongomoe: New wording is much clearer, double LGTM @cbush

PR #1132, opened 2021-06-10
(DOCSP-11954): modify schema properties of a synced realm

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-11954

- [Modify Schema Properties of a Synced Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-11954-Modify-Schema-Properties-Synced-Realm/sync/modify-synced-object-schema/)

#### Notes on SDK pages (it's just the same note literal included):
- [Android SDK > Modify an Object Schema > ~Modify Schema Properties of a Synced Realm ​note](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-11954-Modify-Schema-Properties-Synced-Realm/sdk/android/examples/modify-an-object-schema/)
- [iOS SDK > Modify an Object Schema > ~Modify Schema Properties of a Synced Realm ​note](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-11954-Modify-Schema-Properties-Synced-Realm/sdk/ios/examples/modify-an-object-schema/)
- [.NET SDK > Modify an Object Schema > ~Modify Schema Properties of a Synced Realm note](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-11954-Modify-Schema-Properties-Synced-Realm/sdk/dotnet/examples/modify-an-object-schema/)
- [Node.js SDK > Modify an Object Schema > ~Modify Schema Properties of a Synced Realm note](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-11954-Modify-Schema-Properties-Synced-Realm/sdk/node/examples/modify-an-object-schema/)
- [React Native SDK > Modify an Object Schema > ~Modify Schema Properties of a Synced Realm note](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-11954-Modify-Schema-Properties-Synced-Realm/sdk/react-native/examples/modify-an-object-schema/)

### Review Guidelines


source/includes/note-modify-schema-properties-of-synced-realms.rst
```diff
@@ -0,0 +1,5 @@
+.. note:: Modify Schema Properties of a Synced Realm
+
+   The following page demonstrates how to modify schema properties of a local
+   {+realm+}, learn how to :doc:`modify schema properties of a synced realm
+   </sync/modify-schema-properties-synced-realm>`.
```
nlarew, line 5: Looks like a comma splice here, these should be two separate sentences.

Also, thoughts on phrasing the second sentence a little more like "If you want to modify the schema of a synced realm, see..."?

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
+
+
+.. _modify-schema-properties-synced-realm:
+
+===========================================
+ Modify Schema Properties of a Synced Realm
```
nlarew, line 6:
```suggestion
Modify Schema Properties of a Synced Realm
```
Extra space at the front here - need to also update the `=`

nlarew, line 6: I think the title could be a little more specific - it's properties of a realm object, not the synced realm. Maybe: "Modify a Synced Object Schema"?

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+When developing an application using Realm Sync, you may want to make changes to
+your schema at some point. Realm Database offers an intuitive API for making
```
nlarew, line 20: nit: is it really intuitive? 🙃

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+Additive changes, such as adding a class or adding a field to a class, are
+applied automatically to synced {+realms+}, meaning you can alter the schema
+with no additional configuration. 
+
+.. note:: Removing a Property from Your Schema Is Considered an Additive Change.
+
+   To maintain backward compatibility, removing a field from a schema doesn’t
+   delete the field from the database and instead instructs Realm to ignore that
+   field. New objects retain the removed field, but Realm automatically sets the
+   field's value set to null. Realm sets fields that are non-nullable to an
```
nlarew, line 68:
```suggestion
   field's value to ``null``. Realm sets fields that are non-nullable to an
```

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+   appropriate empty value, such as a 0 for integer values and an empty string
```
nlarew, line 69:
```suggestion
   appropriate empty value, such as a 0 for integer values or an empty string
```

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+-------------------
+Destructive changes to a schema are usually **modifications** of existing fields, such as:
+
+- Changing a property’s type but keeping the same name
+- Changing a primary key
+- Changing a property from optional to required (or vice-versa)
+
+Synchronized {+realms+} only support additive changes to a schema. Attempting to
+perform a destructive change on a synchronized {+realm+} will lead to errors
+like the following:
```
nlarew, line 82: nit: "Attempting to..." => "If you attempt to perform..., you will encounter an error like the following:"

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+   }
+
+
+Using Partner Collections to Modify Properties of a Schema
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you are developing an application using {+sync+}, you can create a second
+version of your collection with a schema containing the modifications to the
+fields you require. Doing so requires data to flow in both directions, meaning
+that when one collection is written to, the other should be as well.
```
nlarew, line 99: Seems like we're trying to use the term "Partner Collection" to describe this pattern. Can we explicitly define it as a term somewhere up front here?

e.g. A partner collection is a second collection that contains the same data as another collection but with a different schema. Partner collections use database triggers that you define to sync changes between each collection.

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+
+To set up your partner collections, copy the data from the initial collection to
+the second collection. You can do this by using the
+:manual:`db.collection.aggregate() <reference/method/db.collection.aggregate>` method
+with the :manual:`Mongo shell <mongo>`.
```
nlarew, line 104: The copied data needs to conform to the modified schema, right? Let's be clear that you need to copy _and transform_ the data before you insert it into the second collection (i.e. why aggregate is the most useful).

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+   schema. 
+   
+   He copies the objects in the first collection to the second
+   collection by performing the following aggregation using the Mongo shell:
+
+   .. code-block:: shell
+      :caption: Match All Documents in the Initial Collection and Output Them to the Partner Collection 
+
+      const collection = client.db("<Your Database Name>").collection("Task");
+      collection.aggregate([{ $match: {} }, { $out: "TaskV2" }]);
```
nlarew, line 128: I don't think this works in `mongo` or `mongosh` so it's not mongo shell code. Looks like node driver/sdk or Realm function JS.

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+   ``Task`` collection, the developer  creates a :ref:`database trigger
+   <create-a-database-trigger>`. The trigger fires whenever {+atlas+} performs an
+   insert, update, delete or replace operation on the ``TaskV2`` collection.
+
+   .. literalinclude:: /examples/Triggers/copy-data-from-second-version-collection.js
+      :language: javascript
+      :caption: copyDataFromTasksV2ToTasks trigger
+      :emphasize-lines: 6, 7
+
+   The developer then navigates to his {+app+}'s :guilabel:`Schema` tab and views his new JSON schemas:
```
nlarew, line 150: Let's remove the mention of the UI (i.e. schema tab) so that this is agnostic to what interface they use

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+      :caption: Task Schema
+
+      {
+        "title": "Task",
+        "bsonType": "object",
+        "required": [
+          "_id",
+          "_partition",
+          "name",
+          "difficulty"
```
nlarew, line 162: thoughts on adding an emphasize-line on the field that we're gonna remove?

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+          },
+          "difficulty": {
+            "bsonType": "long"
+          }
+        }
+      }
+
+   The ``difficulty`` field is required for the ``Task`` collection but is not
+   required for the ``TaskV2`` collection as specified in the schema
+   modification.
```
nlarew, line 209: My intuition is that folks miss small bits of text at the end of a section when they're after a sizable code block or note. Any way we could move this up before the schemas?

source/sync/modify-schema-properties-synced-realm.txt
nlarew, line 128: I don't think this works in `mongo` or `mongosh` so it's not mongo shell code. Looks like node driver/sdk or Realm function JS.

source/examples/Triggers/copy-data-from-initial-collection.js
```diff
@@ -0,0 +1,9 @@
+
+exports = async function(changeEvent) {
+    const db = context.services.get("<Your Mongodb Atlas Service Name>").db("<Your Database Name>"); // replace with your Atlas Service Name and your database name
+    const taskCollection = db.collection("Task");
+    const taskV2Collection = db.collection("TaskV2");
+    const docId = changeEvent.documentKey._id; // _id of changed or inserted document
+    const changedItem = await taskCollection.findOne({_id: docId}); // find the changed or inserted documented
+    await taskV2Collection.insertOne(changedItem); // insert the changed or inserted document into its partnet collection
```
nlarew, line 8:
```suggestion
    await taskV2Collection.insertOne(changedItem); // insert the changed or inserted document into its partner collection
```

nlarew: Don't I need to actually do something other than just directly copy from one collection to the other? I assume the Task -> TaskV2 trigger needs to do some kind of transformation, otherwise why have two collections?

source/examples/Triggers/copy-data-from-second-version-collection.js
```diff
@@ -0,0 +1,8 @@
+exports = async function(changeEvent) {
+    const db = context.services.get("<Your Mongodb Atlas Service Name>").db("<Your Database Name>"); // replace with your Atlas Service Name and your database name
+    const taskV2Collection = db.collection("TaskV2");
+    const taskCollection = db.collection("Task");
+    const docId = changeEvent.documentKey._id; // _id of changed or inserted document
+    const changedItem = await taskV2Collection.findOne({_id: docId}); // find the changed or inserted documented
+    await taskCollection.insertOne(changedItem); // insert the changed or inserted document into its partnet collection
```
nlarew, line 7:
```suggestion
    await taskCollection.insertOne(changedItem); // insert the changed or inserted document into its partner collection
```

nlarew: Same comment as the other trigger - shouldn't this do some transformation to match the other collection's schema?

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+schema changes to synced realms that are backward compatible, allowing old
+clients to sync with newer ones.
+
+.. note::
+
+   This page details how to modify a schema when working with {+sync+}. If you
+   are developing your application without {+sync+}, refer to the modify an
+   object schema page for your SDK:
+
+   .. tabs-realm-sdks::
```
nlarew, line 30: nit: thoughts on removing the tabs and making this a bulleted list w/ the SDK clearly identified for each link? The whole "tabs for a single link" pattern feels a bit silly IMO.

source/sync/modify-schema-properties-synced-realm.txt
```diff
@@ -0,0 +1,221 @@
...
+To set up your partner collections, copy the data from the initial collection to
+the second collection. You can do this by using the
+:manual:`db.collection.aggregate() <reference/method/db.collection.aggregate>` method
+with the :manual:`Mongo shell <mongo>`.
+
+To ensure the collections always have the same data and stay up-to-date when
+data is subsequently changed, create a :ref:`database trigger
+<create-a-database-trigger>` for each respective collection. Within the
+trigger's function, copy the changed data from the collection to its partner
+collection.
```
nlarew, line 110: This should reflect the same transform we do in the initial aggregation

source/sync/modify-schema-properties-synced-realm.txt
nlarew: The aggregation should also transform the Task documents somehow to conform to TaskV2. The example we chose doesn't actually need to change anything, so maybe we should revisit to have a slightly more complex change.

e.g. a V1 model has `userId` as a `number` but the V2 model changes it to `string`

nlarew (changes requested):
A few comments, several revolving around how to do transforms between the collections

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Good catch, fixed locally rather than through the 'Commit suggestion' button. 👍

source/includes/note-modify-schema-properties-of-synced-realms.rst
mongomoe: Fixed the comma issue 👍  

I originally thought about phrasing it that way, but "If you want to modify the schema of a synced realm, see the documentation{link} for modifying schema properties of a synced realm" is so redundant. Alternatively, "If you want to modify the schema of a synced realm, see the documentation{link}" is vaguer and worse for SEO.

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Haha arguably it's intuitive in that we don't even technically let you do destructive changes. But fair enough, "Realm offers an API for making backward-compatible schema changes to synced realms, allowing old clients to sync with newer ones." it is 👍

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Good idea, changed it to that now 👍

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Fixed locally 👍

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Going to keep the current wording since it conveys the same message

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: That's a great idea, I've altered this part. Let me know what you think of the new phrasing. 

Also, would love to use the phrase " to sync changes between each collection." but Ian recommended not to as users might confuse the product sync (which implies sync between clients, or sync between clients and a server) with sync between collections. Instead I used "data flows in both direction"

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Whoops sorry about that, I originally used the atlas UI to perform the aggregation. I've altered it to use mongo shell commands (and I've tested the commands to ensure that they work)

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Nope, the copied data does not need to conform to the modified schema. The new collection needs to contain the existing data, so that newer clients still have access to the old data. 

So to be clear, data does not need to be "transformed" before inserting into the second collection, it should be copied as is. You could also do a mongodump/mongorestore (or various other methods of copying a collection).

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Do you have any suggestions for alternative phrasing?

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Sounds like a good idea, added line emphasize on the "difficulty" field now

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Sure, moved it above the json schema now 👍

source/sync/modify-schema-properties-synced-realm.txt
mongomoe: Going to keep it as is, personally I think the tabs look cleaner than a list (especially since its a long-ish page anyway)

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,176 @@
...
+  :manual:`$merge operator <reference/operator/aggregation/merge/>` to avoid
+  overwriting existing data.  In this example, we wrote the data to the
+  ``TaskV2`` collection.
+
+  .. literalinclude:: /examples/Triggers/copy-data-from-initial-collection.js
+     :language: javascript
+     :caption: copyTaskObjectToTaskV2 trigger
+
+  To copy data from ``TaskV2`` collection to the ``Task`` collection when the
+  trigger fires, alter your trigger function's code. This time, transforming the
```
nlarew, line 171: nit: not really "alter" since it's a separate function entirely and the backward transform may not be identical to the forward transform.

i.e. instead of "alter your trigger function's code..." maybe try "write a function similar to the first trigger...."

nlarew: nit: here we lose the separation between generic advice and example. Can we rework the "This time..." sentence to make it clear that it's in our example?

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,176 @@
...
+  new clients that are using the new collection with the modified schema. 
+
+  To do this, set up a :ref:`database trigger <create-a-database-trigger>` for
+  each respective collection. When data is written to one collection, the
+  trigger's function should perform the write to the partner collection.
+
+  Follow the steps in the :ref:`database trigger <create-a-database-trigger>`
+  documentation to create a trigger to copy data from the ``Task`` collection to
+  the ``TaskV2`` collection. Ensure that you have checkmarked all operation types in
+  the :guilabel:`Operation Type` column.
```
nlarew, line 126: Can we make this less UI-centric? i.e. remove "checkmarked", guilabels, etc

"The trigger should fire for all operation types."

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,176 @@
...
+---
+title: Set up Database Triggers for Your Partner Collections
+ref: set-up-db-triggers-partner-collections
+content: |
+  Once your partner collection is set up, you can use it to read existing data.
+  However, any new writes to the data of the either collection will not be shown
+  on its partner. This will cause the old clients to be out of sync with the
+  new clients that are using the new collection with the modified schema. 
+
+  To do this, set up a :ref:`database trigger <create-a-database-trigger>` for
```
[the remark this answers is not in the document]
mongomoe: Fixed verbatim 👍

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,176 @@
...
+         "_partition": {
+           "bsonType": "string"
+         },
+         "name": {
+           "bsonType": "string"
+         }
+       }
+     }
+---
+title: Set Up a Partner Collection
```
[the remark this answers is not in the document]
mongomoe: Changed instead to "Initialize Your Partner Collection", let me know what your think.

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,176 @@
...
+  Modify the desired fields of the initial collection by using the
+  :manual:`$addFields operator <reference/operator/aggregation/addFields/>`. In
+  the following example, the initial data's ``_id`` field is transformed to a
+  ``string`` type using the :manual:`$toString operator
+  <reference/operator/aggregation/toString/>`.
+
+  Finally, write the transformed data to the partner collection by using the
+  :manual:`$out operator <reference/operator/aggregation/out/>` and specifying
+  the partner collection name. In this example, we wrote the data to a new
+  collection named ``TaskV2``.
```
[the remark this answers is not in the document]
mongomoe: Right now there's no official documentation on copying data from one collection to another collection, outside of this staging link and the deprecated 'copyTo' method (there are other ways such as Mongodump+Mongorestore, looping through data + manually insertOne, Mongoexport + Mongoimport, etc). So since we don't have anywhere to point users to directly, I think vaguely telling them you can use these other ways might lead to more confusion than not. 

For the $addFields, I'll add a line that says you might need to use an alternate operator for your usecase.

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,176 @@
...
+  alter the trigger function's code. 
+  
+  If the :ref:`change event <database-events>` is a delete, meaning the object
+  has been deleted from the collection, delete the object from it's partner
+  collection by performing a :manual:`db.collection.deleteOne()
+  <reference/method/db.collection.deleteOne>` and passing the
+  ``changeEvent.documentKey._id``. In the following example, we call
+  :mdn:`parseInt() <Web/JavaScript/Reference/Global_Objects/parseInt>` on the
+  ``_id`` field since the ``TaskV2`` collection stores the ``_id`` field as an
+  integer.
```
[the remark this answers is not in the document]
mongomoe: This step is dense because it outlines chronologically the steps to create the trigger function. Moving `parseInt` down a bit, would ruin the order of the steps and might be more confusing. There's also other methods that suffer from this same problem (such as deleteOne).

source/includes/steps-perform-destructive-changes-synced-schema.yaml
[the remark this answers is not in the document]
mongomoe: changed the $addFields paragraph to:

"Modify the desired fields of the initial collection by using an aggregation pipeline operator. In the following example, the data is transformed using the $addFields operator. The``_id`` field is transformed to a string type with the $toString operator."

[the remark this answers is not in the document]
mongomoe: Let me know what you think ^

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,176 @@
...
+
+  Modify the desired fields of the initial collection by using the
+  :manual:`$addFields operator <reference/operator/aggregation/addFields/>`. In
+  the following example, the initial data's ``_id`` field is transformed to a
+  ``string`` type using the :manual:`$toString operator <reference/operator/aggregation/toString/>`.
+
+  Finally, write the transformed data to the partner collection, by using the
+  :manual:`$merge operator <reference/operator/aggregation/merge/>` to avoid
+  overwriting existing data.  In this example, we wrote the data to the
+  ``TaskV2`` collection.
```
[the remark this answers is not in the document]
mongomoe: Any specific wording that you would suggest? 

Also, personally I'm not against prescribing a specific way of doing something even if there are alternatives. For this example, using aggregations seems simpler (esp since we just did aggregations earlier and aggregations allow you to not have to do conditional checking on operation type, except delete, to determine what write method to perform)

source/includes/steps-perform-destructive-changes-synced-schema.yaml
mongomoe: Fixed 👍

source/includes/steps-perform-destructive-changes-synced-schema.yaml
mongomoe: Good ideas, fixed now 👍 

"To copy data from TaskV2 collection to the Task collection when the trigger fires, write a function similar to the first trigger's function. In this example, we transform the data of TaskV2 before inserting it into the Task collection by changing the _id from a string to an integer value."

source/includes/steps-perform-destructive-changes-synced-schema.yaml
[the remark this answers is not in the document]
nlarew: Sounds good, already a lot of info so probably okay for now if we don't add more

source/includes/steps-perform-destructive-changes-synced-schema.yaml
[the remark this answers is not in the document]
nlarew: Per my other comment probably okay to leave it out

source/includes/steps-perform-destructive-changes-synced-schema.yaml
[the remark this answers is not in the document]
nlarew: It goes a bit beyond density IMO. We specifically say "In the following example, we call `parseInt()` on the `_id` field since the `TaskV2` collection stores the `_id` field as an integer."

(Side note - this is incorrect because TaskV2 stores `_id` as a string not an integer)

This doesn't feel right because:
- We don't use `parseInt()` in the following code block (it's in the code block _after_ the following)
- We have another sentence after this one that _does_ describe the following code block, i.e. "In the following example, the initial data's `_id` field is transformed to a string type using the `$toString` operator."

I think the issue arising with `deleteOne()` gives us even more reason to evaluate this step critically - there's probably some (relatively small) underlying organizational change we can make to avoid or diminish the problem.

A couple of ideas:
- Be a bit more explicit up front that we're describing the three phases of ETL and then discuss how the trigger will **Extract** from one collection, **Transform** to match the other collection's schema, and then **Load** into the other collection.
- Remove the mention of `parseInt()` from the paragraph and add a code comment about it instead
- Or, move the mention of `parseInt()` to the paragraph after the first trigger code block. If we go this route, we probably also want to mention in a similar way that the first trigger uses `toString()`
- Maybe split the two triggers into separate steps? e.g. **Write a Forward Migration Trigger** followed by **Write a Reverse Migration Trigger**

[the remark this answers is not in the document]
mongomoe: This was a really in-depth and helpful comment! I restructured this bit on trigger functions into two steps and followed a lot of your advice, let me know what you think.

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,180 @@
...
+  :ref:`database trigger <create-a-database-trigger>` for each respective
+  collection. When data is written to one collection, the trigger's function
+  should perform the write to the partner collection.
+
+  Follow the steps in the :ref:`database trigger <create-a-database-trigger>`
+  documentation to create a trigger to copy data from the ``Task`` collection to
+  the ``TaskV2`` collection. The trigger should fire for all operation types. 
+ 
+  Repeat these steps and create a second trigger to copy data from the
+  ``TaskV2`` collection to the ``Task`` collection.
```
nlarew, line 129: nit: probably not a blocker but the timeline of the steps is a little off here if you're using the UI since you define the trigger function when you create it. Thoughts?

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,180 @@
...
+
+  .. figure:: /images/trigger-copyTaskObjectToTaskV2.png
+     :alt: Create a Database Trigger to Copy Data from Task to TaskV2
+     :width: 750px
+     :lightbox:
+---
+title: Create a Forward Migration Function
+ref: create-forward-migration-fn
+content: |
+  To copy any new changes to the initial collection to the partner collection,
```
nlarew, line 139: nit: At first I read this as "copy changes [that were made somewhere] to the initial collection".

Also not sure if "initial collection" is super clear, especially in contrast to "partner collection". My read was that both V1 and V2 are "partners" with eachother and initial could mean "V1 since it is the earlier version" or (in the second trigger) "V2 since it is the collection I'm exporting from'

Maybe we can clear up the wording? (I use "old" and "new" here but there's probably a better choice) e.g.

"A forward migration trigger listens for inserts, updates, and deletes in the old collection, modifies them to reflect the newer collection's schema, and then applies them to the new collection"

source/includes/steps-perform-destructive-changes-synced-schema.yaml
```diff
@@ -0,0 +1,180 @@
...
+
+  .. literalinclude:: /examples/Triggers/copy-data-from-initial-collection.js
+     :language: javascript
+     :caption: copyTaskObjectToTaskV2 trigger
+
+---
+title: Create a Reverse Migration Function
+ref: create-reverse-migration-fn
+content: |
+  To copy any new changes to the second collection to the initial collection,
```
nlarew, line 166: Same thoughts here as in my previous comment re "copy...to" and "initial collection"

nlarew (approved):
A few small wording comments but otherwise looking really good!

source/includes/steps-perform-destructive-changes-synced-schema.yaml
mongomoe: You actually don't need to define the content of the trigger function immediately, you can just create & name the function and leave it blank - which is what the step is saying.

source/includes/steps-perform-destructive-changes-synced-schema.yaml
mongomoe: Personally I don't like 'old' because it makes me think the Task collection is deprecated or unusable compared to the TaskV2 collection. But actually both collections are used simultaneously, Task is used by clients before the schema change, and TaskV2 after the schema change.

I did take your advice on the 'copy changes' and almost verbatimly applied your updated wording:
"A forward migration trigger listens for inserts, updates, and deletes in the initial collection, modifies them to reflect the partner collection's schema, and then applies them to the partner collection."

mongomoe: Personally I don't like 'old' because it makes me think the Task collection is deprecated or unusable compared to the TaskV2 collection. But actually both collections are used simultaneously, Task is used by clients before the schema change, and TaskV2 after the schema change.

I did take your advice on the 'copy changes' and almost verbatim applied your updated wording:
"A forward migration trigger listens for inserts, updates, and deletes in the initial collection, modifies them to reflect the partner collection's schema, and then applies them to the partner collection."

source/includes/steps-perform-destructive-changes-synced-schema.yaml
mongomoe: Altered to: "To listen for changes to the second collection and apply them to the initial collection, write a reverse migration function"

PR #1136, opened 2021-06-11
DOCSP-14053 added note about option to index the partition key

lindseymoore:
### Jira

- Get Started with Sync(https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/indexPartition/sync/get-started/)

### Review Guidelines

### DIFF
--- a/source/includes/steps-enable-sync-ui.yaml
+++ b/source/includes/steps-enable-sync-ui.yaml
@@ -34,8 +34,8 @@ content: |
 
      .. tip:: When to enable automatic creation of an index on the partition key
 
-
         An index will make reads of the data faster but writes slightly slower. You should generally use an index unless:
+
            - You have way more writes than reads
            - You already have an index on the partition key field
 
--- a/source/includes/steps-enable-sync-ui.yaml
+++ b/source/includes/steps-enable-sync-ui.yaml
@@ -29,7 +29,7 @@ content: |
   1. **Select a Cluster to Sync:** {+sync+} applies to the entire cluster. Specify which cluster you want to sync in the dropdown.
   #. **Define a Database Name (Development Mode Tab Only):** If you opted to use Development Mode, {+service-short+} needs to know which database in your cluster to use for new collections it creates for {+sync-short+}. Specify that database name in the input field.
   #. **Choose a Partition Key:** Specify the field name and type to determine which {+realm+} each object belongs to. You can also select the 'Create an index on the partition key' option, to enable Realm to automatically add an index on the partition key field for the synced cluster.  See :ref:`Partition Atlas Data into Realms <partitioning>`.
-     .. tip: When to enable automatic creation of an index on the partition key
+     .. tip:: When to enable automatic creation of an index on the partition key
         An index will make reads of the data faster but writes slightly slower. You should generally use an index unless:
            - You have way more writes than reads
            - You already have an index on the partition key field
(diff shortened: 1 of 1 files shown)


source/includes/steps-enable-sync-ui.yaml
```diff
@@ -28,7 +28,7 @@ content: |
 
   1. **Select a Cluster to Sync:** {+sync+} applies to the entire cluster. Specify which cluster you want to sync in the dropdown.
   #. **Define a Database Name (Development Mode Tab Only):** If you opted to use Development Mode, {+service-short+} needs to know which database in your cluster to use for new collections it creates for {+sync-short+}. Specify that database name in the input field.
-  #. **Choose a Partition Key:** Specify which field to use to determine which {+realm+} each object belongs to. See :ref:`Partition Atlas Data into Realms <partitioning>`.
+  #. **Choose a Partition Key:** Specify the field name and type to determine which {+realm+} each object belongs to. You can also select the 'Create an index on the partition key' option, to allow Sync to automatically handle indexing.  See :ref:`Partition Atlas Data into Realms <partitioning>`.
```
nlarew, line 31: I think we can expand just a little more on what "automatically handle indexing" means.

- If you select the option then Realm adds an index on the partition key field for the synced cluster (if the index doesn't already exist).

We could also give a little tip on when you should or shouldn't create an index (I would literally use a `.. tip:` directive)

- An index will generally make reads faster but writes slightly slower
- Probably a good idea to create the index unless...
  - you have way more writes than reads
  - you already have an index that includes the partition key field

So for example if the partition key field is `_partition` then Realm will create this index on the synced cluster:
```
{ "_partition": 1 }
```
But if your collection already has this index:
```
{ "_partition": 1, "name": 1 }
```
then it would be redundant to create the new one

nlarew: To write the tip you'll need to have some space between the numbered list items and then use a format like this:

```restructuredtext
1. blah blah blah

#.blah blah blah 

#. blah blah blah
   
   .. tip: When to create an index
      
      blah blah blah

#. blah blah blah
```

source/includes/steps-enable-sync-ui.yaml
```diff
@@ -27,8 +27,18 @@ content: |
...
-  #. **Choose a Partition Key:** Specify which field to use to determine which {+realm+} each object belongs to. See :ref:`Partition Atlas Data into Realms <partitioning>`.
+
+  #. **Choose a Partition Key:** Specify the field name and type to determine which {+realm+} each object belongs to. You can also select the 'Create an index on the partition key' option, to enable Realm to automatically add an index on the partition key field for the synced cluster.  See :ref:`Partition Atlas Data into Realms <partitioning>`.
+
+     .. tip:: When to enable automatic creation of an index on the partition key
+
+        An index will make reads of the data faster but writes to the data slightly slower. You should generally use an index unless:
+
+           - You have way more writes than reads
+           - You already have an index on the partition key field
```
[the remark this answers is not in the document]
nlarew: This bullet is good but I think we can also be a bit more specific - "an index on the partition key field" could mean several things:

- ✅  A basic index identical to what this option would create: e.g. `{ _partition: 1 }`
- ✅  A compound index that starts with the partition key: e.g. `{ _partition: 1, date: 1 }`
- 🚫 A compound index that _doesn't_ start with the partition key: e.g. `{ date: 1, _partition: 1 }`

All three of these indexes are "on the partition key field" in some sense but only the first two actually cover the use case due to how indexes work. Do you think we could update to reflect some of this nuance?

nlarew (approved):
LGTM!

PR #1151, opened 2021-06-15
DOCSP-16395 fixed code images to say bsonType instead of type

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-16395

- Define a Custom Resolver (https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16395/graphql/custom-resolvers)

### Review Guidelines


nathan-contino-mongo (changes requested):
These screenshots are so wide that they're difficult to read. Can we replace them with screenshots that are a bit closer to the size of the old screenshots, so they're still legible?
https://docs.mongodb.com/realm/graphql/custom-resolvers/#define-the-payload-type

Additionally, can you change the staging link to point to the page containing your changes? It should point here: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16395/graphql/custom-resolvers/ instead of the root of the staging build. It's not a huge deal, but especially when it comes to `includes` edits, it can be really hard  to figure out which page to look at if it isn't directly linked!

nathan-contino-mongo (approved):
LGTM, nice job! One last thing: if those old images aren't used anywhere else in the repo, you might as well delete them so they don't take up space in the images folder. If deleting them doesn't cause any snooty build errors, you'll know they weren't used anywhere else. [alternatively, you could grep in the repo for usages, but I'm usually lazier than that]

nlarew (approved):
LGTM

PR #1154, opened 2021-06-16
List update: revenge of the forgotten types

nathan-contino-mongo:
### Jira

- no jira today, just a typo i noticed as Steward of the Realm


dacharyc (approved):
Fixin' them forgotten types!

PR #1169, opened 2021-06-23
(DOCSP-16338): Document Realm Admin API App Import endpoint

dacharyc:
## Pull Request Info

Based on conversations since the initial draft, I've updated this PR to mirror the app configuration documentation. We've agreed to only document this endpoint for the current version of the app structure. I've got a few outstanding questions in the linked document below.  

Note: this document renders in a way that requires you to scroll horizontally to view the descriptions and types for each field. We're not able to affect the way this renders on the page, unfortunately.

### Jira

- https://jira.mongodb.org/browse/DOCSP-16338

- [Realm Admin API -> Import an App](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16338/admin/api/v3/#post-/groups/%7Bgroupid%7D/apps/%7Bappid%7D/import)

### Outstanding Questions (Document viewable by MongoDB org members)

- [Outstanding Questions](https://docs.google.com/document/d/1kBKt0ogubZf3YXa-O-wrZ2AmVm9sRTaPP9LCUtL2_SM/edit?usp=sharing)

### Review Guidelines

I'd love reviewers to take a look and see if anything here stands out as obviously incorrect, but I recognize this is a big document and am not requesting a line-by-line read through (unless you really want to!) I would like someone in engineering to confirm the answers to the outstanding questions, so I can incorporate any tweaks.

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -197,6 +197,25 @@ paths:
       - $ref: "#/components/parameters/GroupId"
       - $ref: "#/components/parameters/AppId"
 
+  /groups/{groupId}/apps/{appId}/import:
+    post:
+      tags: ["apps"]
+      operationId: "adminImportApplication"
+      summary: ":ref:`Import <export-realm-app>` an application."
+      requestBody:
+        description: The app to import.
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ApplicationImport'
+      responses:
+        '204':
+          description: "No content"
+    parameters:
+      - $ref: "#/components/parameters/GroupId"
+      - $ref: "#/components/parameters/AppId"
+
   # Triggers
 
   /groups/{groupId}/apps/{appId}/triggers:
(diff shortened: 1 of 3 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -197,6 +197,25 @@ paths:
...
+      tags: ["apps"]
+      operationId: "adminImportApplication"
+      summary: ":ref:`Import <export-realm-app>` an application."
+      requestBody:
+        description: The app to import.
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ApplicationImport'
```
chickenbug, line 211: This is super detailed, but IMO it'll be a pain to maintain, and contains some fields not meant to be manually edited by a user. I feel like we should just link the the Realm Application Configuration [docs](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16338/config/#std-label-app-configuration), and have the user figure it out from there!

source/openapi-admin-v3.yaml
```diff
@@ -197,6 +197,25 @@ paths:
       - $ref: "#/components/parameters/GroupId"
       - $ref: "#/components/parameters/AppId"
 
+  /groups/{groupId}/apps/{appId}/import:
+    post:
+      tags: ["apps"]
+      operationId: "adminImportApplication"
+      summary: ":ref:`Import <export-realm-app>` an application."
```
chickenbug, line 204: I'm unsure if this is the correct link here, since it only talks about exporting an app. No other pages are a perfect fit though. It would be nice if we could update the export-realm-app page to also talk a little bit about importing an app

chickenbug (changes requested):
Hi this looks really great! Thanks for documenting this. I only have one main concern about the level of detail at which the app is documented. I feel like it would be more flexible and maintainable if we pointed to the existing realm Application Configuration docs.

source/openapi-admin-v3.yaml
```diff
@@ -197,6 +197,33 @@ paths:
       - $ref: "#/components/parameters/GroupId"
       - $ref: "#/components/parameters/AppId"
 
+  /groups/{groupId}/apps/{appId}/import:
+    post:
+      tags: ["apps"]
+      operationId: "adminImportApplication"
+      summary: >-
+        :ref:`Import <deploy-cli>` an application. Warning: this endpoint
```
mongomoe, line 205: Is the deploy-cli ref the correct ref?

source/openapi-admin-v3.yaml
```diff
@@ -197,6 +197,33 @@ paths:
       - $ref: "#/components/parameters/GroupId"
       - $ref: "#/components/parameters/AppId"
 
+  /groups/{groupId}/apps/{appId}/import:
+    post:
+      tags: ["apps"]
+      operationId: "adminImportApplication"
+      summary: >-
+        :ref:`Import <deploy-cli>` an application. Warning: this endpoint 
+        requires care and is quite complex. Consider :ref:`using the CLI to
```
mongomoe, line 206: "requires care and is quite complex" seems vague. What does require care mean? How is it complex? Does this endpoint provide any benefits that would outweigh the additional care that a developer would need to take?

source/openapi-admin-v3.yaml
```diff
@@ -2200,6 +2227,698 @@ components:
...
+              description: The application's :guilabel:`App ID`.
+            name:
+              type: string
+              description: The application's name.
+            config_version:
+              type: integer
+              description: >-
+                The {+service+} schema version of the configuration files in the 
+                application. This is distinct from your application's schema
+                version, and tells {+sync+} how to process your application import.
```
mongomoe, line 2250:
```suggestion
                version and tells {+sync+} how to process your application import.
```

unneeded comma

source/openapi-admin-v3.yaml
```diff
@@ -2200,6 +2227,698 @@ components:
...
+                    <config-custom-jwt>`, :ref:`Google <auth-google-configuration>`,
+                    :ref:`Facebook <auth-facebook-configuration>`, and :ref:`Apple 
+                    <configure-apple-auth>`.
+                metadata_fields:
+                  type: array
+                  description: >-
+                    An array of documents where each document defines a metadata
+                    field that describes the user. The existence of this field and
+                    exact format of each depends on the provider ``type``. The
+                    provider ``type``s that use metadata fields are: :ref:`custom
```
mongomoe, line 2346: the word "type" is rendering as "type``s" rather than a formatted version of "type"

mongomoe: probably due to the apostrophe s.

source/openapi-admin-v3.yaml
```diff
@@ -2200,6 +2227,698 @@ components:
...
+            Define and execute server-side logic for your application. For
+            more information on functions, see: :ref:`Functions <functions>`,
+            or for more information on configuring functions, see: 
+            :ref:`Function Configuration Files <appconfig-functions>`.
+          properties:
+            name:
+              type: string
+              description: >- 
+                The name of the function. The name must match the file name
+                of a :ref:`source code file <function-source-code-file>`,
```
mongomoe, line 2425:
```suggestion
                of a :ref:`source code file <function-source-code-file>`,
```
unneeded comma

source/openapi-admin-v3.yaml
```diff
@@ -2200,6 +2227,698 @@ components:
...
+                  type: integer
+                  description: >-
+                    The date and time that sync was last paused or disabled, 
+                    represented by the number of seconds since the Unix epoch 
+                    (January 1, 1970, 00:00:00 UTC).
+        triggers:
+          type: object
+          description: >-
+            :ref:`Triggers <triggers>` listen for application events of 
+            a configured type, and enable you to execute application and
```
mongomoe, line 2794:
```suggestion
            a configured type and enable you to execute application and 
```

mongomoe (commented):
Close to LGTM, just needs some grammar + light copy fixes

source/openapi-admin-v3.yaml
dacharyc: I thought this was the closest we have, because it described importing an app config structure, but a little more digging has found a more direct correlation, so I've changed it to the [Realm CLI Reference -> Import an Application](https://docs.mongodb.com/realm/deploy/realm-cli-reference/#import-an-application). Thank you for asking about this!

source/openapi-admin-v3.yaml
dacharyc: Yes, it is vague. This is a kind of shorthand, vague warning because ideally, we don't want people using this endpoint. Basically, we want to be able to point to it in a very few edge cases, but otherwise want to discourage folks from using it. I didn't get more specific because I don't have a comprehensive list of those edge cases. The thread that spawned this work was related to custom resolvers, but it could also be useful for folks who want to programmatically import apps - and maybe other stuff?

It is complex because it essentially rolls up all of the app configuration objects in one massive payload. It is deeply nested in parts, and there are configuration objects that I can't even include here - I've included placeholder objects, and the descriptions link to the various config options in the corresponding App Configuration documentation sections. The structure and options for those config objects vary depending on which of the possible options in the enum you're configuring. For example, whether you're configuring email/password authentication or Google Auth. I've done my best to be comprehensive, but our openAPI implementation doesn't really support "If you're configuring auth type 1 in this enum of accepted auth types, the structure of the config is..., while auth type 2 has this structure..."

"Requires care" in this case is shorthand for "if you intend to program against this, it will require many hours of your life making sure you're sending the right combination of things, and if you don't send the right combination of things, you may get an error, or you may get a 204 back with no meaningful response body and not know until something blows up in a new and interesting way that something got messed up during import."

I know this seems like a very long rant justifying a short, throwaway phrase, but I genuinely couldn't find a better way to try to convey "Danger, but if you really need it, try not to blow yourself up..." If you have a suggestion for an alternate way to communicate some of these points, that isn't also _way_ too verbose to fit into the summary, I would be happy to incorporate feedback!

source/openapi-admin-v3.yaml
dacharyc: Good catch! I changed a bunch of instances of that after I realized it wasn't rendering nicely, but missed that one. Thanks for spotting it!

source/openapi-admin-v3.yaml
mongomoe: Yeah, I get that, I can see how using that endpoint can be tedious. Have you brought it up with product at all? Bc in reality this sounds more like an issue with the endpoint's usability.

mongomoe: Also do we document anywhere the issue w/ not sending the right combination of things leading to errors, or 204s with no meaningful body?

mongomoe (approved):
LGTM 👍  not sure what to do about this [comment](https://github.com/mongodb/docs-realm/pull/1169/files#r664826582), but I guess its not worth blocking the PR

source/openapi-admin-v3.yaml
dacharyc: Alrighty, @mongomoe - based on your comments here, and Caleb's suggestion at standup, I've reworked the description a little. How do we feel about [this](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16338/admin/api/v3/#post-/groups/%7Bgroupid%7D/apps/%7Bappid%7D/import)? (Alas, our rendering of the openAPI doc strips my formatting - I tried to break that up into paragraphs but no joy.)

chickenbug:
@tsedgwick Just wanted to loop you in on this Docs PR. The idea was to document the App Import endpoint (but with ample warnings) since we document many parts of our admin API. This originated from a user question and discussion documented in [this slack thread](https://mongodb.slack.com/archives/C41SX3C4D/p1622132314167300).

I'm okay documenting the endpoint with adequate warnings, but I have some concerns about the level of detail in which we document the payload. As is I think this could become difficult to maintain as our product grows and changes. In my opinion I think we should document the endpoint, but not the request body, and just refer them to our app configuration docs.

Do you have any thoughts on this?

tsedgwick:
> @tsedgwick Just wanted to loop you in on this Docs PR. The idea was to document the App Import endpoint (but with ample warnings) since we document many parts of our admin API. This originated from a user question and discussion documented in [this slack thread](https://mongodb.slack.com/archives/C41SX3C4D/p1622132314167300).
> 
> I'm okay documenting the endpoint with adequate warnings, but I have some concerns about the level of detail in which we document the payload. As is I think this could become difficult to maintain as our product grows and changes. In my opinion I think we should document the endpoint, but not the request body, and just refer them to our app configuration docs.
> 
> Do you have any thoughts on this?

I agree with you on this @chickenbug. Since this isn't an endpoint we would rather users just interact with using the cli, we can be more vague and point them to the docs that you linked.

dacharyc:
> > @tsedgwick Just wanted to loop you in on this Docs PR. The idea was to document the App Import endpoint (but with ample warnings) since we document many parts of our admin API. This originated from a user question and discussion documented in [this slack thread](https://mongodb.slack.com/archives/C41SX3C4D/p1622132314167300).
> > I'm okay documenting the endpoint with adequate warnings, but I have some concerns about the level of detail in which we document the payload. As is I think this could become difficult to maintain as our product grows and changes. In my opinion I think we should document the endpoint, but not the request body, and just refer them to our app configuration docs.
> > Do you have any thoughts on this?
> 
> I agree with you on this @chickenbug. Since this isn't an endpoint we would rather users just interact with using the cli, we can be more vague and point them to the docs that you linked.

@drewdipalma , looping you in as we had previously discussed publishing docs for this endpoint. Do you have any thoughts on us providing here an endpoint description that points the users at the [App Configuration documentation](https://docs.mongodb.com/realm/config/), instead of documenting the endpoint request body directly?

@chickenbug I'm hoping you can help me understand a nuance I think I'm missing. You've suggested we should send users to the App Configuration documentation instead of documenting the endpoint request body directly. If everything that a developer needs to use this endpoint _is_ present in the App Configuration documentation, I'm missing the downside of documenting it here as a complete payload versus pointing users there to cobble together their own request body. The benefit of documenting it here would seem to be that it provides the correct structure and saves the developer time and potential misunderstandings. 

However, if everything that a developer needs to use this endpoint _is not_ present in the App Configuration documentation, I'm unclear that there's any benefit in adding this endpoint to our documentation at all. In that context, it would seem that we're handing the developers a potentially dangerous tool, and then giving them an incomplete manual. This seems problematic, given the warnings about how using this endpoint improperly can render a Realm app unusable. I'm hoping you or @tsedgwick could give me the engineering context I'm missing - thanks!

chickenbug:
@dacharyc To me, the downside is that we'll have to document and maintain a complex payload for an endpoint we really don't want users to be relying upon. The endpoint duplicates other documentation and would have to be updated when any changes to any other endpoints. I feel like at this level of detail it would become extremely easy to have the documented payload feature incorrect/outdated information.

After thinking about it more though, I do agree that we should probably document the rough shape of the payload to show the path to the individual resources since that will be hard to figure out. However I think we shouldn't document their inner exact details since that would be duplicated by the configuration documentation (eg. document the Functions and GraphQL keys but not nested fields like `private` and `use_natural_pluralization` )

dacharyc:
After discussions with engineering and product, we won't be documenting the Import App endpoint at this time.

PR #1175, opened 2021-06-25
(DOCSP-16511): Update Sync -> Development Mode documentation

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-16511

- [Concepts: Development Mode](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16511/sync/development-mode/): Largely novel content
- [Guides: Enable/Disable Development Mode](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16511/sync/enable-development-mode/): New procedures to disable development mode

### Review Guidelines


source/sync/concepts/development-mode.txt
```diff
@@ -0,0 +1,159 @@
...
+
+Development Mode in Workflows
+-----------------------------
+
+Because development mode disables schema validation and data access rules, 
+use it only in development. Following are {+service-short+} best practices 
+for using development mode in workflows.
+
+Use Separate Realm Apps for Development, Staging, and Production
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nlarew, line 93: The idea of separate apps for each env seems like a production workflow best practice more than it is specific to dev mode. Especially given that app environments are also a feature that you can use independent of dev mode - a "development" app won't necessarily always have "development mode" turned on.

Barring that, it seems like the major point here is similar to the following section, i.e. don't enable dev mode for a production app

source/sync/concepts/development-mode.txt
```diff
@@ -0,0 +1,159 @@
...
+issues can leave your production client applications buggy or unable to
+open.
+
+Do Not Use Production Data in Development
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ensure your development database does not contain production data. When 
+you enable development mode, {+service-short+} does not enforce data access 
+rules. If your development database contains production data, developer
+mode would leave that data vulnerable to inadvertent or malicious access.
```
nlarew, line 112: Also, dev mode in production would let anybody with your app id modify your schemas and break stuff, which is obviously no bueno

source/sync/concepts/development-mode.txt
```diff
@@ -0,0 +1,159 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If your team uses CI/CD workflows, add a step to your workflow to disable 
+development mode. You might set an automation to disable development mode
+when merging from development to testing, and again when merging to production.
+This is an additional safeguard against a developer inadvertently committing
+a config that contains ``developer_mode_enabled``: ``true``.
+
+You can update this in your :ref:`{+sync+} configuration 
+<appconfig-sync>` by setting ``development_mode_enabled`` to ``false``.
```
nlarew, line 124: I think this would be more relevant in the upcoming CI/CD workflow guide since it's advice specific to the CI/CD use case instead of a more general explanation.

source/sync/concepts/development-mode.txt
```diff
@@ -0,0 +1,159 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you enable development mode, {+service-short+} bypasses data access 
+rules and schema validation. This lets developers iterate quickly to develop 
+new features or fix bugs. However, this aspect of development mode means 
+it's vital to disable it in production. Leaving development mode enabled 
+in production can expose your app to security vulnerabilities. 
+
+Development Mode in Workflows
+-----------------------------
```
nlarew, line 86: This section has really good info - I'm wondering if it makes sense on this high level concept page in its current form though. We may be able to eliminate this section and shorten the page if we move some of it around. Check out my other comments on the subsections for more info.

source/sync/concepts/development-mode.txt
```diff
@@ -0,0 +1,159 @@
...
+------------
+
+Development Database
+~~~~~~~~~~~~~~~~~~~~
+
+When you enable Development Mode, you specify a database to store synced 
+objects. {+service-short+} creates new collections in this Development Mode 
+database for every type of synced object.
+
+For example, you specify a Development Mode database of ``myapp``. Your
```
nlarew, line 136: Can we put this in a `.. example::` and move it to the bottom of the subsection?

Right now it splits up the general definition into two halves and it isn't totally clear where the example ends and the general definition comes back in.

source/sync/concepts/development-mode.txt
```diff
@@ -0,0 +1,159 @@
...
+collection. 
+
+{+service-short+} continues creating new server-side schemas and collections 
+for each new object type. If you later add a ``Dog`` object, that object 
+will sync to a new ``myapp.Dog`` collection that {+service-short+} will 
+create.
+
+{+service-short+} does not enforce data access rules while in Development 
+Mode. Your Development Mode Database should not contain any production 
+or user data.
```
nlarew, line 149: Maybe we could cover the "don't enable dev mode for a production app" point here since we mention it again anyway? Could be inline text or a note.

nlarew (changes requested):
Really good stuff, a few comments across the pages and some ideas on how to streamline the concept page. Lemme know your thoughts!

source/sync/guides/enable-development-mode.txt
```diff
@@ -0,0 +1,67 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Overview
+--------
+
+This page contains the procedures to enable or disable development mode.
```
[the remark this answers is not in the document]
dacharyc: I put that there after discovering that directives, if they're the first thing on the page, are the content that shows up in the search results under the page title. I'd rather see a line describing the page contents than a seealso or a warning when I discover the page in the search results. I agree that it's a bit redundant, though - if we could specify this in metadata, that would be ideal. Meanwhile, if you have a suggestion for something else to put here, or if you think we should scrap it anyway and not worry about how the page displays in search results, I'll remove it. Thoughts @nlarew ?

source/sync/concepts/development-mode.txt
```diff
@@ -0,0 +1,159 @@
...
+:ref:`schemas <configure-your-data-model>` directly in client application 
+code.
+
+Design Schemas in Code
+~~~~~~~~~~~~~~~~~~~~~~
+
+With development mode enabled, you can design schemas directly in client 
+code. The objects and relationships that you define in your application 
+become your schema. :ref:`Realm SDKs <realm-sdks>` give you language-specific 
+tools to define and work with these objects.
```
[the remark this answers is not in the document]
dacharyc: I like it. Added!

source/sync/concepts/development-mode.txt
```diff
@@ -0,0 +1,159 @@
...
+In practice, here's how the schema definition process works with 
+development mode:
+
+1. Enable development mode.
+2. Sync {+realm+} files from your local {+client-database+}.
+3. The objects in the files become the definitions for your schema.
+
+After you have synced {+realm+} files to define your schema, you can view
+the schema through the Realm UI. Go to your Realm App -> :guilabel:`Data Access` 
+-> :guilabel:`Schema`.
```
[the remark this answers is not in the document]
dacharyc: That's fair. Linking to a guide makes sense, I think. I wanted to close the loop - i.e. it's good to know from a concept standpoint that once you've created a schema, you can view it and know that it was created correctly, etc. 

We should probably also create a guide for, er, what do we call it - translate your data model to SDKs? I've seen people get excited about the fact that they can generate data models for other languages once they have a schema defined in Sync - i.e. I write a client for my app in Swift, sync it to my Realm app, and then go to Build -> SDKs and copy the data models for JS/TS/Java/Kotlin without having to manually recreate them for cross-platform development. I don't know off the top of my head what the best name for that guide would be.

Should there be new tickets for these guides in the Sync Update epic, or should I put them together for this since it's all sort of related?

source/sync/concepts/development-mode.txt
dacharyc: Yes, that's fair. This section of recommendations was really to set the stage for the following sections, and I was specifically thinking about the case for CI/CD where you're using different apps for different stages in the SDLC, and you use automation to turn off development mode (i.e. have a step to set `development_mode_enabled` to `false` every time someone pushes to a non-development app). So it wasn't necessarily that a development app would always have "development mode" turned on, but more that a non-development app would always have "development mode" turned _off_.

If you think it's too much set up for not enough payout in the context of this page, I'm happy to remove the content.

source/sync/concepts/development-mode.txt
dacharyc: Fair, but this was more about development mode in the context of potentially exposing data that you're using in a development environment, not dev mode in production. I've worked for companies that have used production data in development, so they have "real" representation of what the data looks like... and that seems like a security vulnerability in this context.

source/sync/concepts/development-mode.txt
dacharyc: That's fair, and if the CI/CD workflow guide had been available, I'd have been happy to link to that with a recommendation to use CI/CD to automate turning off dev mode. Because that content doesn't yet exist, I wanted to provide some more recommendations here relevant to this context. I do think it's worth including _some_ recommendations about this on this page, even if it's not in this level of detail, because users should be thinking about this in this context. It also speaks a bit to the "best practices" conversation we keep having with Ian.

source/sync/concepts/development-mode.txt
dacharyc: I do think it's relevant to include some guidance around this in this "development mode" context. I'm happy to shorten the section and link out to other content when it exists, or make a separate ticket to add something like this to this section once that other content does exist, but I do think it should be here in some form.

source/sync/concepts/development-mode.txt
dacharyc: Does that really belong in this section? Since this is specifically about the development database, I wonder if that warning really belongs here again. I'm mostly trying to address the case like I've seen in some of my former companies where the org uses production data in development databases.

source/sync/guides/enable-development-mode.txt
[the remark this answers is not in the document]
nlarew: SGTM would definitely be nice to configure this in metadata down the road

source/sync/concepts/development-mode.txt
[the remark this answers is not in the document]
nlarew: Already have one! https://jira.mongodb.org/browse/DOCSP-16519

source/sync/concepts/development-mode.txt
nlarew: Definitely good advice - I think it'd make sense here but even more sense in a specific CI/CD guide section. I maybe can integrate it into the upcoming CI/CD page but not sure.

nlarew (approved):
LGTM! We can revisit the concept/guides hierarchy once everything fits neatly into it.

PR #1177, opened 2021-06-28
Sync redux

MongoCaleb:
## Pull Request Info

Looking for not only standard review feedback, but also things along the lines of:
- What do we want feature landing pages to look like?
- Is the tone too (in)formal?
- Would an image or two improve things or distract?
- etc.

### Jira

- https://jira.mongodb.org/browse/DOCSP-16510
- 

- https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/sync_redux/sync/

### DIFF
--- a/source/sync.txt
+++ b/source/sync.txt
@@ -63,9 +63,9 @@ the same regardless of network conditions. The :term:`{+service-short+} data mod
 makes it easy to manage a consistent schema of objects in your database
 and across all client platforms.
 
-The {+sync+} Philosophy
--------------------------
-There are 3 pillars to the {+sync+} philosophy:
+How {+sync+} Works
+------------------
+{+sync+} is built on 3 main pillars:
 
 - :ref:`Offline-first functionality <offline-first-overview>`
 - :ref:`Live objects <live-objects-overview>`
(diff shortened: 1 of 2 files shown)


source/sync.txt
```diff
@@ -94,50 +94,46 @@ the local {+realm+}. In effect, you continue to work with the
...
+synchronizing data. The :term:`sync protocol` is the same across platforms. 
+{+service+} handles passing the objects across the network automatically, so 
+you do not need to map from one platform's database to another, or serialize and 
+deserialize objects in order to send and receive them over the network.
+
+The main difficulty with offline-first data is ensuring that devices that are 
+offline for extended periods are able to resolve potential conflicts with 
+changes made to the same data. Ultimately, every client device must
+arrive at the same state when all changes have been synced. This is why {+sync+} 
+includes :ref:`automatic conflict resolution <conflict-resolution-overview>`.
```
cbush, line 112: I know this paragraph was originally here in the offline first section, but would it make more sense to be in the conflict resolution section?

source/sync.txt
```diff
@@ -94,50 +94,46 @@ the local {+realm+}. In effect, you continue to work with the
...
+The offline-first design leads to one of the main benefits of {+service+}: a 
+seamless user experience. Users interact with local data with zero latency 
+regardless of their moment-to-moment connectivity.
+
+Because {+sync+} is built into the {+service-short+} SDK, you do not need to write
+your own synchronization protocol and deal with connectivity issues when 
+synchronizing data. The :term:`sync protocol` is the same across platforms. 
+{+service+} handles passing the objects across the network automatically, so 
+you do not need to map from one platform's database to another, or serialize and 
+deserialize objects in order to send and receive them over the network.
```
cbush, line 106: Does this paragraph fit here or is there somewhere else it could go? I feel this section is running on a bit and might be getting away from being about "offline first".

source/sync.txt
```diff
@@ -28,27 +28,11 @@ Realm Sync Overview
...
-
-The Connectivity Problem
-------------------------
-
+What is {+sync+}?
+-------------------
 :term:`{+client-database+}` is a local, mobile database in
 which your app can store data. Eventually, most apps
-need to share that data with the outside world.
+need to share that data with other apps and the outside world.
```
cbush, line 35: This doesn't answer "what is Realm Sync?". It feels a bit jarring to have the Realm database defined ("is a...") in a section with that title. Would it make sense to open with "Realm Sync is..."?

source/sync.txt
```diff
@@ -158,27 +154,6 @@ example, insertions into lists.
...
-
-Atlas oplog:
-  {+sync+} requires access to the {+atlas+} oplog. For best results,
-  keep 48 hours of oplog for a cluster using {+sync-short+}. 
-
-.. seealso::
-
-   :atlas:`Atlas oplog </reference/atlas-oplog>`
 
 Summary
```
cbush, line 158: I think the summary might need an update to include what Realm Sync is first.

cbush (approved):
I like it a lot, reads much better! A few minor comments, nothing blocking.

source/sync.txt
MongoCaleb: I like it here because it makes a nice bridge to the conflict resolution section. Although this section is quite long...

cbush: That was the original reasoning as well. Ok!

PR #1181, opened 2021-06-29
(DOCSP-17245): MongoDB Realm Root Landing Page Update

nlarew:
### Jira

- (DOCSP-17245): MongoDB Realm Root Landing Page Update

- [MongoDB Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/lp-main/)

### Review Guidelines

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -27,7 +27,7 @@ Realm SDKs and MongoDB Realm application services.
 - MongoDB Realm is a hosted cloud platform that provides services to run
   your application logic and securely access and sync data. Services include
   Realm Sync, user authentication and authorization, serverless functions,
-  database triggers, and dynamic data access rules.
+  database triggers, dynamic data access rules, and more.
 
 .. note:: Feedback Appreciated
 
--- a/source/index.txt
+++ b/source/index.txt
@@ -15,8 +15,20 @@ MongoDB Realm
    MongoDB Stitch (Legacy) </stitch>
    Release Notes </release-notes>
 
+Welcome to the {+service+} documentation! You can use the guides and reference
+materials here to learn how to work with the two key components of Realm: the
+Realm SDKs and MongoDB Realm application services.
 
-Welcome to the {+service+} documentation!
+- The Realm SDKs are a set of free and open source client libraries that include
+  Realm Database, an embedded object database that can be used in place of
+  SQLite to build a mobile, desktop, or IoT application. You can also use the
+  SDKs to connect to MongoDB Realm and sync data between client applications and
+  MongoDB Atlas.
+
+- MongoDB Realm is a hosted cloud platform that provides services to run
+  your application logic and securely access and sync data. Services include
+  Realm Sync, user authentication and authorization, serverless functions,
+  database triggers, and dynamic data access rules.
 
 .. note:: Feedback Appreciated
 
--- a/source/index.txt
+++ b/source/index.txt
@@ -20,10 +20,9 @@ materials here to learn how to work with the two key components of Realm: the
 Realm SDKs and MongoDB Realm application services.
 
 - The Realm SDKs are a set of free and open source client libraries that include
-  Realm Database, an embedded object database that can be used in place of
-  SQLite to build a mobile, desktop, or IoT application. You can also use the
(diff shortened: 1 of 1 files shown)


source/index.txt
```diff
@@ -15,8 +15,20 @@ MongoDB Realm
    MongoDB Stitch (Legacy) </stitch>
    Release Notes </release-notes>
 
-
-Welcome to the {+service+} documentation!
+Welcome to the {+service+} documentation! You can use the guides and reference
```
MongoCaleb, line 18: Personal issue: I don't like welcoming people to the documentation. Maybe because I generally go to documentation to find answers. 🤷

source/index.txt
```diff
@@ -15,8 +15,20 @@ MongoDB Realm
...
+materials here to learn how to work with the two key components of Realm: the
+Realm SDKs and MongoDB Realm application services.
+
+- The Realm SDKs are a set of free and open source client libraries that include
+  Realm Database, an embedded object database that can be used in place of
+  SQLite to build a mobile, desktop, or IoT application. You can also use the
```
MongoCaleb, line 24: Do we need to mention SQLite? Seems like we should just present Realm as a solution in and of itself, rather than as a replacement.

source/index.txt
```diff
@@ -15,8 +15,20 @@ MongoDB Realm
...
+  SDKs to connect to MongoDB Realm and sync data between client applications and
+  MongoDB Atlas.
+
+- MongoDB Realm is a hosted cloud platform that provides services to run
+  your application logic and securely access and sync data. Services include
```
MongoCaleb, line 29: "... to run much of your application logic server-side and ..."

source/index.txt
```diff
@@ -15,8 +15,20 @@ MongoDB Realm
...
+  Realm Sync, user authentication and authorization, serverless functions,
+  database triggers, and dynamic data access rules.
```
mongomoe, line 31: Why the specific mention of database triggers but not authentication or scheduled? 

If it's to keep this sentence short, would any of the following suggestions be more accurate while preventing readers from having to read large and wordy sentence:
1. switch it to a bulleted list
or 
2. change the beginning fragment to "Services include but are not limited to..."
or
3.  change the ending fragment to "... database triggers, dynamic data access rules, and more"?

source/index.txt
```diff
@@ -15,8 +15,20 @@ MongoDB Realm
...
    Release Notes </release-notes>
 
-
-Welcome to the {+service+} documentation!
+Welcome to the {+service+} documentation! You can use the guides and reference
+materials here to learn how to work with the two key components of Realm: the
+Realm SDKs and MongoDB Realm application services.
+
+- The Realm SDKs are a set of free and open source client libraries that include
+  Realm Database, an embedded object database that can be used in place of
```
mongomoe, line 23:
```suggestion
  Realm Database, an embedded object database that can be used as an alternative to
```

Nitpick, but I feel like "can be used in the place of" sounds kind of off (can't exactly put my finger on why)

mongomoe (approved):
LGTM 👍  w/ a couple suggestions

source/index.txt
mongomoe: This is a conflicting opinion to my comment https://github.com/mongodb/docs-realm/pull/1181#discussion_r660855814 (which suggests slightly altering the mention to SQLite rather than removing it), but I'm okay with whatever approach you choose fyi.

source/index.txt
nlarew: Is the thrust of this that we should add "much of" or "server side"?

source/index.txt
nlarew: I could go either way (and we seem to across our various docs projects) but since the text is written in a friendlier way I think I'll leave the greeting for now.

source/index.txt
nlarew: I actually had "and more" for a bit so I'll add it back

source/index.txt
MongoCaleb: both, actually. The way it reads now, it suggests that all of the logic can be run ... somewhere.

PR #1189, opened 2021-07-01
(DOCSP-17280): Configuration Equality Check Error -- Android SDK

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-17280

- [Android SDK - Troubleshooting](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17280/sdk/android/troubleshooting/#configurations-cannot-be-different-if-used-to-open-the-same-file)

### Review Guidelines

### DIFF
--- a/source/sdk/android/troubleshooting.txt
+++ b/source/sdk/android/troubleshooting.txt
@@ -61,12 +61,18 @@ Configurations Cannot be Different if Used to Open the Same File
 ----------------------------------------------------------------
 
 {+client-database+} runs checks whenever you open a {+realm+} file to
-avoid corruption. In most cases, this works well to ensure that you don't
-open one {+realm+} with two different schemas, durability levels, or
-writability settings, since the SDK shares underlying connections to
-the same file. However, some settings are not properly compared by this
-compatibility check, such as ``initialData()`` or ``compactOnLaunch()``.
-To avoid this error when you're using settings like these, you can either:
+avoid corruption. In order to avoid accidentally opening a {+realm+}
+file with incompatible settings, the SDK uses Java's ``equals()`` method
+to compare ``RealmConfiguration`` objects. This prevents the SDK from
+opening a single {+realm+} file with different schemas, durability levels,
+or writability settings. However, configurations that include lambda
+functions, such as those passed to
+:java-sdk:`initialData() <io/realm/RealmConfiguration.Builder.html#initialData-io.realm.Realm.Transaction->`
+and
+:java-sdk:`compactOnLaunch() <io/realm/RealmConfiguration.Builder.html#compactOnLaunch-io.realm.CompactOnLaunchCallback-`,
+can break this ``equals()`` comparison, since two different lambdas are
+never considered equal using Java's built-in comparison.
+To avoid this error when using lambdas, you can either:
 
 1. Store a single configuration statically in your application, so that
    separate {+realm+} instantiations use the exact same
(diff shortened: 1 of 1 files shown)


source/sdk/android/troubleshooting.txt
```diff
@@ -56,3 +56,37 @@ completely anonymous and helps us improve the product by flagging:
...
 the ``REALM_DISABLE_ANALYTICS`` environment variable to ``true``.
+
+Configurations Cannot be Different if Used to Open the Same File
+----------------------------------------------------------------
+
+{+client-database+} runs checks whenever you open a {+realm+} file to
+avoid corruption. In most cases, this works well to ensure that you don't
+open one {+realm+} with two different schemas, durability levels, or
+writability settings, since the SDK shares underlying connections to
+the same file. However, some settings are not properly compared by this
```
cmelchior, line 67: `not properly compared`
This wording is a little wrong, the problem is that we are "too" correct, in the sense that we use Java equal semantics, but people don't understand that creating the same anonymous object twice actually breaks the default `equals` implementation

Suggested change:

In order to avoid accidentally open a Realm file with incompatible settings, Realm uses the Java `equals()` method to compare `RealmConfiguration`s. In most cases this works fine, but for settings that accept a lamda, it can turn problematic as two different lambdas are not considered equal using the standard Java `equal` semantics. Examples of this are `initialData()` and `compactOnLaunch()`

To avoid this error you ...

cmelchior (approved):
I added a suggestion for rewording a smaller detail. Otherwise, it looks fine.

source/sdk/android/troubleshooting.txt
```diff
@@ -56,3 +56,43 @@ completely anonymous and helps us improve the product by flagging:
...
+{+client-database+} runs checks whenever you open a {+realm+} file to
+avoid corruption. In order to avoid accidentally opening a {+realm+}
+file with incompatible settings, the SDK uses Java's ``equals()`` method
+to compare ``RealmConfiguration`` objects. This prevents the SDK from
+opening a single {+realm+} file with different schemas, durability levels,
+or writability settings. However, configurations that include lambda
+functions, such as those passed to
+:java-sdk:`initialData() <io/realm/RealmConfiguration.Builder.html#initialData-io.realm.Realm.Transaction->`
+and
+:java-sdk:`compactOnLaunch() <io/realm/RealmConfiguration.Builder.html#compactOnLaunch-io.realm.CompactOnLaunchCallback-`,
```
biniona-mongodb:
```suggestion
:java-sdk:`compactOnLaunch() <io/realm/RealmConfiguration.Builder.html#compactOnLaunch-io.realm.CompactOnLaunchCallback>`,
```

Issue: I think this link may be broken on the page.

source/sdk/android/troubleshooting.txt
```diff
@@ -56,3 +56,43 @@ completely anonymous and helps us improve the product by flagging:
...
+can break this ``equals()`` comparison, since two different lambdas are
+never considered equal using Java's built-in comparison.
+To avoid this error when using lambdas, you can either:
+
+1. Store a single configuration statically in your application, so that
+   separate {+realm+} instantiations use the exact same
```
biniona-mongodb, line 78:
```suggestion
   separate {+realm+} instances use the exact same
```

Question: Should this be instances? I don't know this content well so I can't say for sure what it should be. If you feel instantiations is better feel free to ignore this comment.

biniona-mongodb (approved):
LGTM with comments!

source/sdk/android/troubleshooting.txt
nathan-contino-mongo: Fixed it already :) thanks!

source/sdk/android/troubleshooting.txt
nathan-contino-mongo: Solid suggestion. This was definitely just me being dumb and using a bigger word because it looked fancier. Thanks!

PR #1203, opened 2021-07-06
(DOCSP-15121): Add note about AWS entry size limit

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-15121

- [Send Trigger Events to AWS EventBridge](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-15121/triggers/eventbridge/#procedure): New Tip under the Procedure heading linked here

### Review Guidelines

### DIFF
--- a/source/triggers/eventbridge.txt
+++ b/source/triggers/eventbridge.txt
@@ -35,9 +35,8 @@ Procedure
 
    For more information, see the `AWS Documentation to calculate Amazon
    PutEvents event entry size <https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-putevent-size.html>`_.
-   You can use :ref:`Project Expression <create-a-database-trigger>`, a 
-   configuration setting on the trigger, to reduce the size of the entry 
-   you send to Eventbridge.
+   You can configure a :ref:`project expression <create-a-database-trigger>` 
+   on the trigger to reduce the size of the entry you send to Eventbridge.
 
 .. include:: /includes/steps/eventbridge.rst
 
--- a/source/triggers/eventbridge.txt
+++ b/source/triggers/eventbridge.txt
@@ -29,6 +29,16 @@ Trigger, and associating the Trigger event source with an event bus.
 Procedure
 ---------
 
+.. note::
+
+   AWS accepts total entry sizes of less than 256 KB.
+
+   For more information, see the `AWS Documentation to calculate Amazon
+   PutEvents event entry size <https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-putevent-size.html>`_.
+   You can use :ref:`Project Expression <create-a-database-trigger>`, a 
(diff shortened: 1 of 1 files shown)


source/triggers/eventbridge.txt
```diff
@@ -29,6 +29,16 @@ Trigger, and associating the Trigger event source with an event bus.
 Procedure
 ---------
 
+.. note::
+
+   AWS accepts total entry sizes of less than 256 KB.
+
+   For more information, see the `AWS Documentation to calculate Amazon
+   PutEvents event entry size <https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-putevent-size.html>`_.
+   You can use :ref:`Project Expression <create-a-database-trigger>`, a
```
nlarew, line 38: nit: the phrasing here feels off. A "project expression" is a singular thing (the expression) as opposed to a proper name for the configuration parameter

"you can configure a project expression on the trigger..."

I'd actually prefer to call it a _projection_ expression but I think they use _project_ in the UI so that's fine

source/triggers/eventbridge.txt
```diff
@@ -29,6 +29,16 @@ Trigger, and associating the Trigger event source with an event bus.
 Procedure
 ---------
 
+.. note::
+
+   AWS accepts total entry sizes of less than 256 KB.
```
nlarew, line 34: nit: this is a per-entry limit with one entry per trigger event right? Maybe:

"The AWS put entry for an EventBridge trigger event must be less than 256 KB"

nlarew (approved):
Some nits otherwise LGTM

source/triggers/eventbridge.txt
dacharyc: The [AWS documentation](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-putevent-size.html) says you can batch events, and the total entry must be less than 256 KB. It's unclear to me if someone using our triggers would be batching events, which is why I left it at "total entry size." If you think that's not likely and it's clearer to change it to your wording, I'm open to that @nlarew - what are your thoughts?

source/triggers/eventbridge.txt
dacharyc: I've updated to your suggested wording here. FWIW, we call it a "Project Expression" in our docs and the linked Atlas docs, which is why I kept that wording; it's more consistent for users who go all the way to the Atlas docs.

source/triggers/eventbridge.txt
nlarew: I don't think we do any batching but it's definitely possible. Can we follow up with eng to find out? If not then "total entry size" makes sense though (separately) we may also want to document batching somewhere.

dacharyc: Engineering confirms that we do not do any batching, so I've incorporated your suggestion here!

PR #1221, opened 2021-07-12
(DOCSP-17507): Kotlin/Android SDK PLP split

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-17507

- [SDK PLP](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17507/sdk/)

### Review Guidelines


source/sdk.txt
```diff
@@ -21,20 +21,20 @@ Realm has open-source SDKs available for most popular languages, frameworks, and
...
-      :url: https://docs.mongodb.com/realm/sdk/kotlin-multiplatform/
-      :icon: /images/icons/kotlin_sdk.svg
-      :icon-alt: Java and Kotlin SDK icon
+      :headline: Android SDK
+      :url: https://docs.mongodb.com/realm/sdk/android/
+      :icon: /images/icons/android_sdk.svg
+      :icon-alt: Android SDK icon
 
-      Build Android and JVM applications. Realm data maps directly to the POJOs and classes in your app.
+      Build applications for Android. Realm data maps directly to the classes in your app.
```
nlarew, line 29: Should we mention that this is the _Java_ SDK (w/ implied Kotlin compatibility)?

Also can you not use POJOs?

source/sdk.txt
```diff
@@ -21,20 +21,20 @@ Realm has open-source SDKs available for most popular languages, frameworks, and
    :columns: 2
 
    .. card::
-      :headline: Java and Kotlin SDKs
-      :url: https://docs.mongodb.com/realm/sdk/kotlin-multiplatform/
-      :icon: /images/icons/kotlin_sdk.svg
-      :icon-alt: Java and Kotlin SDK icon
+      :headline: Android SDK
+      :url: https://docs.mongodb.com/realm/sdk/android/
+      :icon: /images/icons/android_sdk.svg
```
nlarew, line 26: Did you source this icon? This is definitely good for now but I think we should run this by design to see if we already have an android icon in our set or if we need to add one officially.

nlarew (approved):

source/sdk.txt
nathan-contino-mongo: Since there are some arbitrary-ish restrictions on realm object data classes in the SDK (for instance, must extend or implement realmobject or realmmodel, needs to use annotations for a lot of functionality, cannot create class hierarchies) I figure it's better to avoid the POJO term since some folks would call realm object classes not POJO.

nathan-contino-mongo: And as for java/kotlin compatibility, I think it's fair to omit that since that comes with the territory of native Android applications. While the SDK is written in Java, there are also Kotlin extensions so that term is maybe a bit muddy.

source/sdk.txt
nathan-contino-mongo: I'll double-check the centralized icon sets to make sure there isn't one. This one is available for use by anyone in the material UI docs.

nathan-contino-mongo: Wasn't able to find any android icon in our set, guessing Google is the best source for that anyway since it's their IP.

PR #1233, opened 2021-07-15
(DOCSP-16520): [Guide]: Pause or Terminate Sync

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-16520

- [Pause or Terminate Sync](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16520/sync/pause-or-terminate-sync/)

### Review Guidelines


source/sync/pause-or-terminate-sync.txt
```diff
@@ -0,0 +1,219 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+You can temporarily pause, permanently terminate, or re-enable {+sync+}
```
mongomoe, line 19: This overview is confusing to me. 

Are the following choices alternative actions that the reader can perform:
 1. temporarily pause sync
 2. permanently terminate sync
3. re-enable sync 

or is it:
1. temporarily pause sync
2. permanently terminate sync & then re-enable sync 


The first sentence implies the first set of actions. On the other hand, the 2 sentences following a similar structure (you can do {some action} to alter some options), implying that the second set of actions above is more accurate.

source/sync/pause-or-terminate-sync.txt
```diff
@@ -0,0 +1,219 @@
...
+for your {+app+}.
+
+You can temporarily :ref:`pause <pause-sync>` {+sync-short+} before pausing your 
+cluster, and then :ref:`resume <resume-sync>` without resetting your 
+{+sync-short+} configuration or losing {+sync-short+} metadata.
+
+You can permanently :ref:`terminate <terminating-realm-sync>` and 
+:ref:`re-enable <re-enable-realm-sync>` {+sync-short+} for troubleshooting, or
+if you want to change your {+sync-short+} configuration.
```
mongomoe, line 28: is sync configuration the same as sync metadata?

source/sync/pause-or-terminate-sync.txt
```diff
@@ -0,0 +1,219 @@
...
+   If you need to pause your cluster, pause {+sync+} first. Otherwise, you
+   must :ref:`terminate <terminating-realm-sync>` and :ref:`re-enable 
+   <re-enable-realm-sync>` {+sync-short+}, and perform a client reset.
+
+Pausing {+sync-short+} rejects any incoming client connections. This means that 
+pausing {+sync+} for your {+app+} stops syncing changes for all clients. To 
+pause {+sync-short+} from the client side, using logic that situationally 
+pauses {+sync-short+} during a session, see your preferred SDK:
+
+.. tabs-realm-sdks::
```
mongomoe, line 50: Missing React Native SDK here

mongomoe: I know that technically it's the same as the API Node.js SDK but the specific page might have react native specific context

source/sync/pause-or-terminate-sync.txt
```diff
@@ -0,0 +1,219 @@
...
+
+Overview
+--------
+
+You can temporarily pause, permanently terminate, or re-enable {+sync+}
+for your {+app+}.
+
+You can temporarily :ref:`pause <pause-sync>` {+sync-short+} before pausing your 
+cluster, and then :ref:`resume <resume-sync>` without resetting your 
+{+sync-short+} configuration or losing {+sync-short+} metadata.
```
mongomoe, line 24: I'm a bit confused on the difference between sync configuration and sync metadata, I'd imagine readers would be as well. This may be worth clarification.

source/sync/pause-or-terminate-sync.txt
```diff
@@ -0,0 +1,219 @@
...
+
+.. include:: /includes/steps/pause-sync-ui.rst
+
+.. _resume-sync:
+
+Resume Sync
+-----------
+
+After you have paused {+sync+}, you can resume it to begin syncing again for
+all clients. If your :manual:`oplog </core/replica-set-oplog/>` rolls past
```
mongomoe, line 103: What might cause your oplog to roll past the time that you paused sync? Would this occur when trying to resume sync? If that's the case shouldn't we mention it in an early step so users can preemptively avoid this?

mongomoe (commented):
After reading this whole page, I'm kind of confused why "pause & resume sync" and "terminate & reenable sync" aren't two separate pages. Seems like they are two separate sets of actions with two separate results. 

Perhaps these pages can be split up, and then in the overview of "pause & resume sync" you could have an admonition that says 'if you're looking to change sync config settings completely, check out "terminate & reenable sync" & something similar for the other page

source/sync/pause-or-terminate-sync.txt
dacharyc: Sorry about that, @mongomoe ! I copied this from another page and didn't even think to check that all the SDKs were there. But this means it's missing from another page, and I can't remember where I copied it from. I'll add it here!

source/sync/pause-or-terminate-sync.txt
dacharyc: That's fair. It's a bit hand-wavey right now because you can pause and re-enable sync as long as your oplog doesn't roll past the time you paused sync. At that point, you've got no choice but to terminate and re-enable sync. I'll see what I can do about clarifying this without making it too confusing up-front in the intro.

source/sync/pause-or-terminate-sync.txt
dacharyc: Fair point! Especially for users who are coming from the SDK and may think of configuration in that context. I'll add a link and some more details here.

dacharyc:
> After reading this whole page, I'm kind of confused why "pause & resume sync" and "terminate & reenable sync" aren't two separate pages. Seems like they are two separate sets of actions with two separate results.
> 
> Perhaps these pages can be split up, and then in the overview of "pause & resume sync" you could have an admonition that says 'if you're looking to change sync config settings completely, check out "terminate & reenable sync" & something similar for the other page

There are a couple of reasons I went with one page for this. I think users think "I want to stop Sync" and may have a few different ideas of what that looks like - either temporarily pausing it, terminating it, or stopping it for a specific client or based on user logic (i.e. sync for paid users, not syncing free users). I wanted to present all those options so they could decide which one they _actually_ want to do. Additionally, I'm told that we might remove the option to pause sync, so I was trying to avoid the overhead of creating a new page for that and then having to do redirects if that page goes away. If you think it's too confusing as it is, though, I can split it up - what do you think?

source/sync/pause-or-terminate-sync.txt
dacharyc: Fair point, I'll add more details around this - thanks!

source/sync/pause-or-terminate-sync.txt
mongomoe: I see, thanks for clarifying. A smaller nitpicks on the updated intro:

"You might want to temporarily pause Sync if you need to pause your cluster." could be more decisive if you said something like: "Temporarily pause Sync to pause your cluster" (this would also parallel the next sentence on terminating sync, which might further show the readers that these are two alternative paths)

source/sync/pause-or-terminate-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+
+Overview
+--------
+
+If you want to disable {+sync+} for your app, you can temporarily pause or 
+permanently terminate it. You can also re-enable {+sync-short+} after 
+disabling it.
+
+You might want to temporarily :ref:`pause <pause-sync>` {+sync-short+} if 
+you need to pause your cluster. With a short, temporary pause, you can
```
mongomoe, line 24:
```suggestion
you need to pause your cluster. With a temporary pause, you can 
```
nitpick: "short" is not needed here (since it is redundant to temporary, and also is a bit generic since it doesn't specify how long short really is)

source/sync/pause-or-terminate-sync.txt
dacharyc: Got it, I've made that change here, thanks!

source/sync/pause-or-terminate-sync.txt
dacharyc: Fair, I've made that change - thanks!

source/sync/pause-or-terminate-sync.txt
mongomoe: maybe from: https://docs.mongodb.com/realm/reference/terminating-and-reenabling-realm-sync/ ?

source/sync/pause-or-terminate-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+permanently terminate it. You can also re-enable {+sync-short+} after 
+disabling it.
+
+You might want to temporarily :ref:`pause <pause-sync>` {+sync-short+} if 
+you need to pause your cluster. With a short, temporary pause, you can 
+:ref:`resume <resume-sync>` without resetting your :ref:`{+sync-short+} 
+configuration <appconfig-sync>` or losing {+sync-short+} metadata.
+
+You can permanently :ref:`terminate <terminating-realm-sync>` and 
+:ref:`re-enable <re-enable-realm-sync>` {+sync-short+} for troubleshooting, or
```
mongomoe, line 29:
```suggestion
:ref:`re-enable <re-enable-realm-sync>` {+sync-short+} for troubleshooting or
```

unneeded comma

source/sync/pause-or-terminate-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+
+You can pause {+sync+} for a period of time without disabling
+it entirely. When you pause {+sync-short+}, you stop syncing changes between 
+Atlas and your application. 
+
+.. important::
+
+   If you need to pause your cluster, pause {+sync+} first. Otherwise, you
+   must :ref:`terminate <terminating-realm-sync>` and :ref:`re-enable 
+   <re-enable-realm-sync>` {+sync-short+}, and perform a client reset.
```
mongomoe, line 45:
```suggestion
   <re-enable-realm-sync>` {+sync-short+} and perform a client reset.
```

mongomoe (approved):
LGTM 👍  with a couple nitpick suggestions and grammatical fixes

PR #1234, opened 2021-07-15
(DOCSP-17499): Mid-July 2021 Release

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-17499

- [Release Notes](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17499/release-notes/backend/)

### Review Guidelines

### DIFF
--- a/source/release-notes/changelogs/backend/changelog-backend-2021.rst
+++ b/source/release-notes/changelogs/backend/changelog-backend-2021.rst
@@ -4,7 +4,7 @@
 ~~~~~~~~~~~~~~~~~~~~
 
 - Improved usability of the Admin UI Dashboard.
-- Released :ref:`mongodb-cli 2.0 <realm-cli>`.
+- Released :ref:`mongodb-realm-cli 2.0 <realm-cli>`.
 
 02 July 2021 Release
 ~~~~~~~~~~~~~~~~~~~~
--- a/source/release-notes/changelogs/backend/changelog-backend-2021.rst
+++ b/source/release-notes/changelogs/backend/changelog-backend-2021.rst
@@ -1,5 +1,11 @@
 .. _backend_20210702:
 
+15 July 2021 Release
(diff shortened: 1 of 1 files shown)


dacharyc (approved):
Hmm, it's unfortunate that "Improved usability" doesn't have any details, but I guess it's late for us to solicit more info about that given the timing of the release.

LGTM! But going forward, I wonder if we can get more details for something like `released mongodb-cli 2.0`, like a summary of what's in the release, or a blog post or something with more details.

Actually, small nit - wouldn't we prefer `mongodb-realm-cli` (the name of the package) or `realm-cli` (what we call it in the docs)?

nathan-contino-mongo:
> Hmm, it's unfortunate that "Improved usability" doesn't have any details, but I guess it's late for us to solicit more info about that given the timing of the release.

@sumedham Can you provide any more info on the usability improvements? We could always divide this up into 2-3 bullet points if we knew more, and that would be a lot more useful for readers and users.

> I wonder if we can get more details for something like `released mongodb-cli 2.0`, like a summary of what's in the release, or a blog post or something with more details.

I would say that the link to the cli section ought to answer those questions (once we roll out 2.0, that section will update to reflect those changes). Unfortunately it's such a large change I'm not sure we can really describe it all in release notes, plus that might be a level of detail too deep for something non-cloud when these are, after all, the cloud release notes. Personally I would prefer to have that described in mongodb-realm-cli release notes somewhere else, but we haven't got that kind of detail from engineering yet.

> Actually, small nit - wouldn't we prefer `mongodb-realm-cli` (the name of the package) or `realm-cli` (what we call it in the docs)?

Good catch, fixed this. Thanks for noticing!

nathan-contino-mongo:
Chatted with Sumedha, LGTM according to her (she can't currently comment due to some Github authentication issues). Rolling out now since the Cloud release is out.

PR #1235, opened 2021-07-15
Add Pause example for .NET

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-17570

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-17570/sdk/dotnet/examples/sync-changes-between-devices/#pause-or-resume-a-sync-session

### DIFF
--- a/source/sdk/dotnet/examples/sync-changes-between-devices.txt
+++ b/source/sdk/dotnet/examples/sync-changes-between-devices.txt
@@ -74,19 +74,27 @@ The following code creates a new ``Task`` object and writes it to the {+realm+}:
 
 Pause or Resume a Sync Session
 ------------------------------
-
-You can pause (stop) a sync 
-:dotnet-sdk:`Session <reference/Realms.Sync.Session.html>` 
-by calling the 
+Opening a synced {+realm+} starts a {+sync-short+} 
+:dotnet-sdk:`Session <reference/Realms.Sync.Session.html>`. To pause synchronization 
+for a specific session, you can call the 
 :dotnet-sdk:`Stop() <reference/Realms.Sync.Session.html#Realms_Sync_Session_Stop>` 
-method. Calling ``Stop()`` stops any synchronization with the server until the 
-Realm is fully closed and then re-opened. Synchronization can be re-enabled by 
-calling the 
-:dotnet-sdk:`Start()  <reference/Realms.Sync.Session.html#Realms_Sync_Session_Start>` 
-method, as shown in the following code block:
+method on the session.
 
-.. literalinclude:: /examples/generated/dotnet/Examples.codeblock.pause-synced-realm.cs
-   :language: csharp
+When you then call the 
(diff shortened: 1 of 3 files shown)


source/sdk/dotnet/examples/sync-changes-between-devices.txt
```diff
@@ -72,8 +72,22 @@ The following code creates a new ``Task`` object and writes it to the {+realm+}:
...
-.. ------------------------------
+Pause or Resume a Sync Session
+------------------------------
+
+You can pause (stop) a sync 
+:dotnet-sdk:`Session <reference/Realms.Sync.Session.html>` 
+by calling the 
+:dotnet-sdk:`Stop() <reference/Realms.Sync.Session.html#Realms_Sync_Session_Stop>` 
+method. Calling ``Stop()`` stops any synchronization with the server until the 
+Realm is fully closed and then re-opened. Synchronization can be re-enabled by
```
dacharyc, line 83: There's some nuance here that I wonder if we should try to capture... "until the Realm is fully closed and then re-opened" _implies_ that sync automatically starts again when re-opening a realm. Then the next line adds that you can re-enable sync by calling `Start()`. I think that's clear enough for most people that sync starts again when you open the session, _or_ you can call `Start()`, but I wonder if being more explicit might help folks who may have questions based on this language. 

I wonder if it would help to be more explicit about what a sync session is, when it starts, when it stops, and the fact that pausing a sync session for one realm does not pause all sync sessions for the client? Maybe something like:
> Opening a synced {+realm+} starts a {+sync-short+} :dotnet-sdk:`Session <reference/Realms.Sync.Session.html>`. You can call ``Stop()`` to pause the synchronization session with the server. When you call ``Start()``, the {+sync-short+} session resumes. Closing the {+realm+} ends that sync session. 
>
>If you open a new synced {+realm+} while you have another {+sync-short+} session paused, the new {+realm+}'s {+service-short+} session is unaffected; it will not be paused. You must manually call ``Stop()`` and ``Start()`` for every {+realm+} whose {+sync-short+} session you want to pause.

If we decide to adopt language like this, it should probably be ported to all the SDKs. Maybe an `include` that doesn't include specific method names just to communicate this concept?

dacharyc (approved):
LGTM, but seeing this made me wonder if we should improve/clarify a bit more about what exactly this functionality does. Feel free to merge as-is, or take my comment if you think it would benefit from some additional clarity.

source/sdk/dotnet/examples/sync-changes-between-devices.txt
MongoCaleb: @dacharyc - Thanks for the insights. I'd love for you to take another pass at the changes I've made!

dacharyc (approved):
Nice, I think this clarifies perfectly. LGTM!

PR #1255, opened 2021-07-27
Use bluehawk state final in README

cbush:
- This prevents the CI from failing out due to no usage of state final


nathan-contino-mongo (approved):
LMAO LGTM WTF BBQ

PR #1263, opened 2021-07-29
Change monospace to code blocks to improve table layout

dacharyc:

- [Supported Property Types](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/fix-supported-property-types/sdk/ios/data-types/supported-property-types/)

### Review Guidelines


nathan-contino-mongo (approved):
I don't want to say "LGTM" because it's awful that we have to use a hacky workaround instead of actually fixing monospace text in tables. But it is what it is. LG*TM.

* "G" where "G" means something other than "good". I'm gonna go with "gauche" here.

PR #1299, opened 2021-08-16
(DOCSP-17170): [Concept] Client Resets

nlarew:
### Jira

- (DOCSP-17170): [Concept] Client Reset

- [Client Resets](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/sync-client-reset/sync/client-resets/)

### Review Guidelines

### DIFF
--- a/source/sync/client-resets.txt
+++ b/source/sync/client-resets.txt
@@ -1,0 +1,57 @@
+.. _client-resets:
+
+=============
+Client Resets
+=============
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+A **client reset** is an error scenario where a client application that uses a
+Realm SDK must erase all local data and download a new copy of the data from
+MongoDB Atlas before it can continue to sync data. Clients in this state may
+continue to run and save data locally but will not send or receive sync
+changesets until the client reset is complete.
+
+Client reset scenarios are generally rare and occur when the server's copy of a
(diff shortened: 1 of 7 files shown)


source/sync/client-resets.txt
```diff
@@ -0,0 +1,57 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+A **client reset** is an error scenario where a client application that uses a
+Realm SDK must erase all local data and download a new copy of the data from
+MongoDB Atlas before it can continue to sync data. Clients in this state may
+continue to run and save data locally but will not send or receive sync
```
dacharyc, line 21: Nit, but I wonder if we should add a warning here, or change this language a bit... when I read this, I see that data is continuing to be saved locally, and can send and receive sync changesets after the client reset is complete. On the client side, though, we warn that:
> A client reset erases all local data and downloads a new copy of the data stored in MongoDB Atlas. Performing a client reset loses all local changes made since the client last successfully synced.

I see that you mention lower down the page that you have to write a custom client reset error handler if you want to preserve local changes. But that feels to me almost like it conflicts with this, or at least that this intro paragraph doesn't communicate effectively that this is a serious error that could result in client data loss if not handled properly. What do you think?

dacharyc (approved):
LGTM, with a caveat that I think we may want to reword the intro a bit.

source/sync/client-resets.txt
nlarew: Updated with some more info + a spooky warning

dacharyc: Nice, I think that does the job!

PR #1301, opened 2021-08-16
(DOCSP-16514): [Concept] Sync Permissions

nlarew:
### Jira

- (DOCSP-16514): [Concept] Sync Permissions

- [Sync Permissions](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/sync-permissions/sync/permissions/)

### Review Guidelines

### DIFF
--- a/source/sync/permissions.txt
+++ b/source/sync/permissions.txt
@@ -55,17 +55,20 @@ Permission Strategies
 ---------------------
 
 You can structure your read and write permission expressions as a set of
-permission strategies that apply to the your :ref:`partition strategy
+permission strategies that apply to your :ref:`partition strategy
 <partition-strategies>`. The following strategies outline common approaches that
 you might take to define sync read and write permissions for your app.
 
 Global Permissions
 ~~~~~~~~~~~~~~~~~~
 
-You can define global permissions that apply to all users for all
-partitions. To define a global read or write permission, specify a boolean value
-or a :ref:`JSON expression <expressions>` that always evaluates to the same
-boolean value.
+You can define global permissions that apply to all users for all partitions.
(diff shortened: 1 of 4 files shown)


source/sync/permissions.txt
```diff
@@ -15,20 +15,52 @@ Define Sync Permissions
...
+read permissions, so any user that can modify data can also view that data
+before and after it's modified.
+
+.. _sync-permission-strategies:
+
+Permission Strategies
+---------------------
+
+You can structure your read and write permission expressions as a set of
+permission strategies that apply to the your :ref:`partition strategy
```
dacharyc, line 58:
```suggestion
permission strategies that apply to your :ref:`partition strategy
```

source/sync/permissions.txt
```diff
@@ -15,20 +15,52 @@ Define Sync Permissions
...
+<partition-strategies>`. The following strategies outline common approaches that
+you might take to define sync read and write permissions for your app.
 
 Global Permissions
-------------------
+~~~~~~~~~~~~~~~~~~
 
 You can define global permissions that apply to all users for all
```
dacharyc, line 65: Something I wondered here was whether any of these permissions take precedence. Like, can you define a global permission of `false`, which according to the docs means that no user can access data in the partition, and then explicitly give an individual user read or write access to the partition? Or does setting a global permission apply everywhere, and override any individual user permissions? I think if there is any information we have about how these rules interact, we should provide it here.

dacharyc (approved):
Overall, looks great. I did spot an extra word, which I suspect was a remnant of restructuring a sentence. I also had a question about how these permissions interact, and I wonder if there's any additional information we can provide users about that.

PR #1308, opened 2021-08-19
(DOCSP-18233): Provide CLI v2 alternative for deprecated flag when deploying with GitHub

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-18233

- [Deploy Automatically with GitHub](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18233/deploy/deploy-automatically-with-github/#initialize-the-repository): "Important" directive in "Step 3: Initialize the Repository"

### Review Guidelines

### DIFF
--- a/source/includes/steps-github-deploy.yaml
+++ b/source/includes/steps-github-deploy.yaml
@@ -71,19 +71,24 @@ content: |
   The configured branch and directory must contain an :ref:`application
   directory <app-directory>` with configuration files that define your
   application. You can create the application directory manually or :doc:`export
-  the application directory </deploy/export-realm-app>` of an existing app (using the
-  ``--for-source-control`` option).
+  the application directory </deploy/export-realm-app>` of an existing app.
+  
+  .. important::
+   
+     In order to deploy successfully, your git repository *must not* contain 
+     some fields that are present in your application export. 
 
-  .. important:: Export For Source Control
+     If you are using :ref:`realm-cli v1.x <realm-cli-v1>` for the v1 app
+     structure, use the ``--for-source-control`` flag when you export an app.
+     This flag omits the ``name``, ``app_id``, ``location``, and ``deployment_model`` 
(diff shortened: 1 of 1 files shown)


source/includes/steps-github-deploy.yaml
```diff
@@ -71,19 +71,24 @@ content: |
...
+  the application directory </deploy/export-realm-app>` of an existing app.
+  
+  .. important::
+   
+     In order to deploy successfully, your git repository *must not* contain 
+     some fields that are present in your application export. 
+
+     If you are using :ref:`realm-cli v1.x <realm-cli-v1>` for the v1 app
+     structure, use the ``--for-source-control`` flag when you export an app.
+     This flag omits the ``name``, ``app_id``, ``location``, and ``deployment_model``
```
nathan-contino-mongo, line 83: Highly subjective nit: I wonder if, instead of the "some fields" wording above, we could use a bulleted list + "the following fields present in your application export:". And then we can just include two small paragraphs for cli v1 and v2 (with v2 explaining that only app_id matters at this point).

That being said... this is totally fine as is, this is just an alternative that you could explore if you think that flow might work better. If you like this version more, feel free to merge as-is! Super subjective and I have literally no data whatsoever to back up my thought, just a gut thought of "we could do it a different way".

nathan-contino-mongo (approved):
LGTM with a potential different approach, if you're so inclined.

source/includes/steps-github-deploy.yaml
dacharyc: Hmm. I've made that change - I like the idea, but in practice this now looks like a very long callout to me. What do you think, @nathan-contino-mongo ?

[Deploy Automatically with GitHub page on pushless staging](https://docs-mongodbcom-staging.corp.mongodb.com/d4453162/15155c2/realm/docsworker-xlarge/master/deploy/deploy-automatically-with-github/#initialize-the-repository)

PR #1309, opened 2021-08-19
(DOCSP-18184): Add language around anonymous user refreshToken expiration

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-18184

- [Anonymous Authentication](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18184/authentication/anonymous/#account-linking): New paragraph at the beginning of the "Account Linking" section

### Review Guidelines

### DIFF
--- a/source/authentication/anonymous.txt
+++ b/source/authentication/anonymous.txt
@@ -21,6 +21,12 @@ The Anonymous authentication provider allows users to log in to your
 application without providing credentials. Anonymous user objects have a 
 unique ID value but no other metadata fields or configuration options.
 
+.. warning::
+
+   An Anonymous user object is not intended to persist data. Once a user logs 
+   out, either by explicit logout or because the refresh token expires, the 
+   user cannot retrieve any previous user data.
+
 Potential use cases for anonymous authentication include:
 
 - Authenticating the readers of a blog or news service.
@@ -39,11 +45,9 @@ Account Linking
 When someone authenticates anonymously, the provider generates an anonymous 
 user object. If the app does not explicitly log this anonymous user out, 
 the same anonymous user is reused. The anonymous user persists until 
-explicit logout, or until the ``refreshToken`` expires, which is currently 
-60 days. When the ``refreshToken`` expires, login fails.
+explicit logout, or until the refresh token expires, which is currently 
+60 days. When the refresh token expires, login fails.
 
-An Anonymous user object is not intended to be used to persist data, and once 
-a user logs out, they will not be able to retrieve any previous user data.
 To persist data associated with an Anonymous user, you can associate
 that existing Anonymous identity with a user account created by a
 different authentication provider.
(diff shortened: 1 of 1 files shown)


source/authentication/anonymous.txt
```diff
@@ -37,8 +35,15 @@ Potential use cases for anonymous authentication include:
 
 Account Linking
 ---------------
-An Anonymous user object is not intended to be reused, and once a user
-logs out, they will not be able to retrieve any previous user data.
+
+When someone authenticates anonymously, the provider generates an anonymous 
+user object. If the app does not explicitly log this anonymous user out, 
+the same anonymous user is reused. The anonymous user persists until 
+explicit logout, or until the ``refreshToken`` expires, which is currently
```
nathan-contino-mongo, line 42: nit: "refreshToken" is a term engineering uses to refer to an object in the backend/frontend, but as far as I know it's not actually the name of a variable exposed to developers using any SDK. For that reason I think I'd reword this usage and the usage on the next line to "refresh token", since that concept is generally known in development enough that we can just throw it around and folks can look it up if they get curious. Alternatively, you could link to a page that discusses the concept of refresh tokens more thoroughly, such as [this Auth0 page on OAuth 2.0's refresh tokens](https://auth0.com/docs/tokens/refresh-tokens). Though maybe we should find something more generic than that, since I'm not sure if we're actually using OAuth 2.0 under the hood.

source/authentication/anonymous.txt
```diff
@@ -37,8 +35,15 @@ Potential use cases for anonymous authentication include:
...
-An Anonymous user object is not intended to be reused, and once a user
-logs out, they will not be able to retrieve any previous user data.
+
+When someone authenticates anonymously, the provider generates an anonymous 
+user object. If the app does not explicitly log this anonymous user out, 
+the same anonymous user is reused. The anonymous user persists until 
+explicit logout, or until the ``refreshToken`` expires, which is currently 
+60 days. When the ``refreshToken`` expires, login fails.
+
+An Anonymous user object is not intended to be used to persist data, and once
```
nathan-contino-mongo, line 45:
```suggestion
An Anonymous user object is not intended to persist data; once 
```
Seems like this sentence means the same thing if we remove "to be used" here, should we just eliminate it?

Second suggestion: is this a rare situation where we can actually use a semicolon sensibly in the docs? If you think it's too nonstandard (especially for non-english-native readers), we could just leave this as is.

nathan-contino-mongo (approved):
LGTM with two or three not-very-important suggestions.

source/authentication/anonymous.txt
dacharyc: Good note! I'll just leave it generic "refresh token." I think it's enough information to communicate the concept.

source/authentication/anonymous.txt
dacharyc: Hmm. We probably could reasonably use a semicolon here, but I'm going to make it a full stop and newline. I'm also moving this into a warning directive and putting it farther up the page, based on the reporter's feedback. But I've redacted the things you pointed out here - thank you!

PR #1310, opened 2021-08-24
Add page about copying & bundling a realm in .NET

MongoCaleb:
### Jira

- https://jira.mongodb.org/browse/DOCSP-17940

- https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/writeCopy/sdk/dotnet/advanced-guides/bundle-a-realm/

### Review Guidelines


source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,56 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You might want to seed your mobile app with some initial data that will be available
+to users on first launch. To do this, you can make a copy of an existing realm 
+(with only the data you want bundled) and then use Visual Studio to bundle the
```
mongomoe, line 17:
```suggestion
(with only the data you want to be bundled) and then use Visual Studio to bundle the 
```

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,56 @@
...
+Include the Realm File in Your Project
+--------------------------------------
+
+1. Add the copy to your production app's Xamarin project. The simplest way to do 
+   this is as follows:
+
+   a. Navigate to the path you specified for the new realm, and then drag the 
+      newly-created .realm file to the shared Xamarin project in Visual Studio. 
+
+   #. When promoted, choose :guilabel:`Copy the file to the directory`.
```
mongomoe, line 44:
```suggestion
   #. When prompted, choose :guilabel:`Copy the file to the directory`.
```

necessary change: promoted -> prompted

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,56 @@
...
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You might want to seed your mobile app with some initial data that will be available
+to users on first launch. To do this, you can make a copy of an existing realm
```
mongomoe, line 16:
```suggestion
to users on the first launch. To do this, you can make a copy of an existing realm 
```

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,56 @@
...
+
+1. Create a new project with the same data models as your production app. Open 
+   an existing {+realm+} with the data you wish to bundle.
+
+#. Use the 
+   :dotnet-sdk:`WriteCopy() <reference/Realms.Realm.html#Realms_Realm_WriteCopy_Realms_RealmConfigurationBase_>` 
+   method to make a copy of the realm to a new location and/or name. The 
+   following code demonstrates this:
+
+   .. literalinclude:: /examples/generated/dotnet/WriteCopy.codeblock.copy_a_realm.cs
```
mongomoe, line 33: C# Syntax highlighting isn't working in this example snippet. Can you add `:language: csharp` below line 33?

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,56 @@
...
+   #. In the shared project, right-click the realm file you just added, choose 
+      :guilabel:`Build Action`, and then choose :guilabel:`EmbeddedResource`.
+
+2. Update your code to read the realm file, which will be in your app's data 
+   directory.
+
+
+.. note:: Cross-Platform Compatibility
+
+   Realm files are cross-platform compatible, which is why you can bundle the
```
nirinchev, line 55: This is not the case for encrypted Realms, so we should make it clear that when bundling files with the app, the Realm should be unencrypted.

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,56 @@
...
+1. Add the copy to your production app's Xamarin project. The simplest way to do 
+   this is as follows:
+
+   a. Navigate to the path you specified for the new realm, and then drag the 
+      newly-created .realm file to the shared Xamarin project in Visual Studio. 
+
+   #. When promoted, choose :guilabel:`Copy the file to the directory`.
+
+   #. In the shared project, right-click the realm file you just added, choose 
+      :guilabel:`Build Action`, and then choose :guilabel:`EmbeddedResource`.
```
nirinchev, line 47: The database should never be added as an embedded resource as that would significantly increase loading times of the app.

examples/dotnet/Examples/WriteCopy.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            // Create a RealmConfiguration for the *copy*
+            var config = new RealmConfiguration("myRealm_copy.realm")
+            {
+                // optionally encrypt it
+                EncryptionKey = encryptionKey,
+            };
+
+            // Copy the realm
+            realm.WriteCopy(config);
```
mongomoe, line 26: Found this remark in the WriteCopy API link: https://docs.mongodb.com/realm-sdks/dotnet/latest/reference/Realms.Realm.html#Realms_Realm_WriteCopy_Realms_RealmConfigurationBase_

"The destination file cannot already exist." 

Should we do a check to see if the destination file exists ? if (File.Exists(myRealm_copy.realm)) then encrypt + writeCopy

mongomoe (commented):
Close to LGTM just 1 typo fix and a couple other suggestions + 1 question

nirinchev (changes requested):
Please don't merge this yet - there are technical changes that need to be addressed on several levels.

1. @papafe can you work with Caleb to outline the procedure for bundling a Realm file with a Xamarin application.
2. Typically a bundled resource will need to be copied to the data directory so we'd ideally include a code snippet to determine whether to copy the file from app resources or use the one that was copied previously.
3. @DominicFrei can you look into and help provide a summary for how to bundle resources in a Unity application? Even though we don't have a dedicated Unity page, we should make sure to produce all new content with Unity in mind so that whenever we split the docs, we don't have to rewrite everything.
4. Finally, this doesn't cover bundling synchronized Realms which is fairly key. @DominicFrei can you work with Caleb to outline the differences compared to local ones.

DominicFrei:
Sounds good, @nirinchev.
@MongoCaleb Let's have a chat or Zoom and talk this through. 👍

DominicFrei:
> 
> 
> @nirinchev @DominicFrei @papafe
> Out of curiosity, wouldn't it work just as easily to do this:
> 
>     1. Make a copy of the Realm we want bundled.
> 
>     2. Upload the copy to an S3 bucket (or elsewhere, obv.)
> 
>     3. On app load, if the realm file doesn't exist, download it from the bucket.

If I understand correctly this will basically mimick downloading (syncing) the Realm initially so I don't really see an advantage in that.
The idea behind bundling a Realm with the app is that on app start there is no data that needs to be downloaded at all and the app can be used right away (apart from any new data that needs to be synced).

examples/dotnet/Examples/WriteCopy.cs
DominicFrei: You can either check or delete it as I've done in the example that I pushed to the Unity examples repo:
https://github.com/realm/unity-examples-3d-chess/blob/temp/bundled-sync-realm/Bundled%20Realms/Assets/Cube.cs#L47-L48

DominicFrei: Regarding encryption please also refer to @nirinchev's comment in https://github.com/mongodb/docs-realm/pull/1310#discussion_r695221963

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
MongoCaleb: @nirinchev I believe, based on word from @papafe, that bundling it this way is, in fact, the correct thing to do. Just want to confirm before going further with the remaining updates.

nirinchev: Yes, I was mistaken. Ferdinando did test it out and confirmed that embedded resources don't cause excessive memory usage, so this should be just fine.

nirinchev (commented):
Looks good to me. @DominicFrei, @papafe do you think we should show the `using` statements for the code snippets? Or do you think `Path`, `Stream`, `File`, and so on are popular enough API that most people will figure it out?

source/examples/generated/dotnet/Unity.codeblock.read_a_realm_unity.cs
```diff
@@ -0,0 +1,5 @@
+// After copying the above created file to the project folder,
+// we can access it in Applicatoin.dataPath:
```
papafe, line 2: typo in Application

papafe:
> Looks good to me. @DominicFrei, @papafe do you think we should show the `using` statements for the code snippets? Or do you think `Path`, `Stream`, `File`, and so on are popular enough API that most people will figure it out?

I think it makes sense to use them, so we're always on the safe side. And from what I can see about the staged changed, it seems we already have using in all the possible locations.

papafe:
Apart from those few notes from Nikola that still need to be addressed and the typo, the rest looks fine for me.

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,124 @@
...
+
+            Non-encrypted {+realm+} files are cross-platform compatible, which 
+            is why you can bundle the file in the shared project.
+
+   .. tab:: Unity
+      :tabid: unity
+
+      1. Open your production Unity app.
+      
+      2. Copy the new {+realm+} file to the project folder of the app. It will
```
DominicFrei, line 86: There is not really a `project` folder, rather a `project` tab which shows the `Assets` folder and the `Packages` folder.
The bundled Realm File should be copied into the `Assets` folder.

DominicFrei: Especially since we use `Application.dataPath` later in the example, which is specifically the path to the `Assets` folder.

source/examples/generated/dotnet/Unity.codeblock.read_a_realm_unity.cs
```diff
@@ -0,0 +1,5 @@
...
+var realmPath = Application.dataPath + "/copy.realm";
```
DominicFrei, line 3: The Realm that we copied is called `bundled.realm`, here we look for `copy.realm`. I'd suggest having the same name to avoid renaming the file to make it work.

examples/dotnet/Examples/Unity.cs
```diff
@@ -0,0 +1,53 @@
...
+        // 2) Copy the Realm file created by above function into the Assets folder (path will be printed to the console).
+
+        // 3) ReadCopy();
+    }
+
+    private async void CreateCopy()
+    {
+        var app = App.Create("foo");
+        var user = await app.LogInAsync(Credentials.Anonymous());
+        var syncConfiguration = new SyncConfiguration("partition_key", user, "");
```
DominicFrei, line 31:
```suggestion
        var syncConfiguration = new SyncConfiguration("partition_key", user);
```

If you leave the path empty you can just ignore all together, it's the same result.

DominicFrei (changes requested):
Just some small details.

DominicFrei (approved):
LGTM 👍

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,125 @@
...
+
+      1. Navigate to the path you specified for the new realm, and then drag the 
+         newly-created {+realm+} file to the shared Xamarin project in Visual Studio. 
+
+      #. When prompted, choose :guilabel:`Copy the file to the directory`.
+
+      #. In the shared project, right-click the {+realm+} file you just added, choose 
+         :guilabel:`Build Action`, and then choose :guilabel:`EmbeddedResource`.
+
+         .. note:: Cross-Platform Compatibility
```
nirinchev, line 76: That note applies to Unity as well since most games are cross-platform.

examples/dotnet/Examples/Unity.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        var app = App.Create("foo");
+        var user = await app.LogInAsync(Credentials.Anonymous());
+        // :code-block-start: read_a_realm_unity
+        // After copying the above created file to the project folder,
+        // we can access it in Application.dataPath:
+        //:uncomment-start:
+        //var realmPath = Path.Combine(Application.dataPath, "bundled.realm");
+        // // And then we open it like any other realm:
+        //Realm realm = await Realm.GetInstanceAsync(realmPath);
```
nirinchev, line 19:
```suggestion
        //var realm = Realm.GetInstance(realmPath);
```

source/sdk/dotnet/advanced-guides/bundle-a-realm.txt
```diff
@@ -0,0 +1,125 @@
...
+
+      .. literalinclude:: /examples/generated/dotnet/Unity.codeblock.read_a_realm_unity.cs
+         :language: csharp
+
+
+.. important:: Reading a Synced Realm
+   
+      If you are instantiating a new copy of a Synced {+realm+}, ensure that the 
+      {+realm+} is fully synced before using it. As with copying a Synced {+realm+}, 
+      be sure to call ``WaitForUploadAsync()`` and ``WaitForDownloadAsync()``.
```
nirinchev, line 124: This is not necessary - it might be a good idea depending on the application, but for most use cases, users can probably just start using the data that is already in the file.

nirinchev (commented):
A couple final nits from me.

MongoCaleb:
@nirinchev et al - I've updated the page to reflect the fact that we can't bundle Synced realms (for now). Is this enough to cover the issue?

nirinchev (approved):

PR #1328, opened 2021-09-08
(DOCSP-18362): Add PartialKeyPath for version 10.14.0

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-18362

- React to Changes: [Observe Partial Key Paths](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18362/sdk/ios/examples/react-to-changes/#observe-partial-key-paths) (New section)

### Review Guidelines

### DIFF
--- a/examples/ios/Examples/Notifications.swift
+++ b/examples/ios/Examples/Notifications.swift
@@ -97,7 +97,7 @@ func keyPath_objectNotificationExample() {
     try! realm.write {
         dog.favoriteToy = "Frisbee"
     }
-    // When one or more keypaths is specified, changes to other properties
+    // When you specify one or more key paths, changes to other properties
     // do not trigger notifications. In this example, changing the "name"
     // property does not trigger a notification.
     try! realm.write {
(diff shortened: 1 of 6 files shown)


source/examples/generated/code/start/Notifications.codeblock.register-a-partial-keypath-change-listener.swift.code-block.rst
```diff
@@ -0,0 +1,50 @@
...
+           case .deleted:
+               print("The object was deleted.")
+           }
+       })
+
+       // Now update to trigger the notification
+       try! realm.write {
+           dog.favoriteToy = "Frisbee"
+       }
+       // When one or more keypaths is specified, changes to other properties
```
nathan-contino-mongo, line 44:
```suggestion
       // When one or more keypaths are specified, changes to other properties
```
subject/verb agreement? I'm like 90% sure "one or more" is plural. I think.

source/examples/generated/code/start/Notifications.codeblock.register-a-partial-keypath-change-listener.swift.code-block.rst
```diff
@@ -0,0 +1,50 @@
...
+       dog.name = "Max"
+       dog.favoriteToy = "Ball"
+       dog.age = 2
+
+       // Open the default realm.
+       let realm = try! Realm()
+       try! realm.write {
+           realm.add(dog)
+       }
+       // Observe notifications on some of the object's key paths. Keep a strong
```
nathan-contino-mongo, line 24: Some of the comments use the term "key paths", some use "keypaths". Should we keep this more uniform? Does it matter?

nathan-contino-mongo (approved):
LGTM with a couple of nits.

examples/ios/Examples/Notifications.swift
```diff
@@ -111,6 +114,58 @@ class AlternateNotificationExampleKeyPath_Dog: Object {
...
+        }
+    })
+
+    // Now update to trigger the notification
+    try! realm.write {
+        dog.favoriteToy = "Frisbee"
+    }
+    // When one or more keypaths is specified, changes to other properties
+    // do not trigger notifications. In this example, changing the "name"
+    // property does not trigger a notification.
```
nlarew, line 162: I wonder if this comment is worth including in the text above the example as it seems like a pretty important detail. Thoughts?

nlarew (approved):
LGTM w/ one potential change

source/examples/generated/code/start/Notifications.codeblock.register-a-partial-keypath-change-listener.swift.code-block.rst
dacharyc: Weirdly, there is not uniform agreement over whether this phrase should be singular or plural. Apparently it's a grammar rabbit hole! So I'm changing it to "When you specify one or more key paths" to avoid the debate entirely. Good catch!

source/examples/generated/code/start/Notifications.codeblock.register-a-partial-keypath-change-listener.swift.code-block.rst
dacharyc: Swift does the same in the [Apple Developer Documentation for Key Paths](https://developer.apple.com/documentation/swift/swift_standard_library/key-path_expressions). Looks like it should be "key path" in general, concatenating when referring to the class (or in our case, arg), and hyphenating when it's a compound modifier. I'll do a pass to ensure all the usages are correct. Thanks for pointing this out!

examples/ios/Examples/Notifications.swift
dacharyc: Hmm. I call this out in the first example for "[Register a Key Path Change Listener.](https://developer.apple.com/documentation/swift/swift_standard_library/key-path_expressions)" This is another version of the same method with the same caveats, but it takes a PartialKeyPath instead of a string KeyPath.

When I was drafting this, my first pass just included the relevant line:
`objectNotificationToken = dog.observe(keyPaths: [\Dog.favoriteToy, \Dog.age], { change in`

But then I thought that might be a bad user experience, because readers would see this line and then have to scroll for the complete example. So I thought I'd just present the entire example. But you're right about this caveat - I don't want to just reproduce all of the boilerplate - so maybe I should go back to just this line. 🤔

PR #1339, opened 2021-09-13
triggers IA toctree move

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-18327

- [Triggers Section](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18327/triggers/)

### Review Guidelines


cbush (approved):
Looks good! My one thought here is whether the three trigger types pages are dangling out there. If I didn't know what those pages were, I might be overwhelmed. It might be less overwhelming if they were in an explanatory category on the TOC like "Trigger Types".

Also, perhaps out of scope of this PR, I've been wondering if our content is getting lost on the pages that are also categories. In this case, "Triggers" is a top level category but also has content on the page. Elsewhere (e.g. "Atlas Cluster & Data Lakes"), there is no content on the section page. Maybe we should think about making this consistent?

nathan-contino-mongo:
@cbush incorporating your feedback into a new "Trigger Types" page at the top level, and removing the "triggers" landing page entirely from navigation.

PR #1351, opened 2021-09-17
new graphic for custom jwt auth

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCS-14788

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/new_graphic/authentication/custom-jwt/


dacharyc (commented):
Oops, looks like the numbering here needs a little tweak. We're missing a `4` in this diagram. In the original on the [live custom-jwt page](https://docs.mongodb.com/realm/authentication/custom-jwt/), 3. is "(Optional) Perform two-factor authentication", but in this updated diagram, that doesn't have a number - instead, you've made "Service returns JWT" the 3. But that means we're missing a 4.

dacharyc (approved):
Nice - LGTM!

PR #1358, opened 2021-09-22
Update Query Engine page with new info

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-18346

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/18346/sdk/dotnet/fundamentals/query-engine/

### DIFF
--- a/source/sdk/dotnet/fundamentals/query-engine.txt
+++ b/source/sdk/dotnet/fundamentals/query-engine.txt
@@ -26,16 +26,16 @@ Operators
 
 There are several types of operators available to filter a
 :ref:`Realm collection <dotnet-client-collections>`.
-Filters work by **evaluating** an operator expression for
+Filters work by evaluating an operator expression for
 every :term:`object <Realm object>` in the collection being
-filtered. If the expression resolves to ``true``, Realm
-Database includes the object in the results collection.
+filtered. If the expression resolves to ``true``, {+realm+}
+includes the object in the results collection.
 
-An **expression** consists of one of the following:
+An expression consists of one of the following:
 
-- The name of a property of the object currently being evaluated.
-- An operator and up to two argument expression(s).
-- A literal string, number, or date.
+- The name of a property of the object currently being evaluated
+- An operator
+- An object of any type (string, date, number, boolean, etc.)
 
 .. note:: About the examples on this page
 
@@ -55,7 +55,7 @@ An **expression** consists of one of the following:
 
 Comparison Operators
 ~~~~~~~~~~~~~~~~~~~~
-The most straightforward operation in a search is to compare values.
+Value comparisons
 
 .. list-table::
    :header-rows: 1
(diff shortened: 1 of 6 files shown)


source/sdk/dotnet/fundamentals/query-engine.txt
```diff
@@ -55,9 +55,7 @@ An **expression** consists of one of the following:
 
 Comparison Operators
 ~~~~~~~~~~~~~~~~~~~~
-
-The most straightforward operation in a search is to compare
-values.
+The most straightforward operation in a search is to compare values.
```
nathan-contino-mongo, line 58: Should we rephrase this to active voice? Perhaps:
```suggestion
Value comparison is the most straightforward operation in a search.
```

Or, maybe we should just eliminate the sentence, since it's not really telling users anything. Maybe there's some more useful general information we should present about comparisons?

source/sdk/dotnet/fundamentals/query-engine.txt
```diff
@@ -162,19 +166,32 @@ Regex-like wildcards allow more flexibility in search.
        For example, the wildcard string "d?g" matches "dog",
        "dig", and "dug", but not "ding", "dg", or "a dog".
 
-   * - | ``Equals`` (C#)
-     - Evaluates to ``true`` if the left-hand string is lexicographically equal to the right-hand string.
+   * - | ``Equals``
+     - Evaluates to ``true`` if the left-hand string is lexicographically equal
```
nathan-contino-mongo, line 170: It could be a good idea to link to wikipedia here -- most of us are familiar with "lexigraphical," I think, but would struggle to actually define it. Could save some users some pain in the future.

nathan-contino-mongo (approved):
LGTM with a couple of teensy nits.

papafe:
In general it looks fine. I have a couple of notes:

1. In the introduction you say "An expression consists of one of the following .." and the last element is "A literal string, number, or date". This is actually reductive, because we accept not only string, number or dates, but also all the other types (bool, objectId, ...).
2. In the last part where you talk about the aggregates you start using the "Filter" method. It's correct, but it comes out of the blue that we can express queries also with this kind of method that accepts a string query. It would be good to give it an introduction, and eventually also maybe link to a page with the actual syntax. Given that the syntax of these string queries is language independent, it could also be a single page for all SDK. The only difference will be how is the "Filter" method called probably

source/sdk/dotnet/fundamentals/query-engine.txt
```diff
@@ -26,16 +26,16 @@ Operators
...
 
-An **expression** consists of one of the following:
+An expression consists of one of the following:
 
-- The name of a property of the object currently being evaluated.
-- An operator and up to two argument expression(s).
-- A literal string, number, or date.
+- The name of a property of the object currently being evaluated
+- An operator
+- An object of any type (string, date, number, boolean, etc.)
```
papafe, line 38: I would suggest to call it just "value". "Object" seems to be too close to a "RealmObject"

papafe:
Apart from the small comment, it looks good!

source/sdk/dotnet/fundamentals/query-engine.txt
papafe: Maybe something like "A value of any accepted Realm type" or similar

PR #1359, opened 2021-09-23
Docsp 18713: add infobox about pausing syncing

mongodben:
## Pull Request Info
added information about pausing realm sync to SDK examples pages for Sync Changes Between Devices to clarify how it should be used. 

included for:
- Android
- iOS
- .NET
- Node
- React Native

*note*: not relevant for the other SDKs 

given that the pause method has different names depending on SDK, created reusable note with insertable function name per SDK so general content name could stay same with method name changing. 

also added a ref to link to compaction, which is referenced in the note about pausing. 

### Jira

- https://jira.mongodb.org/browse/DOCSP-18713

- [Android SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18713/sdk/android/examples/sync-changes-between-devices/)
- [iOS SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18713/sdk/ios/examples/sync-changes-between-devices/)
- [.NET SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18713/sdk/dotnet/examples/sync-changes-between-devices/)
- [Node SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18713/sdk/node/examples/sync-changes-between-devices/)
- [React Native SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18713/sdk/react-native/examples/sync-changes-between-devices/)

### Review Guidelines

### DIFF
--- a/source/includes/note-details-about-pause.rst
+++ b/source/includes/note-details-about-pause.rst
@@ -1,0 +1,33 @@
+.. note::
+   The ``pause`` function is designed for controlling when devices sync. 
+   
+   Examples of when to use ``pause`` include: 
+   -  Syncing only in the evening
+   - Spreading the sync load out across different times
+   - Conserving battery
+   
+   ## is there an upper bound of how long the sync should take?
+   ## even a general rule of thumb would be OK 
+   ##
+   ## my $.02 is that we should not say more that whatever the general rule of thumb is 
+   ## here, and then put any further discussion in a different page, as this seems like a 
+   ## fairly complex topic. 
+   The ``pause`` function is not intended to determine if a Realm should never sync. 
+   As a general principle, Realm sync pauses should not excede XXX_UNIT_OF_TIME. 
+   Pauses that exceed this much time can cause unexpected issues.  
+
+   ## probable scraps 
+   ## what type of storage issues would these be? 
+   For instance, you use ``pause`` for a prolonged period of time, you could encounter
+   issues with storage when you do attempt to sync because the synced Realm 
+   also stores the operations that created the change. These operations can excede 
+   4x the storage of the actual data stored in the Realm. 
+   These operations get pruned from the Realm after they sync. 
+   
+   Another issue you could run into if you sync a Realm after a long pause is there is a server-side feature know as compaction 
+   which prunes the operations from the server-side partitions. After a set amount of 
+   time it starts pruning old operations if the serverside prunes operations from the 
+   partition which is paused then a client could experience a client reset. This will 
+   probably not be an issue if the partition you are running is a per-user private realm. 
+   It is more of an issue with shared realms.
+
--- a/source/includes/note-details-about-pause.rst
+++ b/source/includes/note-details-about-pause.rst
@@ -15,4 +15,4 @@
 
    - Syncing stores the operations that create the changes, which can quickly grow to 
      become much larger than the actual state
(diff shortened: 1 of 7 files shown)


mongodben:
not sure why the Autobuilder check not successful, as it built to staging.

source/includes/note-details-about-pause.rst
```diff
@@ -0,0 +1,18 @@
+.. note::
+   The |pause_func_name| method gives developers control over what time of day the device sync.
```
MongoCaleb, line 2: We try to avoid the passive voice (an ongoing issue for me). So let's try something like:
"You can use the |pause_func_name| method to control when a device syncs data. You should use it for temporary and short-term sync pauses."

source/includes/note-details-about-pause.rst
```diff
@@ -0,0 +1,18 @@
+.. note::
+   The |pause_func_name| method gives developers control over what time of day the device sync. 
+   It is designed and tested with temporary use cases in mind.
```
MongoCaleb, line 3: This is also passive voice, but I'm having a hard time re-writing it. Perhaps it is not needed since we kinda already say this in the sentence above.

source/includes/note-details-about-pause.rst
```diff
@@ -0,0 +1,18 @@
...
+
+   Examples of when to use |pause_func_name| include: 
+
+   - Syncing only in the evening
+   - Spreading the sync load out across different times
+   - Conserving battery
```
MongoCaleb, line 9: I'd consider a combining and also being a bit more verbose (an unusual thing for me to suggest!):
- Syncing data only at a specified time(s) each day
- Conserving device battery use
(the first one actually covers 1 and 2, I think)

source/includes/note-details-about-pause.rst
```diff
@@ -0,0 +1,18 @@
...
+   
+   The |pause_func_name| function is not intended to determine if a Realm should sync for
```
MongoCaleb, line 11: Passive voice. Try:
You should not use the |pause_func_name| function for long-range (months to years) and indefinite time frames. We did not design or test the function for this use case.

I think the rest of the paragraph (what problems you might see) isn't necessary -- just tell them not to do it. If, however, that request comes from engineering, keep it...but rewrite in active voice.

MongoCaleb (commented):
Most of my comments are around passive voice. Feel free to use my suggestions as starting points for your own writing, or take them as they are, or whatever. :)

source/includes/note-details-about-pause.rst
mongodben: agree on removing the second part of the paragraph w the bulleted list. have chopped that. 

however i kept the passive voice in the sentence "the functionality is not designed and tested..." b/c i don't want the sentence to read like we're blaming ourselves for not testing this functionality properly. i believe this would fall under the penumbra if not the explicit law of this section of the style guide - https://docs.mongodb.com/meta/style-guide/writing/use-active-voice/#when-do-you-use-passive-voice

MongoCaleb: good call!

MongoCaleb (approved):
LGTM!

PR #1398, opened 2021-10-07
Realm Activities and Alerts

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-18949

### PD

https://docs.google.com/document/d/1Sx6mAeW4tOGjGOKbb4aG6f8vhEWdnXipyaqHPVxcgb0/edit#

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/alerts/alerts/

### DIFF
--- a/source/alerts.txt
+++ b/source/alerts.txt
@@ -42,6 +42,7 @@ The feed list automatically updates with each change you make.
 
 .. figure:: /images/filter-realm-activities.png
    :alt: Filter Realm activities in the Atlas UI
+   :figwidth: 300px
    :width: 300px
 
 .. note:: Filtering By {+service-short+} Application


source/alerts.txt
```diff
@@ -0,0 +1,136 @@
...
+   * - :ref:`Deployment <application-deployment>` 
+     - Deployment Failure
+     - A user (or Code Deployment) deployment has failed 
+   
+   * - :ref:`Limit <mongodb-service-limitations-request-traffic>`
+     - Request Rate Limit
+     - An overall request rate limit is hit
+   
+   * - :atlas:`Limit </reference/free-shared-limitations/>`
+     - Free Tier Usage
```
drewdipalma, line 102: This is not included, it will be added later as part of a different project.

source/alerts.txt
```diff
@@ -0,0 +1,124 @@
...
+that displays events for the project. {+service-short+} activities are included 
+in this feed, and you can access them through the {+atlas-short+} UI and the 
+{+atlas-short+} API.
+
+{+atlas-short+} Activity Feed
+-----------------------------
+The {+atlas-short+} Activity Feeds show events that have occurred at either the 
+organization level or project level. Within the Project feed, you can find events 
+related to your {+service-short+} applications. By using the filters, you can view 
+and/or download only {+service-short+} events.
```
jeff-allen-mongo, line 28: [suggestion]

I feel like there is a way we can combine this information with the Introduction, since it seems like some information is repeated but with slightly different words.

I'm also slightly confused because in the intro we said that the Activity Feed showed events for the project, but here we're talking about organizations.

source/alerts.txt
```diff
@@ -0,0 +1,124 @@
...
+
+Access the Project Activity Feed
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+There are two ways to access the Project Activity Feed: through the 
+{+atlas-short+} UI, and via the Atlas API. The following {+atlas-short+}
```
jeff-allen-mongo, line 33:
```suggestion
{+atlas-short+} UI, and through the Atlas API. The following {+atlas-short+} 
```

"via" is listed as a word to avoid per our [style guide](https://docs.mongodb.com/meta/style-guide/terminology/general-term-guidelines/avoid-obscure-words/)

source/alerts.txt
```diff
@@ -0,0 +1,124 @@
...
+
+- A {+service-short+} application hits an overall request rate limit
+
+Custom {+service-short+} Alerts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+For any activity that is not mapped to a default alert, you can set a custom 
+alert at the Project level.
+
+You create {+service-short+} alerts in the same way you 
+`create other {+atlas-short+} alerts <https://docs.atlas.mongodb.com/configure-alerts/>`_.
```
cbush, line 120: Do we not have an `:atlas:` role?

source/alerts.txt
```diff
@@ -0,0 +1,124 @@
...
+{+service-short+} Alerts
+------------------------
+There are default alerts that map to four of the activities listed above. In 
+addition, you can create custom alerts for the other {+service-short+} 
+activities.
+
+Default Alerts
+~~~~~~~~~~~~~~~
+Default {+service-short+} alerts trigger a notification and email to the 
+Project Owners. The four default {+service-short+} alerts occur when:
```
cbush, line 106: Still four?

source/alerts.txt
```diff
@@ -0,0 +1,124 @@
...
+
+- A {+sync-short+} process fails and cannot be restarted  
+
+- A trigger fails and cannot be restarted
+
+- A {+service-short+} application hits an overall request rate limit
```
jeff-allen-mongo, line 112: Slightly confused here. Which four activities? We've listed three items in the bulleted list, and I'm having trouble making the connection to the activities.

source/alerts.txt
```diff
@@ -0,0 +1,124 @@
...
+   (and its corresponding alert) provides context and routing information for the 
+   appropriate application.
+
+
+{+service-short+} Activities
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+{+service-short+} activities are a subset of 
+:atlas:`{+atlas-short+} alerts </alerts/>`. The **Target** for 
+all {+service-short+} activities is "Realm"; within that target, there are
```
ccho-mongodb, line 61: A couple items:

1) Is there an action that you can recommend to select the "Target" or is that a column label or a button and would it be worth identifying that to the reader (similar to how you've identified how to filter activities in the UI)? Or does that label change form in the UI too often?

2) This also leaves me wondering how I can identify the event types. Is that information you can provide or does the UI change too often for it to be worthwhile?

source/alerts.txt
ccho-mongodb, line 112: I'm confused by the "default alerts that map to four activities" and "four default Realm alerts". Are these the same thing? If not, how are they different?
It looks like there are three items in the list to me, but maybe I'm missing something.

"There are four lights!" -Jean-Luc Picard

source/alerts.txt
ccho-mongodb, line 120: Is there any further customization they can do other than Realm as the target? Can they selectively alert on event type? If so, would it be worth showing an example?

source/alerts.txt
```diff
@@ -0,0 +1,124 @@
...
+     - A Sync process has failed and cannot be restarted  
+   
+   * - :ref:`Trigger <trigger-types>` 
+     - Trigger Failure
+     - A Trigger has failed and cannot be restarted 
+   
+   * - :ref:`Deployment <static-hosting>` 
+     - | URL Confirmation 
+         (Static Hosting)
+     - A Custom URL is confirmed for static hosting
```
ccho-mongodb, line 83: Nitpick: the other items use past tense in the description. I wonder if it would be helpful to keep all the descriptions in the same tense.

source/alerts.txt
```diff
@@ -0,0 +1,124 @@
...
+     - Sync Failure
+     - A Sync process has failed and cannot be restarted  
+   
+   * - :ref:`Trigger <trigger-types>` 
+     - Trigger Failure
+     - A Trigger has failed and cannot be restarted 
+   
+   * - :ref:`Deployment <static-hosting>` 
+     - | URL Confirmation 
+         (Static Hosting)
```
ccho-mongodb, line 82: This left me wondering whether there is an event for dynamic host url confirmation. Or whether the parenthetical is necessary since you describe the event as applicable to "static hosting"

source/alerts.txt
MongoCaleb: I can't do math.

PR #1420, opened 2021-10-14
(DOCSP-19020): React Native JS SDK refresh of page "Open a Synced Realm"

mongodben:
## Pull Request Info

Refresh page content for React native SDK page "Open a Synced Realm". 

This is largely a copy and paste job from the newly created Node docs (https://jira.mongodb.org/browse/DOCSP-17709). It's a separate PR w the content copy-pasted instead of just using an include b.c the refs need to change. 

Content changes focus on using the method `Realm.open()` to open a new realm instead of `new Realm()`, as this is the currently preferred method per the JS SDK team.

Additional changes: 
* In addition to main changes on the page `source/sdk/react-native/examples/open-and-close-a-realm.txt`, there are updates to the refs in the page and renaming the text in links to the page. 
* redirect from old page URL to the new

### Jira

- https://jira.mongodb.org/browse/DOCSP-19020

- [Open & Close a Realm (React Native)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19020/sdk/react-native/examples/open-and-close-a-realm/)

### Review Guidelines

### DIFF
--- a/source/includes/react-native-open-a-synced-realm.rst
+++ b/source/includes/react-native-open-a-synced-realm.rst
@@ -1,0 +1,5 @@
+If you're opening a synced {+realm+}, the configuration you use depends on whether 
+or not the user's device is connected to the Internet. Learn how to handle different
+connection states in the :ref:`Open a Synced Realm While Online <react-native-open-a-synced-realm-while-online>` 
+and :ref:`Open a Synced Realm While Offline <react-native-open-a-synced-realm-while-offline>` 
+sections.
--- a/source/includes/react-native-open-a-synced-realm.rst
+++ b/source/includes/react-native-open-a-synced-realm.rst
@@ -1,5 +1,12 @@
-When opening a synced {+realm+}, the configuration you use depends on whether 
-or not the user's device is connected to the Internet. Learn how to handle different
-connection states in the :ref:`Open a Synced Realm While Online <react-native-open-a-synced-realm-while-online>` 
-and :ref:`Open a Synced Realm While Offline <react-native-open-a-synced-realm-while-offline>` 
-sections.
+To open a synced {+realm+}, call :js-sdk:`Realm.open() <Realm.html#.open>`. 
+Pass in a :js-sdk:`Configuration <Realm.html#~Configuration>`
+object, which must include the ``sync`` property defining a 
+:js-sdk:`SyncConfiguration <Realm.App.Sync.html#~SyncConfiguration>` object. 
+
+When opening a synced {+realm+}, the configuration you use depends on the behavior you want
+for how the client to sync with the server. On a high level, the ways to connect 
+to a synced {+realm+} are: 
+
+- :ref:`Syncing all data before returning <react-native-sync-all-data-before-returning>`
+- :ref:`Returning after a timeout with background sync <react-native-return-after-timeout-with-background-sync>`
+- :ref:`Opening immediately with background sync <react-native-open-immediately-with-background-sync>`
--- a/source/includes/react-native-open-a-synced-realm.rst
+++ b/source/includes/react-native-open-a-synced-realm.rst
@@ -2,6 +2,7 @@ To open a synced {+realm+}, call :js-sdk:`Realm.open() <Realm.html#.open>`.
 Pass in a :js-sdk:`Configuration <Realm.html#~Configuration>`
 object, which must include the ``sync`` property defining a 
 :js-sdk:`SyncConfiguration <Realm.App.Sync.html#~SyncConfiguration>` object. 
+In the SyncConfiguration, you must include include ``user`` and ``partitionValue``.
 
 When opening a synced {+realm+}, the configuration you use depends on the initial sync behavior you want. You can connect to a synced {+realm+} in the following ways: 
 
--- a/source/includes/react-native-open-a-synced-realm.rst
+++ b/source/includes/react-native-open-a-synced-realm.rst
@@ -3,10 +3,8 @@ Pass in a :js-sdk:`Configuration <Realm.html#~Configuration>`
 object, which must include the ``sync`` property defining a 
 :js-sdk:`SyncConfiguration <Realm.App.Sync.html#~SyncConfiguration>` object. 
 
-When opening a synced {+realm+}, the configuration you use depends on the behavior you want
-for syncing the client with the server. The ways to connect 
-to a synced {+realm+} are: 
+When opening a synced {+realm+}, the configuration you use depends on the initial sync behavior you want. You can connect to a synced {+realm+} in the following ways: 
 
-- :ref:`Syncing all data before returning <react-native-sync-all-data-before-returning>`
-- :ref:`Returning after a timeout with background sync <react-native-return-after-timeout-with-background-sync>`
-- :ref:`Opening immediately with background sync <react-native-open-immediately-with-background-sync>`
+- :ref:`Sync all data before returning <react-native-sync-all-data-before-returning>`
+- :ref:`Return after a timeout with background sync <react-native-return-after-timeout-with-background-sync>`
+- :ref:`Open immediately with background sync <react-native-open-immediately-with-background-sync>`
--- a/source/includes/react-native-open-a-synced-realm.rst
+++ b/source/includes/react-native-open-a-synced-realm.rst
@@ -8,4 +8,4 @@ When opening a synced {+realm+}, the configuration you use depends on the initia
 
 - :ref:`Sync all data before returning <react-native-sync-all-data-before-returning>`
 - :ref:`Return after a timeout with background sync <react-native-return-after-timeout-with-background-sync>`
-- :ref:`Open immediately with background sync <react-native-open-immediately-with-background-sync>`
+- :ref:`Return immediately with background sync <react-native-open-immediately-with-background-sync>`
--- a/source/includes/react-native-open-a-synced-realm.rst
+++ b/source/includes/react-native-open-a-synced-realm.rst
@@ -4,7 +4,7 @@ object, which must include the ``sync`` property defining a
 :js-sdk:`SyncConfiguration <Realm.App.Sync.html#~SyncConfiguration>` object. 
 
 When opening a synced {+realm+}, the configuration you use depends on the behavior you want
-for how the client to sync with the server. On a high level, the ways to connect 
+for syncing the client with the server. The ways to connect 
 to a synced {+realm+} are: 
 
 - :ref:`Syncing all data before returning <react-native-sync-all-data-before-returning>`
--- a/source/includes/react-native-open-a-synced-realm.rst
+++ b/source/includes/react-native-open-a-synced-realm.rst
@@ -1,4 +1,4 @@
-If you're opening a synced {+realm+}, the configuration you use depends on whether 
+When opening a synced {+realm+}, the configuration you use depends on whether 
 or not the user's device is connected to the Internet. Learn how to handle different
 connection states in the :ref:`Open a Synced Realm While Online <react-native-open-a-synced-realm-while-online>` 
 and :ref:`Open a Synced Realm While Offline <react-native-open-a-synced-realm-while-offline>` 
--- a/source/sdk/react-native/examples/open-and-close-a-realm.txt
+++ b/source/sdk/react-native/examples/open-and-close-a-realm.txt
@@ -1,8 +1,8 @@
 .. _react-native-open-close-realm:
 
-=============================================
-Open & Close a Local Realm - React Native SDK
-=============================================
+=======================================
+Open & Close a Realm - React Native SDK
+=======================================
 
 .. default-domain:: mongodb
 
--- a/source/sdk/react-native/examples/open-and-close-a-realm.txt
+++ b/source/sdk/react-native/examples/open-and-close-a-realm.txt
@@ -60,12 +60,11 @@ Return After Timeout with Background Sync
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you don't need to sync all data before before returning the {+realm+}, configure the
-``newRealmFileBehavior`` and ``existingRealmFileBehavior`` fields of your
+``newRealmFileBehavior`` and ``existingRealmFileBehavior`` properties of your
 SyncConfiguration to automatically open the realm after a timeout period elapses. You
-can use a single :js-sdk:`OpenRealmBehaviorConfiguration
-<Realm.App.Sync.html#~OpenRealmBehaviorConfiguration>`for both
-``newRealmFileBehavior`` and ``existingRealmFileBehavior`` that contains the following
-field values:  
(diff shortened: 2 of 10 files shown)


source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,150 @@
...
+Return After Timeout with Background Sync
+`````````````````````````````````````````
+
+If you don't need to sync all data before before returning, configure your SyncConfiguration
+to include ``newRealmFileBehavior`` and ``existingRealmFileBehavior`` each with
+the same :js-sdk:`OpenRealmBehaviorConfiguration <Realm.App.Sync.html#~OpenRealmBehaviorConfiguration>`.  
+
+Set the OpenRealmBehaviorConfiguration to have ``type: 'downloadBeforeOpen'``, 
+``timeOut: <time in milliseconds>`` and ``timeOutBehavior: 'openLocalRealm'``.
+These properties make ``Realm.open()`` return at the end of the ``timeOut`` period, and open a local {+realm+}.
```
nathan-contino-mongo, line 67:
```suggestion
These properties force the ``Realm.open()`` method to return an open {+realm+} after ``timeOut`` milliseconds, or when the {+realm+} has completely downloaded -- whichever comes first. 
```

Not sure what the term "local realm" means in this context -- in the other SDKs, realms are either synced or local, they never flux between the two. It sounds like this actually opens a synced realm with incomplete data, not a local realm... so we should probably avoid confusing readers with the "local realm" term.

nathan-contino-mongo: Additionally, we should make it clear that if the realm downloads completely before the timeout, it'll open before the timeout is reached.

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,150 @@
...
+`````````````````````````````````````````
+
+If you don't need to sync all data before before returning, configure your SyncConfiguration
+to include ``newRealmFileBehavior`` and ``existingRealmFileBehavior`` each with
+the same :js-sdk:`OpenRealmBehaviorConfiguration <Realm.App.Sync.html#~OpenRealmBehaviorConfiguration>`.  
+
+Set the OpenRealmBehaviorConfiguration to have ``type: 'downloadBeforeOpen'``, 
+``timeOut: <time in milliseconds>`` and ``timeOutBehavior: 'openLocalRealm'``.
+These properties make ``Realm.open()`` return at the end of the ``timeOut`` period, and open a local {+realm+}. 
+Syncing between local data and the server continues in the background. For example:
```
nathan-contino-mongo, line 68:
```suggestion
If the {+realm+} doesn't finish downloading before the timeout, the initial realm sync continues in the background. For example: 
```
Just to match up with the explanation in the above sentence, which now talks about timeout and complete download situations.

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,150 @@
...
+These properties make ``Realm.open()`` return at the end of the ``timeOut`` period, and open a local {+realm+}. 
+Syncing between local data and the server continues in the background. For example: 
+
+.. literalinclude:: /examples/generated/node/open-and-close-a-realm.codeblock.open-synced-realm-config.js
+  :language: javascript
+
+This configuration is also useful if you're attempting to sync in an environment 
+where it's uncertain if the user has an Internet connection. The client attempts 
+to sync until the end of the the ``timeOut`` period, and then opens the local {+realm+},
+even if the it never connects to the server.
```
nathan-contino-mongo, line 76: Personally I'd say that this paragraph just retreads information we've already conveyed -- we can probably eliminate it entirely. Applications are either OK with incomplete data, or they aren't -- whether the user may or may not have an internet connection doesn't impact that design choice. And the second sentence is now explained above.

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,150 @@
...
+  ``existingRealmFileBehavior`` and ``newRealmFileBehavior`` set to allow the user 
+  to work with the local {+realm+} if the device is doesn't fully sync within the ``timeOut``
+  period of 1000 milliseconds.  
+
+  .. literalinclude:: /examples/generated/node/open-and-close-a-realm.codeblock.open-synced-realm-offline-with-car-schema.js
+    :language: javascript
+
+.. warning:: 
+
+   If you are attempting to synchronize a {+realm+} while not connected to
```
nathan-contino-mongo, line 93:
```suggestion
   When attempting to synchronize a {+realm+} while not connected to 
```
Passive -> active voice.

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,150 @@
...
+  .. literalinclude:: /examples/generated/node/open-and-close-a-realm.codeblock.open-synced-realm-offline-with-car-schema.js
+    :language: javascript
+
+.. warning:: 
+
+   If you are attempting to synchronize a {+realm+} while not connected to 
+   the Internet, you must pass ``Realm.open()`` a :js-sdk:`Configuration <Realm.html#~Configuration>`
+   object with ``newRealmFileBehavior`` and ``existingRealmFileBehavior``. If you 
+   don't include these properties, ``Realm.open()`` continues trying
+   to connect to the server indefinitely without opening.
```
nathan-contino-mongo, line 97: I wonder if we should consider moving this information to the introduction of this section, instead of hosting it in a warning after a lengthy example and a significant amount of body text. Maybe we can combine it with the default behaviour explanation above?

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,150 @@
...
+
+.. _react-native-open-a-synced-realm-while-offline:
+
+Open a Synced Realm While Offline
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If a user's device is not connected to the Internet or you're not certain if it's 
+connected, you must use  a a :js-sdk:`OpenRealmBehaviorConfiguration
+<Realm.App.Sync.html#~OpenRealmBehaviorConfiguration>` that includes ``newRealmFileBehavior`` and ``existingRealmFileBehavior`` 
+properties.
```
nathan-contino-mongo, line 110: Interesting alternative example -- I wonder if the "while offline" and "after timeout" cases are misleading users here, though. The way I see it, they really both belong in the "while offline" category.... but we should recommend "with timeout" for applications that would *prefer* fetching of initial data, and "immediately" for applications that don't need initial data (for instance, an application that doesn't need anything from the server, like a notepad or a todo list).

Mainly because right now they draw a false dichotomy -- I don't think these cases are truly that distinct, so we should use examples to make it clearer to users where they should use which behaviour.

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,150 @@
...
+include the ``sync`` property defining a :js-sdk:`SyncConfiguration
+<Realm.App.Sync.html#~SyncConfiguration>` object. Set this
+``OpenRealmBehaviorConfiguration`` object as the value for
+the ``newRealmFileBehavior`` and ``existingRealmFileBehavior`` fields of the
+``SyncConfiguration``. 
+
+.. literalinclude:: /examples/generated/node/sync-changes-between-devices.codeblock.sync-changes-between-devices-sync-changes-in-the-background-create-config.js
+      :language: javascript
+
+Finally,  call :js-sdk:`Realm.open() <Realm.html#.open>`
```
nathan-contino-mongo, line 131:
```suggestion
Finally, call :js-sdk:`Realm.open() <Realm.html#.open>`
```
Extra space.

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,150 @@
...
+
+.. .. _react-native-local-realm-configuration:
+
+.. Local Realm Configuration
+.. -------------------------
+
+.. .. _react-native-provide-a-subset-of-classes-to-a-realm:
+
+.. Provide a Subset of Classes to a Realm
+.. ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nathan-contino-mongo, line 150: Can we delete these commented out sections? We can always pull them from git history if we need them.

source/sdk/react-native/examples/open-and-close-a-realm.txt
mongodben: yep i think your suggestions render this bit irrelevant

source/sdk/react-native/examples/open-and-close-a-realm.txt
mongodben: moving to the end of the intro for `Open a Synced Realm`, right before the `Open a Synced Realm While Online` content.

mongodben:
@nathan-contino-mongo implemented the changes w the exception of the above comment abt synced realm while online - https://github.com/mongodb/docs-realm/pull/1420#discussion_r729827641

lmk what you think for that comment

source/sdk/react-native/examples/open-and-close-a-realm.txt
nathan-contino-mongo: I think this is a good way of breaking this concept down. When it comes to options like this, I think the tasks we chose were a little too abstract for the topic. This seems like a significant improvement -- take a stab at refactoring it to match this!

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,138 @@
...
+properties. This configuration is useful if a user's device is not connected to the Internet or you're not certain if it's 
+connected. 
+
+Create a :js-sdk:`OpenRealmBehaviorConfiguration
+<Realm.App.Sync.html#~OpenRealmBehaviorConfiguration>` object and set its
+``type: "openImmediately"``. This configuration opens a local {+realm+} 
+immediately without attempting to sync with the server. The client continues  
+attempting to connect with the server in the background.
+
+.. literalinclude:: /examples/generated/node/sync-changes-between-devices.codeblock.sync-changes-between-devices-sync-changes-in-the-background-create-OpenRealmBehaviorObject.js
```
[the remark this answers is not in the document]
mongodben: i removed this as a separate example, and made it part of the example below in the section

source/includes/react-native-open-a-synced-realm.rst
```diff
@@ -0,0 +1,11 @@
...
+Pass in a :js-sdk:`Configuration <Realm.html#~Configuration>`
+object, which must include the ``sync`` property defining a 
+:js-sdk:`SyncConfiguration <Realm.App.Sync.html#~SyncConfiguration>` object. 
+In the SyncConfiguration, you must include include ``user`` and ``partitionValue``.
+
+When opening a synced {+realm+}, the configuration you use depends on the initial sync behavior you want. You can connect to a synced {+realm+} in the following ways: 
+
+- :ref:`Sync all data before returning <react-native-sync-all-data-before-returning>`
+- :ref:`Return after a timeout with background sync <react-native-return-after-timeout-with-background-sync>`
+- :ref:`Open immediately with background sync <react-native-open-immediately-with-background-sync>`
```
nathan-contino-mongo, line 11:
```suggestion
- :ref:`Return immediately with background sync <react-native-open-immediately-with-background-sync>`
```
Suggestion: use the verb "return" here to make parallelism more obvious between these bullet points.

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,129 @@
...
+Open a Synced Realm
+-------------------
+
+.. include:: /includes/react-native-open-a-synced-realm.rst
+
+.. warning:: 
+
+   When attempting to synchronize a {+realm+} while not connected to 
+   the Internet, you must pass ``Realm.open()`` a :js-sdk:`Configuration <Realm.html#~Configuration>`
+   object with ``newRealmFileBehavior`` and ``existingRealmFileBehavior``. If you
```
nathan-contino-mongo, line 33:
```suggestion
   object that contains the ``newRealmFileBehavior`` and ``existingRealmFileBehavior`` properties. If you 
```
Nitpick: "with" is not as clear as it could be, consider using "that contains... properties" instead

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,129 @@
...
+These properties force the ``Realm.open()`` method to return an open {+realm+} after ``timeOut`` milliseconds, or when the {+realm+} has completely downloaded, whichever comes first. 
+If the {+realm+} doesn't finish downloading before the timeout, the initial realm sync continues in the background. For example: 
+
+.. literalinclude:: /examples/generated/node/open-and-close-a-realm.codeblock.open-synced-realm-config.js
+  :language: javascript
+
+This configuration is also useful if you're attempting to sync in an environment 
+where it's uncertain if the user has an Internet connection. The client attempts 
+to sync until the end of the the ``timeOut`` period, and then opens the local {+realm+},
+even if the it never connects to the server.
```
nathan-contino-mongo, line 82: Looks like the second sentence is still here, can we remove it?

source/sdk/react-native/examples/open-and-close-a-realm.txt
```diff
@@ -0,0 +1,129 @@
...
+This configuration is also useful if you're attempting to sync in an environment 
+where it's uncertain if the user has an Internet connection. The client attempts 
+to sync until the end of the the ``timeOut`` period, and then opens the local {+realm+},
+even if the it never connects to the server. 
+
+.. example::
+  
+  The following example shows opening a synced {+realm+} with a 
+  ``Configuration`` object using a predefined ``Car`` :doc:`schema </sdk/node/examples/define-a-realm-object-model>`, 
+  the currently logged in user, a partition value of "MyPartition", and a
```
nathan-contino-mongo, line 88:
```suggestion
  the currently logged in user, a partition value of "myPartition", and a 
```
The partition value mentioned here still does not match the value in the example.

source/sdk/react-native/examples/open-and-close-a-realm.txt
nathan-contino-mongo, line 82:
```suggestion
where it's uncertain if the user has an Internet connection.
```
Adding this suggestion so you can incorporate all of these suggestions via the GitHub UI if you'd like.

mongodben: thanks, i think i missed the 2nd part of your suggestion 😅

nathan-contino-mongo:
Feel free to squash & merge whenever -- looks good!

PR #1424, opened 2021-10-18
(DOCSP-19134): Refactor Node "Open & Close a Realm" to follow style of React Native

mongodben:
## Pull Request Info

Refactor content on the Open & Close a Realm page for Node SDK to follow the same formatting as the Open & Close a Realm page for React Native. These pages are the same exact content with different Refs. So the only thing was copy-pasting over the content, and changing the refs. 

For original work on the React Native page, see https://github.com/mongodb/docs-realm/pull/1420

### Jira

- https://jira.mongodb.org/browse/DOCSP-19134

- [Open & Close a Realm (Node)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19134/sdk/node/examples/open-and-close-a-realm/)

### Review Guidelines

### DIFF
--- a/source/sdk/node/examples/open-and-close-a-realm.txt
+++ b/source/sdk/node/examples/open-and-close-a-realm.txt
@@ -26,61 +26,64 @@ Open a Synced Realm
 
 .. include:: /includes/node-open-a-synced-realm.rst
 
-.. _node-open-a-synced-realm-while-online:
+.. warning:: 
 
-Open a Synced Realm While Online
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+   When attempting to synchronize a {+realm+} while not connected to 
+   the Internet, you must pass ``Realm.open()`` a :js-sdk:`Configuration <Realm.html#~Configuration>`
+   object that contains the ``newRealmFileBehavior`` and ``existingRealmFileBehavior`` properties. If you 
+   don't include these properties, ``Realm.open()`` continues trying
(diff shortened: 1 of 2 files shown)


source/sdk/node/examples/open-and-close-a-realm.txt
```diff
@@ -89,63 +92,36 @@ even if the it never connects to the server.
...
-   
-   Refer to the :ref:`Open a Synced Realm While Offline <node-open-a-synced-realm-while-offline>`
-   documentation for more information about connecting to your offline {+realm+}. 
-
-.. _node-open-a-synced-realm-while-offline:
+.. _node-open-immediately-with-background-sync:
 
-Open a Synced Realm While Offline
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Open Immediately With Background Sync
```
nathan-contino-mongo, line 97: Nitpick (both here and in the react native docs): we use "return" in the bullet points above for consistency, but "open" here. Should we switch both titles to "Return Immediately ..." instead?

nathan-contino-mongo: I agree with that. However, I think some parallelism in the titles/bullet points would be helpful for users -- if we use a different verb for this section, they might not understand how the three sections relate. So I think we should try switching *each* of the sections to "open" instead of return, and the bullet points. This page is about opening and closing a realm, after all, so it does make sense.

PR #1427, opened 2021-10-19
(DOCSP-16273): Backend Compaction / Client Maximum Offline Time

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-16273

- [Optimize Sync Atlas Usage](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16273/sync/optimize-sync-atlas-usage/)

### Review Guidelines


source/sync/client-maximum-offline-time.txt
```diff
@@ -0,0 +1,100 @@
...
+      1. Select the :guilabel:`Sync` menu in the sidebar.
+
+      2. Click the :guilabel:`Client Maximum Offline Time (Optional)` dropdown
+         in the "Configuration" tab.
+
+         .. figure:: /images/client-maximum-offline-time.png
+           :width: 750px
+           :alt: The client maximum offline time configuration in a disabled state.
+
+      3. Click the :guilabel:`Enable Max Offline Time` toggle.
```
dacharyc, line 69: When I got to this step, I wondered what the behavior is if we don't enable this setting. If there is no "max offline time" - does it just keep storing changeset indefinitely, until the device runs out of memory, or...?

I think that because this is an optional setting that is not enabled by default, we should probably more explicitly describe the benefit of enabling this feature, and what happens if users don't enable it. I _think_ based on reading above starting at around line 37, we're saying that if a client doesn't check in for a long period of time (i.e. the client changeset predates the backend history), the dev must perform a client reset... but if devs enable this "max offline" setting, the client must _still_ perform a client reset if it's offline longer than the max specified here. So what is the benefit of this setting, if either case results in a client reset? 

I'd love to see a very simple description (maybe in the overview?) of:
- If you don't enable this setting, the default behavior is:
- When you enable this setting, the behavior is:
- This is good because:

source/sync/get-started.txt
```diff
@@ -172,6 +172,12 @@ Define a client reset handler:
   histories diverge, you should define a :ref`client reset <client-resets>`
   handler when you open each synced realm with an SDK.
 
+Define a client maximum offline time:
+  {+service-short+} uses a history of changes to underlying data to
+  synchronize clients. Configure a :ref:`client maximum offline time
+  <client-maximum-offline-time>` to control the number of days of
+  history stored by your {+app+}.
```
dacharyc, line 179: Reading this on this page also made me want to know why this is important. If I don't do anything with this setting, does it just stop storing history at some point? If I _do_ set this for something like storing "8 days" of history, what happens on day 9? I know we cover that on the page itself, but it would probably be good to add a detail or two here. 

Also, I see we say `{+app+}` here, but the setting says client, so I'm a little confused about where this setting applies - are we saying it's the number of days of history the backend app stores, or the client app?

source/sync/client-maximum-offline-time.txt
```diff
@@ -0,0 +1,100 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+The **client maximum offline time** controls how long a client can
+remain offline between synchronization sessions with the {+backend+}
+backend. Clients that do not synchronize with the backend for more than the specified
```
dacharyc, line 21: My first thought on reading this is: maybe we should specify _when_ the client reset happens - i.e. something like: "When a client attempts to synchronize with the backend after the maximum number of days have passed, it must perform a client reset." I know that's implied, but I generally prefer being explicit for clarity.

After thinking about this more, though, I'm also wondering _how_ a client knows that it has been longer than the maximum offline time and it must perform a client reset. Is there an error code we're passing that we should specify on this page somewhere? And/or should we add corresponding info to the SDK docs on the client reset pages about the maximum offline time?

source/sync/client-maximum-offline-time.txt
```diff
@@ -0,0 +1,100 @@
...
+number of days must perform a :ref:`client reset <client-resets>`.
+
+Key Concepts
+------------
+
+{+service-short+} keeps a history of :ref:`changes to underlying data
+<changesets>`, similar to the :manual:`MongoDB oplog
```
dacharyc, line 28: Is this history of changes "similar to the oplog" or are we explicitly using the oplog for this? On the [Get Started with Sync](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16273/sync/get-started/#best-practices) page, we say:
> Realm Sync requires access to the oplog of your synced cluster. For best results, keep 48 hours of oplog for a cluster using Sync.

I think that, and reading in other places, makes me think we're explicitly using the oplog, although maybe I'm wrong about that - but it's certainly implied in that other bullet.

Thinking about that - does that mean the max offline time and the oplog retention time should be the same? i.e. if the max offline time is to prevent the client changeset from exceeding the oplog, do we need to match them?

source/sync/client-maximum-offline-time.txt
```diff
@@ -0,0 +1,100 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
```
dacharyc, line 16: Overall, I've noticed this page has a lot of long, difficult-to-read sentences. I've started a Readability epic for next quarter, and this page scores low enough for readability that it would fall into that epic to refactor. Maybe you can rework it a bit using something like [Hemingway App](https://hemingwayapp.com/) to reduce the number of long sentences and improve readability?

source/sync/client-maximum-offline-time.txt
nathan-contino-mongo: Hm, I think that the backend still runs compaction, but with a 30 day default if you don't have this feature enabled -- reaching out to some engineers to see if they know any more details about this. Definitely something we should cover here.

source/sync/client-maximum-offline-time.txt
nathan-contino-mongo: "Falling off the end of history" is a condition that could already cause a client reset -- and it's the most common condition at that. This setting makes that "end of history" a variable that users can control in their app. Past that, the client reset works exactly the same -- the backend notices that a client reset is necessary when a client attempts to synchronize. The backend sends the client reset signal over the sync protocol. The client performs the reset in the built-in client reset handler, or using your custom code block.

When I first wrote this, I felt like those internal client reset details belong on the client reset page -- if the user knows that a client is forced to perform a client reset, they can presumably read that page to learn more about those details. I still feel like covering those details is out of scope, and that's why I only cover client resets in brief here.

I'm open to adding more details, but I don't want to bloat this page with too much detail about other features. So maybe some of these details are things we should present better in the client reset page? Spitballing here.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,216 @@
...
+
+Key Concepts
+~~~~~~~~~~~~
+
+In order for synchronization to converge at the same end state
+as the backend and all other clients, a synchronization client requires
+the full history of changes beginning immediately after the client's last
+synchronization session. When a client does not check in for a long
+period of time, the end of history stored on that client can predate
+the beginning of the history of changes stored in the backend. When the
```
kmorkos, line 124: as discussed on our call earlier these last two sentences are not accurate. Prior to this max offline feature, it was possible for a client to be offline for an unbounded duration of time and still catch up to history assuming sync was not terminated and re-enabled while they were offline

source/sync/get-started.txt
```diff
@@ -172,6 +182,14 @@ Define a client reset handler:
   histories diverge, you should define a :ref`client reset <client-resets>`
   handler when you open each synced realm with an SDK.
 
+Define a client maximum offline time:
+  The {+service-short+} backend uses a history of changes to underlying
+  data to synchronize clients. Configure a :ref:`client maximum offline time
+  <client-maximum-offline-time>` to control the number of days of
+  history stored by your {+app+}. Clients that have not synchronized in
```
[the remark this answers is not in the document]
nathan-contino-mongo: A fair distinction. The question is: is the net effect the same from the user's perspective? Would this only differ in cases where a user connects, but doesn't fully sync a realm?

source/sync/get-started.txt
```diff
@@ -172,6 +182,14 @@ Define a client reset handler:
...
+  more than that number of days must perform a client reset the next
```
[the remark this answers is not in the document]
nathan-contino-mongo: This matches up with the phrasing used in the rest of the docs -- note that the *client* has to perform a reset, not the *user*, and we actually want to make it more obvious that client resets exist in the docs. So this should be fine.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,216 @@
...
+
+The lower the client maximum offline time, the less space {+sync+} uses
+to store that history in your synced Atlas clusters.
+
+Key Concepts
+~~~~~~~~~~~~
+
+In order for synchronization to converge at the same end state
+as the backend and all other clients, a synchronization client requires
+the full history of changes beginning immediately after the client's last
```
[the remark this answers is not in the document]
nathan-contino-mongo: I wouldn't worry about this since this section is scoped under the client maximum offline time, not compaction. We'll see if we need to reword this depending on how we frame CMOT.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+uses this history to synchronize data between the backend and clients.
+{+service-short+} stores history in your synced Atlas cluster.
+
+.. _backend-compaction:
+
+Backend Compaction
+------------------
+
+**Backend compaction** is a maintenance process that automatically
+runs for all {+app+}s. During backend compaction, {+service-short+}
```
andrewlaguna824, line 40: Sentence might be a bit confusing "Realm runs through a realm's history"

Maybe "The backend runs through..."?

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+
+      CREATE table1.object1
+      UPDATE table1.object1.x = 42
+
+Compaction removes unnecessary steps from history to reduce space
+usage in your Atlas cluster. Additionally, compaction reduces the time
+and data usage of sync. Clients must download fewer historical steps to
+converge on the same result.
+
+Compaction removes any instruction that is later
```
dacharyc, line 68: This line seems redundant - we say basically the same thing in the opening paragraph, and it's also illustrated in the example and then almost the same thing is stated in the paragraph above.

I do think this is clearer than the final line of this section's intro paragraph, so I would probably replace "Compaction removes changes whose effects are later overwritten..." with "Compaction removes any instruction that is later overwritten by a newer instruction." - and then probably remove this paragraph, since you say above that compaction is a maintenance process that automatically runs for all Realm apps.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+overwritten by a future instruction. Compaction runs regularly on all
+synced clusters as part of {+sync+}.
+
+.. _advanced-backend-compaction:
+
+Advanced Backend Compaction
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Advanced backend compaction** uses client sync states to optimize
```
dacharyc, line 77: I had to read this a couple of times to parse it. I think we can communicate this more concisely and a bit more clearly. What about something like:

> Advanced backend compaction uses client sync states to further optimize a changeset. However, this optimization can prevent convergence for clients that have only seen a subset of history.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+   in the first place ignore the DELETE.
+
+   The following history uses advanced backend compaction to converge on
+   the same state:
+
+   .. code-block::
+
+      CREATE table1.obj1
+
+   Because ``object2`` was deleted, it has no impact on the current state
```
dacharyc, line 117: I think I'm missing something in this example. This CREATE is for object 1 on table 1, but object 2 was on table 2. So I'm not sure how object 2 being deleted relates to table 1/object 1.

(Also, here we're saying `table1.obj1` but above we said `table1.object1` - is this meant to be a new example? or should this example be instead `table1.object1`?)

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+
+The **client maximum offline time** controls the age limit of history
+optimized with :ref:`advanced backend compaction <advanced-backend-compaction>`.
+This indirectly changes how long a client can remain offline between
+synchronization sessions with the backend. Clients that do not
+synchronize for more than the specified number of days
+may experience a :ref:`client reset <client-resets>` the next time they
+connect with the backend.
+
+The lower the client maximum offline time, the sooner compaction can
```
dacharyc, line 137: Ahh, this paragraph is what I was missing before about why we'd want to use this. Nice!

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+
+Client Maximum Offline Time
+---------------------------
+
+The **client maximum offline time** controls the age limit of history
+optimized with :ref:`advanced backend compaction <advanced-backend-compaction>`.
+This indirectly changes how long a client can remain offline between
+synchronization sessions with the backend. Clients that do not
+synchronize for more than the specified number of days
+may experience a :ref:`client reset <client-resets>` the next time they
```
dacharyc, line 134: Do we need to go into more detail about why it's "may" experience a client reset?

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+
+As a result, the client must complete a client reset before it can
+resume synchronization. In a client reset scenario, the client deletes
+the client-local copy of a {+realm+} and downloads the current state of
+that {+realm+} from the backend. Synchronization then resumes using the
+new copy of the {+realm+}.
+
+The client maximum offline time controls how long your backend waits
+before applying advanced backend compaction. After the specified number
+of days without syncing, clients experience a client reset the next time
```
dacharyc, line 191: Above we say "may" experience a client reset, but here we seem to indicate that clients will definitely experience a client reset. Makes me wonder which is accurate.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+they connect with the backend.
+
+Applications that do not specify the client maximum offline time never
```
dacharyc, line 194: I think we can drop some unnecessary words here to shorten some of these sentences and improve readability. I fiddled with this and got to something like this:

> Applications that do not specify the client maximum offline time never apply advanced backend compaction. This means that clients can connect after any period of time offline -- weeks, months, or even years -- and synchronize changes. As time passes, frequently-edited realms may reflect many changes. With a large changeset, synchronization requires more time and data usage.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+      2. Click the :guilabel:`Client Maximum Offline Time (Optional)` dropdown
+         in the "Configuration" tab.
+
+         .. figure:: /images/client-maximum-offline-time.png
+           :width: 750px
+           :alt: The client maximum offline time configuration in a disabled state.
+
+      3. Click the :guilabel:`Enable Max Offline Time` toggle.
+
+      4. Enter a number greater than or equal to 1 specifying a number of
```
dacharyc, line 220:
```suggestion
      4. Enter a number greater than or equal to 1 specify a number of
```

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+
+   .. code-block::
+
+      CREATE table1.obj1
+
+   Because ``object2`` was deleted, it has no impact on the current state
+   of the {+realm+}. As a result, advanced backend compaction can
+   completely remove it from history.
+
+Advanced backend compaction only runs on clusters that specify a
```
andrewlaguna824, line 121:
```suggestion
Advanced backend compaction only runs on realm apps that specify a
```

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+
+         .. figure:: /images/client-maximum-offline-time.png
+           :width: 750px
+           :alt: The client maximum offline time configuration in a disabled state.
+
+      3. Click the :guilabel:`Enable Max Offline Time` toggle.
+
+      4. Enter a number greater than or equal to 1 specifying a number of
+         days for your application's client maximum offline time. The default
+         value is 30 days.
```
dacharyc, line 222: I think we're missing a step - it looks like users have to press the `Update Offline Time` button to accept/confirm the changes.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+   
+   .. tab::
+      :tabid: cli
+      
+      1. Pull a local copy of the latest version of your app with the
```
dacharyc, line 227: Might be good to link to link to the realm-cli pull page here so users can easily see any additional options they might want to specify, and link to push below for the same reason.

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+Key Concepts
+~~~~~~~~~~~~
+
+Sync should always converge at the same end state on all clients. In
+order to converge during a sync, clients require the full history of
+changes beginning immediately after their last sync. When a client does
+not sync for a long period of time, advanced backend compaction can
+alter the history in ways that prevent the client from converging. Since
+synchronization relies on all clients converging on a common result,
+such a client cannot synchronize.
```
mongodben, line 150: can you rephrase this line? "such a client" seems overly vague to me. took me a couple of readings of the paragraph to understand what you meant here

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+           :alt: The client maximum offline time configuration in a disabled state.
+
+      3. Click the :guilabel:`Enable Max Offline Time` toggle.
+
+      4. Enter a number greater than or equal to 1 specifying a number of
+         days for your application's client maximum offline time. The default
+         value is 30 days.
+   
+   .. tab::
+      :tabid: cli
```
dacharyc, line 225: I'm wondering this because we recently created several sets of Realm CLI instructions for the "Manage Realm Apps" section - should we add as prerequisites for this step having a copy of `realm-cli` and an API key, ala: [Deploy Changes with Realm CLI Prerequisites](https://docs.mongodb.com/realm/manage-apps/deploy/manual/deploy-cli/#prerequisites)?

Also, this made me wonder if the [Admin API/Sync API](https://docs.mongodb.com/realm/admin/api/v3/#sync-apis) needs to be updated - it looks like it only reflects the `service_id` and `partition_fields`, but I wonder if you've tried this endpoint to see if the max client offline time is reflected in the GET response?

(Unrelated to this PR, I've noticed that where we do specify CLI prerequisites, they're worded inconsistently - we should probably just make them an include so we're using the same prerequisites everywhere we provide CLI & API instructions. Maybe a maintenance task?)

source/sync/optimize-sync-atlas-usage.txt
mongodben: also could be good to talk about the implications of this here. as i understand it from reading further below, the implication is that there must be a client reset. 

might be better to put the example running lines 152-181 *after* the text running lines 181-199. this way the concept is explained fully before the example

source/sync/optimize-sync-atlas-usage.txt
```diff
@@ -0,0 +1,249 @@
...
+{+service-short+} stores history in your synced Atlas cluster.
+
+.. _backend-compaction:
+
+Backend Compaction
+------------------
+
+**Backend compaction** is a maintenance process that automatically
+runs for all {+app+}s. During backend compaction, {+service-short+}
+runs through a {+realm+}'s history. Compaction
```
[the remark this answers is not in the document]
nathan-contino-mongo: Already reworded this a little thanks to Andrew's suggestions, but I'm stealing the "the process" phrasing and "newer changes" change from here -- definite improvements. Thanks!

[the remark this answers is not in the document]
nathan-contino-mongo: Already reworded this a little thanks to Andrew's suggestions, but I'm stealing the "the process" phrasing and "newer changes" change from here -- definite improvements. Thanks!

source/sync/optimize-sync-atlas-usage.txt
nathan-contino-mongo: Reworded this to make this clearer. You are supposed to compare example 1) (no compaction) to example 2) (compaction) to example 3) (advanced compaction). If you think this is still unclear, happy to take restructuring/rewording suggestions.

source/sync/optimize-sync-atlas-usage.txt
nathan-contino-mongo: Not sure if this suggestion is proper english -- reworded this significantly instead:

```
         Specify a number of days for your application's client maximum
         offline time. The default value is 30 days. The minimum value is
         1.
```

source/sync/optimize-sync-atlas-usage.txt
nathan-contino-mongo: Couldn't agree more on the prerequisites bit here -- I'm planning a PD for refactoring all of these UI/CLI commands to use common parts next quarter. Stay tuned :) 

And adding this to the Admin API, which does include this field.

PR #1444, opened 2021-10-27
(DOCSP-16147): Node/RN - Update traversing set documentation

mongodben:
## Pull Request Info

Update info for traversing sets. Added **Traverse a Set** section. 

### Jira

- https://jira.mongodb.org/browse/DOCSP-16147

- [Sets (React Native)](https://docs-mongodborg-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16147/sdk/react-native/data-types/sets/)
- [Sets (Node)](https://docs-mongodborg-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-16147/sdk/node/data-types/sets/)

### Review Guidelines

### DIFF
--- a/examples/node/Examples/data-types.js
+++ b/examples/node/Examples/data-types.js
@@ -510,7 +510,8 @@ describe("Node.js Data Types", () => {
     };
     // :code-block-start: make-array-with-insertion-order-from-set
     let levelsCompletedInOrder = [];
-    async function onCharacterChange(character, changes) {
+    function onCharacterChange(character, changes) {
+      console.log("char set lvls completed", [...character.levelsCompleted]);
       if (
         !changes.deleted &&
         changes.changedProperties?.includes("levelsCompleted")
@@ -531,7 +532,7 @@ describe("Node.js Data Types", () => {
         schema: [characterSchema],
       });
 
-      await realm.write(() => {
+      realm.write(() => {
         playerOne = realm.create("Character", {
           _id: new BSON.ObjectId(),
           name: "PlayerOne",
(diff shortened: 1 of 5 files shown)


mongodben:
@cbush pls re-review when you get the chance

examples/node/Examples/data-types.js
```diff
@@ -471,4 +496,125 @@ describe("Node.js Data Types", () => {
...
+      });
+      // :code-block-start: traverse-a-set
+      playerOne.inventory.forEach((item) => {
+        console.log(item);
+      });
+      // :code-block-end:
+    } catch (err) {
+      console.error("error is", err);
+    } finally {
+      let totItems = 0;
```
cbush, line 533: If the realm fails to open (and throws an exception), this code will still run. The purpose of `finally` is to do cleanup that must happen whether the try block succeeded or failed.

examples/node/Examples/data-types.js
```diff
@@ -463,6 +463,31 @@ describe("Node.js Data Types", () => {
...
+      playerTwo.inventory.add(additions[4]);
+    });
+    console.log(playerTwo.inventory.values());
+
+    // size property counts number of items in set
+    expect(playerTwo.inventory.size).toBe(5);
+
+    // convert to array with Array.from()
+    const setAsArr = Array.from(playerTwo.inventory);
+    expect(Array.isArray(setAsArr)).toBe(true);
```
cbush, line 482: Since Array.from() always returns an array, is this really testing anything (other than the standard library)? What about `expect(setAsArr.length).toBe(playerTwo.inventory.size)`?

examples/node/Examples/data-types.js
```diff
@@ -463,6 +463,31 @@ describe("Node.js Data Types", () => {
...
+
+    // also convert to array with [...set]
+    const setAsArrAlt = [...playerTwo.inventory];
+    expect(setAsArr).toStrictEqual(setAsArrAlt);
+
+    // conversion to array **does't** guarantee insertion order
```
cbush, line 488:
```suggestion
    // conversion to array **doesn't** guarantee insertion order
```

examples/node/Examples/data-types.js
mongodben: i see you point. adjusting.

mongodben:
> Sorry for the delayed review. Still seeing a few issues. Let me know if you want to pair on the try/catch block.

thanks chris. let's pair on the try/catch block. i think 15-20 min of working thru it together would provide a lot of good reference for me going forward. will put some time on the calendar soon.

examples/node/Examples/data-types.js
mongodben: waiting on our pair programming to adjust.

mongodben:
@cbush ready for re-review

cbush (approved):
I can see this is building on some previous code where I would have done it differently. I think each test should open and close its own realm. You should avoid using bluehawk `uncomment` to replace lines as it invalidates the guarantee that the code is tested. The afterEach block should open the default realm and delete all of its contents (or alternatively delete the realm file from the fs).

Approving because I don't want to block you anymore and the user-facing parts look good, but we should really refactor these tests.

PR #1450, opened 2021-11-02
(DOCSP-17052): Create Projection documentation and code examples

dacharyc:
## Pull Request Info

This feature is not yet live. Do not merge until [PR #7375](https://github.com/realm/realm-cocoa/pull/7375/files) is merged and released. Once the release is live, resolve these items before merging:

- [x] Update `..versionadded` directive with release version
- [x] Update `Podfile` to point at release version instead of git repo/branch 
- [x] Add a link to Projection in Jazzy docs

### Jira

- https://jira.mongodb.org/browse/DOCSP-17052

- [Fundamentals](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17052/sdk/swift/fundamentals/object-models-and-schemas/#build-view-models-with-realm): New section: `Build View Models with Realm`
- [Define and Use Class Projections](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17052/sdk/swift/examples/define-and-use-class-projections/): New page
- [Test and Debug](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17052/sdk/swift/test-and-debug/#simplify-testing-with-class-projections): New section: `Simplify Testing with Class Projections`

### Review Guidelines


examples/ios/Examples/Projection.swift
```diff
@@ -0,0 +1,138 @@
...
+    @Persisted var city: String = ""
+    @Persisted var country = ""
+}
+// :code-block-end:
+
+// :code-block-start: declare-projection
+class ProjectionExample_PersonProjection: Projection<ProjectionExample_Person> {
+    @Projected(\ProjectionExample_Person.firstName) var firstName // Passthrough from original object
+    @Projected(\ProjectionExample_Person.address?.city) var homeCity // Rename and access embedded object property through keypath
+    @Projected(\ProjectionExample_Person.friends.projectTo.firstName) var firstFriendsName: ProjectedList<String> // Collection mapping
```
cbush, line 28: Neat

examples/ios/Examples/Projection.swift
```diff
@@ -0,0 +1,138 @@
...
+}
+// :code-block-end:
+
+ class ProjectionExample: XCTestCase {
+     override func setUp() {
+         Realm.Configuration.defaultConfiguration = Realm.Configuration(inMemoryIdentifier: "ProjectionExample")
+         let realm = try! Realm()
+
+         let jason = ProjectionExample_Person(value: ["firstName": "Jason",
```
cbush, line 37: The formatting in this section looks wrong. Can you use a swift formatter (or indent one level on the next line)

examples/ios/Examples/Projection.swift
```diff
@@ -0,0 +1,138 @@
...
+         XCTAssert(person.homeCity == "Zurich")
+         // :hide-end:
+         // Update projection property in a write transaction
+         try! realm.write {
+             person.firstName = "David"
+         }
+         // :code-block-end:
+         XCTAssert(person.firstName == "David")
+     }
+     // :code-block-start: test-with-projection
```
cbush, line 83: Should the codeblock be within the function?

examples/ios/Examples/Projection.swift
```diff
@@ -0,0 +1,138 @@
...
+     func testWithProjection() {
+         let realm = try! Realm()
+         // Create a Realm object, populate it with values
+         let jasonBourne = ProjectionExample_Person(value: ["firstName": "Jason",
```
cbush, line 87: Formatting is incorrect?

examples/ios/Examples/Projection.swift
```diff
@@ -0,0 +1,138 @@
...
+                                                       "lastName": "Bourne",
+                                                       "address": [
+                                                         "city": "Zurich",
+                                                         "country": "Switzerland"]])
+         try! realm.write {
+             realm.add(jasonBourne)
+         }
+         
+         // Retrieve the Realm object as a projection
```
cbush, line 96: Can this be clarified?

source/sdk/ios/examples/define-and-use-projections.txt
```diff
@@ -0,0 +1,149 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+.. versionadded:: 10.?
+
+A projection is a class that passes through or transforms some or all of your
+{+realm+} object's properties. Projection enables you to build view models
```
cbush, line 21: Capital-R "Realm" in "Realm object" ({+client-database+})

source/sdk/ios/examples/define-and-use-projections.txt
```diff
@@ -0,0 +1,149 @@
...
+--------
+
+.. versionadded:: 10.?
+
+A projection is a class that passes through or transforms some or all of your
+{+realm+} object's properties. Projection enables you to build view models 
+that use an abstraction of your object model. This simplifies using and testing
+{+client-database+} objects in your application.
+
+With projection, you can use a subset of your object's properties in the UI,
```
cbush, line 25: Subject did not change, so no comma
```suggestion
With projection, you can use a subset of your object's properties in the UI 
```

source/sdk/ios/examples/define-and-use-projections.txt
```diff
@@ -0,0 +1,149 @@
...
+
+.. versionadded:: 10.?
+
+A projection is a class that passes through or transforms some or all of your
+{+realm+} object's properties. Projection enables you to build view models 
+that use an abstraction of your object model. This simplifies using and testing
+{+client-database+} objects in your application.
+
+With projection, you can use a subset of your object's properties in the UI, 
+or change them into another form before displaying them. When you use a
```
cbush, line 26: "change them into another form" -> "transform"?

source/sdk/ios/examples/define-and-use-projections.txt
```diff
@@ -0,0 +1,149 @@
...
+   <realm/realm-cocoa/tree/master/examples#projections>`.
+
+.. _ios-react-to-changes-to-a-projection:
+
+React to Changes to a Projection
+--------------------------------
+
+Like other {+realm+} objects, you can :ref:`react to changes <ios-react-to-changes>`
+to a projection. When you register a projection change listener, you see 
+notification for changes made through the projection object directly.
```
cbush, line 134:
```suggestion
notifications for changes made through the projection object directly. 
```

source/sdk/ios/fundamentals/object-models-and-schemas.txt
```diff
@@ -163,6 +163,26 @@ boilerplate query code and capacity for error.
...
    :ref:`ios-declare-relationship-properties`.
 
+.. _ios-build-view-models-with-realm:
+
+Build View Models with Realm
+----------------------------
+
+.. versionadded:: 10.?
+
+You can work with a subset of your {+realm+} object's properties by creating
```
cbush, line 173:
```suggestion
You can work with a subset of your {+client-database+} object's properties by creating 
```

source/sdk/ios/examples/define-and-use-projections.txt
```diff
@@ -0,0 +1,149 @@
...
+
+.. _ios-define-a-projection:
+
+Define a Projection
+-------------------
+
+Define a projection by creating a class of type ``Projection``. Specify
+the :swift-sdk:`Object <Extensions/Object.html>` or 
+:swift-sdk:`EmbeddedObject <Extensions/EmbeddedObject.html>` base that the
+projection should derive from. Use the ``@Projected`` property wrapper to
```
cbush, line 58: Can you base a projection on another projection? If not, should that be mentioned here?

I don't know about the word "derive". Implies subclassing and I don't know if that's actually the case. Maybe like "should be a projection of" or something?

cbush (approved):

examples/ios/Examples/Projection.swift
dacharyc: For this example, I was trying to show explicitly that it was a test func, so I put the code block markup outside of the function. If you think that's not necessary, I can move the Bluehawk commands so I'm just showing the func contents.

source/sdk/ios/examples/define-and-use-projections.txt
```diff
@@ -0,0 +1,148 @@
...
+view models that use an abstraction of your object model. This simplifies 
+using and testing {+client-database+} objects in your application.
+
+With projection, you can use a subset of your object's properties directly 
+in the UI or transform them. When you use a projection for this, you get 
+all the benefits of {+client-database+}'s live objects:
+
+- The projected object live updates
+- You can observe it for changes
+- You can do derived queries on top of results
```
pavel-ship-it, line 31: This is not available in this version. The workaround is to run a query against Realm objects and then protect the result.

dacharyc: Thanks, @pavel-ship-it ! Removed the line. Do you think it's important to call out this workaround in the docs? I'm wondering if developers will expect to be able to do this and need to be told how to work around the lack of this feature, or if people won't expect to be able to do this we can just remove the mention.

pavel-ship-it: I'd say the users will try to query projections and results could be unexpected. Querying field with the same name and type as in the original object will work, if name/type has changed it will fail

dacharyc: Alright, thanks! I've added a note later on the page about this, which says:

> Don't do derived queries on top of projection results. Instead, run a query against the Realm object directly and then project the result. If you try to do a derived query on top of projection results, querying a field with the same name and type as the original object works, but querying a field with a name or type that isn't in the original object fails.

Do you think this covers the explanation, or should we also provide an example? I got a little stuck when trying this out myself - I wondered if there was an example of doing this in any of the projection tests?

examples/ios/Examples/Projection.swift
```diff
@@ -0,0 +1,143 @@
...
+    @Persisted var city: String = ""
+    @Persisted var country = ""
+}
+// :code-block-end:
+
+// :code-block-start: declare-projection
+class ProjectionExample_PersonProjection: Projection<ProjectionExample_Person> {
+    @Projected(\ProjectionExample_Person.firstName) var firstName // Passthrough from original object
+    @Projected(\ProjectionExample_Person.address?.city) var homeCity // Rename and access embedded object property through keypath
+    @Projected(\ProjectionExample_Person.friends.projectTo.firstName) var firstFriendsName: ProjectedList<String> // Collection mapping
```
pavel-ship-it, line 28: There is a change in collection projections - both List and MutableSet will be projected as ProjectedCollection type.
will look like that:
```
class PersistedCollections: Object {
    @Persisted public var list: MutableSet<CommonPerson>
    @Persisted public var set: List<CommonPerson>
}

class ProjectedCollections: Projection<PersistedCollections> {
    @Projected(\PersistedCollections.list.projectTo.firstName) var list: ProjectedCollection<String>
    @Projected(\PersistedCollections.set.projectTo.firstName) var set: ProjectedCollection<String>
}
```
Basically, just replace `ProjectedList` with `ProjectedCollection`

dacharyc: Thanks for calling this out, @pavel-ship-it - just added updated this to add those details 👍

PR #1454, opened 2021-11-04
(DOCSP-17188): Update OpenAPI spec with add template for creating new application

mongodben:
## Pull Request Info

Update the OpenAPI spec route for POST `/groups/{groupId}/apps` to include information about adding a template to app creation. 

Note that while updating the spec, a bug with our OpenAPI spec renderer was discovered in which the `data_source` field is marked as required when it's not, presumably because some of its own fields are required _iff_ the data_source is included. However there is no adverse effect of of always the user to always include the data_source field. 

### Jira

- https://jira.mongodb.org/browse/DOCSP-17188

- [Realm Administration API](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17188/admin/api/v3/#post-/groups/%7Bgroupid%7D/apps)

### Review Guidelines

### DIFF
--- a/source/openapi-admin-v3.yaml
+++ b/source/openapi-admin-v3.yaml
@@ -2494,13 +2494,12 @@ components:
         name:
           type: string
           description: Can only contain ASCII letters, numbers, underscores, and hyphens.
+          required: true
         template_id:
           type: string
           description: Name of :ref:`supported template app <template-apps>` to serve as base template.
         data_source: 
-            $ref: '#/components/schemas/DataSource'
-      required:
-        - name
+          $ref: '#/components/schemas/DataSource'
     DataSource: 
       type: object
       properties: 
@@ -2510,22 +2509,22 @@ components:
           example: mongodb-atlas
           enum: 
             - mongodb-atlas
+          required: true
         type: 
           type: string
           description: Should be `"mongodb-atlas"`.
           example: mongodb-atlas
           enum: 
             - mongodb-atlas
+          required: true
         config:
           type: object
           properties:
             clusterName: 
               type: string
-              description: Name of {+atlas-short+} cluster associated with the {+app+}. 
-      required:
-        - name
-        - type
-        - config
+              description: Name of {+atlas-short+} cluster associated with the {+app+}.
+              required: true
+          required: true
     Service:
       properties:
         _id: {type: string}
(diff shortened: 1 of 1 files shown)


source/openapi-admin-v3.yaml
```diff
@@ -2494,7 +2494,40 @@ components:
...
+        data_source: 
+          $ref: '#/components/schemas/DataSource'
+      required:
+      - name
+    DataSource: 
+      type: object
+      properties: 
+        name: 
+          type: string
+          description: Should be `"mongodb-atlas"`.
```
nathan-contino-mongo, line 2509: Do all template apps assume this datasource name? If so, this is fine; if not, we might want to explain how the name is customizable.

source/openapi-admin-v3.yaml
```diff
@@ -2494,7 +2494,40 @@ components:
...
+          example: mongodb-atlas
+          enum: 
+            - mongodb-atlas
+        type: 
+          type: string
+          description: Should be `"mongodb-atlas"`.
```
nathan-contino-mongo, line 2515: Similarly, Realm supports other data sources, including (and maybe limited to?) data lake -- if that's a possibility for template apps, we should mention it. I don't think non-atlas data sources support sync, so *probably not*, but worth considering if you haven't already.

nathan-contino-mongo (approved):
LGTM with a couple of small questions.

source/openapi-admin-v3.yaml
mongodben: there was actually a unit test in the source code for the API explicitly prohibiting data lake as a data source. not sure abt others, if they exist

source/openapi-admin-v3.yaml
mongodben: if it's customizable, i'm not sure how it would be. based on your knowledge of realm/atlas, what else would this be?

source/openapi-admin-v3.yaml
nathan-contino-mongo: Seems fair then!

mongodben:
thx for review @nathan-contino-mongo. merging changes. pls see additional comments. while this change is an improvement on that which was there, seems like there could be some additional improvements per your questions

PR #1456, opened 2021-11-04
Create a Collection/Role/Filter

nlarew:
## Pull Request Info

- Splits the **Define Roles & Permissions** page into smaller guides on each concept's page
- Refocus **Define Roles & Permissions** to **Configure a Collection**

### Jira

- (DOCSP-18473): Create a Role
- (DOCSP-18474): Create a Filter

- [Configure a Collection](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/rules-config/mongodb/configure-a-collection/)
- [Create a Role](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/rules-config/rules/roles/#create-a-role)
- [Create a Filter](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/rules-config/rules/filters/#create-a-filter)

### Review Guidelines


source/includes/steps-configure-a-collection.yaml
```diff
@@ -0,0 +1,88 @@
...
+           the process of configuring role-based permissions. Each template is a
+           pre-configured set of roles that represents a common use case and
+           data access pattern. You can customize each template to match your
+           data model or select :guilabel:`No Template` to start with a single
+           default role.
+     
+     .. tab::
+        :tabid: cli
+        
+        You configure collections in namespaced directories within a
```
MongoCaleb, line 31: What are "namespaced directories"? (also, Namespaced, AFAIK, is not a word)

source/rules/roles.txt
```diff
@@ -241,19 +241,14 @@ A role applies to a given document if its Apply When :ref:`expression
...
 
-.. note::
+Create a Role
+-------------
 
-   To learn how to configure and deploy a role in your app, see
-   :ref:`define-roles`.
+.. note::
    
    For examples of expressions and roles for common scenarios, see
```
MongoCaleb, line 251: the term "expressions" has not be introduced on this page prior to this note, so I think you need a bit more explanation and/or call it a "apply When expression" or just remove it, since step 3 in this section explains them.

source/includes/steps-configure-a-collection.yaml
```diff
@@ -0,0 +1,88 @@
...
+  create, read, modify, and search data in the collection. You can use
+  rules to reshape and omit data as well as enforce role-based permissions
+  for entire documents and individual fields within each document.
+  
+  To learn how to define rules, see :ref:`roles` and :ref:`filters`.
+  
+  .. important:: Sync Rules
+     
+     Synced clusters use Realm's partition-based read & write rules instead of
+     role-based access permissions. For more information, see :ref:`sync-rules`.
```
MongoCaleb, line 68: perhaps "instead of the role-based access permissions defined in Atlas"? or "Atlas's role-based..."?

source/includes/steps-create-a-filter.yaml
```diff
@@ -0,0 +1,180 @@
...
+---
+title: Specify a Query Filter
+ref: specify-a-query-filter
+content: |
+  .. tabs-realm-admin-interfaces::
+     
+     .. tab::
+        :tabid: ui
+        
+        In the :guilabel:`Query` input box, specify a document that contains
```
MongoCaleb, line 97: "specify a document" might be confusing, as people might think you mean an existing document. Perhaps "create a json object that contains the additional query predicates needed to filter the data"

source/includes/steps-create-a-filter.yaml
```diff
@@ -0,0 +1,180 @@
...
+           { "score": { "$gt": 20 } }
+        
+        .. figure:: /images/filters-query.png
+           :alt: A filter's query filter input
+     
+     .. tab::
+        :tabid: cli
+        
+        In the ``query`` field, specify a MongoDB query filter to add to
+        filtered incoming queries. If you do not want to add additional query
```
MongoCaleb, line 113: If they are already "filtered incoming queries", why are we specifying a filter?

source/includes/steps-create-a-filter.yaml
```diff
@@ -0,0 +1,180 @@
...
+title: Specify a Projection
+ref: specify-a-projection
+content: |
+  .. tabs-realm-admin-interfaces::
+     
+     .. tab::
+        :tabid: ui
+        
+        In the :guilabel:`Projection` input box, specify a MongoDB
+        :manual:`projection </tutorial/project-fields-from-query-results>` add
```
MongoCaleb, line 134: "*to* add to"

source/includes/steps-create-a-role.yaml
```diff
@@ -0,0 +1,231 @@
...
+        .. code-block:: json
+           
+           {
+             "name": "Owner"
+           }
+---
+title: Specify an Apply When Expression
+ref: specify-an-apply-when-expression
+content: |
+  A role's Apply When :ref:`expression <expressions>` evaluates to ``true`` when
```
MongoCaleb, line 62: Just [pondering if the "Apply When" should be in quotes, bolded, or something to indicate it as a "thing" and just just capitalized? Dunno, just a thought to consider.

MongoCaleb (commented):
Back to you for a few fairly minor things

source/includes/steps-configure-a-collection.yaml
nlarew: This is actually referring to Realm's role-based access permissions - will try to clarify

source/includes/steps-create-a-role.yaml
nlarew: It's tricky, ultimately I wish it was just one word. We used to always wrap it in a guilabel but that only really makes sense for UI stuff. Recently I've been just capitalizing without any other styles.

That said Moe has also called this point out. Do you think it's a matter of "we're used to it being a guilabel" or is this a real point of confusion in general?

MongoCaleb: Valid points. It's not a gui element, and I now don't thin k it should be bold. What if we add a descriptor: 
"A role's Apply When [field] or [definition] or [schema] or [json bloberooni]"
If all else fails, leave it as is. I can live with it.

MongoCaleb (approved):

PR #1460, opened 2021-11-08
(DOCSP-19554): Add tip around SPM with Xcode 13.1

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19554

- [Install Realm for iOS, macOS, etc.](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19554/sdk/ios/install/#specify-options): In the SwiftPM tab, step 3, added a `Tip` with the relevant info

### Review Guidelines

### DIFF
--- a/source/includes/steps-install-swiftpm.yaml
+++ b/source/includes/steps-install-swiftpm.yaml
@@ -26,8 +26,11 @@ content: |
   .. tip::
 
      If you are using Xcode version 13.1 to install the RealmSwift SDK,
-     you may run into an issue where ``Up to Next Major`` installs an old
-     version of the SDK. You can resolve this issue in one of two ways:
+     you may run into an issue where ``Up to Next Major`` installs version
+     ~5.x instead of ~10.x. This is related to the way Xcode 13.1 handles
+     versions, and will hopefully be resolved in a future Xcode release.
+     
+     Meanwhile, you can bypass this issue in one of two ways:
 
      - Set the version as ``Up to Next Major`` and set the minimum version to
        ``10.19.0`` (or the minimum version you want to install)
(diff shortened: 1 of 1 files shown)


source/includes/steps-install-swiftpm.yaml
```diff
@@ -23,6 +23,16 @@ content: |
   In the options for the ``realm-cocoa`` package which comes up, leave the default 
   ``Dependency Rule`` of ``Up to Next Major``, then click ``Add Package``.
 
+  .. tip::
+
+     If you are using Xcode version 13.1 to install the RealmSwift SDK,
+     you may run into an issue where ``Up to Next Major`` installs an old
+     version of the SDK. You can resolve this issue in one of two ways:
```
mongodben, line 30: how does the user know if the wrong version of the RealmSwift SDK is being installed? what are the expected errors to be seen if the wrong version is installed? 

i think it'd be helpful to the reader if another sentence or 2 were added abt:
1. what version of the RealmSwift SDK should be installed
2. what it looks like/what will happen if an old version of the SDK is installed

mongodben (commented):
hi looks good but could use a bit of expansion on what'll happen if the error occurs. see comment on text for more.

source/includes/steps-install-swiftpm.yaml
dacharyc: Added more details around the issue. "What it looks like" is that the old version that I've added here is installed, instead. Does this help?

mongodben: def better! is there any expected error behavior that happens when the old version is installed? would be good to note here if so

dacharyc: If they're on this page because they're installing Realm for the first time, they won't see any errors because they'll be starting from a new project. It won't have any code. They'll just see version 5.x installed instead of 10.x.

mongodben:
lgtm

mongodben (approved):

PR #1463, opened 2021-11-09
(DOCSP-17705): Document user metadata for iOS SDK

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-17705

- [User Metadata - iOS SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17705/sdk/ios/examples/users/user-metadata/): New page with completely novel content around using user metadata in the SDK. Based on the custom user data page.
- [Work with Users - iOS SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17705/sdk/ios/examples/work-with-users/): New page collecting the `User` examples into a single section, with a description of the user object and what you can do with it
- [Enable User Metadata](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-17705/users/enable-user-metadata/): New backend page describing how to configure user metadata in the auth provider on the backend

### Review Guidelines


source/sdk/ios/examples/work-with-users.txt
```diff
@@ -0,0 +1,81 @@
...
+to :ref:`access custom user data <ios-read-a-users-custom-data>`. 
+
+To :ref:`create <ios-create-a-users-custom-data-document>` and :ref:`update 
+<ios-update-a-users-custom-data>` custom user data, you must access 
+your MongoDB data source directly. {+backend-short+} does not offer a SDK 
+method to create or update this custom user data; it's a read-only property.
+
+.. _ios-app-read-user-metadata-from-login-providers:
+
+Read User Metadata from Login Providers
```
cbush, line 74: Why are we using Login Providers here instead of Authentication Providers? Not against it, just curious.

cbush (approved):
Looks great. A few nitpicks.

source/sdk/ios/examples/users/user-metadata.txt
```diff
@@ -0,0 +1,72 @@
...
+   :language: swift
+
+User metadata options vary depending on which provider you're using, and
+:ref:`which metadata fields you enable <configure-user-metadata-on-the-backend>`. 
+
+.. warning:: User Metadata May Be Stale
+   
+   {+backend+} fetches the most recent version of user metadata whenever a 
+   user refreshes their :ref:`access token <user-sessions>`, which is used 
+   by most SDK operations that contact the {+backend+} back end. If the token
```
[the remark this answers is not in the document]
dacharyc: That's fair. I do have to check the snooty.toml sometimes multiple times in a PR to figure out which substitution to use. (This one renders as `MongoDB Realm` so it would be "MongoDB Realm back end.")

I suppose we could change the naming of these to be more descriptive to a reviewer. Maybe something like `{+backend-MongoDB-Realm+}` (can you have caps in these?) But then if the name changes, instead of just changing the source constant, we'd have to do a find and replace and replace all of the invocation instances, thus defeating the purpose.

source/sdk/ios/examples/work-with-users.txt
dacharyc: This is kind of a nit of mine. `Authentication Providers` feels like jargon, and it's also a lot of syllables which negatively impacts readability. Plus, higher on the page, we have a very simple heading "Log Users In and Out" - so this parallels that language. If we say "Authenticate Users" in the first heading so we can use "Authentication Providers" here, we don't cover all the use cases described in the section following the first heading (i.e. log the user out, remove the user), and that also gets back to "Authenticate" feeling like jargon to me.

I know "authentication" is a common term for developers and it has a specific meaning, so if we feel strongly that this ought to be "Authentication", I'll change it. But my personal inclination is to talk like a human in docs, and this is one of those areas where I think we can tip the balance.

cbush: I don't feel strongly. I can see your point. I hate to favor consistency over good. I wonder then if we should think about renaming "Authentication Providers" entirely to Login Providers (though this would clash with the UI)

PR #1475, opened 2021-11-15
(DOCSP-19239): Add AsyncOpen property wrapper to SwiftUI quick start

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19239

- [Quick Start with SwiftUI](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19239/sdk/ios/swiftui/)

### Review Guidelines


source/examples/generated/swiftui/sync/QuickStart.codeblock.complete-swiftui-quick-start.swift
```diff
@@ -76,68 +76,87 @@ struct LocalOnlyContentView: View {
...
-// The view for opening a realm and setting up a group
-struct RealmContentView: View {
-    
-    // Observe a realm that may be opened after login.
-    @State var realm: Realm?
+/// This view opens a synced realm.
+struct OpenSyncedRealmView: View {
+    // Use AsyncOpen to download the latest changes from
+    // your Realm app before opening the realm.
+    @AsyncOpen(appId: YOUR_REALM_APP_ID_HERE, partitionValue: "", timeout: 4000) var asyncOpen
```
cbush, line 114: When is partitionValue actually passed to this?

dacharyc: Ahh, right, I wonder if this needs some more explanation? It's being passed in as an environment object from ln 102 above:

`OpenSyncedRealmView().environment(\.partitionValue, user.id)`

But my memory is a little fuzzy on the details of why we implemented this in this way. I _think_ I remember Diana saying that if we pass it in like this, with an empty string in the AsyncOpen call, it will take the partition value from the environment object.

dacharyc: Yeah, I've confirmed with the devs that it is not currently possible to remove `partitionValue` from the initializer, so this is the recommended way of using this. I've added some more explicit details around this aspect of it in my most recent commit, including a couple of one-line code examples to illustrate this as I state it explicitly. Do you think that clarifies things well enough?

cbush: It would be nice to also add a comment above this line that "the partitionValue is actually read from the environment" or something along those lines. Otherwise LGTM

cbush (approved):

PR #1486, opened 2021-11-22
(DOCSP-15317): Create Docs on InMemory Realms + Upgrade Docs Unit Test Suite to RealmJSv10.10.1

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-15317

- [Open & Close a Realm - Node.js SDK # Open an In-Memory Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-15317-InMemory-Realms-JS/sdk/node/examples/open-and-close-a-realm/#open-an-in-memory-realm)
- [Open & Close a Realm - React Native SDK # Open an In-Memory Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-15317-InMemory-Realms-JS/sdk/react-native/examples/open-and-close-a-realm/#open-an-in-memory-realm)

### Review Guidelines


source/includes/js-open-an-in-memory-realm.rst
```diff
@@ -0,0 +1,13 @@
+Open an In-Memory Realm
+-----------------------
+To create a realm that runs entirely in memory without being written to a file,
```
dacharyc, line 3: Feel free to take this or leave this, but in the iOS SDK, we're a bit more explicit:

> You can open a realm entirely in memory, which will not create a `.realm` file or its associated auxiliary files. Instead the SDK stores objects in memory while the realm is open and discards them immediately when all instances are closed.

The iOS SDK has a section about auxiliary files, which this links to. I think it's a bit more complete, but am uncertain whether that extra info is actually needed. Feel free to ignore this note if you think it adds unnecessary verbiage that isn't really valuable here.

dacharyc (approved):
LGTM overall! Just one note about where we might want to be more explicit, but feel free to take it or leave it.

source/includes/js-open-an-in-memory-realm.rst
mongomoe: JS actually does create those auxiliary files as of right now. The only file not created is the .realm file unless memory is low, in which case the .realm file is emptied after the realm is closed. Going to merge this as is. Thanks for the review!

PR #1487, opened 2021-11-22
(DOCSP-19805): Change Array documentation to exclude required values with `!`

mongodben:
## Pull Request Info

This pull request updates the GraphQL documentation to make it so that arrays are not required to have values in them. 

It should be noted that this is actually contrary to the GraphQL specification. However this change updates the documentation to reflect the current state of our GraphQL implementation. For more information on the subject, see https://jira.mongodb.org/browse/REALMC-9030. 

### Jira

- https://jira.mongodb.org/browse/DOCSP-19805

- [GraphQL Types, Resolvers, and Operators](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19805/graphql/types-and-resolvers/)

### Review Guidelines

### DIFF
--- a/source/graphql/types-and-resolvers.txt
+++ b/source/graphql/types-and-resolvers.txt
@@ -92,7 +92,7 @@ is specified.
      rated: String
      runtime: Int
      director: String
-     cast: [String!]
+     cast: [String]
    }
 
 .. _graphql-field-mapping:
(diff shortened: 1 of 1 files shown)


dacharyc (approved):
LGTM with one question: the change itself seems fine, but because this is actually contrary to the GraphQL spec, should we add a callout box explicitly stating more info about this? i.e. 

> Unlike the GraphQL spec, we do not require `!` when defining arrays.

Something noting that this is not just an oversight in the docs, but is actually how we implement this in Realm.

mongodben:
i don't think this one is necessary to call out. the way we're doing it is a superset of the graphql spec, so all valid graphql is also valid for our implementation. and our graphql schema is generated so user wouldn't actually be defining.

PR #1488, opened 2021-11-22
(DOCSP-19807): [iOS] Add User-defined Object @Persisted syntax details

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19807

- [Supported Property Types](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19807/sdk/ios/data-types/supported-property-types/): New row at bottom of table, note below table.

### Review Guidelines

### DIFF
--- a/source/sdk/ios/data-types/supported-property-types.txt
+++ b/source/sdk/ios/data-types/supported-property-types.txt
@@ -154,6 +154,12 @@ Property Cheat Sheet
 
                 @Persisted var value: AnyRealmValue
            - N/A
+         * - User-defined :swift-sdk:`Object <Extensions/Object.html>`
+           - N/A
+           - .. code-block:: swift
+                :copyable: false
+           
+                @Persisted var myObject: MyClass?
       
       Additionally:
 
@@ -162,6 +168,9 @@ Property Cheat Sheet
 
       ``CGFloat`` properties are discouraged, as the type is not
       platform independent.
+
+      To use Key-Value Coding with a user-defined object in the ``@Persisted`` 
+      syntax, add the ``@objc`` attribute: ``@Persisted @objc var myObject: MyClass?``
 
       Setting Default Values
       ~~~~~~~~~~~~~~~~~~~~~~


source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -163,6 +169,9 @@ Property Cheat Sheet
       ``CGFloat`` properties are discouraged, as the type is not
       platform independent.
 
+      To use Key-Value Coding with a user-defined object in the ``@Persisted`` 
+      syntax, add the ``@objc`` attribute: ``@Persisted @objc var myObject: MyClass?``
```
cbush, line 173: Does myObject _have_ to be optional?

cbush (approved):

source/sdk/ios/data-types/supported-property-types.txt
```diff
@@ -154,6 +154,12 @@ Property Cheat Sheet
 
                 @Persisted var value: AnyRealmValue
            - N/A
+         * - User-defined :swift-sdk:`Object <Extensions/Object.html>`
+           - N/A
+           - .. code-block:: swift
+                :copyable: false
+           
+                @Persisted var myObject: MyClass?
```
cbush, line 162: Should this have `@objc` too?

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: I believe so. It's the `@Persisted` version of the `@objc dynamic` declaration, and that is optional only.

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: I left it out here because it _works_ when you use `@Persisted` without `@objc` - _except_ that KVC does not work unless you add the `@objc`. This is based on a member of the Cocoa team first telling a dev to use `@Persisted var myObject: MyClass?`, and then adding `@objc` when the dev reported their KVC unit tests weren't working. I gather the preferred syntax is plain `@Persisted` and adding the `@objc` is an additional step for some use cases. 

I'm not aware of another declaration where we have this, so I'm undecided whether to present it as "always use this syntax even though it doesn't match others" or "only use this additional attribute if you want to use KVC."

source/sdk/ios/data-types/supported-property-types.txt
cbush: Should we say so here?

source/sdk/ios/data-types/supported-property-types.txt
cbush: No, better to leave it off

source/sdk/ios/data-types/supported-property-types.txt
dacharyc: Hmm. To me, that seems like an additional detail that is already covered by the table above, and isn't really part of the info we're communicating here - i.e. this is about adding the `@objc` attribute to make KVC work with a user-defined object. But if you feel it's also worth calling out here that it's only optional, I can add a note stating that.

cbush: I think it's fine then.

PR #1501, opened 2021-11-29
Add info on bundling a synced realm; add new code snippet; replace deprecated GetSession()

MongoCaleb:
### Jira
- [Bundle Synced Realm](https://jira.mongodb.org/browse/DOCSP-16502)
- [Replace GetSession()](https://jira.mongodb.org/browse/DOCSP-19890)

- https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/bundledsyncrealm/sdk/dotnet/advanced-guides/bundle-a-realm/
- https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/bundledsyncrealm/sdk/dotnet/examples/sync-changes-between-devices

### DIFF
--- a/examples/dotnet/Examples/BundleARealmExamples.cs
+++ b/examples/dotnet/Examples/BundleARealmExamples.cs
@@ -35,18 +35,30 @@ namespace Examples
         [Test]
         public async Task TestWriteCopySynced()
         {
-            var app = App.Create(Config.appid);
-            var user = app.CurrentUser;
+            var appConfig = new AppConfiguration(Config.appid);
+            var app = App.Create(appConfig);
+            var user = app.LogInAsync(Credentials.Anonymous()).Result;
 
             // :code-block-start: copy_a_synced_realm
+
             // open an existing realm
-            var realm = await Realm.GetInstanceAsync();
+            // :uncomment-start:
+            // var existingConfig = new SyncConfiguration("myPartition", user);
+            // :uncomment-end:
+            // :hide-start:
+            var existingConfig = new SyncConfiguration("myPartition", user)
+            {
+                Schema = new[] { typeof(Examples.Models.User) }
+            };
+            // :hide-end:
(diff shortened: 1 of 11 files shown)


examples/dotnet/Examples/BundleARealmExamples.cs
```diff
@@ -30,6 +32,49 @@ public void TestWriteCopy()
...
+            var user = app.LogInAsync(Credentials.Anonymous()).Result;
+
+            // :code-block-start: copy_a_synced_realm
+
+            // open an existing realm
+            // :uncomment-start:
+            // var existingConfig = new SyncConfiguration("myPartition", user);
+            // :uncomment-end:
+            // :hide-start:
+            var existingConfig = new SyncConfiguration("myPartition", user)
```
cbush, line 49: Is there a way to set a default configuration for the test so we don't have to do this uncomment/hide stuff?

cbush (approved):
One comment, not really blocking but something to think about. I'd prefer not to over-use Bluehawk this way.

examples/dotnet/Examples/BundleARealmExamples.cs
MongoCaleb: Not really. We need to filter out all of the test objects we use, but I don't want to show that "fluff" to the reader.

nirinchev (commented):
LGTM - the only nit on my end is that the [Using the Copied Realm File](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/bundledsyncrealm/sdk/dotnet/advanced-guides/bundle-a-realm/#using-the-copied-realm-file) section only shows how to use the bundled local Realm but not the synchronized Realm. It may trick people into thinking they no longer need to login a user or specify a partition when using bundled Realm files, but they still do.

Issue #1498, opened 2021-11-26
Realm node.js app tutorial: email must be of type 'string', got ([object Object])


mangkoran:
First of all, thank you for the amazing tutorial!

However, I got an error when trying to run then node.js tutorial app.

![image](https://user-images.githubusercontent.com/57625126/143572906-edc0dfdf-8fb7-4b9b-8931-6e58c1960fa2.png)

I tried with [`final`](https://github.com/mongodb-university/realm-tutorial-node-js/tree/final) branch too, but still got the same error.

mangkoran:
After looking the source code, the attached diff solved the issue.

```diff
diff --git a/users.js b/users.js
index ae150c6..35c8ef8 100644
--- a/users.js
+++ b/users.js
@@ -72,10 +72,7 @@ async function registerUser() {
   ]);

   try {
-    await app.emailPasswordAuth.registerUser({
-      email: input.email,
-      password: input.password
-    });
+    await app.emailPasswordAuth.registerUser(input.email, input.password);
     const credentials = Realm.Credentials.emailPassword(
       input.email,
       input.password
```

PR #1502, opened 2021-11-29
(DOCSP-19905): Fix realm JS SDK versions update Node Tutorial user custom data access, and eslint for web tutorial 

mongodben:
## Pull Request Info

This PR was initiated by this issue - https://github.com/mongodb/docs-realm/issues/1498 

While addressing the issue by bumping the Realm JS version in the web tutorial and across the docs-realm repo, a couple of other issues were encountered and addressed: 
- the Node tutorial `getProjects()` function for getting custom user data was updated to use the `refreshCustomData()` method and have error handling for slow to fire triggers. 
- changed the ESLint version of the web tutorial to be compatible with Create React App

### Jira

- https://jira.mongodb.org/browse/DOCSP-19905 

- [NodeJS CLI Tutorial](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/fix_realm_js_versions/tutorial/nodejs-cli/)

### Review Guidelines

### DIFF
--- a/source/tutorial/generated/node/projects.codeblock.getProjects.js
+++ b/source/tutorial/generated/node/projects.codeblock.getProjects.js
@@ -1,7 +1,15 @@
 async function getProjects() {
-  const realm = await index.getRealm(`user=${users.getAuthedUser().id}`);
-  const currentUser = users.getAuthedUser().id;
-  const user = realm.objectForPrimaryKey("User", currentUser);
-  const projects = user.memberOf;
-  return projects;
+  const user = users.getAuthedUser();
+  try {
+    const { memberOf: projects } = await user.refreshCustomData();
+
+    // Make sure that the user object has been created
+    if (!projects) {
+      output.error("The user object hasn't been created yet. Try again soon.");
+      throw new Error("No projects for user");
+    }
+    return projects;
+  } catch (err) {
+    output.error("There was a problem accessing custom user data");
+  }
 }
--- a/source/tutorial/nodejs-cli.txt
+++ b/source/tutorial/nodejs-cli.txt
@@ -286,8 +286,8 @@ In ``projects.js``:
 * ``getProjects``
   
   As defined by our data model, ``projects`` are an embedded object of the ``users`` object. 
-  To get all projects the user is a part of, we'll have to use the ``objectForPrimaryKey``
-  method to get the current user and then access the current user's ``memberOf`` property.
+  To get all projects the user is a part of, we need to get the current user.
+  Then we'll use the ``refreshCustomData`` method to get the current user's ``memberOf`` property.
 
   .. literalinclude:: /tutorial/generated/node/projects.codeblock.getProjects.js
      :language: javascript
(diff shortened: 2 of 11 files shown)


source/tutorial/generated/node/projects.codeblock.getProjects.js
```diff
@@ -1,7 +1,15 @@
 async function getProjects() {
-  const realm = await index.getRealm(`user=${users.getAuthedUser().id}`);
-  const currentUser = users.getAuthedUser().id;
-  const user = realm.objectForPrimaryKey("User", currentUser);
-  const projects = user.memberOf;
-  return projects;
+  const user = users.getAuthedUser();
+  try {
```
cbush, line 3: Not sure if this really needs a try/catch block. What about returning an empty array in the 'warning' case?

cbush (approved):

source/tutorial/generated/node/projects.codeblock.getProjects.js
```diff
@@ -1,7 +1,15 @@
...
+    const { memberOf: projects } = await user.refreshCustomData();
+
+    // Make sure that the user object has been created
+    if (!projects) {
+      output.error("The user object hasn't been created yet. Try again soon.");
+      throw new Error("No projects for user");
+    }
+    return projects;
+  } catch (err) {
+    output.error("There was a problem accessing custom user data");
```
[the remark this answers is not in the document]
mongodben: sure, only helps to add. adding another `output.error(err)` before the one that's already there

source/tutorial/generated/node/projects.codeblock.getProjects.js
mongodben: returning an empty array would just result in an error downstream in the function `selectProject`, so i'd rather catch the error here. not the most robust error handling, but aligns with what's already being done in the project, where errors are presented and the program must be exited. think ok for now, though certainly something that could be improved upon in future iterations.

mongodben: also including the try/catch catches error returned by the `refreshCustomData()` function in the event that there is no custom user data yet and it throws an error, which tom on the JS team believes to be the outcome if it's called before custom user data is created.

cbush: By catching you're just returning undefined. And that has to be handled by the caller. Is that being done?

mongodben: i see what you're saying. as understand the code in this project, this issue of unhandled errors persists persists across the whole project. 

to remedy, you think it'd make sense to have a error handler function that can be used in catch statements across the project?

cbush: Why not just let the exception go unhandled and bubble up to the top of the program, thereby killing it?

mongodben: that would certainly work too, and be the easiest way of going about this. created this ticket to address the issue - https://jira.mongodb.org/browse/DOCSP-19916

PR #1504, opened 2021-11-30
(DOCS-14138): Add SDK-agnostic Realm Query Language reference

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCS-14138

- [Realm Query Language](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCS-14138/reference/realm-query-language/): New page, based on iOS NSPredicate page, updated to remove SDK-specific references/examples, provide JSON (BSON) schema, and add SORT/DISTINCT/LIMIT

### Review Guidelines

### DIFF
--- a/source/reference/realm-query-language.txt
+++ b/source/reference/realm-query-language.txt
@@ -16,12 +16,15 @@ Overview
 --------
 
 {+client-database+} supports a string-based query language to constrain 
-searches when retreiving objects. Queries evaluate a predicate for every 
+searches when retrieving objects. Queries evaluate a predicate for every 
 object in the collection being queried. If the predicate resolves 
 to ``true``, {+client-database+} includes the object in the results 
 collection.
 
-You can use Realm Query Language in every SDK with your SDK's filter method.
+You can use Realm Query Language in most Realm SDKs with your SDK's filter 
+or query methods. The Swift SDK is the exception, as it uses the NSPredicate 
+query API.
+
 You can also use Realm Query Language to browse for data in :ref:`Realm Studio 
 <realm-studio>`.
 
@@ -32,7 +35,9 @@ Filters consist of **expressions** in a predicate. An expression consists of
 one of the following:
 
 - The name of a property of the object currently being evaluated.
-- An operator and up to two argument expression(s).
+- An operator and up to two argument expression(s). For example, in the 
+  expression ``A + B``, the entirety of ``A + B`` is an expression, but ``A``
+  and ``B`` are also argument expressions to the operator ``+``. 
 - A value, such as a string (``'hello'``) or a number (``5``).
 
 .. literalinclude:: /examples/generated/realm-query-language/realm-query-language.codeblock.predicate.js
@@ -70,6 +75,12 @@ You can iterate through a collection property with another query using the
   variable name specified by ``variableName`` to refer to the currently-iterated
   element.
 
+A subquery takes a collection, then iterates through each object (as 
+variableName) checking the predicate against that object. Subqueries work well 
+when you have a collection (A) objects, and each object has a collection (B) 
+other objects. Use a subquery if you're trying to filter A based on 2 or more 
+varying attributes of B.
+
 .. example::
 
    Running the following filter on a ``projects`` collection returns projects
@@ -77,6 +88,9 @@ You can iterate through a collection property with another query using the
 
    .. literalinclude:: /examples/generated/realm-query-language/realm-query-language.codeblock.subquery.js
       :language: javascript
+    
+   SUBQUERY(...) returns an array. We need to check if its count > 0 to 
+   return the true or false value predicate expects.
 
 .. note:: About the Examples On This Page
 
@@ -88,42 +102,21 @@ You can iterate through a collection property with another query using the
 
    See the schema for these two classes, ``Project`` and ``Task``, below:
 
-   .. code-block:: json
+   .. code-block:: 
 
-      {
-        "title": "task",
-        "required": [
-          "_id",
-          "name",
-          "isComplete",
-          "priority",
-          "progressMinutes"
-        ],
-        "properties": {
-          "_id": { "bsonType": "objectId" },
-          "name": { "bsonType": "string" },
-          "isComplete": { "bsonType": "bool" },
-          "assignee": { "bsonType": "string" },
-          "priority": { "bsonType": "int" },
-          "progressMinutes": { "bsonType": "int" }
-        }
-      },
-      {
-        "title": "project",
-        "required": [
-          "_id",
-          "name",
-          "tasks"
-        ],
-        "properties": {
-          "_id": { "bsonType": "objectId" },
-          "name": { "bsonType": "string" },
-          "tasks": { "bsonType": "array",
-            "tasks": {
-              "bsonType": "objectId"
-            }
-          }
-        }
+      class Task {
+          ObjectId id;
+          string name;
+          bool isComplete;
+          string? assignee;
+          int priority;
+          int progressMinutes;
+      }
+
+      class Project {
+          ObjectId id;
+          string name;
+          Task[] tasks;
       }
 
 Operators
@@ -215,15 +208,15 @@ You can make compound predicates using logical operators.
    * - Operator
      - Description
 
-   * - | ``and``
+   * - | ``AND``
        | ``&&``
      - Evaluates to ``true`` if both left-hand and right-hand expressions are ``true``.
 
-   * - | ``not``
+   * - | ``NOT``
        | ``!``
      - Negates the result of the given expression.
 
-   * - | ``or``
+   * - | ``OR``
        | ``||``
      - Evaluates to ``true`` if either expression returns ``true``.
 
@@ -398,10 +391,18 @@ results collection.
    * - ``SORT``
      - Specify the name of the property to compare. You can optionally 
        specify ascending (``ASC``) or descending (``DESC``) order.
+       If you specify multiple SORT fields, the query sorts by the first
+       field, and then the second. For example, if you ``SORT (priority, name)``, 
+       the query returns sorted by priority, and then by name when priority
+       value is the same.
 
    * - ``DISTINCT``
      - Specify a name of the property to compare. Remove duplicates 
-       for that property in the results collection.
+       for that property in the results collection. If you specify multiple 
+       DISTINCT fields, the query removes duplicates by the first field, and 
+       then the second. For example, if you ``DISTINCT (name, assignee)``, 
+       the query only removes duplicates where the values of both properties
+       are the same.
 
    * - ``LIMIT``
      - Limit the results collection to the specified number.
@@ -411,7 +412,7 @@ results collection.
    We use the query engine's sort, distinct, and limit operators to find: 
    
    - Tasks where the assignee is Ali
-   
+
      - Sorted by priority in descending order
      - Enforcing uniqueness by name
      - Limiting the results to 5 tasks
--- a/source/reference/realm-query-language.txt
+++ b/source/reference/realm-query-language.txt
@@ -80,26 +80,35 @@ predicate against each object in the collection. The predicate may refer
 to the current iterated object with the variable name passed to
 ``SUBQUERY()``.
 
-{+client-database+} only supports the ``@count`` :ref:`aggregate
-operator <rql-aggregate-operator>` on the result of a subquery. This
-allows you to count how many objects in the subquery input collection
-matched the predicate.
+A subquery expression resolves to an array. {+client-database+} only
+supports the ``@count`` :ref:`aggregate operator
+<rql-aggregate-operator>` on this result. This allows you to count how
+many objects in the subquery input collection matched the predicate.
+
+You can use the count of the subquery result as you would any other
+number in a valid expression. In particular, you can compare the count
+with a number literal (such as ``0``) or another property (such as
+``quota``).
 
 .. example::
 
-   Running the following filter on a ``projects`` collection returns projects
-   with tasks that have not been completed by a user named Alex.
+   The following example shows two filters on a ``projects`` collection.
+   
+   - The first returns projects with tasks that have not been completed by a user named Alex.
+   - The second returns the projects where the number of completed tasks is greater than or equal to the project's quota value.
 
    .. literalinclude:: /examples/generated/realm-query-language/realm-query-language.codeblock.subquery.js
       :language: javascript
 
 .. note:: About the Examples On This Page
 
-   The examples in this page use a simple data set for a task list app. The 
-   two Realm object types are ``Project`` and ``Task``. A ``Task`` has a 
-   name, assignee's name, and completed flag. There is also an arbitrary 
-   number for priority -- higher is more important -- and a count of minutes 
-   spent working on it. A ``Project`` has zero or more ``Tasks``.
+   The examples in this page use a simple data set for a task list app.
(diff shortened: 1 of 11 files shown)


source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,435 @@
...
+          }
+        }
+      }
+
+Operators
+~~~~~~~~~
+
+There are several types of operators available to filter a
+Realm collection. Filters work by **evaluating** an operator expression 
+for every :term:`object <Realm object>` in the collection being
```
cbush, line 136: The section is about operators but this paragraph really explains filters, which were already explained above. Can we clarify?

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,435 @@
...
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Operator
+     - Description
+
+   * - ``SORT``
+     - Specify the key path of the property to compare. You can optionally 
+       specify ascending (``ASC``) or descending (``DSC``) order.
```
cbush, line 424: Is it definitely "DSC" and not "DESC"?

cbush (approved):

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,435 @@
...
+   :local:
+   :backlinks: none
+   :depth: 3
+   :class: singlecol
+
+Overview
+--------
+
+{+service+} supports a string-based query language to constrain searches 
+for a fetch or for in-memory filtering. Queries evaluate an operator
```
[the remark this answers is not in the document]
dacharyc: I don't actually _know_ - that came from the Node.js page, I think. What about:

> constrain searches when retrieving objects.

tgoyne:
The aggregates section is missing an example of aggregates on a primitive collection (i.e. `values.@sum > 10` on a `List<Int>` property). Otherwise looks good to me.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+   :local:
+   :backlinks: none
+   :depth: 3
+   :class: singlecol
+
+Overview
+--------
+
+{+client-database+} supports a string-based query language to constrain 
+searches when retreiving objects. Queries evaluate a predicate for every
```
nirinchev, line 19:
```suggestion
searches when retrieving objects. Queries evaluate a predicate for every 
```

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+Overview
+--------
+
+{+client-database+} supports a string-based query language to constrain 
+searches when retreiving objects. Queries evaluate a predicate for every 
+object in the collection being queried. If the predicate resolves 
+to ``true``, {+client-database+} includes the object in the results 
+collection.
+
+You can use Realm Query Language in every SDK with your SDK's filter method.
```
nirinchev, line 24: I don't think the iOS SDK exposes this - they only expose the NSPredicate query API, which while similar, are different from RQL.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+<realm-studio>`.
+
+Expressions
+~~~~~~~~~~~
+
+Filters consist of **expressions** in a predicate. An expression consists of
+one of the following:
+
+- The name of a property of the object currently being evaluated.
+- An operator and up to two argument expression(s).
```
nirinchev, line 35: I'm not sure I understand this - what are the argument expressions?

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+property of that address: 
+
+.. code-block:: javascript
+
+   workplace.address.zipcode == 10012
+
+Subqueries
+~~~~~~~~~~
+
+You can iterate through a collection property with another query using the
```
nirinchev, line 61: Is there an expectation that the subquery is followed by an aggregation operator, such as `@count`, `@avg`, and so on to be used in the query? I think we should provide more examples as this is a fairly complex concept and we should be explicit about what's possible and what isn't.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+
+   The examples in this page use a simple data set for a task list app. The 
+   two Realm object types are ``Project`` and ``Task``. A ``Task`` has a 
+   name, assignee's name, and completed flag. There is also an arbitrary 
+   number for priority -- higher is more important -- and a count of minutes 
+   spent working on it. A ``Project`` has zero or more ``Tasks``.
+
+   See the schema for these two classes, ``Project`` and ``Task``, below:
+
+   .. code-block:: json
```
nirinchev, line 93: I think using json schema to explain the data structure is going to be foreign for the majority of Realm users. We can instead use a C-based pseudocode language to convey the types without the complexities and verbosity of json schema. Like:

```
class Task {
    ObjectId id;
    string name;
    bool isComplete;
    string? assignee;
    int priority;
    int progressMinutes;
}

class Project {
    ObjectId id;
    string name;
    Task[] tasks;
}
```

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+
+   .. code-block:: javascript
+
+      "priority > 5"
+
+      "progressMinutes > 120"
+
+      "assignee == nil"
+
+      "assignee IN {'Ali', 'Jamie'}"
```
nirinchev, line 213: Is this really supported?

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+   :widths: 40 60
+
+   * - Operator
+     - Description
+
+   * - | ``beginsWith``
+     - Evaluates to ``true`` if the left-hand string expression begins with the right-hand string expression. This is similar to ``contains``, but only matches if the right-hand string expression is found at the beginning of the left-hand string expression.
+
+   * - | ``contains``, ``in``
+     - Evaluates to ``true`` if the left-hand string expression is found anywhere in the right-hand string expression.
```
nirinchev, line 280: I don't think this is correct. `CONTAINS` will return `true` if the left hand expression contains the right-hand one, while I would expect `IN` to be the opposite.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+      "tasks.@min.priority > 5"
+
+      "tasks.@count > 5"
+
+      "tasks.@sum.progressMinutes > 100"
+
+Set Operators
+`````````````
+
+A **set operator** uses specific rules to determine whether
```
nirinchev, line 377: Is "set" a common terminology for these operators? Initially I thought those only applied to collections that are of type Set, but they apply to any collection - i.e. set, list, backlinks. Additionally, this should explain how these operators interact with dictionaries which have both a key and a value.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+Filters consist of **expressions** in a predicate. An expression consists of
+one of the following:
+
+- The name of a property of the object currently being evaluated.
+- An operator and up to two argument expression(s).
+- A value, such as a string (``'hello'``) or a number (``5``).
+
+.. code-block:: javascript
+
+   "progressMinutes > 1 AND name == %@", "Ali"
```
nirinchev, line 40: I don't think %@ is valid value substitution here.

nirinchev (changes requested):
Looks good so far. Some notable things that are missing are:

1. Information about how queries on dictionary properties behave.
2. Information about how data is serialized in a query. Obviously you can use value substitution and pass $0, $1, and so on, but sometimes people want to go the hard route and stringify the date/objectId/uuid themselves.
3. Information about backlink queries.
4. Information about comparing two properties. E.g. `SUBQUERY(tasks, $t, $t.isCompleted).@count > minQuota` where `minQuota` and `tasks` are properties on the same object.

It seems like you may have tested this using the iOS SDK which is precisely the SDK that doesn't use RQL, so it might be a good idea to re-run your tests with the JS SDK or even better - create actual unit tests like the ones the docs team maintains for the SDK-specific docs. That way it will be easier to add and test new queries.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+collection.
+
+You can use Realm Query Language in every SDK with your SDK's filter method.
+You can also use Realm Query Language to browse for data in :ref:`Realm Studio 
+<realm-studio>`.
+
+Expressions
+~~~~~~~~~~~
+
+Filters consist of **expressions** in a predicate. An expression consists of
```
cmelchior, line 31: This sentence makes sense to me, but the following details, don't sound right, how can just "42" be a valid expression  or "myFieldName"?

source/reference/realm-query-language.txt
cmelchior: Yes, that looks like the iOS syntax. `$1` is used in RQL.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
+.. _realm-query-language:
+
+====================
+Realm Query Language
```
cmelchior, line 4: I might be missing it, but I don't see any examples using `@LinkingObjects`?

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+   example, comparing an ObjectId with string will result in a precondition 
+   failure with a message like: 
+   
+   .. code-block::
+      :copyable: false
+   
+      "Expected object of type object id for property 'id' on object of type 
+      'User', but received: 11223344556677889900aabb (Invalid value)"
+   
+   You can compare any numeric type with any other numeric type.
```
cmelchior, line 162: Maybe specifically spell out if this is true for decimal, float and Decimal128 as well?

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+You can make compound predicates using logical operators.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 70
+
+   * - Operator
+     - Description
+
+   * - | ``and``
```
cmelchior, line 229: Also add the upper case `AND` here and elsewhere since that is what we are using in all our examples?

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Operator
+     - Description
+
+   * - | ``@avg``
+     - Evaluates to the average value of a given numerical property across a collection.
```
cmelchior, line 333: We need to describe semantics if there is `null` values in here. I do not recall if it ignore `null` values in the count or include them.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+results collection.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Operator
+     - Description
+
+   * - ``SORT``
```
cmelchior, line 426: We can also add multiple SORT fields, which will be used in case of same values, e.g `SORT (name, age)` will first sort by name and for equal names use age.

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,439 @@
...
+     - Specify the name of the property to compare. You can optionally 
+       specify ascending (``ASC``) or descending (``DESC``) order.
+
+   * - ``DISTINCT``
```
cmelchior, line 430: Same for DISTINCT, it allow multiple values.

source/reference/realm-query-language.txt
dacharyc: Nope, I think you're right, @nirinchev - when I converted these to JS-tested code examples, I realized this does not work in JS, so I've removed it from the examples.

source/reference/realm-query-language.txt
dacharyc: Added some language to clarify - let me know if this resolves the confusion, @nirinchev :

> An operator and up to two argument expression(s). For example, in the 
  expression ``A + B``, the entirety of ``A + B`` is an expression, but ``A``
  and ``B`` are also argument expressions to the operator ``+``.

source/reference/realm-query-language.txt
dacharyc: Added some more details to clarify (these come from academy.realm.io - can't take credit for it!)

> A subquery takes a collection, then iterates through each object (as 
variableName) checking the predicate against that object. Subqueries work well 
when you have a collection (A) objects, and each object has a collection (B) 
other objects. Use a subquery if you're trying to filter A based on 2 or more 
varying attributes of B.

And then in the example itself:
> SUBQUERY(...) returns an array. We need to check if its count > 0 to return the true or false value predicate expects.

source/reference/realm-query-language.txt
cbush: I was thinking of "set" in the mathematical sense, set_intersection etc.

source/reference/realm-query-language.txt
cbush: Yes I think this is backwards.

source/reference/realm-query-language.txt
cbush: Not to throw a new term out there but we could also say "and two operands"

source/reference/realm-query-language.txt
cbush: `42` or `"myFieldName"` may not be a valid _predicate_ on their own but they are still expressions in the general sense (expression of a value)

source/reference/realm-query-language.txt
cbush: We could say "collection operators" instead.

source/reference/realm-query-language.txt
nirinchev: Looks better, but I think we should be explicit about the operators one can use in combination with a subquery. I guess that's all the aggregation operators as those will convert the collection to a single value that can then be compared to something else to return a boolean. Still, since I'm kind of uncertain about it, my guess is that users will be even more so.

source/reference/realm-query-language.txt
cbush: We have many examples using collections and nested object. I'm open to it if so, but do we need an explicit `@LinkingObjects` example?

source/reference/realm-query-language.txt
```diff
@@ -0,0 +1,444 @@
...
+  element.
+
+A subquery iterates through the given collection and checks the given
+predicate against each object in the collection. The predicate may refer
+to the current iterated object with the variable name passed to
+``SUBQUERY()``.
+
+A subquery expression resolves to an array. {+client-database+} only
+supports the ``@count`` :ref:`aggregate operator
+<rql-aggregate-operator>` on this result. This allows you to count how
```
dacharyc, line 85: @cbush Looks like the snooty build log CI is failing because this ref is missing an `s` at the end - the anchor is `rql-aggregate-operators` in line 310.
```suggestion
<rql-aggregate-operators>` on this result. This allows you to count how
```

cbush:
@nirinchev, could you clarify point 3, "information on backlink properties"? Added your point 4. I don't think we want to cover your point 2 unless people frequently ask for it. Looking into point 1.

cbush:
@nirinchev never mind, I just discovered `@links` https://github.com/realm/realm-core/blob/master/test/test_parser.cpp#L256

cbush:
Dictionary-specific operators will be done in this ticket: https://jira.mongodb.org/browse/DOCSP-20137

PR #1511, opened 2021-12-01
(DOCSP-19467): Refactor NodeJS data type tests

mongodben:
## Pull Request Info

Refactored NodeJS data type unit tests. High level summary of changes:
- Made tests more easily idempotent by deleting realm files after each test
- defined realm variable w/in each test to avoid potentially passing wrong reference around and not require as much bluehawkification
- misc bluehawk cleanup and removing unused code

### Jira

- https://jira.mongodb.org/browse/DOCSP-19467

N/A

### Review Guidelines

### DIFF
--- a/examples/node/Examples/data-types.js
+++ b/examples/node/Examples/data-types.js
@@ -152,14 +152,12 @@ describe("Node.js Data Types", () => {
     // :code-block-start: remove-fields-of-the-dictionary
     realm.write(() => {
       // remove the 'windows' and 'doors' field of the Summerhill House.
-      // :uncomment-start:
-      // summerHillHouse.remove(["windows", "doors"]);
-      // :uncomment-end:
+      summerHillHouse.remove(["windows", "doors"]);
     });
     // :code-block-end:
 
-    // expect(summerHillHouse.windows).toBe(undefined); // since windows has been removed as a field, it should be undefined
-    // expect(summerHillHouse.doors).toBe(undefined); // since doors has been removed as a field, it should be undefined
+    expect(summerHillHouse.windows).toBe(undefined); // since windows has been removed as a field, it should be undefined
+    expect(summerHillHouse.doors).toBe(undefined); // since doors has been removed as a field, it should be undefined
   });
   test("should work with Mixed Type", async () => {
     // :code-block-start: define-mixed-in-schema
@@ -202,8 +200,9 @@ describe("Node.js Data Types", () => {
     // :code-block-start: query-objects-with-mixed-values
     // To query for Blaise's birthDate, filter for his name to retrieve the realm object.
     // Use dot notation to access the birthDate property.
-    let blaiseBirthDate = realm.objects("Dog").filtered(`name = 'Blaise'`)[0]
-      .birthDate;
+    let blaiseBirthDate = realm
+      .objects("Dog")
+      .filtered(`name = 'Blaise'`)[0].birthDate;
     console.log(`Blaise's birth date is ${blaiseBirthDate}`);
     // :code-block-end:
     expect(blaiseBirthDate).toEqual(new Date("August 17, 2020"));
@@ -321,13 +320,7 @@ describe("Node.js Data Types", () => {
       },
     };
 
-    // :hide-start:
-    realm = await Realm.open({
-      // realm = await Realm.open({
-      // :hide-end:
-      // :uncomment-start:
-      // const realm = await Realm.open({
-      // :uncomment-end:
+    const realm = await Realm.open({
       schema: [ProfileSchema],
     });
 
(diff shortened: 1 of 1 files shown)


examples/node/Examples/data-types.js
```diff
@@ -53,12 +53,19 @@ const PetOwnerSchema = {
...
-    realm?.write(() => {
-      realm.deleteAll();
+    const toRemove = [
+      "default.realm",
+      "default.realm.lock",
+      "default.realm.note",
+      "default.realm.management",
+    ];
+    toRemove.forEach((fileOrDirectory) => {
+      fs.rmSync(fileOrDirectory, { force: true, recursive: true });
```
mongomoe, line 65: A few thoughts here:
1. Why use FS here instead of Realm JS's ``deleteFile()`` method? (https://docs.mongodb.com/realm-sdks/js/latest/Realm.html#.deleteFile)
2. Also, deleting the ``.realm`` files after each test may lead to issues in other unit tests files, because Jest tests are run in parallel. 
3. Were these tests initially failing on consecutive runs & that is why you would want to make them more idempotent? If so, some alternatives to deleting the .realm & realm auxilery files that don't impact other unit test files are:

         a. Specify a specific realm path for this file that wont be used in other files (for example "data-types-path" which would open "data-types-path.realm")
         b. Deleting the specific objects that you create after each test
         c. Changing object names to a unique name (for instance, Person => Person2 or PersonFromDataTypesTest) // Person2 is a little nicer looking, whereas PersonFromDataTYpesTest is more unique sounding), and then deleting all of that object after the test

examples/node/Examples/data-types.js
```diff
@@ -1,7 +1,7 @@
+import fs from "fs";
 import Realm from "realm";
 import BSON from "bson";
 
-let realm;
```
mongomoe, line 4: Although this is a personal preference, having a top-level realm and closing it at the end may be more preferable IMO. This is because if someone forgets to close a realm within a test block and a new realm is opened with the same path (or the default path), then they may run into errors (I think the error they would get is something like 'realm already opened on current thread')

examples/node/Examples/data-types.js
mongodben: 1. use FS b.c the `afterEach` function doesn't have access to the specific realms' configs, which is required for the `deleteFile()` method per the API. 
2. interesting, didn't know jest runs tests in parallel. i think your suggestion of specifying a unique path and then bluehawk removing it would resolve this
3. tests weren't failing IIRC. i believe the idea was to make them idempotent w/o having to do manual cleanup

examples/node/Examples/data-types.js
mongodben: i wonder if the deleting the realm files would render the risk of someone forgetting to close realm irrelevant.

also wonder if there's a way to inject the line `realm?.close()` at the end of each test to close realm using the locally scope variable.

will do a bit of investigation on these

mongodben: looks like no good way to inject code at end of test in jest :/

mongomoe: yeah afterEach would be the closest thing to force realm.close() but realm would have to be defined either globally or in the afterEach block. It also feels like a pretty roundabout way to do it.

examples/node/Examples/data-types.js
mongomoe: 1. Makes sense, thanks for answering that
2. Would definitely prefer the unique path approach, of course this could lead to the potential issue where a developer adds a new test on this file that doesn't specify a unique path, and then it is not idempotent.
3. I see, I don't remember them failing either. I also don't have any issue with manually cleaning up by deleting objects. In the realm-js repo we have examples of this as well: https://github.com/realm/realm-js/blob/master/tests/js/set-tests.js#L100

Correct me if I'm wrong: from the original PR (https://github.com/mongodb/docs-realm/pull/1444/files), it seems the issue Chris is having (that lead to the refactoring ticket?) is more with the use of "uncomment-start" (see: https://github.com/mongodb/docs-realm/pull/1444#pullrequestreview-797092930). I don't see how this solves that issue, since we don't delete objects manually within a "uncomment" block. The main thing we were previously doing w/ the uncomment blocks was making "realm = await Realm.open({" look like "const realm = await Realm.open({".

mongomoe: Overall, I think maybe we keep the manual cleanup's & don't delete the file at all (IMO this also sets a precedent of having a .realm file for each unit test file, which may get unwieldy so its definitely a tradeoff).

mongomoe: If you want to feel free to have the realm declared & defined in each test method (the [thread](https://github.com/mongodb/docs-realm/pull/1511/files#r762100623) above is more of a personal preference so this should be fine + it would get rid of a bunch of the "uncomment-start"s)

mongodben:
@mongomoe i redid this refactor to simplify things. 

there's a bit more manual clean up, but i think the net result is positive, removing uncomments and keeping things simple & straightforward.

mongomoe (approved):
LGTM but I'd suggest to switch from deleteAll to deleting individual objects, or switching from the default.realm to a specific realm for this file and then you can deleteAll (as discussed via slack)

PR #1515, opened 2021-12-03
(DOCSP-19891): Create Query-Based Sync content in Realm Sync -> Define Data Access Patterns

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19891

- [Choose Your Sync Mode](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19891/sync/data-access-patterns/sync-mode/): New page consisting of entirely novel content. Examples are blank - reviewer, I'd love a good idea for examples that illustrate a good use case for each.
- [Flexible Sync](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19891/sync/data-access-patterns/query-based-sync/): New page consisting of entirely novel content. The links to the SDK pages do not currently work, but will be added in later Jira tickets.
- [Sync Rules & Permissions](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19891/sync/data-access-patterns/permissions/#query-based-sync-rules-and-permissions): Existing page, divided into two sections for Partition-Based Sync and Query-Based Sync. Content in Flexible Sync is entirely novel. There's an example here under `Roles` that is blank, and I'd love a suggestion.
- [Partition-Based Sync](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19891/sync/data-access-patterns/partitions/): New page combining "[Partitions](https://docs.mongodb.com/realm/sync/data-access-patterns/partitions/)" and "[Partition Strategies](https://docs.mongodb.com/realm/sync/data-access-patterns/partition-strategies/)" pages. No new content here so it doesn't need a thorough review, but I'd appreciate a check to make sure it looks reasonable.

### Review Guidelines


source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,84 @@
...
+
+When you enable {+sync+}, you must specify a {+sync-short+} mode for your 
+{+app+}. The {+sync-short+} mode determines how users can access data 
+from the {+service+} SDK in the client app. {+service-short+} offers two 
+{+sync-short+} modes:
+
+- Partition-Based Sync
+- Query-Based Sync (beta)
+
+You configure and use each type of {+sync-short+} mode with different data.
```
mongomoe, line 26: "You configure and use each type of {+sync-short+} mode with different data." sounds to me like it implies that different data sets require different sync modes. But I think you meant, only one type of sync mode can be applied per application. Is that correct? If so can you improve this sentence so it reflects that

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,84 @@
...
+You can't mix Partition-Based Sync and Query-Based Sync within the same 
+{+app+}, so you must decide which type of {+sync+} to use in a given
```
mongomoe, line 28:
```suggestion
{+app+}, so you must decide which mode of {+sync+} to use in a given 
```

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,84 @@
...
+You can't mix Partition-Based Sync and Query-Based Sync within the same 
+{+app+}, so you must decide which type of {+sync+} to use in a given 
+application.
+
+Partition-Based Sync
+~~~~~~~~~~~~~~~~~~~~
+
+:ref:`Partition-Based Sync <partition-based-sync>` works best for 
+applications where a subset of the documents in a synced cluster are 
+related in some way and have the same read/write permissions for a given
```
mongomoe, line 36: "... are related in some way" begs the question: "what way?"

This question doesn't get answered until the Partitions Page, in which we go into various partition strategies. Suggestion: Can we alter this sentence to be a bit more descriptive? Perhaps replace "in some way" with examples of ways, such as by user, team, region, etc?

source/sync/data-access-patterns/query-based-sync.txt
```diff
@@ -0,0 +1,217 @@
...
+
+When a client makes a query, {+service-short+} searches the server-side 
+data set for documents matching the query. Then, the sync engine applies 
+a set of rules to determine which documents in the query the user should 
+be able to access. {+service-short+} returns only the matching documents 
+where the user has read or read/write access, and assigns the user
+the appropriate permissions.
+
+Client applications can add, update, or remove queries dynamically at 
+run-time from the device.
```
mongomoe, line 37: Should you add a warning or an additional sentence that states that removing queries in this way will trigger a client reset?

mongomoe:
Can you add a redirect for partition strategies + any other moved/deleted content

source/sync/data-access-patterns/query-based-sync.txt
```diff
@@ -0,0 +1,217 @@
...
+
+{+service+} Query-Based Sync lets you define a query in the client, and 
+sync only the objects that match the query. With client-side subscriptions, 
+client applications can:
+
+- Maintain queries
+- React to changes
+- Add, change, or delete queries
+
+With :ref:`rules and permissions <sync-rules>`, this provides a powerful
```
mongomoe, line 26: what is "this" in the sentence "provides a powerful and flexible way to configure ..."?

Currently it ambiguously means one of the following:
* rules and permissions (in general) 
* client-side subscriptions 
* query-base sync

Can you clarify + update the sentence?

source/sync/data-access-patterns/query-based-sync.txt
```diff
@@ -0,0 +1,217 @@
...
+run-time from the device.
+
+.. include:: /includes/note-qbs-beta.rst
+
+.. _queryable-fields:
+
+Queryable Fields
+----------------
+
+When you configure Query-Based Sync, you specify which fields your client
```
mongomoe, line 46: [](url)
```suggestion
When you configure Query-Based Sync, you specify a set of fields your client 
```
the above suggestion might make it more clear that you can include multiple fields

source/sync/data-access-patterns/query-based-sync.txt
```diff
@@ -0,0 +1,217 @@
...
+.. _qbs-eligible-field-types:
+
+Eligible Field Types
+~~~~~~~~~~~~~~~~~~~~
+
+Any top-level primitive field with a scalar type can be a queryable field. 
+Query-Based Sync does not currently support embedded objects or arrays as 
+queryable fields.
+
+Performance and Storage
```
mongomoe, line 69: This section is awesome haha might steal a lot of the wording from this section for the "Optimize Sync Atlas Usage"

source/sync/data-access-patterns/query-based-sync.txt
```diff
@@ -0,0 +1,217 @@
...
+:ref:`rules and permissions <query-based-sync-rules-and-permissions>` for 
+more granular control over who can access which collections.
+
+.. example::
+
+   Your {+app+} may contain 20 or 30 collections, but you can only configure
+   10 queryable fields. You must re-use queryable fields across collections 
+   in order to sync objects from every collection.
+
+For best performance, use a broad Sync query, and slice off subqueries
```
mongomoe, line 88: Question: If you use a broad query how would you slice off subqueries? Did you mean, use a specific query and then slice off subqueries in order to make it a broader query to optimize for storage space? Or did I misunderstand?

mongomoe: I think my confusion may be around the term "working set". Can you define that for me in this thread? I see that in a couple spots of this documentation, but haven't seen it elsewhere in our docs. Guessing you don't mean: [Wikipedia:working set](https://en.wikipedia.org/wiki/Working_set#:~:text=Working%20set%20is%20a%20concept,in%20a%20given%20time%20interval.)?

source/sync/data-access-patterns/permissions.txt
```diff
@@ -285,6 +289,279 @@ cannot express solely in JSON expressions.
...
+session.
+
+Because roles apply at the collection level, a user could have different roles
+for different collections.
+
+.. example::
+
+   
+
+Roles "fall through" until the user's metadata evaluates to ``true`` for
```
mongomoe, line 349: Nitpick: Can we alter the term "fall through"? It's not a technical term we can link through and it's also vague.

Perhaps "Roles resolve to a lower level of permissions until the user's metadata evaluates to ``true`` ...".

source/sync/data-access-patterns/permissions.txt
```diff
@@ -285,6 +289,279 @@ cannot express solely in JSON expressions.
...
+Roles "fall through" until the user's metadata evaluates to ``true`` for 
+a given role. At that point, the user receives the corresponding permissions
+for the duration of their session.
+
+Default Roles
+`````````````
+
+You can create one or more default roles that apply across all collections. 
+If a collection does not have any custom roles defined, role resolution 
+falls through to default roles.
```
mongomoe, line 358:
```suggestion
defaults to default roles. 
```
similar to https://github.com/mongodb/docs-realm/pull/1515/files#r768725918

mongomoe (commented):
The content itself is really good on a technical level & some of the best explanations I've seen of QBS + how it relates to permissions that I've seen. 

I had a bunch of comments + questions, that are frequently about how the use of certain wording made the subject confusing. Let me know if you want to zoom about any, I realize it was a fair amount of comments.

source/sync/data-access-patterns/query-based-sync.txt
dacharyc: That's a good question, @mongomoe  - I've waffled a bit with whether to include warnings like that here, or whether they belong in the configuration documentation. Since this is intended to be high-level concept information, I didn't include that in my initial draft here. But if you think it's an important omission that we should cover in both places, I can add it. What do you think?

source/sync/data-access-patterns/query-based-sync.txt
dacharyc: That's a fair point - do you think we should remove this content from here and repurpose it for the "Optimize Sync Atlas Usage" page, and/or configuration details? I did wonder if this was too configuration-specific to belong in concept content, or if it's helpful to get people thinking about these details while they're learning more about QBS.

source/sync/data-access-patterns/query-based-sync.txt
dacharyc: I don't actually know, @mongomoe - I had some questions about this terminology, too. I took it from a conversation in the outline document, and had previously asked for more clarification there, but hadn't gotten a response yet: https://docs.google.com/document/d/1YR_KRkpXUwmgBbhj4G6_8aMMgAN7ujWmotezXMX3biE/edit?disco=AAAASXEP8I4

I'll bump the comment again and see if I can get some more information.

source/sync/data-access-patterns/sync-mode.txt
dacharyc: What I meant was that different types of data & data access patterns fit each type of sync naturally - i.e. PBS might still be a better fit for syncing data where you don't need to create overlapping result sets and your data structure supports permissions at a partition level. But I get into more detail about that in the following sections, so I don't think this sentence adds anything and it is vague and confusing. I'll just remove it 👍

source/sync/data-access-patterns/query-based-sync.txt
dacharyc: Alright, I got a little more clarification and have revised this paragraph to reflect the additional info. Let me know if this is clear enough, @mongomoe , or if I should rework it further.

source/sync/data-access-patterns/sync-mode.txt
mongomoe: 👍

source/sync/data-access-patterns/query-based-sync.txt
mongomoe: Good point, let's leave it as is, it does makes more sense to leave the mention to client reset out of this sentence since this content is more conceptual.

👍

source/sync/data-access-patterns/query-based-sync.txt
mongomoe: Nice rewording! 👍

source/sync/data-access-patterns/query-based-sync.txt
mongomoe: I think it's fine for two places to have the same info, we can swap it to a literal include if we need

mongomoe: Although I do think the Optimize Sync Atlas Usage has the opportunity to go into more depth + provide additional details, whereas this provides a great high-level passage.

dacharyc: That's fair! Alright, I'll leave this section here. 👍

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+- Maintain queries
+- React to changes
+- Add, change, or delete queries
+
+Together with :ref:`rules and permissions <sync-rules>`, Flexible Sync 
+provides a powerful way to configure sync around your business logic.
+
+When a client makes a query, {+service-short+} searches the server-side 
+data set for documents matching the query. Then, the sync engine applies 
+a set of rules to determine which documents in the query the user should
```
mongomoe, line 31:
```suggestion
a set of rules to determine which documents in the results set the user should 
```

mongomoe: Wording mistake ^, documents are returned by realm through the use of a query, but the query itself does not contain documents

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,162 @@
...
+<flexible-sync-rules-and-permissions>` to determine which documents 
+in the query the user can access. Realm maps the documents that a user can 
+access to a synced ``.realm`` file. Each object in a synced realm has a 
+corresponding document in the {+app+}.
+
+Choose Flexible Sync when:
+
+- You need to create overlapping query and result sets
+- You want to apply permissions at the document level
+- You need a flexible permissioning system to apply more granular permissions
```
mongomoe, line 92:
```suggestion
- You need a flexible permissions system to apply more granular permissions
```
permissioning isn't an actual world

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+{+service+} Flexible Sync lets you define a query in the client, and 
+sync only the objects that match the query. With client-side subscriptions, 
+client applications can:
+
+- Maintain queries
+- React to changes
+- Add, change, or delete queries
+
+Together with :ref:`rules and permissions <sync-rules>`, Flexible Sync 
+provides a powerful way to configure sync around your business logic.
```
mongomoe, line 27:
```suggestion
provides a powerful way to configure Sync around your business logic.
```

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+----------------
+
+When you configure Flexible Sync, you specify one or more fields 
+from your :ref:`Sync Schema <sync-schema-overview>` that your client 
+application can query. 
+
+.. example::
+
+   In a task tracker app, you might set ``assignee`` or ``owner`` 
+   as queryable fields. On the client side, you can then query for tasks
```
mongomoe, line 53:
```suggestion
   as queryable fields. On the client-side, you can then query for tasks 
```

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+
+Performance and Storage
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Queryable fields use storage on the backing Atlas cluster. The more 
+queryable fields you configure, the more storage you use on the backing 
+cluster.
+
+You can configure up to 10 queryable fields per application for Flexible 
+Sync. If you have a large number of collections in an {+app+}, you may need to
```
mongomoe, line 78:
```suggestion
Sync. If you have a large number of collections in a {+app+}, you may need to
```

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+
+   :ref:`enable-sync`
+
+Select Fields to Query
+``````````````````````
+
+When you enable Flexible Sync, you can select queryable fields in two ways:
+
+- If you have defined your schema, you can select queryable fields from 
+  :ref:`eligible fields <flexible-sync-eligible-field-types>` in your schema
```
mongomoe, line 145:
```suggestion
  :ref:`eligible fields <flexible-sync-eligible-field-types>` in your schema. 
```
The second bullet ends with a period so I would match the punctuation.

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+You can configure up to 10 queryable fields per {+app+}. These 
+fields apply across all of the collections in your application.
+
+Define Roles
+````````````
+
+When a user submits a query to the {+app+}, Sync determines which role 
+applies for the user. 
+
+Roles consists of:
```
mongomoe, line 159:
```suggestion
Roles consist of:
```

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,222 @@
...
+
+Create Queries in Your Client Application
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With Flexible Sync enabled, you can start creating queries from 
+your client application. 
+
+The {+service+} SDKs provide methods to create, update, and remove queries 
+from the client application. The SDKs use subscriptions to maintain those 
+queries on the client side. Through these subscriptions, your applications
```
mongomoe, line 184:
```suggestion
queries on the client-side. Through these subscriptions, your applications
```
In a few places, the PR uses 'client side' but others use 'client-side', this should be consistent.

mongomoe:
Suggestion: can we change the "flexible sync is in preview" admonition to an 'important' rather than a 'warning'? To be more in line with it being a preview rather than a beta, this may improve Flexible Sync's perception as something that is a major part of Realm Sync, rather than something that is new, experimental, and potentially risky.

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,162 @@
...
+Choose Partition-Based Sync when:
+
+- You don't need to create overlapping result sets
+- Your data structure supports permissions at a partition level
+- You want to release a production app quickly, while Flexible Sync
+  is still in preview
+
+.. example::
+
+   Partition-Based Sync might be a good fit for a store inventory application.
```
mongomoe, line 54: A couple suggestions/not required for LGTM:

- "a store inventory application" kind of makes me think it's to define the inventory for 1 store, but the store number makes it seem like multiple stores. Can we call this a "franchise inventory application" or a "an application to manage multiple stores" or something along those lines?
- The flexible sync example is a bit more verbose, partially because it slightly touches on how the example is impacted by permissions. Perhaps we can include a bullet about this. For instance, maybe discussing the app could use permissions to grant customers only read access to store 42's inventory, but giving read/write access to store employees.

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,162 @@
...
+                 name: "provider", 
+                 applyWhen: { "%%user.custom_data.isProvider": true},
+                 read: {} // evaluates to the same as %%true
+                 write: {}
+               },
+               {
+                 name: "self",
+                 applyWhen: {},
+                 read: {"patient_id": "%%user.id"}
+                 write: {}
```
mongomoe, line 137: question: is this saying that a patient can write to all data?

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,162 @@
...
+             "roles": [
+               {
+                 name: "provider", 
+                 applyWhen: { "%%user.custom_data.isProvider": true},
+                 read: {} // evaluates to the same as %%true
+                 write: {}
+               },
+               {
+                 name: "self",
+                 applyWhen: {},
```
mongomoe, line 135: shouldn't this be:
 ` "apply_when": { "email": "%%user.data.email" }`
or 
`  "apply_when": { "patient_id": "%%user.id" },`

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,162 @@
...
+                 name: "billingSpecialist", 
+                 applyWhen: { "%%user.custom_data.isBillingSpecialist": true},
+                 read: {} // evaluates to the same as %%true
+                 write: {}
+               },
+               {
+                 name: "self",
+                 applyWhen: {},
+                 read: {"patient_id": "%%user.id"}
+                 write: {}
```
mongomoe, line 153: should
```suggestion
```
shouldn't write be false, so that user's cannot write to their own or anyone else's bills?

source/sync/data-access-patterns/flexible-sync.txt
dacharyc: The PR uses hyphenated `client-side` when it is a compound adjective; i.e. `client-side query`, but does not use the hyphen when it's jus a phrase - i.e. `on the client side.` I did just re-check the instances and find one hyphenated instance that shouldn't be hyphenated, though, so it's a good note to enforce them consistently!

source/sync/data-access-patterns/flexible-sync.txt
dacharyc: This isn't being used as a compound adjective, so this instance of `client side` should not be hyphenated.

source/sync/data-access-patterns/sync-mode.txt
dacharyc: I actually did mean `permissioning` here, which is a system or process for granting authorization. Spell-checkers don't like it, though, so I understand the flag.

source/sync/data-access-patterns/sync-mode.txt
dacharyc: For the first item, I did a quick search and it looks like the keyword for this type of app is "inventory management application," so I've changed it from "store inventory application" to "inventory management application."

I also added a note about customers having read-only access.

I'm not super tied to this example, though - I had a lot of trouble brainstorming examples, so I'm happy to change it if something else better illustrates a use case for Partition-Based Sync.

source/sync/data-access-patterns/sync-mode.txt
dacharyc: Good call, I missed that - updated 👍

source/sync/data-access-patterns/sync-mode.txt
dacharyc: I don't think that's required here. By leaving the `applyWhen` empty, it evaluates to true, which lets this role apply to anyone. The read permission restricts access to documents where the patient ID matches their user ID.

source/sync/data-access-patterns/sync-mode.txt
dacharyc: Yup, I missed the write permissions for both of these `self` roles - I just set them to false. Good catch!

source/sync/data-access-patterns/query-based-sync.txt
cbush: LGTM.

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+{+app+}. The {+sync-short+} mode determines how users can access data 
+from the {+service+} SDK in the client app. {+service-short+} offers two 
+{+sync-short+} modes:
+
+- Partition-Based Sync
+- Flexible Sync (preview)
+
+You can't mix Partition-Based Sync and Flexible Sync within the same 
+{+app+}, so you must decide which {+sync+} mode to use in a given 
+application.
```
tkaye407, line 28: It may be worth highlighting that you cannot switch back and forth between the two without terminating sync?

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+
+Partition-Based Sync
```
tkaye407, line 30: Not sure if this is possible, but these headers kind of seem smaller than they should be IMO. The overview is larger than the Partition-Based Sync and Flexible Sync header

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+   number is Store 42.
+
+   Customers, meanwhile, could have read-only access to store inventory.
+
+Flexible Sync
+~~~~~~~~~~~~~
+
+.. include:: /includes/note-flexible-sync-preview.rst
+
+:ref:`Flexible Sync <flexible-sync>` works best for applications
```
tkaye407, line 74: This seems to inject permissions in here without really explaining it and then moving to the query bit (which I think is the more important bit). Another potential wording of this could be

Flexible Sync works best for applications that need to define a query in the client application, and Sync only the documents that match the query. Realm translates those documents to objects in the client application, making it easy to work with objects in your app. Additionally, it allows permissions to be applied on a per-document basis.

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+.. include:: /includes/note-flexible-sync-preview.rst
+
+:ref:`Flexible Sync <flexible-sync>` works best for applications 
+where you need to define permissions on a per-document basis. With 
+Flexible Sync, you can define a query in the client application, and 
+Sync only the documents that match the query. {+service-short+} translates 
+those documents to objects in the client application, making it easy to
+work with objects in your app.
+
+In Flexible Sync, you define :ref:`queryable fields <queryable-fields>`
```
tkaye407, line 81: you define a set of queryable fields

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+
+:ref:`Flexible Sync <flexible-sync>` works best for applications 
+where you need to define permissions on a per-document basis. With 
+Flexible Sync, you can define a query in the client application, and 
+Sync only the documents that match the query. {+service-short+} translates 
+those documents to objects in the client application, making it easy to
+work with objects in your app.
+
+In Flexible Sync, you define :ref:`queryable fields <queryable-fields>` 
+that you can query against from the client application. You can then
```
tkaye407, line 82: is against the normal word to use with queries?

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+work with objects in your app.
+
+In Flexible Sync, you define :ref:`queryable fields <queryable-fields>` 
+that you can query against from the client application. You can then 
+subscribe to queries against those fields from the client application. 
+When a client starts a Sync session, {+service-short+} evaluates :ref:`rules 
+<flexible-sync-rules-and-permissions>` to determine which documents 
+in the query the user can access. Realm maps the documents that a user can 
+access to a synced ``.realm`` file. Each object in a synced realm has a 
+corresponding document in the {+app+}.
```
tkaye407, line 88: There is a corresponding document in Atlas, not sure what it means that there is a corresponding document in the "Realm app"

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+those documents to objects in the client application, making it easy to
+work with objects in your app.
+
+In Flexible Sync, you define :ref:`queryable fields <queryable-fields>` 
+that you can query against from the client application. You can then 
+subscribe to queries against those fields from the client application. 
+When a client starts a Sync session, {+service-short+} evaluates :ref:`rules 
+<flexible-sync-rules-and-permissions>` to determine which documents 
+in the query the user can access. Realm maps the documents that a user can 
+access to a synced ``.realm`` file. Each object in a synced realm has a
```
tkaye407, line 87: I know the goal of this is to be succinct, but two cool things to potentially add in here would be: 

1. The way that sync works is there is permissions and there is the query/subscription. And the subset of data that physically makes it to the device can be thought of as the the && of those two things. IE permissions is essentually another query / subset of data that the user has access to, so the real data that gets to the device is essentially: `{$and: [ { permissions...}, {subscription/filter} ]}`
2. It may be worth pointing out later the purpose / reason that these need to be defined up front (which is that they each require a cost of storing and processing additional metadata)

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+subscribe to queries against those fields from the client application. 
+When a client starts a Sync session, {+service-short+} evaluates :ref:`rules 
+<flexible-sync-rules-and-permissions>` to determine which documents 
+in the query the user can access. Realm maps the documents that a user can 
+access to a synced ``.realm`` file. Each object in a synced realm has a 
+corresponding document in the {+app+}.
+
+Choose Flexible Sync when:
+
+- You need to create overlapping query and result sets
```
tkaye407, line 92: Its not just "overlapping", but its more expressive. Partition based sync is functionally defining a single field that you can only express an equality query on, flexible sync is letting you define many fields (10) that you can define mostly any query you could in mongodb ($eq, $gt, $lt, $in, etc)

source/sync/data-access-patterns/sync-mode.txt
```diff
@@ -0,0 +1,164 @@
...
+                 name: "self",
+                 applyWhen: {},
+                 read: {"patient_id": "%%user.id"},
+                 write: false
+               }
+             ]
+           }
+         }
+
+   The queryable fields might be ``patient_id`` and ``procedure_date``.
```
tkaye407, line 161: I think this example is great! I think my only concern with this page is that is seems like a majority of the focus is on permissions (which is important) and seems to gloss over the whole "user/sdk can define a query to sync down data". I guess they are functionally the same thing as you know and this example drives home that the client can be as stupid as possible and rely on permissions to do everything for them (which should be the case IMO), but I think the average user also has a hard time with understanding permissions and I suspect people will initially focus on defining arbitrary queries in the SDK (similar to how they do using a local realm), and having it just sync down that subset of data.

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,219 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+{+service+} Flexible Sync lets you define a query in the client, and
```
tkaye407, line 18: I think that this is a better overview in someways than the one in the overview page that focuses more on permissions.

source/sync/data-access-patterns/flexible-sync.txt
```diff
@@ -0,0 +1,219 @@
...
+data set for documents matching the query. Then, the sync engine applies 
+a set of rules to determine which documents in the results set the user 
+should be able to access. {+service-short+} returns only the matching 
+documents where the user has read or read/write access, and assigns the 
+user the appropriate permissions.
+
+Client applications can add, update, or remove queries dynamically at 
+run-time from the device.
+
+.. include:: /includes/note-flexible-sync-preview.rst
```
tkaye407, line 39: weirdly this bit says "beta", but the others say "preview", is there one that we are going with or are they interchangeable?

source/sync/data-access-patterns/sync-mode.txt
tkaye407: It actually seems like both of these are in the Queryable Fields section 👍

source/sync/data-access-patterns/permissions.txt
```diff
@@ -10,23 +10,15 @@ Sync Rules and Permissions
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 2
+   :depth: 3
```
tkaye407, line 13: I think it would be helpful to have a page table of contents or something here so that you can easily skip to the QBS rules section if you want

tkaye407 (commented):
Looks great! Have a few thoughts but looks largely good and easy to follow!

source/sync/data-access-patterns/sync-mode.txt
dacharyc: Updating this to remove the preposition in this instance, and change it to "on" in the next one.

source/sync/data-access-patterns/sync-mode.txt
dacharyc: Hmm. The focus of this doc PR is the backend concept stuff they need to know about choosing Flexible Sync. There's a separate "Fundamentals" page in the SDK (I'm working on a draft for that in #1542 ) that focuses more on the query/SDK side and hand-waves permissions, and then the actual SDK usage examples is all about the queries (PR for one of the SDK pages is #1540 ). We've historically separated the concerns this way between Cloud and SDK docs. You're right about this page only telling part of the story when it comes to understanding the Sync modes, though. I'll have a chat with the team and a think and see what we might do to improve this.

source/sync/data-access-patterns/flexible-sync.txt
dacharyc: That's weird! It was initially beta, but we changed it to preview. Looks like that is still staged from the earlier version of the note. In the merged branch, it's displaying as expected (using "preview"): https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/query-based-sync/sync/data-access-patterns/flexible-sync/

source/sync/data-access-patterns/permissions.txt
dacharyc: Hmm, looks like an old staged version of the page, again - in the version merged to the feature branch, this is at `:depth: 3`: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/query-based-sync/sync/data-access-patterns/permissions/

dacharyc:
Thanks for the review, @tkaye407 ! I've started a new PR to address many of your comments here: https://github.com/mongodb/docs-realm/pull/1546

(Should have done that earlier, apologies!)

There are a couple of broader things I'll discuss with the team (overview, the permissions vs. query content) and will potentially make some bigger changes. I appreciate your time and 👀 !

PR #1538, opened 2022-01-05
(DOCSP-19892): Configure Realm Sync section

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19892

- [Enable Realm Sync](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19892-config-realm-sync-section/sync/configure/enable-sync/#enable-flexible-sync) (The "Enable Flexible Sync" page-section is the only new content on this page and the only content that requires review)
- [Alter Your Sync Configuration](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19892-config-realm-sync-section/sync/configure/alter-sync-configuration/) (The entire page is new content & requires review)
- [Optimize Sync Atlas Usage](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19892-config-realm-sync-section/sync/configure/optimize-sync-atlas-usage/) (The "Optimizing Performance and Storage When Using Flexible Sync" page-section is new content but you may have already seen this content in Dachary's PR: https://github.com/mongodb/docs-realm/pull/1515, in the "Flexible Sync" page under the ["Performance & Storage")](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19891/sync/data-access-patterns/flexible-sync/#performance-and-storage) section, this has been moved here)

### Review Guidelines


source/includes/steps-configure-flexible-sync-ui.yaml
```diff
@@ -0,0 +1,80 @@
...
+  .. note:: {+sync-short+} Rules Cannot Change
+     
+     You define {+sync-short+} rules at the same time as you enable {+sync-short+}. Once {+sync-short+} is
+     enabled, you can't change your app's {+sync-short+} rules unless you terminate {+sync-short+}
+     and re-enable it with new rules.
+---
+title: Select Development Details
+ref: select-development-details
+content: |
+  You can configure various development details under the :guilabel:`Select
```
MongoCaleb, line 16: Suggest re-ordering this sentence to "you can configure various development details, such as ..., under the :guilabel:..."

source/includes/steps-configure-flexible-sync-ui.yaml
```diff
@@ -0,0 +1,80 @@
...
+  Development Details` heading, such as {+sync-short+} types, development mode,
+  and a cluster to {+sync-short+}.
+
+  Below the :guilabel:`{+sync-short+} types` subheading, select the :guilabel:`Flexible`
```
MongoCaleb, line 20: Suggest removing "and send queries from a client using a set of
  queryable fields."

source/includes/steps-configure-flexible-sync-ui.yaml
```diff
@@ -0,0 +1,80 @@
...
+---
+title: Select Queryable Fields
+ref: select-queryable-fields
+content: |
+  When configuring Flexible {+sync-short+}, specify one or more fields from your
+  {+sync-short+} Schema that your client application can query. 
+  
+  To configure queryable fields for your application, navigate to the
+  :guilabel:`Select Queryable Fields` heading. Select up to 10 existing fields
+  from your schema to query on. If you wish to create a new field(s) to
```
MongoCaleb, line 46: Suggest: "If you want to create one or more new fields to..." (I think we avoid "wish")

source/sync/configure/alter-sync-configuration.txt
```diff
@@ -0,0 +1,101 @@
...
+
+.. alter-partition-based-sync-config:
+
+Alter Your Partition-Based Sync Configuration
+---------------------------------------------
+
+You must :ref:`Terminate {+sync-short+} <terminating-realm-sync>` and :ref:`Re-enable {+sync-short+}
+<re-enable-realm-sync>` to make changes to your Partition-Based {+sync-short+}
+Configuration. While you are re-enabling {+sync+}, you can specify a new
+Partition Key, or changes to your Read/Write Permissions. Making changes to your
```
MongoCaleb, line 35: Suggest: "you can specify a new Partition Key or make changes..."

source/sync/configure/alter-sync-configuration.txt
```diff
@@ -0,0 +1,101 @@
...
+{+sync-short+} configuration while terminating and re-enabling {+sync-short+} will trigger a client
+reset. To learn more about handling client resets, read the :ref:`client reset
+<client-resets>` documentation.
```
MongoCaleb, line 38: I think elsewhere in the docs, we use this style:

"To learn more about handling client resets, see :ref:`client resets <client-resets>`."

source/sync/configure/alter-sync-configuration.txt
```diff
@@ -0,0 +1,101 @@
...
+
+.. image:: /images/FS-remove-queryable-field.png
+   :alt: Remove a Queryable Field
+
+.. flexible-sync-permission-changes:
+
+Permission Changes
+~~~~~~~~~~~~~~~~~~
+Making changes to permissions, whether implicitly through user metadata, or
+explicitely through changing configuration filters, will trigger a client reset
```
MongoCaleb, line 95: "explicitely" ==> "explicitly"

source/sync/configure/alter-sync-configuration.txt
```diff
@@ -0,0 +1,101 @@
...
+and re-download of all {+sync-short+} data with the new permissions applied. 
+
+A client reset will not be triggered under the following cases: 
+
+- Adding a new collection to a {+sync-short+} schema, and defining permissions for the new namespace or using default roles. This will not trigger a client reset because permissions have not been applied previously.
+- Configuring custom permissions for the new collection in the same :ref:`draft <deploy-ui>` as the new schema. Alternatively, deploying a draft with permission changes after you deploy the schema will lead to a client reset, since default permissions were applied in the initial deployment.
```
MongoCaleb, line 101: "alternatively" should be "Conversely" or "However", I think.

source/sync/configure/optimize-sync-atlas-usage.txt
```diff
@@ -298,15 +300,50 @@ Set the Client Maximum Offline Time
 
             realm-cli push --remote="<Your App ID>"
 
+.. _optimizing-performance-and-storage-flexible-sync:
+
+Optimizing Performance and Storage When Using Flexible Sync
```
MongoCaleb, line 305: I'm now wondering about this title being "Flexible Sync Considerations"

MongoCaleb (approved):
It's not in this PR, but I'm questioning the need for the "Key Concepts" heading in the Optimize page. I think all it serves is to break up the main part of the Max Offline section, which starts off with Key Concepts. If there is a difference I'm missing, I'd consider a different heading.

source/includes/steps-configure-flexible-sync-api.yaml
```diff
@@ -0,0 +1,158 @@
...
+  you continue.
+
+  You'll need the cluster's service configuration file to configure {+sync-short+}. You
+  can find the configuration file by listing all services through the Admin API:
+
+  .. code-block:: shell
+     
+     curl https://realm.mongodb.com/api/admin/v3.0/groups/{GROUP_ID}/apps/{APP_ID}/services \
+       -X GET \
+       -h 'Authorization: Bearer <Valid Access Token>'
```
cbush, line 16: Should we link to how to get a valid access token? Otherwise someone might be lost.

source/includes/steps-configure-flexible-sync-api.yaml
```diff
@@ -0,0 +1,158 @@
...
+                 "queryable_fields_names": []
+             }
+         }
+     }
+
+  Replace the ``database_name`` field's value with the name of your database.
+
+  .. note::
+    
+    The default role applies to any collection where you do not define custom roles.
```
cbush, line 48: What does the default role in this example do? Also, why is defaultRoles plural and an array, does that imply you can have multiple default roles? How do they interact in that case?

source/includes/steps-configure-flexible-sync-api.yaml
```diff
@@ -0,0 +1,158 @@
...
+  .. note::
+    
+    The default role applies to any collection where you do not define custom roles. 
+    Since no custom roles are defined in the above example, the default role 
+    applies to all collections.
+---
+title: Choose Queryable Fields
+ref: choose-queryable-fields
+content: |
+  The ``queryable_fields_names`` property allows you to define a set of
```
cbush, line 55:
```suggestion
  The ``queryable_fields_names`` property defines a set of
```

source/includes/steps-configure-flexible-sync-api.yaml
```diff
@@ -0,0 +1,158 @@
...
+  queryable fields for the device should query on. A queryable field can be any
```
cbush, line 56:
```suggestion
  queryable fields that the device may query on. A queryable field can be any
```

source/includes/steps-configure-flexible-sync-api.yaml
```diff
@@ -0,0 +1,158 @@
...
+  top-level primitive field with a scalar type, excluding embedded objects or arrays.
```
cbush, line 57: Isn't scalar type (and primitive field) by definition non-objects and arrays?

source/includes/steps-configure-flexible-sync-api.yaml
```diff
@@ -0,0 +1,158 @@
...
+             }
+         }
+     }
+---
+title: Define Read & Write Permissions
+ref: define-read-write-permissions
+content: |
+  {+service+} allows you to define :ref:`JSON expressions <expressions>` that it
+  dynamically evaluates whenever a user opens a {+realm+} to determine if the user
+  has read or write permissions for data in the document.
```
cbush, line 92: This sentence looks fairly complex and could be broken up:

>You can define JSON expressions to tell Realm whether a user has read or write permissions on a given document. Realm evaluates these expressions for a user when that user opens a realm.

source/includes/steps-configure-flexible-sync-api.yaml
```diff
@@ -0,0 +1,158 @@
...
+
+  {+sync-short+} rule expressions have access to :json-expansion:`%%user`, which resolves
+  to the user that opened the {+realm+} You can also use operators, including
```
cbush, line 95:
```suggestion
  to the user that opened the {+realm+}. You can also use operators, including
```

source/includes/steps-configure-flexible-sync-api.yaml
```diff
@@ -0,0 +1,158 @@
...
+  :json-operator:`%function`, to handle more complex cases. For examples, see
+  :ref:`{+sync-short+} Permissions <sync-permissions>`.
+
+  Once you've determined how to decide a user's read and write permissions for a
+  given document, define the corresponding JSON expressions in the
+  ``read`` and ``write`` fields of ``config.sync.partition.permissions``.
```
cbush, line 101: "determined how to decide" is confusing. I think you're getting at the different types of expressions (complex ones using operators or simple ones), but I think it would be more straightforward to remove the first part of this sentence and just say where to write the JSON expressions.

source/includes/steps-configure-flexible-sync-cli.yaml
```diff
@@ -0,0 +1,154 @@
...
+                 },
+                 "queryable_fields_names": ["age", "name"]
+             }
+         }
+     }
+---
+title: Define Read & Write Permissions
+ref: define-read-write-permissions
+content: |
+  Realm allows you to define :ref:`JSON expressions <expressions>` that it
```
cbush, line 91: Same comments as above apply here.

source/includes/steps-configure-flexible-sync-ui.yaml
```diff
@@ -0,0 +1,80 @@
...
+  the :guilabel:`Select Cluster To {+sync-short+}` dropdown menu.
+
+  .. figure:: /images/flex-sync-select-development-details.png
+     :width: 750px
+     :alt: Select Development Details Menu
+---
+title: Select Queryable Fields
+ref: select-queryable-fields
+content: |
+  When configuring Flexible {+sync-short+}, specify one or more fields from your
```
tkaye407, line 41: [suggest]  This should probably link to the queryable fields page I think, no?

source/includes/steps-configure-flexible-sync-ui.yaml
```diff
@@ -0,0 +1,80 @@
...
+---
+title: Define Read & Write Permissions
+ref: define-read-write-permissions
+content: |
+  The :guilabel:`Define Permissions` section allows you to define :ref:`JSON
+  expressions <expressions>` that {+service+} dynamically evaluates to determine a
+  user's read and write permissions for the data in a given partition.
+  
+  {+sync-short+} rule expressions have access to :json-expansion:`%%user`, which resolves
+  to the user that opened the {+realm+}, and :json-expansion:`%%partition`, which
```
tkaye407, line 61: This section shouldnt mention partitions IMO, also permissions don't have access to this. The basic gist is that 
(a) the apply when has access to the "user" only 
(b) the permissions have access to the user and the document being passed through sync

source/includes/steps-configure-flexible-sync-ui.yaml
```diff
@@ -0,0 +1,80 @@
+title: Navigate to the {+sync-short+} Configuration Screen
+ref: navigate-to-the-{+sync-short+}-configuration-screen
+content: |
+  To define {+sync-short+} rules and enable {+sync-short+} for your application, navigate to the
+  :guilabel:`{+sync-short+}` configuration screen through the {+leftnav+}.
+
+  .. note:: {+sync-short+} Rules Cannot Change
```
tkaye407, line 7: Is this the case? I can double check but I think that you can change the rules (your other page even seems to suggest this)

source/sync/configure/alter-sync-configuration.txt
```diff
@@ -0,0 +1,101 @@
...
+   :depth: 1
+   :class: singlecol
+
+Overview
+--------
+
+You may want to alter your {+sync-short+} configuration after you have already
+:ref:`enabled {+sync-short+} <enable-realm-sync>`. 
+
+If you are using Partition-Based {+sync-short+}, and want to change the ``Partition Key``,
```
tkaye407, line 21: The overview seems a bit odd to me because it seems like these two sentences shouldnt be here, but curious what you think

source/sync/configure/alter-sync-configuration.txt
```diff
@@ -0,0 +1,101 @@
...
+you can terminate and re-enable {+sync-short+}.
+
+If you are using Flexible {+sync-short+} and want to add or remove queryable
+fields, you can do so immediately, without terminating {+sync-short+}.
+
+.. alter-partition-based-sync-config:
+
+Alter Your Partition-Based Sync Configuration
+---------------------------------------------
```
tkaye407, line 30: You can also enable / disable dev mode without terminating sync

tkaye407: May also be wokrth talking about schema changes. A change to the schema will result in all of the documents for that new schema being found in MongoDB and re-inserted into Realm (to pick up new objects / fields)

source/sync/configure/optimize-sync-atlas-usage.txt
tkaye407, line 305: I almost think we would want the PBS / QBS to be totally separate pages / toggleable since they have very different perf considerations (compaction doesnt exist for QBS for example)

source/sync/configure/optimize-sync-atlas-usage.txt
```diff
@@ -298,15 +300,50 @@ Set the Client Maximum Offline Time
...
+-----------------------------------------------------------
+
+For Flexible Sync configuration, the amount of Atlas storage space used is
```
tkaye407, line 308: I would say it is equally correlated with: 
1. Amount of data in the system: number of docs * size of docs 
2. Amount of changes flowing through the system: more changes = more history 
3. Number of queryable fields (have to store additional metadata about these)

source/sync/configure/optimize-sync-atlas-usage.txt
```diff
@@ -298,15 +300,50 @@ Set the Client Maximum Offline Time
...
+more refined queries to expose targeted sets of data in the client 
+application. Slicing off working sets from a broad query provides
+better performance than opening multiple synced {+realm+}s using more 
+granular queries.
+
+When you configure queryable fields, consider the broad queries you use 
+for Sync, and select fewer fields that support those broad queries.
+
+.. example::
+
```
tkaye407, line 335: Curious what @mpobrien thinks, but I think it wouldnt be a bad idea to also briefly discuss how sync is designed for the happy case of a change generally doesnt have too many documents shuffling in / out of a query (ie query on { numLikes: 100 } and num likes goes like [99, 100, 90, 119, 80, 120, etc]. We will handle this, but there will be additionally cost and latency for having objects moving in and out of view of the query / permissions frequently. I think it is potentially something worth highlighting as it has design considerations

source/includes/steps-configure-flexible-sync-ui.yaml
mongomoe: Good suggestion, slightly altered it to "Under the :guilabel:`Select Development Details` heading, you can configure
  various development details, such as {+sync-short+} types, development mode,
  and a cluster to {+sync-short+}."

👍

source/includes/steps-configure-flexible-sync-ui.yaml
mongomoe: Fair 👍

source/includes/steps-configure-flexible-sync-ui.yaml
mongomoe: Replaced the word wish with want👍

source/sync/configure/alter-sync-configuration.txt
mongomoe: Fixed 👍

source/sync/configure/alter-sync-configuration.txt
mongomoe: Replaced w/ Conversely 👍

source/sync/configure/optimize-sync-atlas-usage.txt
mongomoe: Resolving this since I think this is a bit more subjective. Personally, I think the current one is a bit more specific than "Considerations"

source/sync/configure/alter-sync-configuration.txt
mongomoe: Can you describe what seems odd about the overview as it is? Happy to change it, The first sentence already existed & the second two sentences are brief descriptions on how to change a sync config depending on which sync mode/type you are using.

source/sync/configure/optimize-sync-atlas-usage.txt
mongomoe: Would you recommend not using ``numLikes`` in that case? Or not using similar integer fields as queryable fields where documents are likely to move in/out of the query?

source/sync/configure/optimize-sync-atlas-usage.txt
mongomoe: The first two are true of PBS as well, this section focuses more on making optimizations based on the number of queryable fields.

source/sync/configure/optimize-sync-atlas-usage.txt
mongomoe: The first two are true of PBS as well, this section focuses more on making optimizations based on the number of queryable fields.

source/sync/configure/alter-sync-configuration.txt
mongomoe: There are some things that do apply to both, for instance, sync history. Maybe a note under the "backend compaction" (and "Client Maximum Offline Time") that says it is not supported by Flexible Sync would be better than splitting this content into two pages? It would definitely be the quicker change as well.

Alternatively, what about an "Optimizing Performance and Storage When Using Partition-Based Sync" section, where backend compaction and client maximum offline time could be two subsections?

source/sync/configure/alter-sync-configuration.txt
mongomoe: * Can enabling/disabling dev without terminating sync be done for both Flexible Sync? If so, that info should be on the development mode page
* Not sure if we necessarily want to put schema change information in the PBS config section, since this can apply to both sync modes. Also, since this page deals with the sync configuration screen, would that info be better in another page? Maybe: https://docs.mongodb.com/realm/schemas/

source/includes/steps-configure-flexible-sync-ui.yaml
mongomoe: Can you describe what seems odd about the overview as it is? Happy to change it, The first sentence already existed & the second two sentences are brief descriptions of how to change a sync config depending on which sync mode/type you are using.

source/includes/steps-configure-flexible-sync-ui.yaml
mongomoe: I don't think it is in retrospect, I've removed it now.  👍

source/includes/steps-configure-flexible-sync-ui.yaml
mongomoe: Sorry, this looks like a copy & paste mixup on my part. I've altered this step now to remove any mention of partitions. The text now is:
```
  The Define Permissions section allows you to define JSON expressions that
  Realm dynamically evaluates to determine a user's read and write permissions
  for the documents being passed through sync.

  The Apply When expression determines if a role applies to a given document. 
```

source/sync/configure/optimize-sync-atlas-usage.txt
mongomoe: There are some things that do apply to both, for instance, sync history. Maybe a note under the "backend compaction" (and "Client Maximum Offline Time") that says it is not supported by Flexible Sync would be better than splitting this content into two pages? It would definitely be the quicker change as well.

Alternatively, what about an "Optimizing Performance and Storage When Using Partition-Based Sync" section, where backend compaction and client maximum offline time could be two subsections?

source/sync/configure/alter-sync-configuration.txt
tkaye407: To the first question the answer is yes. As for the second question I guess this is more a question for product / you guys, but it seems like we are taking the approach of having kind of a "shared", "partition", and "flexible" logical separation of the docs. I guess this makes sense, but I also think its likely that people who come to use flexible sync will not read the section about partition sync and vice versa. I kind of feel like long term having all of these pages be toggleable / completely separate (ie not having a PBS and QBS section on every page about a shared topic like this causing you to scroll to find the part you care about).

source/sync/configure/optimize-sync-atlas-usage.txt
tkaye407: That seems reasonable. Client Maximum offline time is something we are going to support pretty quickly so not sure it is worth removing mention of it from Flexible Sync. I guess my question just is the same as the above, do we want a user to have to navigate the docs to figure out which section is relevant to them pbs/qbs or just have only the imformation relevant to their sync type on the page / toggleable. Sounds like if we did this it would be a more extensive change so just calling this out as a nit / future improvement / discussion.

source/sync/configure/optimize-sync-atlas-usage.txt
tkaye407: Same as the above. If I were I user, I would scroll right down to this section and not read through the partition based sync section. I also think this wording makes it seem like the number of queryable fields is the most important variable in this equation when I would guess / suspect it is the least important of these 3

source/sync/configure/optimize-sync-atlas-usage.txt
tkaye407: Fair point. I think showing that these queries can be done are cool / important. I guess I would suggest: 
(a) we highlight that users can now use the (almost full) mongo query language, so an example with $in, $gt, etc would be cool since $eq is functionally just partition based sync 
(b) You are correct, I would not recommend this, but it is also perfectly fine and supported just with some perf considerations. I guess what I am trying to get at is just that I think it's worth calling out that while this works, users who are concerned about performance should try to craft their permissions/client queries to avoid lots of objects moving in / out of the view of the expression

tkaye407 (approved):
Looks good to me mod some things. I do think a question we will later want to answer is whether we want to put Partition Sync and Flexible Sync docs together for all of these sections or have it organized in a way where you can easily choose that you only care about one of them and dont keep having to scroll past the other to get to the information you care about

source/includes/steps-configure-flexible-sync-api.yaml
mongomoe: Fixed: Added a note linking to documentation on how to get a valid access token 👍

source/includes/steps-configure-flexible-sync-api.yaml
mongomoe: This note is referring to the default role specified in that example (all users can read/write all data). You can create multiple default roles (although you need at least 1) to apply across collections. I'll add a link here to the "default roles" section in case anyone is confused, let me know if that works for you.

source/includes/steps-configure-flexible-sync-api.yaml
mongomoe: Yes, that's true, I was trying to be very explicit that you could not use embedded objects and arrays, but my previous wording was a bit redundant. I've reworded it to the following, which is still a bit redundant but hopefully this way no one is confused & we're not implying that a scalar type can be an array/embedded object outside of a queryable field:

```
A queryable field can be any top-level primitive field with a scalar type. 
A queryable field cannot be an embedded object or array.
```

source/includes/steps-configure-flexible-sync-api.yaml
mongomoe: Good point - your sentence sounds much more simple, I've replaced my old sentence w/ the suggested one now 👍

source/includes/steps-configure-flexible-sync-api.yaml
mongomoe: Fixed 👍  removed that first part of the sentence

source/includes/steps-configure-flexible-sync-cli.yaml
mongomoe: Fixed 👍

cbush (approved):
Looks great.

source/sync/configure/alter-sync-configuration.txt
mongomoe: Deciding to leave out the enable/disabling dev mode bit, since it seems unnecessary for this specific page. But we do specify how to enable/disable dev mode here - https://docs.mongodb.com/realm/sync/data-model/development-mode/#enable-or-disable-development-mode

As for the approach of showing PBS & FS on the same page, this will likely be altered by the official release when FS becomes the main way to sync. Right now we want to show both ways since FS is new & there may be issues users run into while developing a FS application. Showing both on the same page allows users to compare alternative approaches. However, if we get feedback requesting for it to be split, I'm happy to split the two modes into two separate pages.

source/sync/configure/optimize-sync-atlas-usage.txt
mongomoe: Added a note about client max offline time + backend compaction not currently being supported by flex sync

PR #1540, opened 2022-01-05
(DOCSP-19418): Swift SDK Flexible Sync

dacharyc:
## Pull Request Info

Note for reviewer: the Flexible Sync examples can't currently be tested in our Swift SDK unit test framework. I created an independent project and verified that the syntax compiles, but haven't tested these code examples against a backend. We'll need to update with tested code examples after the feature is live.

### Jira

- https://jira.mongodb.org/browse/DOCSP-19418

- [Flexible Sync](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19418/sdk/swift/examples/flexible-sync/): Net new content about managing subscriptions using the Swift SDK API
- [Sync Changes Between Devices](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19418/sdk/swift/examples/sync-changes-between-devices/): Removed content about opening a realm, match other SDK versions of the page
- [Configure and Open a Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19418/sdk/swift/examples/configure-and-open-a-realm/): Moved content here from Sync Changes Between Devices page to match the other SDKs, add an example of opening a Flexible Sync configuration

### Review Guidelines

### DIFF
--- a/source/sdk/swift/examples/configure-and-open-a-realm.txt
+++ b/source/sdk/swift/examples/configure-and-open-a-realm.txt
@@ -178,7 +178,7 @@ to open a synced realm.
    let app = App(id: YOUR_APP_ID_HERE)
    let user = try await app.login(credentials: Credentials.emailPassword(email: "email", password: "password"))
    let config = user.flexibleSyncConfiguration()
-   let realm = try await Realm.init(configuration: config, downloadBeforeOpen: .always)
+   let realm = try await Realm(configuration: config, downloadBeforeOpen: .always)
 
 .. _ios-specify-download-behavior:
 
--- a/source/sdk/swift/examples/configure-and-open-a-realm.txt
+++ b/source/sdk/swift/examples/configure-and-open-a-realm.txt
@@ -71,6 +71,139 @@ Open a Local Realm
       .. literalinclude:: /examples/generated/code/start/OpenCloseRealm.codeblock.open-local-realm.m
          :language: objectivec
 
+.. _ios-login-and-open-realm:
+
+Open a Synced Realm
+-------------------
+
+The typical flow for opening a synced {+realm+} involves:
+
+1. :ref:`Authenticating the user <ios-authenticate-users>`.
+#. Creating a sync configuration.
+#. Opening the user's synced {+realm+} with the configuration.
+
+At authentication, we cache user credentials in a ``sync_metadata.realm`` 
+file on device.
+
+When you open a synced {+realm+} after authenticating, you can bypass the 
+login flow and go directly to opening the synced {+realm+}, using the same 
+sync configuration you already created.
+
+With cached credentials, you can:
+
+- Open a synced {+realm+} immediately with the data that is on the device.
+  You can use this method offline or online.
+- Open a synced {+realm+} after downloading changes from your {+app+}. 
+  This requires the user to have an active internet connection.
+
+Synced Realms vs Local Realms
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Synced {+realms+} differ from local {+client-database+} in a few ways:
+
+- Synced {+realms+} attempt to sync changes with your backend {+app+},
+  whereas local {+realms+} do not.
+- Synced {+realms+} can be accessed by authenticated users, while local 
+  {+realms+} do not require any concept of users or authentication.
+- With synced {+realms+}, you can :ref:`specify the download behavior 
+  <ios-specify-download-behavior>` to download updates before opening a 
+  {+realm+}. However, this requires users to be online. Local {+realms+} - 
+  with no sync capability - can always be used offline.
+
+You can copy data from a :ref:`local {+client-database+} <ios-open-a-local-realm>` 
+to a synced {+realm+}, but you cannot sync a local {+client-database+}.
+
+.. _ios-partition-based-sync-open-realm:
+
+Open a Synced Realm with Partition-Based Sync
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Initialize a synced {+realm+} with a :swift-sdk:`sync configuration <Extensions/User.html#/s:So7RLMUserC10RealmSwiftE13configuration14partitionValueAC0B0V13ConfigurationVx_tAC4BSONRzlF>`.
+This enables you to specify a partition value whose data should sync to the realm.
+
+.. tabs-realm-languages::
+   
+   .. tab::
+      :tabid: swift
+
+      .. versionchanged:: 10.15.0
+
+      .. seealso::
+         
+         This example uses async/await introduced in Swift 5.5 for iOS 15. If you're
+         using an older version of Swift or need to support older platforms, see:
+         :ref:`Legacy Realm Sync Open Methods <ios-realm-open-legacy>`. 
+
+      Pass a logged-in user's :swift-sdk:`configuration <Structs/Realm/Configuration.html>` 
+      object with the desired :ref:`partition value <partition-value>` to 
+      :swift-sdk:`{+realm+} initializers 
+      <Structs/Realm.html#/s:10RealmSwift0A0V13configuration5queueA2C13ConfigurationV_So012OS_dispatch_D0CSgtKcfc>`.
+
+      You can optionally :ref:`specify whether a {+realm+} should download 
+      changes before opening <ios-specify-download-behavior>`. If you do not
+      specify download behavior, this opens a {+realm+} with data that is on
+      the device, and attempts to sync changes in the background.
+
+      .. literalinclude:: /examples/generated/code/start/Sync.codeblock.open-synced-realm.swift
+         :language: swift
+
+   .. tab::
+      :tabid: objective-c
+
+      The first time you log in and open a synced {+realm+}, you'll log in the
+      user, and pass the user's :objc-sdk:`RLMSyncConfiguration 
+      <Classes/RLMRealmConfiguration.html#/c:objc(cs)RLMRealmConfiguration(py)syncConfiguration>` 
+      object with the desired :objc-sdk:`partitionValue 
+      <Classes/RLMSyncConfiguration.html#/c:objc(cs)RLMSyncConfiguration(py)partitionValue>` 
+      to :objc-sdk:`+[RLMRealm realmWithConfiguration:error:]
+      <Classes/RLMRealm.html#/c:objc(cs)RLMRealm(cm)realmWithConfiguration:error:>`.
+
+      This opens a synced {+realm+} on the device. The {+realm+} 
+      attempts to sync with your {+app+} in the background to check for changes 
+      on the server, or upload changes that the user has made.
+
+      .. literalinclude:: /examples/generated/code/start/Sync.codeblock.init-synced-realm.m
+         :language: swift
+
+.. _ios-flexible-sync-open-realm:
+
+Open a Synced Realm with a Flexible Sync Configuration
+------------------------------------------------------
+
+When you use Flexible Sync, use the ``flexibleSyncConfiguration()``
+to open a synced realm. 
+
+.. code-block:: swift
+
+   let app = App(id: YOUR_APP_ID_HERE)
+   let user = try await app.login(credentials: Credentials.emailPassword(email: "email", password: "password"))
+   let config = user.flexibleSyncConfiguration()
+   let realm = try await Realm.init(configuration: config, downloadBeforeOpen: .always)
+
+.. _ios-specify-download-behavior:
+
+Download Changes Before Open
+----------------------------
+
+.. versionadded:: 10.15.0
+
+When you open a synced {+realm+} with the Swift SDK, you can pass the 
+``downloadBeforeOpen`` parameter to specify whether to download the 
+changeset from your {+app+} before opening the {+realm+}. This parameter 
+accepts a case from the ``OpenBehavior`` enum:
+
+- ``never``: Immediately open the {+realm+} on the device. Download changes 
+  in the background when the user has internet, but don't block opening
+  the {+realm+}.
+- ``always``: Check for changes every time you open the {+realm+}.
+  Requires the user to have an active internet connection.
+- ``once``: Download data before opening a {+realm+} for the first time, but
+  open it without downloading changes on subsequent opens. This lets you 
+  populate a {+realm+} with initial data, but enables offline-first 
+  functionality on subsequent opens.
+
+.. include:: /examples/generated/code/start/Sync.codeblock.specify-download-behavior.swift.code-block.rst
+
 .. _ios-close-a-realm:
 
 Close a Realm
--- a/source/sdk/swift/examples/configure-and-open-a-realm.txt
+++ b/source/sdk/swift/examples/configure-and-open-a-realm.txt
@@ -96,15 +96,15 @@ With cached credentials, you can:
 - Open a synced {+realm+} after downloading changes from your {+app+}. 
   This requires the user to have an active internet connection.
 
-Synced Realms vs Local Realms
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Synced Realms vs. Local Realms
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Synced {+realms+} differ from local {+client-database+} in a few ways:
 
 - Synced {+realms+} attempt to sync changes with your backend {+app+},
   whereas local {+realms+} do not.
 - Synced {+realms+} can be accessed by authenticated users, while local 
-  {+realms+} do not require any concept of users or authentication.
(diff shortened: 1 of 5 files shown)


source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+You can update subscriptions using ``update``. In this example, we 
+search for a subscription matching our query, and then update that 
+subscription with a new query. 
+
+.. code-block:: swift
+
+   let subscriptions = realm.subscriptions
+   let query = { QuerySubscription<Team> { $0.name == "Developer Education" } }
+   if let subscription = subscriptions.first(where: query) {
+      try subscriptions.write {
```
cbush, line 160: Don't these need to be `try!` or else have a catch somewhere? (but for examples we can just force try)

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+         subscription.update {
+            QuerySubscription<Team>(name: "docs-team") {
+               $0.name == "Documentation"
```
cbush, line 163: What consequence does this have on the name of the query? If `subscriptions` is a live object, when we commit this, shouldn't `subscription` no longer actually contain a result (since the query no longer applies)? Something worth noting?

In other words, if we use `subscription` after this, would we be using an invalid object? Or is it still referring to the correct instance, just with a new name?

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+
+.. _ios-remove-subscriptions:
+
+Remove Subscriptions
+~~~~~~~~~~~~~~~~~~~~
+
+To remove subscriptions, you can:
+
+- Remove a single subscription query
+- Remove all subscriptions of a specific type
```
cbush, line 177: Ambiguous between specific type of _subscription_ (what's that?) and object type. Can we say something like:
```suggestion
- Remove all subscriptions from a specific object type
```

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+
+When you remove a subscription query, the server asynchronously removes
```
cbush, line 179: How does the server remove anything from the client?

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+- Remove all subscriptions of a specific type
+
+When you remove a subscription query, the server asynchronously removes 
+synced data from the client device.
+
+Remove a Single Subscription
+````````````````````````````
+
+You can remove a specific subscription query in a subscription write block 
+using ``remove``. Specify the query by name, or use the query as a string
```
cbush, line 186:
```suggestion
using ``remove``. Specify the query by name or use the query as a string 
```

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+You can remove a specific subscription query in a subscription write block 
+using ``remove``. Specify the query by name, or use the query as a string 
+name to find the appropriate query subscription to remove.
+
+.. code-block:: swift
+
+   let subscriptions = realm.subscriptions
+   if let subscription = subscriptions.first(where: { $0.name == "docs-team" }) {   
+      try subscriptions.write {
+         try subscriptions.remove(subscription)
```
cbush, line 194: Does this API need try?

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+If you want to remove all subscriptions to a specific object type, use the 
+``removeAll`` method in a subscription write block.
+
+.. code-block:: swift
+
+   let subscriptions = ream.subscriptions
+   try await subscriptions.write {
+      try subscriptions.removeAll(ofType: Team.self)
+   }
+   try await subscriptions.waitforAsync()
```
cbush, line 210: Where is `waitforAsync()` explained?

source/sdk/swift/examples/configure-and-open-a-realm.txt
```diff
@@ -71,6 +71,139 @@ Open a Local Realm
...
+sync configuration you already created.
+
+With cached credentials, you can:
+
+- Open a synced {+realm+} immediately with the data that is on the device.
+  You can use this method offline or online.
+- Open a synced {+realm+} after downloading changes from your {+app+}. 
+  This requires the user to have an active internet connection.
+
+Synced Realms vs Local Realms
```
cbush, line 99: Shouldn't that be `vs.`?

source/sdk/swift/examples/configure-and-open-a-realm.txt
```diff
@@ -71,6 +71,139 @@ Open a Local Realm
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Synced {+realms+} differ from local {+client-database+} in a few ways:
+
+- Synced {+realms+} attempt to sync changes with your backend {+app+},
+  whereas local {+realms+} do not.
+- Synced {+realms+} can be accessed by authenticated users, while local 
+  {+realms+} do not require any concept of users or authentication.
```
cbush, line 107: It feels like we're mixing two ideas in this, "do not require" and "any concept of". Maybe it should say:
```suggestion
  {+realms+} have no concept of users or authentication.
```

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+
+You can:
+
+- Add subscriptions
+- Check subscription state
+- Update subscriptions with new queries
+- Remove individual subscriptions or all subscriptions of a type
+
+Data matching the subscription, where the user has the appropriate 
+permissions, syncs between devices and the backend application.
```
cbush, line 59: One thing I'm left wondering is how to set up permissions. In particular, how does a given query match permissions? (This might be answered elsewhere, but maybe we should link to it from here).

cbush (approved):
Several comments that I think should be addressed, but overall it's looking great!

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+
+A Realm subscription has:
+
+- Identifier
```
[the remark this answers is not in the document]
dacharyc: I think I'm going to delete these, actually; the optional `name` is exposed, but these other components aren't exposed in the API, so I don't think it's helpful for devs to know about them.

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+- Name
+- Created and updated dates
+
+You can specify a string name for your subscription. If you do not give your
+subscription a name, it uses the value of the query as a string name.
+
+When you create a subscription, Realm looks for data matching a query on a
+specific object type. You can have subscriptions on several different object 
+types, or several queries on the same object type, in your Flexible Sync
```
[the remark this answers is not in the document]
dacharyc: I actually think I'll break the sentence apart to emphasize the difference. Something like:

> You can have multiple subscription sets on different object types. You can also have multiple queries on the same object type.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: Hmm, good point, this is a poor example. I'll change it.

cbush: I actually think the question would be worth answering in the docs.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: This is a fair point. I had a similar comment on a permissions-related PR about queries. I've spoken with Moe and am pondering additional info/examples/links.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: Nope, error on my part. I'm updating all of the API code examples based on my experimenting with flex sync in development.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: It's not - it was in the original tech design code example, but I'm removing it from the updated examples based on my experimentation.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: I think this particular question no longer applies, as there has been a change to the API (I think? At least in the other SDKs - haven't seen the updated Swift SDK lately). `name` no longer returns a stringified version of the query - if you don't set an explicit name, `name` returns null.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: I think this was a method in the original tech design that does not exist in the API. I've added some verbiage to the "check state" section that explains something similar.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: Added this to a ..seealso:: in the Overview 👍

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+- Trigger error handling
+- Show a progress indicator while data is downloading
+- Find out when a subscription set is superseded and you should obtain a
+  new instance of the subscription set to write a subscription change
+
+.. code-block:: swift
+
+   let subscriptions = realm.subscriptions
+   try await subscriptions.write {
+      try subscriptions.remove {
```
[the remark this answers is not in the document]
dacharyc: It doesn't _really_ matter - any change to a subscription will trigger a state notification - and this is what was in the tech design. But if you think another type of change would make more sense from an example standpoint, I'm happy to change it.

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+Check Subscription State
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can watch the state of the subscription set with the ``SyncSubscriptionState``
+enum. You can use subscription state to:
+
+- Trigger error handling
+- Show a progress indicator while data is downloading
+- Find out when a subscription set is superseded and you should obtain a
+  new instance of the subscription set to write a subscription change
```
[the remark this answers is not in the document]
dacharyc: Fair point - I've shortened this bullet point, and have added a section talking about `superceded` state (which will likely need more updating before we publish as there has been some debate about what this means/should entail).

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+   if subscriptions.isEmpty {
+      try await subscriptions.write {
+         try subscriptions.append {
+            QuerySubscription<Team> {
+               $0.name == "Developer Education"
+            }
+         }
+      }
+      try await subscriptions.waitForAsync()
+   }
```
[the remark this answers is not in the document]
dacharyc: I've added a paragraph about name/query value uniqueness, but will want to update this more once we have a final version I can test with (I don't know if this will throw or just quietly ignore a dupe).

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,210 @@
...
+   ``$0.name == "Developer Education"``.
+
+   .. code-block:: swift
+
+      QuerySubscription<Team> {
+         $0.name == "Developer Education"
+      }
+
+   You can also create a subscription with an explicit name. Then, you can
+   search for that subscription by name to update or remove it.
```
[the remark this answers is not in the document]
dacharyc: Added a paragraph about this - good note 👍

mongomoe (approved):
LGTM but added a couple comments based on the last commit

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,284 @@
...
+
+   let subscriptions = realm.subscriptions
+   try! subscriptions.write {
+      subscriptions.remove {
+         QuerySubscription<Task> {
+            $0.assignee == "Joe Doe"
+         }
+      }
+   }
+   subscriptions?.observe { state in
```
jsflax, line 176: This API is changing pretty notably.

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,284 @@
...
+      if case .complete = state {
+            // Do something when state is complete
+      }
+   }
+
+   // After the state change, wait for the synced realm to update data
+   waitForDownloads(for: realm)
```
jsflax, line 183: Where are you getting this method from?

source/sdk/swift/examples/configure-and-open-a-realm.txt
```diff
@@ -71,6 +71,139 @@ Open a Local Realm
...
+
+When you use Flexible Sync, use the ``flexibleSyncConfiguration()``
+to open a synced realm. 
+
+.. code-block:: swift
+
+   let app = App(id: YOUR_APP_ID_HERE)
+   let user = try await app.login(credentials: Credentials.emailPassword(email: "email", password: "password"))
+   let config = user.flexibleSyncConfiguration()
+   let realm = try await Realm.init(configuration: config, downloadBeforeOpen: .always)
```
jsflax, line 181: can remove the `.init`, generally considered "unconventional"

jsflax (commented):
There is notably no mention of waiting for the `subscription.write`s to sync with the server, which is an important part of the feature. It will look like this for pre async/await:
```swift
subcriptions.write({
    ...
}, onComplete: { error in // error is optional
   ...
})
```
And this for async/await:
```swift
try await subcriptions.write {
    ...
}
```

source/sdk/swift/examples/flexible-sync.txt
jsflax: See my top review comment.

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -93,6 +97,10 @@ When you create a subscription, Realm looks for data matching a query on a
 specific object type. You can have multiple subscription sets on different 
 object types. You can also have multiple queries on the same object type.
 
+Subscription names must be unique. If you do not explicitly name a 
+subscription, and instead subscribe to the same unnamed query more than
+once, Realm ignores the duplicate query subscriptions.
```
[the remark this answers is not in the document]
dacharyc: I've expanded this section and made it a `note` that contains additional details I was able to glean from testing. I may make some more changes once this is live and we have a working iOS test for it, but I think it contains enough relevant details for the launch.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: Gotcha - I've looked at the changes to the Flex Sync PR and have updated this example, and the other examples. Thanks for the info!

Is `superceded` state something we want to advise devs to watch for directly, or can devs just get an error if they try to write to a subscription set that is superceded? I can see a case for `.pending` if apps may be syncing large subscription sets, but I'm wondering about how much guidance we want to give devs for `.superceded`.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: Ah, my error - I was misreading some of the older tests. I've updated this based on the info you've provided and the updates to the PR.

source/sdk/swift/examples/flexible-sync.txt
dacharyc: Updated - thanks for the info, @jsflax !

cbush (approved):

source/sdk/swift/examples/flexible-sync.txt
```diff
@@ -0,0 +1,358 @@
...
+task list app. The two Realm object types are ``Team``
+and ``Task``. A ``Task`` has a ``taskName``, assignee's name, and
+completed flag. There is also an arbitrary number for
+priority -- higher is more important -- and a count of
+minutes spent working on it. A ``Team`` has a ``teamName``, 
+zero or more ``Tasks``, and a list of ``members``.
+
+.. code-block:: swift
+
+   class Task: Object {
```
jsflax, line 59: Is there any other name we could pick for this class? `Task` is the [fundamental type](https://developer.apple.com/documentation/swift/task) for Swift's async/await feature.

dacharyc: Started talking with the team about this and we're thinking about this change while looking holistically at the use of `Task` across the entire docs set.

PR #1543, opened 2022-01-05
(DOCSP-19920): Add Flexible Sync RQL Limitations to RQL page

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19920

- [Realm Query Language](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19920/reference/realm-query-language/#flexible-sync-rql-limitations): This PR adds a section for Flexible Sync RQL Limitations

### Review Guidelines

### DIFF
--- a/source/reference/realm-query-language.txt
+++ b/source/reference/realm-query-language.txt
@@ -457,6 +457,9 @@ that determines which data to sync, the server does not support these
 query operators. However, you can still use the full range of RQL features
 to query the synced data set in the client application.
 
+Unsupported Query Operators in Flexible Sync
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 .. list-table::
    :header-rows: 1
    :widths: 40 60
@@ -464,29 +467,30 @@ to query the synced data set in the client application.
    * - Operator Type
      - Unsupported Operators
 
-   * - Embedded or Linked Objects
-     - Flexible Sync does not support querying on properties in Embedded 
-       Objects or links. For example, ``obj1.field = “foo”``.
-
    * - Comparison Operators
-     - Flexible Sync does not support BETWEEN. Use range operators for this 
-       type of query, instead. For example, ``field > 100 AND field < 200``.
+     - ``BETWEEN``, ``IN``
     
    * - Case-Insensitive Queries
-     - Flexible Sync does not support modifiers for case insensitivity.
+     - ``[c]``
 
    * - String Operators
-     - BEGINSWITH, ENDSWITH, CONTAINS, LIKE
+     - ``beginsWith``, ``endsWith``, ``contains``, ``like``, ``in``
 
    * - Aggregate Operators
-     - Flexible Sync does not support property operators.
+     - ``@avg``, ``@count``, ``@max``, ``@min``, ``@sum``
 
    * - Collection Operators
-     - ANY, SOME, ALL, NONE
+     - ``ANY``, ``SOME``, ``ALL``, ``NONE``
 
    * - Query Suffixes
-     - DISTINCT, SORT, LIMIT
+     - ``DISTINCT``, ``SORT``, ``LIMIT``
 
-   * - IN
-     - Flexible Sync does not support the ``IN`` operator, whether using
-       it for comparison or as a string operator.
+The recommended workaround for the ``BETWEEN`` comparison operator is to
+combine range operators with ``AND`` for this type of query, instead. For 
+example, ``field > 100 AND field < 200``.
+
+Embedded or Linked Objects
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Flexible Sync does not support querying on properties in Embedded Objects 
+or links. For example, ``obj1.field = “foo”``.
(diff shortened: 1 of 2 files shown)


mpobrien:
The contents about the flexible sync limitations are correct; however, I think it's important to make clear to the user that these limitations only apply to the query that is used to _open_ the sync session. It's still possible to query the synced data set locally using the full set of RQL features available on the client. I'm sure this will become more careful as we add examples of how to use the SDK as well.

source/reference/realm-query-language.txt
```diff
@@ -440,3 +445,48 @@ results collection.
...
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Operator Type
+     - Unsupported Operators
+
+   * - Embedded or Linked Objects
+     - Flexible Sync does not support querying on properties in Embedded 
+       Objects or links. For example, ``obj1.field = “foo”``.
```
nathan-contino-mongo, line 469: Maybe this is just pre-coffee brain, but it's unclear if this means you can't query *directly* on embedded/linked objects (that's what it seems to say right now) or if you just can't subquery on those objects, e.g. `obj.embeddedObject.field = "foo"`.

If both aren't allowed, I would suggest adding examples of *both* to make it clear that you can't do either. Unless this is already covered somewhere else I'm not thinking of!

source/reference/realm-query-language.txt
```diff
@@ -440,3 +445,48 @@ results collection.
...
+
+   * - Comparison Operators
+     - Flexible Sync does not support BETWEEN. Use range operators for this
```
nathan-contino-mongo, line 472:
```suggestion
     - Flexible Sync does not support BETWEEN. Combine range operators with AND for this 
```

Range operators sound like half the story, but it seems equally important to mention stitching them together with AND.

source/reference/realm-query-language.txt
```diff
@@ -440,3 +445,48 @@ results collection.
...
+     - Flexible Sync does not support property operators.
+
+   * - Collection Operators
+     - ANY, SOME, ALL, NONE
+
+   * - Query Suffixes
+     - DISTINCT, SORT, LIMIT
+
+   * - IN
+     - Flexible Sync does not support the ``IN`` operator, whether using
```
nathan-contino-mongo, line 491: If we're going to monospace "IN" here, maybe we should monospace all of the operators in the previous table rows?

source/reference/realm-query-language.txt
```diff
@@ -440,3 +445,48 @@ results collection.
...
+that determines which data to sync, the server does not support these
+query operators. However, you can still use the full range of RQL features
+to query the synced data set in the client application.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Operator Type
+     - Unsupported Operators
```
nathan-contino-mongo, line 465:
```suggestion
     - Description
```

Unfortunately it looks like the "Unsupported Operators" cells have expanded beyond lists of operators and include other details and even workarounds now. I don't really like the term "description" here, but it seems slightly better suited to the current contents than "Unsupported Operators."

source/reference/realm-query-language.txt
```diff
@@ -440,3 +445,48 @@ results collection.
...
+     - BEGINSWITH, ENDSWITH, CONTAINS, LIKE
+
+   * - Aggregate Operators
+     - Flexible Sync does not support property operators.
+
+   * - Collection Operators
+     - ANY, SOME, ALL, NONE
+
+   * - Query Suffixes
+     - DISTINCT, SORT, LIMIT
```
nathan-contino-mongo, line 488: The "String Operators", "Collection Operators" and "Query Suffixes" descriptions are all just lists of operators. It's pretty jarring to come across these -- I did a bit of a double take at first to make sure that these were still exclusive lists of unsupported operators. I think we should either:

- expand these cells into "Flexible Sync does not support ..." prefixed sentences
- maintain a "Unsupported Operators" column that's JUST a list of operators, and add a second column that's titled something like "Suggested Workarounds" or "Details" to handle situations where we need to elaborate

Or maybe something else -- these are just the solutions that popped out to me without diving too deeply into what a solution would actually look like. But I do think we have a bit of a problem with column content uniformity that we ought to address here.

nathan-contino-mongo (commented):
A few nitpicks and thoughts on consistency.

source/reference/realm-query-language.txt
dacharyc: Updated monospace and capitalization to match the prior instances on the page. Good call!

source/reference/realm-query-language.txt
dacharyc: Fair. I've updated the table to explicitly list all of the unsupported operators in the table, instead of some of the blanket "Flexible Sync doesn't support this category of operator" that I'd been using as shorthand. Probably better to be explicit.

I moved the one workaround to below the table - am pondering adding an asterisk to indicate there's a footnote. And I moved the "Embedded Object" callout to its own section, because it's not really an operator. I think this cleans things up.

nathan-contino-mongo (approved):
LGTM!

PR #1545, opened 2022-01-05
(DOCSP-18468,18470): Relationships

nlarew:
### Jira

- (DOCSP-18468): [Schemas - Guide] Define a Relationship
- (DOCSP-18470): [Schemas - Concept] Relationships

- [Relationships](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/relationships/schemas/relationships/)

### Review Guidelines

### DIFF
--- a/source/schemas/relationships.txt
+++ b/source/schemas/relationships.txt
@@ -15,11 +15,10 @@ Relationships
 Overview
 --------
 
-A relationship is a reference that links individual documents in a
-collection to other documents based on the values that each document
-contain. Relationships allow you to fluently reference and query related
-documents in read and write operations, even if the documents are in
-separate databases or collections.
+A relationship is a connection between two documents. Relationships
+allow you to reference and query related documents in read and write
+operations, even if the documents are in separate databases or
+collections.
 
 You define a relationship for a "source" MongoDB collection and link to
 documents in a "foreign" collection. Realm automatically resolves
@@ -93,8 +92,10 @@ relationships.
         }
       }
    
-   With this relationship defined, Realm can fluidly return a customer
-   and all of their accounts in the same GraphQL query.
+   With this relationship defined, Realm can return a customer and all
+   of their accounts in the same GraphQL query. Without a relationship,
+   the query would return a list of just ``account_id`` values instead
+   of the full ``Account`` objects.
    
    .. code-block:: text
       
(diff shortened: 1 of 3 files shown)


source/schemas/relationships.txt
```diff
@@ -0,0 +1,343 @@
...
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+A relationship is a reference that links individual documents in a
+collection to other documents based on the values that each document
```
cbush, line 19: It feels overly complex for a first sentence. What about something like:

>A relationship is a connection between two documents.

You later specify the connection can exist within the same collection or even across databases, so it can be left out.

Not sure what you mean by "based on the values that each document contains".

source/schemas/relationships.txt
```diff
@@ -0,0 +1,343 @@
...
+contain. Relationships allow you to fluently reference and query related
```
cbush, line 20:
```suggestion
contains. Relationships allow you to fluently reference and query related
```

cbush, line 20: Is there a special meaning to "fluently" (like Java-style "fluent interface pattern")? If so I would clarify - otherwise it feels wordy so I would remove.

```suggestion
contain. Relationships allow you to reference and query related
```

source/schemas/relationships.txt
```diff
@@ -0,0 +1,343 @@
...
+      
+      {
+        "accounts": {
+          "ref": "#/realm/mongodb-atlas/sample_analytics/accounts",
+          "foreign_key": "account_id",
+          "is_list": true
+        }
+      }
+   
+   With this relationship defined, Realm can fluidly return a customer
```
cbush, line 96: It feels like the meaning of "fluidly" here could be teased out a bit more: I think we're not explicitly saying you WOULD have to populate all the references in another system, but Realm is doing that automagically for you, saving you tons of boilerplate (without the marketing spin)

source/schemas/relationships.txt
```diff
@@ -0,0 +1,343 @@
...
+that it can reference. Realm supports two relationship cardinalities:
+"to-one" and "to-many".
+
+.. _relationships-to-one:
+
+To-One
+~~~~~~
+
+A to-one relationship links each source document with either a single
+document from the foreign collection.
```
cbush, line 124: Missing the "or" in "Either... or"?

cbush (approved):
It looks great. A few suggestions on the wording.

PR #1548, opened 2022-01-07
quickfix - add note about transitioning between sync modes

mongomoe:
### Jira

N/A

- [Define Data Acess Patterns > Choose Your Sync Mode](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/quick-fix-add-note-for-switching-between-sync-modes/sync/data-access-patterns/sync-mode/) (Includes a usage of the new note)

### Review Guidelines


source/includes/note-transition-between-sync-modes.rst
```diff
@@ -0,0 +1,6 @@
+.. note::
+
+   While Flexible Sync is in *preview*, you cannot use Flexible Sync &
+   Partition-Based Sync in the same {+app+}, so you must decide which {+sync+}
```
dacharyc, line 4: Nit: this is kind of a long sentence - would be better for readability to break it up.

```suggestion
   Partition-Based Sync in the same {+app+}. You must decide which {+sync+}
```

source/includes/note-transition-between-sync-modes.rst
```diff
@@ -0,0 +1,6 @@
...
+   mode to use in a given application. To switch between sync modes, you must
+   :ref:`terminate & re-enable sync <terminating-and-reenabling-sync>`.
```
dacharyc, line 6: Should we add to this note that terminating & re-enabling sync triggers a client reset? Something like:

```suggestion
   :ref:`terminate & re-enable sync <terminating-and-reenabling-sync>`, which triggers a :ref:`client reset <client-resets>`. 
```

I know that information is available on the terminate & re-enable Sync page, but I worry that for folks who don't know how this works, it isn't obvious that turning Sync off and back on again requires additional handling. At least adding the "client reset" info here makes it more explicit.

dacharyc, line 6: I think we use the proper noun when referring to the feature name?

```suggestion
   mode to use in a given application. To switch between Sync modes, you must
   :ref:`terminate & re-enable Sync <terminating-and-reenabling-sync>`. 
```

dacharyc (approved):
LGTM, mod a couple of grammar tweaks and a suggestion.

source/includes/note-transition-between-sync-modes.rst
mongomoe: Added the suggestion locally + committed + pushed. Thanks!

PR #1552, opened 2022-01-11
Update iOS CI

dacharyc:
I had to update the iOS CI to resolve CI issues on PR #1450 , but have been asked to not merge that PR yet. These three commits should fix the iOS CI so I can use the fixes on other branches.


examples/ios/Examples/ObjectModels.swift
```diff
@@ -200,11 +200,11 @@ class ObjectModels: XCTestCase {
         //        self.collection = collection
         //    }
 
-            let collection: AnyRealmCollection<ObjectModelsExamples_MyModel>
-
-            init<C: RealmCollection>(collection: C) where C.ElementType == ObjectModelsExamples_MyModel {
-                self.collection = AnyRealmCollection(collection)
-            }
+//            let collection: AnyRealmCollection<ObjectModelsExamples_MyModel>
```
cbush, line 203: Why not comment this out using block comments so as not to interfere with bluehawk?

cbush (approved):

PR #1557, opened 2022-01-13
DOCSP-19799: Add Redoc OpenAPI spec renderer

mongodben:
## Pull Request Info

primary changes: 
1. Refactored API spec page to make it compatible with Redoc
2. made spec contain all content (before some stuff was in RST
3. use the redoc openapi component to render the spec instead of the former fully homegrown solution.
4. update links in other docs pages to point to correct parts of the spec (b.c doesn't use refs)

<s>note: autobuilder is saying there are errors with links b/c it doesn't correctly parse anchor links on the api spec page. not sure if need to address or not. will these errors show up on all future builds that include these changes? if so, we should try to find a solution (open to ideas on that). if not, i think we can leave as is. </s> (edit: found a workaround to the autobuilder error)

### Jira

- https://jira.mongodb.org/browse/DOCSP-19799

- [API Reference](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19799/admin/api/v3/)

### Review Guidelines


cbush:
I think the build errors need to be fixed or quashed somehow.

mongodben:
> I think the build errors need to be fixed or quashed somehow.

will ask #snooty-platform on monday AM

mongodben:
@cbush ready for re-review. 
refactored to use the snooty role `admin-api-endpoint` so now it passes tests.

cbush (approved):

PR #1578, opened 2022-01-21
(DOCSP-19410): Flexible Sync - .NET SDK Documentation

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19410

- [Usage Examples > Flexible Sync - .NET SDK # Open a Synced Realm with a Flexible Sync Configuration](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-19410-fs-dotnet/sdk/dotnet/examples/open-a-realm/#open-a-synced-realm-with-a-flexible-sync-configuration) # Note: only the section entitled: "Open a Synced Realm with a Flexible Sync Configuration" needs to be reviewed
- [Usage Examples > Flexible Sync - .NET SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-19410-fs-dotnet/sdk/dotnet/examples/flexible-sync/)
- [Fundamentals > Realm Sync ](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-19410-fs-dotnet/sdk/dotnet/fundamentals/realm-sync/) # Note: this is a port over from the Swift/Node/Java page, and has already been copy reviewed

### Review Guidelines


source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+Get Subscriptions
+~~~~~~~~~~~~~~~~~
+When using a flexible synced realm, you can access a ``SubscriptionSet``, a collection of subscriptions, through the
+``Realm.Subscriptions`` property.
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.get-subscriptions.cs
+   :language: csharp
+
+You can use this snapshot to add queries to this list of subscriptions and
```
MongoCaleb, line 89: I don't think we previously define a snapshot; is that even the correct term? It implies a static collection (to me, at least)

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+When you create a subscription, Realm looks for data matching a query on a
+specific object type. In your Flexible Sync subscriptions, you can have
+subscriptions on several different object types or several queries on the same
+object type.
+
+
+Get Subscriptions
+~~~~~~~~~~~~~~~~~
+When using a flexible synced realm, you can access a ``SubscriptionSet``, a collection of subscriptions, through the
```
MongoCaleb, line 83: "When using flexible sync..."

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+You can add a single query, or batch multiple queries within a ``SubscriptionSet.Update``
+transaction. In the example below, we subscribe to three queries.
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.update-subscriptions.cs
+   :language: csharp
+
+Wait for Subscription Changes to Sync
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Writing an update to the subscription set locally is only one part
```
MongoCaleb, line 121: "After updating the subscription locally, Realm Sync resolves data changes with the updates, which may result in Realm adding or removing data from the synced realm"

MongoCaleb (commented):
General comment (made specifically in 2 places below): update all text referring to methods, properties, etc. to point to the appropriate pages in the SDK docs.

source/sdk/dotnet/examples/open-a-realm.txt
```diff
@@ -119,6 +120,19 @@ device already has a ``user``, it opens the synced {+realm+} with that user:
...
+Open a Synced Realm with a Flexible Sync Configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 10.9.0
+
+When you use Flexible Sync, pass the current user to the ``FlexibleSyncConfiguration()``
+to open a synced realm. 
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.open-a-flexible-synced-realm.cs
+   :language: swift
```
nirinchev, line 134:
```suggestion
   :language: csharp
```

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+   :class: singlecol
+
+
+Overview
+--------
+
+Flexible Sync uses subscriptions and permissions to determine which
+data to sync with your {+app+}.
+
+To use Flexible Sync in a .NET client:
```
nirinchev, line 22: Maybe this?

```suggestion
To use Flexible Sync in a .NET application:
```

source/sdk/dotnet/examples/flexible-sync.txt
nirinchev: Agreed - while the collection is indeed static in some sense, it's also dynamic - i.e. if the subscriptions changed on a bg thread, when you access the property again, you'll get the updated collection.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+When using a flexible synced realm, you can access a ``SubscriptionSet``, a collection of subscriptions, through the
+``Realm.Subscriptions`` property.
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.get-subscriptions.cs
+   :language: csharp
+
+You can use this snapshot to add queries to this list of subscriptions and
+update existing subscriptions, as shown in the examples below.
+
+Add a Query to the List Of Subscriptions
```
nirinchev, line 92: We prefer to use `Set` instead of `List` terminology here to convey the fact that subscriptions will be deduplicated on insertion. I.e. if you add the same query multiple times, only one element will end up in the collection.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
+Within a Flexible Sync ``transaction`` you can add queries to your list of
```
nirinchev, line 94: What does a flexible sync transaction mean here? I don't think we should bring in additional terminology that does not match directly the API. How about we call out that all mutations on the collection must happen within an update block?

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+subscriptions.
+
+Queries must be added to the list of existing subscriptions within a
+``transaction``. Create a transaction by calling ``SubscriptionSet.Update``.
+Within the callback function, call the ``Add()`` method on the
+``SubscriptionSet`` to add a query to the subscription.
+
+The ``SubscriptionSet.Add()`` method takes a ``query``, and a
+subscription options object. The subscription options include:
```
nirinchev, line 103: We should call out that providing subscription options is optional and subscriptions will be unnamed if you leave it as default.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+- a ``Name`` string field
+- a ``UpdateExisting`` boolean field.
+
+If ``UpdateExisting`` is true, adding a subscription with an
+existing name will replace the existing query with the new query. However, if
+you set ``UpdateExisting`` to false, adding a subscription with an existing name
+but a different query throws an exception. ``UpdateExisting`` defaults to false.
+
+You can add a single query, or batch multiple queries within a ``SubscriptionSet.Update``
```
nirinchev, line 113: We should clearly state that batching queries is by far the preferred way of doing as every update is quite expensive on the server.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+Within the callback function, call the ``Add()`` method on the
+``SubscriptionSet`` to add a query to the subscription.
+
+The ``SubscriptionSet.Add()`` method takes a ``query``, and a
+subscription options object. The subscription options include:
+
+- a ``Name`` string field
+- a ``UpdateExisting`` boolean field.
+
+If ``UpdateExisting`` is true, adding a subscription with an
```
nirinchev, line 108: We should call out the set semantics of the subscriptions collection - namely, that adding multiple unnamed subscriptions with the same query will de-duplicate them, as well as adding multiple identical named subscriptions, regardless of whether `UpdateExisting` is true or false, will be a no-op.

examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -0,0 +1,122 @@
...
+            var subscriptions = realm.Subscriptions;
+            // :code-block-end:
+
+            // :code-block-start: update-subscriptions
+            subscriptions.Update(() =>
+            {
+                // subscribe to all long running tasks, and give the subscription the name 'longRunningTasksSubscription'
+                var longRunningTasksQuery = realm.All<MyTask>().Where(t => t.Status == "completed" && t.ProgressMinutes > 120 ); 
+                var longRunningTasksSubscriptionOptions = new SubscriptionOptions() { Name = "longRunningTasksSubscription" };
+                var longRunningTasksSubscription = subscriptions.Add(longRunningTasksQuery, longRunningTasksSubscriptionOptions);
```
nirinchev, line 35: This can be simplified with the C# 10 syntax to:

```suggestion
               subscriptions.Add(longRunningTasksQuery, new() { Name = "longRunningTasks" });
```

There's no need to include `subscription` in the name, as it's fairly obvious that everything in the subscriptions collection is a subscription. There's also no need to assign the return value of `.Add()` if you're not going to use it.

examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                // subscribe to all of Ben's Task objects
+                var bensTasks = subscriptions.Add(realm.All<MyTask>().Where(t => t.Owner == "Ben"));
+
+                // subscribe to all Teams, and give the subscription the name 'teamsSubscription' and throw an error if a new query is added to the team subscription
+                var teamsSubscriptionOptions = new SubscriptionOptions() { Name = "teamsSubscription", UpdateExisting = false };
+                var teamsSubscription = subscriptions.Add(realm.All<Team>(), teamsSubscriptionOptions);
```
nirinchev, line 42:
```suggestion
                var teamsSubscriptionOptions = new SubscriptionOptions() { Name = "teamsSubscription", UpdateExisting = false };
                subscriptions.Add(realm.All<Team>(), new() { Name = "teams", UpdateExisting = false }); 
```

examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -0,0 +1,122 @@
...
+            // :code-block-start: update-subscriptions
+            subscriptions.Update(() =>
+            {
+                // subscribe to all long running tasks, and give the subscription the name 'longRunningTasksSubscription'
+                var longRunningTasksQuery = realm.All<MyTask>().Where(t => t.Status == "completed" && t.ProgressMinutes > 120 ); 
+                var longRunningTasksSubscriptionOptions = new SubscriptionOptions() { Name = "longRunningTasksSubscription" };
+                var longRunningTasksSubscription = subscriptions.Add(longRunningTasksQuery, longRunningTasksSubscriptionOptions);
+
+                // subscribe to all of Ben's Task objects
+                var bensTasks = subscriptions.Add(realm.All<MyTask>().Where(t => t.Owner == "Ben"));
```
nirinchev, line 38:
```suggestion
               subscriptions.Add(realm.All<MyTask>().Where(t => t.Owner == "Ben"));
```

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+you set ``UpdateExisting`` to false, adding a subscription with an existing name
+but a different query throws an exception. ``UpdateExisting`` defaults to false.
+
+You can add a single query, or batch multiple queries within a ``SubscriptionSet.Update``
+transaction. In the example below, we subscribe to three queries.
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.update-subscriptions.cs
+   :language: csharp
+
+Wait for Subscription Changes to Sync
```
nirinchev, line 119: Can we group these sections logically - I'd prefer to have Add, Update, Remove one after another and then Wait to wrap them up.

examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                // subscribe to all Teams, and give the subscription the name 'teamsSubscription' and throw an error if a new query is added to the team subscription
+                var teamsSubscriptionOptions = new SubscriptionOptions() { Name = "teamsSubscription", UpdateExisting = false };
+                var teamsSubscription = subscriptions.Add(realm.All<Team>(), teamsSubscriptionOptions); 
+            });
+            // :code-block-end:
+
+
+            // :code-block-start: wait-for-synchronization
+            // Wait for the server to acknowledge the subscription and return all objects
```
nirinchev, line 48:
```suggestion
            // Wait for the server to acknowledge the subscription change and return all objects
```

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Writing an update to the subscription set locally is only one part 
+of changing a subscription. After the local subscription change, the realm 
+synchronizes with the server to resolve any updates to the data due to 
+the subscription change. This could mean adding or removing data from the 
+synced realm. 
+
+Use the ``SubscriptionSet.WaitForSynchronizationAsync()`` method to wait for the server to
```
nirinchev, line 127: This section should also explain that an exception will be thrown if the server rejects the change.

examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -0,0 +1,122 @@
...
+            // matching the query
+            await subscriptions.WaitForSynchronizationAsync();
+            // :code-block-end:
+
+            // :code-block-start: update-a-subscription
+            subscriptions.Update(() =>
+            {
+                var updatedLongRunningTasksQuery = realm.All<MyTask>().Where(t => t.Status == "completed" && t.ProgressMinutes > 130);
+                var longRunningTasksSubscriptionOptions = new SubscriptionOptions() { Name = "longRunningTasksSubscription" };
+                var longRunningTasksSubscription = subscriptions.Add(updatedLongRunningTasksQuery, longRunningTasksSubscriptionOptions);
```
nirinchev, line 58:
```suggestion
                var updatedLongRunningTasksQuery = realm.All<MyTask>().Where(t => t.Status == "completed" && t.ProgressMinutes > 130);
                subscriptions.Add(updatedLongRunningTasksQuery, new() { Name = "longRunningTasks" });
```

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+acknowledge this set of subscriptions.
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.wait-for-synchronization.cs
+   :language: csharp
+
+
+Update Subscriptions with a New Query
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+You can update a named subscription with a new query. To update a subscriptions
```
nirinchev, line 136: We should probably explicitly state that updating unnamed subscriptions is not possible and the only way to do it is to remove the previous one and add the new one.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+.. note:: Realm .NET SDK Version Requirement
+
+   In addition to the requirements above, you must use :github:`Realm .NET version
+   10.9.0 <realm/realm-dotnet/releases>` or greater in order to use Flexible Sync
+   in your .NET client application.
+
+.. _dotnet-sync-subscribe-to-queryable-fields:
+
+Subscribe to Queryable Fields
```
nirinchev, line 46: The wording of this sections is odd - you don't subscribe to fields, you subscribe to queries.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+.. note::
+
+   Attempting to update a subscription that has the
+   ``SubscriptionOptions.UpdateExisting`` field set to false will throw an exception.
+
+Remove Subscriptions
+~~~~~~~~~~~~~~~~~~~~
+To remove subscriptions from the subscription set, you can:
+
+- Remove a single subscription with the given query
```
nirinchev, line 158: You can also remove a specific subscription if you have a reference to it - either by storing the returned value from `.Add` or by looking it up in the collection.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+- Remove all subscriptions
+
+When you remove a subscription query, the server also removes synced data from
+the client device.
+
+
+Remove a Subscription by Query
+``````````````````````````````
+You can remove a specific subscription by query by executing a transaction on
+the subscriptions set. Pass the query to the ``remove()`` method on the mutable
```
nirinchev, line 170:
```suggestion
the subscriptions set. Pass the query to the ``Remove()`` method on the mutable
```

nirinchev, line 170: There's no mutable subscription instance in .NET.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.remove-subscription-by-query.cs
+   :language: csharp
+
+
+Remove a Subscription by Name
+`````````````````````````````
+To remove a specific subscription by name, execute a transaction on
+the subscriptions set. Within the transaction, pass the name to the
+``removeByName()`` method on the mutable subscriptions instance.
```
nirinchev, line 184: Need to update the method name and again, no mutable subs in .NET.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.remove-subscription-by-name.cs
+   :language: csharp
+
+
+Remove All Subscriptions of an Object Type
```
nirinchev, line 190: We should note that `RemoveAll` have an optional `removeNamed` argument which controls whether named subscriptions will be removed or only the anonymous ones.

nirinchev (changes requested):
Apart from the comments, there is some information missing:
1. Examples iterating the subscriptions collection.
2. Examples of using `realm.Subscriptions.Find` to look up a subscription by query or name. This can be used in combination with update - e.g.:
    ```csharp
    if (realm.Subscriptions.Find("teams") == null)
    {
        // We don't have the teams data yet, subscribe and wait for the server to send it
        realm.Subscriptions.Update(() =>
        {
             realm.Subscriptions.Add(realm.All<Team>(), new() { Name = "teams" });
        });
        await realm.Subscriptions.WaitForSynchronizationAsync();
    }

    // At this point we have teams locally - either because the server had previously
    // sent the data, or because we just subscribed and waited for it.
    ```
3. Examples of error handling - rejected subscription are going to be fairly common early on, so we should show users how to handle the exceptions from `WaitForSynchronizationAsync` or access the subscription state and error properties.
4. Explanation of the different states the subscription set can be in.

examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -0,0 +1,122 @@
...
+            var realm = Realm.GetInstance(config);
+            // :code-block-end:
+    
+
+            // :code-block-start: get-subscriptions
+            var subscriptions = realm.Subscriptions;
+            // :code-block-end:
+
+            // :code-block-start: update-subscriptions
+            subscriptions.Update(() =>
```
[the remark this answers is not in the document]
mongomoe: Makes sense, I've replaced the uses of `subscriptions.{someAPI}()` with `realm.Subscriptions.{someAPI}()` everywhere

examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -0,0 +1,122 @@
...
+            // :code-block-start: remove-all-subscriptions-of-object-type
+            subscriptions.RemoveAll("Team");
+            // :code-block-end:
+
+            // :code-block-start: remove-all-subscriptions
+            subscriptions.RemoveAll();
+            // :code-block-end:
+        }
+    }
+    class MyTask : RealmObject
```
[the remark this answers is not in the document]
mongomoe: 👍

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+- :ref:`Add subscriptions to the client application <dotnet-sync-subscribe-to-queryable-fields>`
+
+You can add, update, and remove query subscriptions to determine which data 
+syncs to the client device.
+
+.. include:: /includes/note-flexible-sync-preview.rst
+
+.. include:: /includes/note-flexible-sync-prerequisite.rst
+
+.. note:: Realm .NET SDK Version Requirement
```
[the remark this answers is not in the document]
mongomoe: Yeah the 3 warnings is a bit ugly for me as well, but the 3 separate warnings are used in different places, so changing 1 would have ramifications on multiple pages.

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: On the backend, you select up to 10 queryable fields (which is what the following sentences indicate: `"When configuring Flexible Sync on the backend, you specify which fields your client application can query. In the client application, use the subscriptions API to manage a set of subscriptions to specific queries on queryable fields."`)  

I've taken this format from the other SDKs as well: 
- https://docs.mongodb.com/realm/sdk/swift/examples/flexible-sync/#subscribe-to-queryable-fields
- https://docs.mongodb.com/realm/sdk/node/examples/flexible-sync/#subscribe-to-queryable-fields
- https://docs.mongodb.com/realm/sdk/android/examples/flexible-sync/#subscribe-to-queryable-fields 

which all use "Subscribe to Queryable Fields". Let me know if you still want me to change this (if that's the case we should also consider changing those other pages)

nirinchev: Yeah, I think we should change the title here because it's ambiguous - it implies an API where you subscribe to a field rather than subscribe to a query on that field. The last sentence of that quote is actually fairly precise - you "manage a set of subscriptions to [...] queries".

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+- Get a list of all subscriptions
+- Add subscriptions
+- Check subscription state
+- Update a subscription with a new query
+- Remove individual subscriptions or all subscriptions of a type
+
+When data matches the subscription and has appropriate permissions, it syncs
+between devices and the backend application.
+
+A Realm subscription in .NET has:
```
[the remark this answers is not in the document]
mongomoe: No, you don't necessarily need to provide the following fields, this is just a description of the Subscription Class. Although you can set the name & query field through the subscriptions.add method. I'd be happy to remove this though if you want, I'm not in love w/ this bit of the page.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+- A created and updated date
```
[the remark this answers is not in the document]
mongomoe: Exactly, when the subscription was created and updated

source/sdk/dotnet/examples/flexible-sync.txt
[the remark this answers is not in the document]
mongomoe: Update: I just removed this bit altogether. Originally, I had it to parity the swift version, but it looks like Dachary's removed it from hers as well, so it makes sense here to not have it as well.

source/sdk/dotnet/examples/flexible-sync.txt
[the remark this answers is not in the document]
mongomoe: Resolving, see: https://github.com/mongodb/docs-realm/pull/1578/files#r796021477

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+- An object type
+- A query
+
+You can specify a string name for your subscription. If you do not give your
+subscription a name, the name is set to null.
+
+When you create a subscription, Realm looks for data matching a query on a
+specific object type. In your Flexible Sync subscriptions, you can have
+subscriptions on several different object types or several queries on the same
+object type.
```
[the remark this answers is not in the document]
mongomoe: In the "Add a Query to the List Of Subscriptions" subsection below, there's an example of subscription to 2 different object types (task & team) & several queries on a single object type.

source/sdk/dotnet/examples/flexible-sync.txt
```diff
@@ -0,0 +1,207 @@
...
+
+
+Get Subscriptions
+~~~~~~~~~~~~~~~~~
+When using a flexible synced realm, you can access a ``SubscriptionSet``, a collection of subscriptions, through the
+``Realm.Subscriptions`` property.
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.get-subscriptions.cs
```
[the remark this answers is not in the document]
mongomoe: Can you try hard refreshing it? It should be there. Cmd+shift+r

Screen Shot 2022-01-31 at 3 13 36 PM

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: Sorry, that's leftover from when I copied the JS content. I've changed it now to "You can use the subscription set to  ..."

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: I've removed the word 'transaction' now. The new sentence reads: `"You must perform all mutations on the subscriptions set within an update block.To create an update block, call the SubscriptionSet.Update() method"`

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: Sure, how would you recommend saying that? Maybe: "We recommend batching queries together to decrease server cost"?

nirinchev: We can be even more forceful here. "Performing query updates is an expensive operation on the server, so it is strongly advised that you design your app in a way that minimizes updates - for example by creating all subscriptions in a single `Update` block the first time the user launches the app and batching any follow-up changes to the subscription set."

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: I'm going to roughly keep it as is, but change "writing an update to the subscription set" to "mutating the subscription set". The version I had is taken from the swift/node versions + is a bit more accurate than your suggestion (for instance, in the fragment: 'Realm Sync resolves data changes' , there's actually no changes to the data in the realm, but rather what data is being synchronized due to the mutated subscription).

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: Added a note explicitly stating that 👍

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: How about as a title: "Manage Your Subscriptions"? Which is broad enough that it encompasses all of the examples

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: Added the following note, let me know if the wording is okay for you:

```

.. note:: {+service-short+} Automatically Removes Duplicate Subscriptions 

   {+service-short+} will automatically remove duplicate unnamed subscriptions
   with the same query. Also, adding multiple identical named subscriptions,
   regardless of the value of ``UpdateExisting``, is a :wikipedia:`no-op <NOP_(code)>`.
   
 ```

File: source/sdk/dotnet/examples/flexible-sync.txt (hunk 9 above)
Comment (mongomoe) [reply source/sdk/dotnet/examples/flexible-sync.txt]: Thanks! I've took that suggestion, but just altered the wording slightly to be a bit more readable:

"You can add a single query, or batch multiple queries within a ``SubscriptionSet.Update``
block. Performing query updates is an expensive operation on the server. We
strongly advise designing your application to minimize updates. You can do this
by creating all subscriptions in a single update block the first time the user
launches the app and batching any follow-up changes to the subscription set."

File: source/sdk/dotnet/examples/flexible-sync.txt (hunk 16 above)
Comment (mongomoe) [reply source/sdk/dotnet/examples/flexible-sync.txt]: Added the sentence: "If the server rejects the change, the ``SubscriptionSetState`` will be an error state, and
an exception will be thrown."

File: source/sdk/dotnet/examples/flexible-sync.txt [hunk 31]
```diff
@@ -0,0 +1,207 @@
...
+
+.. literalinclude:: /examples/generated/dotnet-flexible-sync/FlexibleSyncExamples.codeblock.remove-all-subscriptions-of-object-type.cs
+   :language: csharp
+
+
+Remove All Subscriptions
+````````````````````````
+To remove all subscriptions from the subscriptions set, execute a transaction on
+the subscriptions set. Call the ``removeAll()`` method
+on the mutable subscriptions instance within the transaction
```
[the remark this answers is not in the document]
mongomoe: Sorry for all these wrong name + mutable subs, a copy & paste error from when I ported over the JS SDK content. It should be fixed now, everywhere.

source/sdk/dotnet/examples/flexible-sync.txt
mongomoe: Changed the sentence to "The ``Add()`` method takes a ``query``, and an optional subscription options object" 

The sentence "If you do not give your subscription a name, the name is set to null." is already in the beginning of the page (under the section currently titled "Subscribe to Queryable Fields")

nirinchev (approved):
I think we can merge this (after updating the `Subscribe to Queryable Fields` section wording) and file tickets for the follow-up work, some of which probably needs to be done across all SDK pages. Off the top of my head, here is what's missing:
1. Explanation of the fields in the `Subscription` class. Those do have API docs, but it would still be a good idea to highlight the key ones.
2. Looking up a subscription by name/query.
3. Removing a subscription directly (i.e. after looking it up).
4. Explanation of the subscription set states.
5. Replace the obsoleted `SyncConfiguration` with `PartitionSyncConfiguration` across the docs.

Thanks for moving this forward and I'm sorry I've been a PITA during the review. We'll try to figure out a process in engineering where a more thorough eng review will be done for major new features for all SDKs to avoid ending up in a situation where copy-pasting some already approved content results in tens of comments and a week of back-and-forth.

source/sdk/dotnet/examples/flexible-sync.txt
MongoCaleb: I'd avoid using any term that requires a link to Wikipedia -- suggests that many/most of our users won't understand. 
Isn't this really saying "If you add a named subscription with the same name, {+service-short+} will ignore it"?

MongoCaleb (approved):
LGMT % my comment about using a term like no-op.

PR #1597, opened 2022-01-28
(DOCSP-19895): Add flexible sync config object and definitions

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19895

- [Realm Sync Configuration Files](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19895/manage-apps/configure/config/sync/)

### Review Guidelines

### DIFF
--- a/source/manage-apps/configure/config/sync.txt
+++ b/source/manage-apps/configure/config/sync.txt
@@ -21,8 +21,13 @@ directory:
    └── sync/
        └── config.json
 
-Configuration
--------------
+.. _partition-based-sync-config-object:
+
+Partition-Based Sync Configuration
+----------------------------------
+
+When you use :ref:`Partition-Based Sync <partition-based-sync>`, your Realm 
+app uses this Sync configuration:
 
 .. code-block:: json
    :caption: sync/config.json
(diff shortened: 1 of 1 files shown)


source/manage-apps/configure/config/sync.txt
```diff
@@ -125,3 +130,188 @@ Configuration
...
+       user has permission to read objects in the collection. If the 
+       expression evaluates to ``false``, {+service-short+} does not allow 
+       the user to read an object or its properties.
+
+   * - | ``rules.<Coll>.write``
+       | JSON Expression
+     - An :ref:`expression <expressions>` that evaluates to ``true`` when a 
+       user has permission to write objects in the collection. If the 
+       expression evaluates to ``false``, {+service-short+} does not allow 
+       the user to modify an object or its properties.
```
cbush, line 257: Question: Does "write" permission imply "read" permission or can they be kept separate?

cbush (approved):

source/manage-apps/configure/config/sync.txt
```diff
@@ -125,3 +130,188 @@ Configuration
...
+   * - | ``development_mode_enabled``
+       | Boolean
+     - If ``true``, :ref:`development mode <enable-development-mode>` is enabled
+       for the application. While enabled, Realm automatically stores synced
+       objects in a specific database (specified in ``database_name``) and
+       mirrors objects types in that database's collection schemas.
+       
+       .. important::
+          
+          Realm does not enforce rules when development mode is enabled.
```
cbush, line 201: Is this actually true? There used to be a bug where the UI did not show the permissions configuration when development mode was enabled, but you could still set permissions. Or am I misremembering?

source/manage-apps/configure/config/sync.txt
```diff
@@ -125,3 +130,188 @@ Configuration
...
+---------------------------
+
+When you use :ref:`Flexible Sync <flexible-sync>`, your Realm app uses this 
+Sync configuration:
+
+.. code-block:: json
+   :caption: sync/config.json
+
+   {
+     "type": <"flexible" | "partition">,
```
nlarew, line 146: Given that this is specifically a config for flex sync, should we change this to be `"type": "flexible"` and also add `"type": "partition"` to the config in the other section?

source/manage-apps/configure/config/sync.txt
dacharyc: Seems this is not true, so I'm removing it from here and will make a separate ticket to remove it from elsewhere in our docs where we state this.

source/manage-apps/configure/config/sync.txt
dacharyc: Currently, this isn't _actually_ implemented and the partition-based sync config has not yet changed. I've taken your suggestion here about changing to `"type": "flexible"`, and will make a separate update when the PBS sync config changes.

source/manage-apps/configure/config/sync.txt
dacharyc: Write does imply read, but that verbiage wasn't here already for Partition-Based Sync so I didn't add it for Flexible Sync. It's probably good to note it here, though, so I'll add that - thanks for the suggestion!

PR #1600, opened 2022-01-28
(DOCSP-20528): global explanation of client reset strategies

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-20528

- [Client Resets](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-20528/sync/error-handling/client-resets/)

### Review Guidelines

### DIFF
--- a/source/sync/error-handling/client-resets.txt
+++ b/source/sync/error-handling/client-resets.txt
@@ -33,35 +33,105 @@ synced realm file has an older version than the client's realm file. The most
 common causes of client resets are:
 
 - **Realm server crashes**: A server may restore from a backup that has an earlier
-  version. An automatic client reset for this scenario would reset the client to
+  version. An client reset for this scenario would reset the client to
   the earlier version and lose any changes that were saved on the client but not
   yet synced with the server.
+
+- **Disabling and re-enabling sync**: Turning Sync off, then on again in
+  the Realm UI causes all clients to perform a client reset.
+
+- **Client/Backend schema mismatch**: If a client application attempts
+  to synchronize with the backend while using a Realm Object schema that
+  does not exist in the backend, that client must perform a client reset.
+  This only applies in one direction: if the backend schema contains a
+  Realm Object class not used in a client, that client will *not*
+  experience a client reset.
+
+- **Client maximum offline time**: When a client has not synchronized with
+  the backend in more than :ref:`client maximum offline time
+  <client-maximum-offline-time>` days, that client can no longer synchronize
+  unsynced local changes with the backend. The client must discard all
+  local changes since the last sync and download the current status of
+  the realm from the backend.
 
 - **Destructive schema changes**: A :ref:`destructive change
   <destructive-changes-synced-schema>`, like changing a property type or a
   primary key, requires you to terminate and re-enable sync. This creates a new
-  synced realm file with a version unrelated to the client's file.
-
-- Clients exceeding the :ref:`client maximum offline time
-  <client-maximum-offline-time>`.
+  synced realm file with a version unrelated to the client's file. Because
+  this scenario requires a new realm file, you can only handle client resets
+  caused by destructive changes using the :ref:`manually recover unsynced changes
+  <>` strategy.
 
 .. _handle-a-client-reset:
 
 Handle a Client Reset
 ---------------------
 
-The SDKs can detect and perform automatic client resets on application startup.
-They also expose an interface for you to add custom recovery code where you can
-triage and save local data before the reset occurs.
+The SDKs detect the need for client resets on application startup. You
+can handle client resets in your application using one of the available
+**client reset strategies**:
 
-.. important::
-   
-   Automatic client resets lose any changes saved in the client's realm file
-   since the client last successfully synced. To preserve these changes, you
-   must write a custom client reset error handler.
-   
-   To see how, check out a client reset usage example for your SDK:
-   
+.. _discard-unsynced-changes:
+
+Discard Unsynced Changes
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The **discard unsynced changes** strategy automatically handles client
+reset events without the need for app-specific logic. During a client
+reset that uses this strategy, the client:
+
+1. Downloads a fresh copy of the realm from the backend.
+2. Uses a diff to bring the original (local) copy of the
+   realm to the same state as the fresh copy.
+3. Discards the fresh copy. The app continues to use
+   the original copy of the realm with the diff applied.
+
+This allows users to perform a client reset without closing the original
+realm or missing notifications on new changes.
+
+.. warning:: Discard Unsynced Changes Deletes Unsynced Local Changes Permanently
+
+   The "discard unsynced changes" client reset strategy throws away any
+   unsynced local changes. Do not use this client reset strategy if your
+   application needs to preserve these unsynced changes.
+
+You can use this strategy to resolve a client reset caused by anything
+*except* a destructive schema change.
+
+.. _manually-recover-unsynced-changes:
+
+Manually Recover Unsynced Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The **manually recover unsynced changes** client reset strategy gives
+the app developer complete control over handling client reset events.
+With this client reset strategy, you can choose to either:
+
+1. Initiate the client reset immediately upon receiving a client reset error,
+   after closing all realms.
+2. To wait until all realms have been closed and the app connection is
+   re-established. The easiest way to do this is restarting the app.
+
+During a client reset that uses this strategy, the client:
+
+1. Makes a backup of the original (local) copy of the realm.
+2. Downloads a fresh copy of the realm from the backend. This fresh copy
+   replaces the original local realm.
+3. Provides a reference to the local realm backup and the fresh copy of
+   the realm.
+
+With the local realm backup and the fresh copy in hand, you can triage
+and save local data by copying it from the local realm backup to the
+fresh copy of the realm. Recovering unsynced changes is highly dependent
+on your schema and application logic, so you may need to supplement
+your schema with "last modified" timestamps for best results.
+
+Examples
+~~~~~~~~
+
+For more information on performing a client reset, check out the client
+reset examples for your SDK:
+
    - :ref:`Android <android-client-resets>`
    - :ref:`iOS, macOS, tvOS, and watchOS <ios-client-resets>`
    - :ref:`.NET <dotnet-client-resets>`
--- a/source/sync/error-handling/client-resets.txt
+++ b/source/sync/error-handling/client-resets.txt
@@ -32,7 +32,7 @@ Client reset scenarios occur when the server's history is incompatible
 with the client's history. The most common causes of client resets are:
 
 - **Realm server crashes**: A server may restore from a backup that has an earlier
-  version. An client reset for this scenario would reset the client to
+  version. A client reset for this scenario would reset the client to
   the earlier version and lose any changes that were saved on the client but not
   yet synced with the server.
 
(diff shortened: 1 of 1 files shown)


source/sync/error-handling/client-resets.txt
```diff
@@ -33,35 +33,105 @@ synced realm file has an older version than the client's realm file. The most
...
 
 .. _handle-a-client-reset:
 
 Handle a Client Reset
 ---------------------
 
-The SDKs can detect and perform automatic client resets on application startup.
-They also expose an interface for you to add custom recovery code where you can
-triage and save local data before the reset occurs.
+The SDKs detect the need for client resets on application startup. You
```
ironage, line 70: Maybe not super important to call out here, but a client reset error may also occur at any time after application startup. Eg. if sync is disabled and enabled on the server while the app is running. Application startup would be the most common time that it happens though.

ironage (commented):
> Client reset scenarios are generally rare and occur when the server's copy of a
synced realm file has an older version than the client's realm file.

Not sure what "copy" is referring to here. To be precise, client resets occur when the server's history is incompatible with the client's history.

source/sync/error-handling/client-resets.txt
nathan-contino-mongo: Good call on both of these -- the first sentence was a bit of a holdover from the OG Realm days, when Realm Cloud essentially was just another client with another copy of the realm, and this was just a logical error on my part. Corrected both!

ironage (approved):
🎉

ironage (approved):

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
 
-Client reset scenarios are generally rare and occur when the server's copy of a
-synced realm file has an older version than the client's realm file. The most
-common causes of client resets are:
+Client reset scenarios occur when the server's history is incompatible
+with the client's history. The most common causes of client resets are:
 
 - **Realm server crashes**: A server may restore from a backup that has an earlier
-  version. An automatic client reset for this scenario would reset the client to
+  version. An client reset for this scenario would reset the client to
```
cbush, line 35:
```suggestion
  version. A client reset for this scenario would reset the client to
```

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
   the earlier version and lose any changes that were saved on the client but not
   yet synced with the server.
 
+- **Disabling and re-enabling sync**: Turning Sync off, then on again in
```
cbush, line 39:
```suggestion
- **Disabling and re-enabling sync**: Turning Sync off then on again in
```

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
+  the Realm UI causes all clients to perform a client reset.
```
cbush, line 40:
```suggestion
  the Realm UI causes all clients to require a client reset.
```

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
+
+- **Client/Backend schema mismatch**: If a client application attempts
+  to synchronize with the backend while using a Realm Object schema that
+  does not exist in the backend, that client must perform a client reset.
+  This only applies in one direction: if the backend schema contains a
+  Realm Object class not used in a client, that client will *not*
+  experience a client reset.
```
cbush, line 47: I think we should be consistent about how "client reset" is presented as a concept. Is it something a client "experiences" or is it something a client "needs to do"?
```suggestion
  need to client reset.
```

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
+
+- **Client maximum offline time**: When a client has not synchronized with
+  the backend in more than :ref:`client maximum offline time
+  <client-maximum-offline-time>` days, that client can no longer synchronize
```
cbush, line 51: Should this be a ref or a source constant?

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
+
+The **discard unsynced changes** strategy automatically handles client
+reset events without the need for app-specific logic. During a client
+reset that uses this strategy, the client:
+
+1. Downloads a fresh copy of the realm from the backend.
+2. Uses a diff to bring the original (local) copy of the
+   realm to the same state as the fresh copy.
+3. Discards the fresh copy. The app continues to use
+   the original copy of the realm with the diff applied.
```
cbush, line 86: I don't understand this at all. The title implies that you are losing local changes, but this talks about applying the diff between the local realm and the latest on the backend. So what are the "unsynced changes" that are being discarded?

Also, does the user actually need to write anything to do this or can this be done automatically by realm? Either way, it needs more clarity because it would be frustrating to read that this is how it's supposed to go, but not have any guidance as to where to actually do these operations.

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The **manually recover unsynced changes** client reset strategy gives
+the app developer complete control over handling client reset events.
+With this client reset strategy, you can choose to either:
+
+1. Initiate the client reset immediately upon receiving a client reset error,
+   after closing all realms.
+2. Wait until all realms have been closed and the app connection is
+   re-established. To easily do this, restart the app.
```
cbush, line 112:
```suggestion
   re-established. To do this, restart the app.
```

source/sync/error-handling/client-resets.txt
```diff
@@ -28,42 +28,111 @@ until the client reset is complete.
...
+
+During a client reset that uses this strategy, the client:
+
+1. Makes a backup of the original (local) copy of the realm.
+2. Downloads a fresh copy of the realm from the backend. This fresh copy
+   replaces the original local realm.
+3. Provides a reference to the local realm backup and the fresh copy of
```
cbush, line 119: I think I would feel better about these strategy descriptions if they were more active guides than passive descriptions. "The user does this..." feels like it could be incredibly frustrating to read. Having the SDK links below the fold also doesn't help, but I know there's not much we can do with how the site is structured. I would rather read something like: "You can discard the local changes or manually recover the changes. To manually recover, do this, this, and this. *link to SDKs*".

cbush (changes requested):
It's a great page but I think the strategy section still needs some clarification.

source/sync/error-handling/client-resets.txt
nathan-contino-mongo: The whole point is that this happens automatically, without the user doing anything. AND it uses the same realm file as before the client reset -- previously, client resets always created a new file, which meant at the very least you had to close all open realms, close the app connection, etc. Usually it was safer to just restart the app and handle the reset, and then most likely restart it again. With this approach you just let the client reset "happen" to your realm, and you can leave realms open the whole time, see changes happen via notifications, etc.

I'll see if I can clarify this more because this was a big sticking point for me when I first heard about this originally. Took me a while to grok the whole thing, and if that's happening for you, too, it's a sign that this is both a) complicated and b) i'm not explaining it quite enough.

source/sync/error-handling/client-resets.txt
nathan-contino-mongo: Personally I feel this is fine as a ref -- you kinda need to provide a link for users to understand the concept. I tried playing around with a hovertext but it's just too complex a concept to explain in one or two sentences. And I don't think our users are trained very well to use hovertexts.

source/sync/error-handling/client-resets.txt
nathan-contino-mongo: This is meant to explain the difference between the strategy implementations so you can choose which you prefer from an informed perspective -- as your other comment demonstrates, there are some gaps in understanding the differences right now so I don't think we can just "do away" with the internals explanation. I'm revamping the explanations right now, so maybe we can revisit this afterwards. I do think we should provide links to each SDK's explanation in each section, instead of just at the end, but since none of the SDKs actually demonstrate this right now I can't do that yet. Another problem with SDKs rolling out features on completely separate schedules!

source/sync/error-handling/client-resets.txt
```diff
@@ -28,42 +28,133 @@ until the client reset is complete.
...
-   - :ref:`Android <android-client-resets>`
-   - :ref:`iOS, macOS, tvOS, and watchOS <ios-client-resets>`
-   - :ref:`.NET <dotnet-client-resets>`
-   - :ref:`React Native <react-native-client-resets>`
-   - :ref:`Node.js <node-client-resets>`
+.. note:: Destructive Changes Require Manual Recovery
+
+   You can only handle a client reset triggered by destructive changes
+   using the :ref:`manually recover unsynced changes
+  <manually-recover-unsynced-changes>` strategy.
```
cbush:
```suggestion
   <manually-recover-unsynced-changes>` strategy.
```

source/sync/error-handling/client-resets.txt
```diff
@@ -28,42 +28,153 @@ until the client reset is complete.
...
+reset events without the need for app-specific logic. During a client
+reset that uses this strategy, the client:
+
+1. Downloads a fresh copy of the realm from the backend.
+#. Performs a diff to figure out the set of steps required to
+   bring the original (local) copy of the realm to the same state as
+   the fresh copy of the backend.
+#. Applies that set of steps to transform the local realm into a state
+   where it can sync with the backend. This includes deleting all local
+   changes that haven't yet synced with the backend, as well as any
```
cbush, line 100: It could be read as "includes deleting all X as well as Y" which makes it sound like you're deleting the inserts and updates. I would just split the sentences. And reverse them. "Steps to transform... It applies all inserts, updates, and deletes that happened on the backend that haven't yet synced to the client. It also deletes local changes that haven't synced to the backend."

cbush (approved):

source/sync/error-handling/client-resets.txt
nathan-contino-mongo: Done!

source/sync/error-handling/client-resets.txt
nlarew, line 51: "more than client maximum offline time days" doesn't read great to me

maybe: "more than the number of day specified as your client maximum offline time"

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
   <destructive-changes-synced-schema>`, like changing a property type or a
   primary key, requires you to terminate and re-enable sync. This creates a new
-  synced realm file with a version unrelated to the client's file.
-
-- Clients exceeding the :ref:`client maximum offline time
-  <client-maximum-offline-time>`.
+  synced realm file with a version unrelated to the client's file. Because
+  this scenario requires a new realm file, you can only handle client resets
+  caused by destructive changes using the :ref:`manually recover unsynced changes
+  <>` strategy.
```
nlarew:
```suggestion
  <manually-recover-unsynced-changes>` strategy.
```

source/sync/error-handling/client-resets.txt
```diff
@@ -28,40 +28,109 @@ until the client reset is complete.
...
+
+1. Makes a backup of the original (local) copy of the realm.
+2. Downloads a fresh copy of the realm from the backend. This fresh copy
+   replaces the original local realm.
+3. Provides a reference to the local realm backup and the fresh copy of
+   the realm.
+
+With the local realm backup and the fresh copy in hand, you can triage
+and save local data by copying it from the local realm backup to the
+fresh copy of the realm. Recovering unsynced changes is highly dependent
```
nlarew, line 154: This first sentence seems like it could be a 4th step maybe?

PR #1610, opened 2022-01-31
(DOCSP-19838): Document translator fatal error / ChangeStreamHistoryLost

cbush:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19838

- [Sync Errors](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/fatal-sync-error/sync/error-handling/errors/)

### Review Guidelines

### DIFF
--- a/source/sync/error-handling/errors.txt
+++ b/source/sync/error-handling/errors.txt
@@ -125,16 +125,23 @@ and MongoDB Atlas.
 
    * - TranslatorFatalError - ChangeStreamHistoryLost
      - This error occurs when old entries in the :manual:`oplog
-       </core/replica-set-oplog/>` have been erased but the server-side
-       "translator" process still needed them to bring the MongoDB
-       cluster and the Realm object server to an equivalent state. This
-       can happen when Sync is paused for too long. The free tier has a
-       shared oplog and is therefore particularly susceptible to this
-       error. Dropping a collection or the MongoDB cluster being
-       unreachable for too long could also cause the translator to give
-       up, which could eventually result in this error. The only
-       resolution is to :ref:`terminate <terminating-realm-sync>` and
-       :ref:`re-enable Sync <re-enable-realm-sync>`.
+       </core/replica-set-oplog/>` have expired before the server-side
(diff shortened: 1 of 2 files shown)


source/sync/error-handling/errors.txt
```diff
@@ -123,6 +123,19 @@ and MongoDB Atlas.
        primary key value. Alternatively, ensure that every synced object has a
        unique primary key, even across partitions.
 
+   * - TranslatorFatalError - ChangeStreamHistoryLost
+     - This error occurs when old entries in the :manual:`oplog
+       </core/replica-set-oplog/>` have been erased but the server-side
+       "translator" process still needed them to bring the MongoDB
```
dacharyc, line 129: I'm finding "still needed them" a little difficult to parse here. What about something like:

> but the server-side "translator" process _was still working through them_ to bring the MongoDB cluster and the Realm object server to an equivalent state.

source/sync/error-handling/errors.txt
```diff
@@ -123,6 +123,19 @@ and MongoDB Atlas.
...
+       cluster and the Realm object server to an equivalent state. This
```
dacharyc, line 130: These next few lines are kind of a wall of text. What about breaking it up into bullet points for readability? Something like:

This can happen when:

- Sync is paused long enough for the entries to fall off the oplog
- You drop a collection
- The MongoDB cluster becomes unreachable for too long

Because the free tier has a shared oplog, it is more vulnerable to this error.

To resolve this error, :ref:`terminate <terminating-realm-sync>` and :ref:`re-enable Sync <re-enable-realm-sync>`.

dacharyc (approved):
LGTM, but I had to stop and think about the wording or really focus to parse some of the text, so I might suggest simplifying the text a bit to improve readability. Just a couple of thoughts.

source/sync/error-handling/errors.txt
cbush: Trying something, but "still working through" is less ideal because the problem is usually the translator being suspended in some way

PR #1614, opened 2022-02-01
(DOCSP-19901): Update Sync Errors page for Flexible Sync

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19901

- [Sync Errors](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19901/sync/error-handling/errors/): Moves the Partition-Based Sync error into a new section, and adds a section for Flexible Sync errors

### Review Guidelines

### DIFF
--- a/source/sync/error-handling/errors.txt
+++ b/source/sync/error-handling/errors.txt
@@ -67,16 +67,6 @@ errors and how to handle them. {+service+} reports errors in your
        To recover from this error, perform a :ref:`client reset
        <client-resets>`.
 
-   * - ErrorIllegalRealmPath
-     - This error indicates that the client attempted to open a {+realm+} with a
-       partition value of the wrong type. For example, you might see the error
-       message "attempted to bind on illegal realm partition: expected partition
-       to have type objectId but found string".
-       
-       To recover from this error, ensure that the type of the partition value
-       used to open the realm matches the partition key type in your
-       {+sync-short+} configuration.
-
    * - ErrorPermissionDenied
      - This error occurs when a user's data access permissions are not
        sufficient for a given request. This can occur if a user attempts to open
@@ -90,6 +80,81 @@ errors and how to handle them. {+service+} reports errors in your
      - This error indicates an internal failure that is not covered by a more
        specific error. For example, this might occur when you hit the storage
        limit of a free tier Atlas cluster.
+
+.. _partition-based-sync-errors:
+
+Partition-Based Sync Errors
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following error may occur when your {+app+} uses :ref:`Partition-Based Sync 
+<partition-based-sync>`.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 70
+
+   * - ErrorIllegalRealmPath
+     - This error indicates that the client attempted to open a {+realm+} with a
+       partition value of the wrong type. For example, you might see the error
+       message "attempted to bind on illegal realm partition: expected partition
+       to have type objectId but found string".
+       
+       To recover from this error, ensure that the type of the partition value
+       used to open the realm matches the partition key type in your
+       {+sync-short+} configuration.
+
+.. _flexible-sync-errors:
+
+Flexible Sync Errors
+~~~~~~~~~~~~~~~~~~~~
+
+The following errors may occur when your {+app+} uses :ref:`Flexible Sync 
+<flexible-sync>`.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 70
+
+   * - ErrorBadQuery
+     - This error indicates the client query is invalid or malformed. This 
+       error includes a message that provides details about why the query
+       is invalid.
+
+       To recover from this error, you may need to verify the :ref:`query 
+       syntax <realm-query-language>` is correct, and that you are using 
+       :ref:`query operators that are supported on the server 
+       <flexible-sync-rql-limitations>`. Additionally, confirm you are querying
+       a :ref:`queryable field <queryable-fields>` in your Flexible Sync 
+       configuration.
+
+   * - ErrorServerPermissionsChanged
+     - This error indicates that server permissions for the file ident have changed
+       since the last time it was used.
+
+       To recover from this error, perform a :ref:`client reset
+       <client-resets>`.
+
+   * - ErrorInitialSyncNotCompleted
+     - This error indicates that the client tried to open a session before the 
+       initial sync was complete.
+
+       To resolve this error, check Sync status in the client application 
+       before attempting to read from or write to the realm. Consult your 
+       preferred SDK's Flexible Sync documentation for details.
+    
+   * - ErrorWriteNotAllowed
+     - This error occurs when a client attempts a write that is disallowed 
+       by permissions, or if the write modifies an object outside of the 
+       subscription query. 
+       
+       Consider an example where the client application opens a subscription 
+       on a ``Places`` object type with the query ``name = "NY"``. If the 
+       client then tries to insert a ``Places`` object with a field 
+       ``name: "CT"``, the write triggers an ``ErrorWriteNotAllowed`` because 
+       the object is not within the subscription query.
+
+       To recover from this error, perform a :ref:`client reset
+       <client-resets>`.
 
 .. _mongodb-translator-errors:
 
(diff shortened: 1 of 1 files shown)


source/sync/error-handling/errors.txt
```diff
@@ -91,6 +81,87 @@ errors and how to handle them. {+service+} reports errors in your
...
+<partition-based-sync>`.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 35 65
+
+   * - Error Name
+     - Description
+
+   * - ErrorIllegalRealmPath
```
nathan-contino-mongo, line 99: Personally I find this hard to read as it renders in Akzidenz -- maybe this is a valid use-case for monospace text? Or maybe the `:file:` role (like monospace, but without the silly border).

Probably not a blocker before merge, and a bit subjective. But "Ill" just jars me so much I thought I'd point it out.

source/sync/error-handling/errors.txt
```diff
@@ -91,6 +81,87 @@ errors and how to handle them. {+service+} reports errors in your
        specific error. For example, this might occur when you hit the storage
        limit of a free tier Atlas cluster.
 
+.. _partition-based-sync-errors:
+
+Partition-Based Sync Errors
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following error may occur when your {+app+} uses :ref:`Partition-Based Sync
```
nathan-contino-mongo, line 89:
```suggestion
The following errors may occur when your {+app+} uses :ref:`Partition-Based Sync 
```

Even though it's just one right now, I assume this list could grow in the future. "errors" doesn't sound wrong when we're referring to a table of one, but will sound wrong and could easily be forgotten about when we add more entries to this table.

nathan-contino-mongo: Or we could rewrite this to match the Sync Protocol intro and use "table" instead.

source/sync/error-handling/errors.txt
```diff
@@ -91,6 +81,87 @@ errors and how to handle them. {+service+} reports errors in your
...
+       To recover from this error, ensure that the type of the partition value
+       used to open the realm matches the partition key type in your
+       {+sync-short+} configuration.
+
+.. _flexible-sync-errors:
+
+Flexible Sync Errors
+~~~~~~~~~~~~~~~~~~~~
+
+The following errors may occur when your {+app+} uses :ref:`Flexible Sync
```
nathan-contino-mongo, line 114: Similar thought here -- maybe we should intro with "the following table describes"?

source/sync/error-handling/errors.txt
```diff
@@ -91,6 +81,87 @@ errors and how to handle them. {+service+} reports errors in your
...
+     - Description
+
+   * - ErrorBadQuery
+     - This error indicates the client query is invalid or malformed. This 
+       error includes a message that provides details about why the query
+       is invalid.
+
+       To recover from this error, you may need to verify the :ref:`query 
+       syntax <realm-query-language>` is correct, and that you are using 
+       :ref:`query operators that are supported on the server
```
nathan-contino-mongo, line 131:
```suggestion
       :ref:`query operators that are supported on the backend 
```
Nit: I usually use the term "backend" since Realm is serverless by nature and the term "server" sort of clashes with that idea, even if it's accurate. This applies to all usage of server on this page, honestly. Not a huge deal though.

source/sync/error-handling/errors.txt
```diff
@@ -91,6 +81,87 @@ errors and how to handle them. {+service+} reports errors in your
...
+       <flexible-sync-rql-limitations>`. Additionally, confirm you are querying
+       a :ref:`queryable field <queryable-fields>` in your Flexible Sync 
+       configuration.
+
+   * - ErrorServerPermissionsChanged
+     - This error indicates that server permissions for the file ident have changed
+       since the last time it was used.
+
+       To recover from this error, perform a :ref:`client reset
+       <client-resets>`.
```
nathan-contino-mongo, line 141: Interesting -- does this error mean that a client reset should *also* be on its way to the client? Usually you can only trigger a client reset if you get a client reset error from the backend, so it might be important to mention that here. Otherwise folks might wonder how they trigger a client reset, which could lead to some nasty hacks!

source/sync/error-handling/errors.txt
```diff
@@ -91,6 +81,87 @@ errors and how to handle them. {+service+} reports errors in your
...
+
+   * - ErrorInitialSyncNotCompleted
+     - This error indicates that the client tried to open a session before the 
+       initial sync was complete.
```
nathan-contino-mongo, line 145: Wild. Does this mean that all SDKs should default to waiting for initial sync to fully download data from the backend when opening a flexible sync realm? Or does this just mean there's an initial handshake required?

nathan-contino-mongo (approved):
LGTM with a few random nits that might inspire elaboration or small changes.

source/sync/error-handling/errors.txt
dacharyc: Hmm. I don't think monospace text would work well for this, because I already had to change the table widths just to accommodate the length of some of these error names - our monospace styling would probably make these errors so wide they'd have to be on a separate line, or the description would be comically small, or the error names themselves would have to wrap.

It's a good point, though, and I wish we could get some design option for a monospace that isn't so WIDE and renders a little better on the page.

source/sync/error-handling/errors.txt
dacharyc: Blarg, I hate the "table describes" verbiage - it's so many extra unnecessary words. Your point is valid, though, I'll make this "errors."

source/sync/error-handling/errors.txt
dacharyc: Hmm, there are a lot of instances of `server` already on the page, and even some error message names that refer to server (i.e. `ErrorServerPermissionsChanged`). I think that ship has sailed, at least on this page.

source/sync/error-handling/errors.txt
dacharyc: Yeah, this error triggers a client reset. I took the "perform a client reset" verbiage from other errors that were already on the page - i.e. ln 42 and ln 60. Neither of those errors mentions anything about triggering a client reset - just that you need to "perform a client reset" to recover from the errors. So I thought this verbiage was consistent with those.

Curious what else you'd envision for this verbiage, instead.

nathan-contino-mongo: I'd probably adjust it to "this error triggers a client reset error. Check out the [docs for your SDK] for more information on how to perform a client reset.

nathan-contino-mongo: Just to make it clearer that you don't need to handle the client reset bit manually when you see this error -- the config's client reset handler ought to handle it.

source/sync/error-handling/errors.txt
dacharyc: There's [a section about waiting for subscription changes to sync in the Swift SDK](https://docs.mongodb.com/realm/sdk/swift/examples/flexible-sync/#wait-for-subscription-changes-to-sync); the recommendation is to wait for the `onComplete` block or use the async/await method to wait for data changes to sync after changing a query. I guess we did not explicitly state to wait for data to sync when opening a Realm. Jason mentioned that waiting for data to sync was important with Flex Sync, but it might be good to get info specifically around this.

nathan-contino-mongo: Makes sense -- maybe this is mostly an iOS thing? Didn't come up in the technical review for android, but that's no guarantee it isn't an issue.

I feel like this is mostly an implementation problem, though. The user should never be able to open a realm in a way that is broken, and they certainly shouldn't have to explicitly wait for data *every time they open a flexible sync realm*. That's just boilerplate for no reason. The default config should just handle that implicitly, and maybe let you explicitly skip the wait if there's a legitimate reason for that.

dacharyc: Alright, great question! I misunderstood this error. After asking for clarification, I've gotten more info and have updated the text. (Short answer: no! If this error occurs, Realm will just retry until it is able to connect normally, and then the error goes away.)

source/sync/error-handling/errors.txt
dacharyc: Alrighty, I've updated this to make things clearer. Text is now:
> This error triggers a client reset. To recover from this error, perform a :ref:`client reset <client-resets>`.

Hopefully someday soon there will be less "performing" involved!

PR #1628, opened 2022-02-08
(DOCSP-19593): Seamless client resets, android examples

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19593

- [Reset a Client Realm - Android SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19593/sdk/android/examples/reset-a-client-realm/)

### Review Guidelines


source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,345 @@
...
+       :tabid: java
+
+       .. literalinclude:: /examples/generated/android/sync/ClientResetTest.codeblock.handle-manual-reset.java
+         :language: java
+
+.. note:: This Example is Simplified
+
+   This example keeps track of the last time each *object* was updated. As
+   a result, the recovery operation overwrites the entire object in the
+   new realm if any field was updated after last last successful sync
```
cbush, line 340:
```suggestion
   new realm if any field was updated after the last successful sync
```

cbush (approved):
The docs looks great within the bounds of what our client reset support looks like, but I'm, I don't know, flabbergasted(?) at the sanctioned workarounds we're providing here. While I appreciate that some guidance is better than nothing, I'm curious if there are plans from the product side to make this better for the user because having to do all this state management for reset recovery is going to be very error-prone for users. It also includes a timeout to avoid an infinite loop etc. Do you see this as an issue? Do they (product/eng)?

source/examples/generated/android/sync/ClientResetTest.codeblock.handle-manual-reset.java
```diff
@@ -0,0 +1,220 @@
...
+            Log.w("EXAMPLE", "Inserting: " + potato.getString("species"));
+            newRealm.executeTransaction(transactionRealm ->
+                    transactionRealm.insertOrUpdate(new Potato(potato)));
+        }
+        for(DynamicRealmObject onion : onionQuery.findAll()) {
+            Log.w("EXAMPLE", "Inserting: " + onion.getString("varietal"));
+            newRealm.executeTransaction(transactionRealm ->
+                    transactionRealm.insertOrUpdate(new Onion(onion)));
+        }
+        for(DynamicRealmObject rice : riceQuery.findAll()) {
```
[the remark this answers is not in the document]
nathan-contino-mongo: Hmm, I use the built-in Kotlin formatter in Android Studio, so inconsistency is surprising. But that formatter is also chock full of bugs (constantly messes up tabbing, even at the same scope level) so I am not surprised. Will do a once-over to fix glaring inconsistencies.

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,345 @@
...
+   option, you can call :java-sdk:`SyncSession.downloadAllServerChanges()
+   <io/realm/mongodb/sync/SyncSession.html#downloadAllServerChanges(long,java.util.concurrent.TimeUnit)>`
+   after opening the realm.
+
+#. Use the :java-sdk:`getBackupRealmConfiguration()
+   <io/realm/mongodb/sync/ClientResetRequiredError.html#getBackupRealmConfiguration()>`
+   method of the provided ``ClientResetRequiredError`` to open an
+   instance of the client realm file from the backup file. You must open
+   this instance as a :java-sdk:`DynamicRealm <io/realm/DynamicRealm.html>`,
+   a type of realm that uses text field lookups for all data access.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Yeah, the SDK actually throws an exception if you try to open it with anything other than a dynamic realm -- presumably it's not actually necessary if your client reset was caused by the max offline time or something like that, but the SDK enforces it anyway!

Kind of like the cached configuration incompatibility issue, where the Java SDK complains if you try to open one realm with slightly different settings (for instance, one with the "wait for initial sync" setting, and one without). There's niche cases where these things can cause bugs, but we're maybe a bit overzealous in enforcing the rules to a point where it causes inconvenience.

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,345 @@
...
+and your schema. However, there are a few techniques that can help with
+most manual recoveries. The following example implementation demonstrates
+one method of recovering unsynced changes from a backup realm.
+
+.. _android-manual-recovery-implementation-last-updated-time:
+
+Last Updated Time
+~~~~~~~~~~~~~~~~~
+
+Ordinarily, there is now way to detect when a Realm object was last
```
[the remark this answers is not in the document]
nathan-contino-mongo: I suspect it exists in core, or it'll have to exist for the "automatic recovery" strategy that's coming in a few months. Whether or not it's exposed to the end user is another matter. I very much hope we can push for it to aid the manual recovery process.

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,345 @@
...
+
+Last Synced Time
+~~~~~~~~~~~~~~~~
+
+Just knowing when objects were changed isn't enough to recover data
+during a client reset. You also need
+to know when the realm last completed a sync successfully. This example
+implementation uses a singleton object called ``LastSynced`` in the
+realm, paired with an upload progress listener, to record whenever a
+realm finishes syncing successfully.
```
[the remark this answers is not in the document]
nathan-contino-mongo: I'm pretty sure I wrote this down in a notebook somewhere about this entire feature 🤣

examples/android/sync/app/src/main/java/com/mongodb/realm/examples/model/java/Potato.kt
```diff
@@ -0,0 +1,38 @@
...
+// :code-block-start: potato-kotlin-definition
+import io.realm.DynamicRealmObject
+import io.realm.RealmObject
+import io.realm.annotations.PrimaryKey
+import org.bson.types.ObjectId
+
+open class Potato : RealmObject {
+    @PrimaryKey
+    var _id: ObjectId? = null
+        private set
```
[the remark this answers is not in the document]
cbush: Yes but there's no empty line between this line and the next line, so a casual reader has to think for a sec about whether this modifies _id or the next property.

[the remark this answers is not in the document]
nathan-contino-mongo: But I'll remove both instances of this pattern because it's not really that important to protect against _id changes and it's probably more confusing to even the Kotlin-literata than it's worth.

source/sdk/android/examples/reset-a-client-realm.txt
nathan-contino-mongo: Fixed in the larger comments commit.

examples/android/sync/app/src/main/java/com/mongodb/realm/examples/model/java/LastSynced.kt
```diff
@@ -0,0 +1,21 @@
...
+// :code-block-start: last-synced-kotlin-definition
+import io.realm.RealmObject
+import io.realm.annotations.PrimaryKey
+import org.bson.types.ObjectId
+
+open class LastSynced : RealmObject {
+    var timestamp: Long? = null
+
+    @PrimaryKey
+    var _id: ObjectId? = null
```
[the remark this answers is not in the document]
nathan-contino-mongo: Looks like creating the same ObjectId multiple times is... difficult, because they're inherently supposed to have a bit of randomness when created from a seed. I *think* creating an objectId repeatedly from an array of 12 zero bytes consistently creates the same object id... but in the end, it felt simpler to just use "null" as the singleton objectid value, because it:

- captures the idea that there really *isn't* a primary key here -- it's a singleton!
- is simpler and shorter to write in code than creating an array of 12 zero bytes in java (roughly 40 characters)

Thanks for this idea!

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,129 @@
...
+- ``onError()``: mimics the "manually recover unsynced changes"
+  strategy. The SDK only calls this method when your application
+  experiences a destructive schema change. For more information on
+  how to handle a this scenario, see the
+  :ref:`Manually Recover Unsynced Changes
+  <android-manually-recover-unsynced-changes>` strategy documentation.
+
+The following example implements this strategy:
+
+.. tabs-realm-languages::
```
dacharyc, line 74: Do we have an example of something that ``onBeforeReset()`` and ``onAfterReset()`` might do besides just printing a message? I presume these methods are provided so you can take care of some sort of business logic, if needed, before and after doing the client reset - it would be great to provide an example, or maybe even just a comment with something like:

> // Complete any operations you need to do before deleting the Realm

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,129 @@
...
+       .. literalinclude:: /examples/generated/android/sync/ClientResetTest.codeblock.client-reset-discard-unsynced-changes.kt
+         :language: kotlin
+
+   .. tab::
+       :tabid: java
+
+       .. literalinclude:: /examples/generated/android/sync/ClientResetTest.codeblock.client-reset-discard-unsynced-changes.java
+         :language: java
+
+Manual Fallback for Destructive Schema Changes
```
dacharyc, line 90: Hmm. The title "Manual Fallback for Destructive Schema Changes" does not communicate to me that I'm handling the error state for discarding unsynced changes, or that this manual fallback also discards unsynced changes. I wonder if there's something else we can call this that is easier to parse for users who are quickly scanning the page? Maybe "Discard Unsynced Changes for Destructive Schema Changes"? Or maybe "Discard Unsynced Changes on Error" and then point out that destructive changes will always trigger the error, and therefore, this handler?

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,129 @@
...
+the client realm file, but has not yet synced to the backend.
+
+The "discard unsynced changes" strategy can handle every kind of client
+reset error *except* for client resets triggered by
+:ref:`destructive schema changes <destructive-changes-synced-schema>`.
+If your application experiences a destructive schema change, this strategy
+will fall back to a mode that mimics the "manually recover unsynced changes"
+strategy.
+
+To handle client resets with the "discard unsynced changes" strategy,
```
dacharyc, line 51: Does this require a specific version of the SDK? I wonder if we should add a `..versionadded::` directive?

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,295 @@
...
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. warning:: Advanced Concept
+
+   Manual recovery requires significant amounts of code, schema concessions,
+   and custom conflict resolution logic. If your application can accommodate
+   losing unsynced data in the rare event of a client reset, try the
```
dacharyc, line 19: Nit: "the rare event of a client reset" sounds a little disingenuous, given that people will likely encounter this if they're iterating on their app _or_ will have to do major contortions involving partner collections to avoid it. Maybe remove the "rare"?

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,295 @@
...
+   :ref:`discard unsynced changes <android-discard-unsynced-changes>`
+   client reset strategy instead.
+
+The **manually recover unsynced changes** client reset strategy gives
+developers the opportunity to recover data that has been written to
+the client realm file, but has not yet synced to the backend. The
+following steps demonstrate the process at a high level:
+
+1. Your application receives a client reset error code from the backend.
```
dacharyc, line 28: This was a lot of text to parse, and I felt like I had to keep a lot of state in my head while reading this section. Maybe we could preface each step with a concise summary of what the step entails? i.e.:

1) **Client Reset Error**: Your application receives...
2) **Call error handler**: SDK calls your...
3) **Close realm**: Close all open instances...
4) **Backup the realm**: move the current copy of the client realm to a backup file using...
5) **Open a new synced realm**: Open a new instance...
6) **Download synced data**: download the entire data set...
7) **Open the backed-up realm**: open an instance of the client realm from the backup file...
8) **Query for data to recover**: find the data sets you need to recover
9) **Apply changes to the synced realm**: Insert, update, etc.

source/examples/generated/android/sync/ClientResetTest.codeblock.handle-manual-reset.kt
```diff
@@ -0,0 +1,290 @@
+fun handleManualReset(app: App, session: SyncSession?, error: ClientResetRequiredError) {
+    Log.w("EXAMPLE", "Beginning manual reset recovery.")
+
+    // close all instances of your realm -- this application only uses one
```
dacharyc, line 4: If you gave a concise heading to each step in the "broad overview" set of numbered steps at the top of the page, you could add those headings and number the steps in the comments here, to make it easier for users to follow the logic. So this could be:

> // 3) Close realm: close all instances of...

source/examples/generated/android/sync/ClientResetTest.codeblock.handle-manual-reset.kt
```diff
@@ -0,0 +1,290 @@
...
+    globalRealm!!.close()
+
+    try {
+        Log.w("EXAMPLE", "About to execute the client reset.")
+        // execute the client reset, moving the current realm to a backup file
```
dacharyc, line 9: And this could be:

> // 4) Backup the realm: execute the...

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,295 @@
...
+   below for an :ref:`example implementation <android-manual-recovery-implementation>`.
+
+The specifics of manual recovery depend heavily upon your application
+and your schema. However, there are a few techniques that can help with
+most manual recoveries. The following example implementation demonstrates
+one method of recovering unsynced changes from a backup realm.
+
+.. _android-manual-recovery-implementation-last-updated-time:
+
+Last Updated Time
```
dacharyc, line 108: This isn't really a meaningful heading in the context of "On This Page" and the overall process of recovering unsynced changes. I wonder if changing it to something like "Add Last Updated Time to Object Model" or something similar would make it clearer that this is a step you're taking to prepare your app to handle client recovery?

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,295 @@
...
+   .. tab::
+       :tabid: java
+
+       .. literalinclude:: /examples/generated/android/sync/RiceJava.codeblock.rice-java-definition.java
+         :language: java
+         :caption: Rice.java
+
+.. _android-manual-recovery-implementation-last-synced-time:
+
+Last Synced Time
```
dacharyc, line 167: Same with this heading - you might change it to a verb to make it clear this is a step you're doing to prepare the app for client reset. Maybe something like: "Watch for Last Synced Time"

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,295 @@
...
+
+.. note:: ``handleManualReset()`` Implementation
+
+   This client reset example calls a separate method that handles the
+   specific logic of the client reset. Continue reading the sections
+   below for an :ref:`example implementation <android-manual-recovery-implementation>`.
+
+The specifics of manual recovery depend heavily upon your application
+and your schema. However, there are a few techniques that can help with
+most manual recoveries. The following example implementation demonstrates
```
dacharyc, line 103: You might add a short paragraph here describing what this example will show. Something like:

> This example adds a "Last Updated Time" to the object model to track the most recent update to an object. We'll watch the realm for the "Last Synced Time" to determine at what point the last successful Sync occurred. Then, we can implement logic to look for objects that have been updated since the last Sync time, and update those objects after we download a new Synced realm.

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,295 @@
...
+   of the backup realm. This could overwrite fields updated by other
+   clients with old data from this client. If your realm objects contain
+   multiple fields containing important data, consider keeping track of
+   the last updated time of each *field* instead, and recovering each
+   field individually.
+
+Alternative Implementations
+---------------------------
+
+Other possible implementations include:
```
dacharyc, line 272: I might add a short phrase to each implementation to concisely describe the different strategies here. I.e.:

- **Overwrite all synced data.** If you don't want to track "Last Updated Time" and "Last Synced Time", you can...
- **Track updates by field.** Instead of tracking for every object...
- **Create an Updates object.** Instead of tracking...

Alternately, you might lay out a brief use case as an initial sentence. i.e.:

- **User only writes to their own realm**: ``insertOrUpdate()``...

dacharyc (approved):
This looks great! I think you've captured here a moderate example of the major effort that is a manual client reset. I'm able to follow the code examples, chonky though they are, and overall I understand what's represented here. There are some things I think we could do to make the information easier to digest/more scannable, so I've made a few suggestions, but even if you don't make any of these changes, I think it's a huge improvement over what we've had to date.

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,129 @@
...
+comes with one large caveat: it deletes all changes made since the last
+successful sync from your realm. Do not use the "discard unsynced changes"
+strategy if your application cannot lose data that has been written to
+the client realm file, but has not yet synced to the backend.
+
+The "discard unsynced changes" strategy can handle every kind of client
+reset error *except* for client resets triggered by
+:ref:`destructive schema changes <destructive-changes-synced-schema>`.
+If your application experiences a destructive schema change, this strategy
+will fall back to a mode that mimics the "manually recover unsynced changes"
```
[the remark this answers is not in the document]
nathan-contino-mongo: ha, i anticipated this comment with a prescient fix :)

source/sdk/android/examples/reset-a-client-realm.txt
nathan-contino-mongo: Technically you can access a frozen copy of the unsyncable realm in the `onAfterReset()` method to copy data over, but as it isn't a persistent backup, you run the risk of losing it (and any unsynced data within) if the app crashes for any reason. Honestly I couldn't figure out any real uses for those methods beyond printing out logs, but I'll see if @cmelchior has any suggestions on this.

source/sdk/android/examples/reset-a-client-realm.txt
nathan-contino-mongo: Using "Discard Unsynced Changes after Destructive Schema Changes", agreed that this was a little confusing!

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
nathan-contino-mongo: Arg, marketing speak coming out. I actually *removed* a "rare" from the backend version of this page when I rewrote it, should have known better.

Removed it.

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
nathan-contino-mongo: Fantastic, this is a huge improvement. TY!

source/examples/generated/android/sync/ClientResetTest.codeblock.handle-manual-reset.kt
nathan-contino-mongo: I actually originally did this, but when I added more steps, it turned into a nightmare. I will use the "concise summaries" instead, I think that's a good compromise to avoid writing a million numbers down :)

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
nathan-contino-mongo: I guess it's just called "last updated time" because it's unlikely that you'd actually follow this example exactly -- it's not steps. Instead, this just demonstrates one possible solution. I'll change it to "Keep track of object updates" to make the concept clearer.

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
nathan-contino-mongo: same as above!

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -103,10 +107,17 @@ and your schema. However, there are a few techniques that can help with
 most manual recoveries. The following example implementation demonstrates
 one method of recovering unsynced changes from a backup realm.
 
-.. _android-manual-recovery-implementation-last-updated-time:
+This example adds a "Last Updated Time" to each object model to track
+the when each object last changed. We'll watch the realm for the
```
dacharyc, line 111: Oops! "the when"

nathan-contino-mongo: LOL thank you.

nathan-contino-mongo: Sounds like the villain from The Neverending Story.

"What's that?"

"THE WHEN"

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,311 @@
...
+
+   Manual recovery requires significant amounts of code, schema concessions,
+   and custom conflict resolution logic. If your application can accommodate
+   losing unsynced data during a client reset, try the
+   :ref:`discard unsynced changes <android-discard-unsynced-changes>`
+   client reset strategy instead.
+
+The **manually recover unsynced changes** client reset strategy gives
+developers the opportunity to recover data already written to
+the client realm file, but not yet synced to the backend. The
```
cbush, line 25:
```suggestion
the client realm file but not yet synced to the backend. The
```

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,311 @@
...
+and your schema. However, there are a few techniques that can help with
+most manual recoveries. The following example implementation demonstrates
+one method of recovering unsynced changes from a backup realm.
+
+This example adds a "Last Updated Time" to each object model to track
+when each object last changed. We'll watch the realm for the
+"Last Synced Time" to determine when the realm last uploaded its state
+to the backend. Then, we can find objects that were deleted, created,
+or updated since the last sync with the backend, and copy that data from
+the backup realm to the new realm.
```
cbush, line 115: Should this section go below the next heading? Feels disconnected, I'm not sure which "example" you're referring to.

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,311 @@
...
+
+- **Overwrite the entire backend with the backup state**:
+  with no "last updated time" or "last synced time", ``insertOrUpdate()``
+  all objects from the backup realm into the new realm. There is no way
+  to recovered unsynced deletions with this approach. This approach
+  overwrites all data written to the backend by other clients since the
+  last sync. Recommended for applications where only one user writes
+  to each realm.
+
+- **Track changes by field**:
```
cbush, line 293: I don't think this or the next ("separately by objects") should be suggested in the docs.

cbush (commented):
I still have a problem with the "advanced guide" because we are basically endorsing this as a viable solution when the guidance should be: **don't make destructive schema changes and expect to recover unsynced data**.

I don't think it's reasonable to have database meta info in the model or expect users to maintain their own form of changelog just to handle a situation that would be avoided if they heeded the real recommendation: **don't make destructive schema changes and expect to recover unsynced data**.

I would prefer if the tone of the advanced guide were more like "look at what you'd have to do to break this rule" rather than "if you're advanced, you can try this cool trick...". And, of course, the PR should state and restate the rule: **don't make destructive schema changes and expect to recover unsynced data**.

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,143 @@
...
+
+The **discard unsynced changes** client reset strategy helps you perform
+a client reset with minimal code and minimal disruption to your application
+workflow. This strategy restores your local realm to a syncable state without
+closing the realm or missing notifications.
+
+This strategy comes with a downside: it permanently deletes all
+local unsynced changes made since the last successful sync of the realm.
+Do not use the "discard unsynced changes" strategy if your application
+cannot lose data already written to the client realm file, but not yet
```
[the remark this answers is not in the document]
nathan-contino-mongo: fixed locally

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -0,0 +1,311 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. warning:: Advanced Concept
```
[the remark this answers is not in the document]
nathan-contino-mongo: renamed

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
nathan-contino-mongo: fixed locally

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
nathan-contino-mongo: Adding an example section and grouping those other sections beneath, good point. This will serve as the introduction paragraph.

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
nathan-contino-mongo: Why not? Not sure what makes these any different from the example above.

nathan-contino-mongo:
@cbush I would appreciate your eyes on this one last time, to make sure all the changes based on your suggestions look good. Thanks!

source/includes/destructive-schema-change-app-update.rst
```diff
@@ -0,0 +1,5 @@
+.. note:: Destructive Schema Changes Require an App Schema Update
+
+   If you make a destructive schema change in your backend schema, client
+   apps must update their local object models before they can resume
```
cbush, line 4: Sorry, this is actually too strong. They only "must" if the apps currently use the part of the schema that was destructively affected, right?

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -12,17 +12,26 @@ Manual Client Reset Data Recovery - Android SDK
    :depth: 2
    :class: singlecol
 
-.. warning:: Advanced Concept
+.. note:: Manual Recovery is Manual
```
cbush, line 15: Haha. I like the tautology to make a point but I wonder if people will get it.

source/sdk/android/advanced-guides/manual-client-reset-data-recovery.txt
```diff
@@ -12,17 +12,26 @@ Manual Client Reset Data Recovery - Android SDK
...
    losing unsynced data during a client reset, try the
    :ref:`discard unsynced changes <android-discard-unsynced-changes>`
    client reset strategy instead.
 
+.. warning:: Do Not Make Destructive Schema Changes
+
+   This guide demonstrates one method of recovering *some* unsynced
+   changes after a destructive schema change. But no manual recovery
+   method is perfect. The best way to preserve user data is to never
+   make a destructive schema change at all.
```
cbush, line 28: I preferred my phrasing because this version makes me think you should never make a destructive schema change at all, which isn't true. (In development etc you can do it to your heart's content and just delete the app). Maybe:

>Avoid making destructive schema changes in production.
>If you must make a destructive schema change, do not expect to recover unsynced data.

cbush (commented):
Better!

cbush (approved):

examples/android/sync/app/src/main/java/com/mongodb/realm/examples/model/java/LastSynced.kt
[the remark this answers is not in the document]
cmelchior: It doesn't have to be an ObjectId? You could also just use a long?

source/includes/destructive-schema-change-app-update.rst
```diff
@@ -0,0 +1,6 @@
+.. note:: Destructive Schema Changes Require an App Schema Update
+
+   After a destructive schema change, clients must update their local
+   object models before they can resume syncing. Only clients that use
```
cmelchior, line 4: This isn't entirely true....since all data is being synced, even that not part of the client schema, if you modify the serverside schema, all connected clients will see a Client Reset.

source/sdk/android/examples/reset-a-client-realm.txt
```diff
@@ -0,0 +1,149 @@
...
+------------------------
+
+.. versionadded:: 10.10.0
+
+The **discard unsynced changes** client reset strategy helps you perform
+a client reset with minimal code and minimal disruption to your application
+workflow. This strategy restores your local realm to a syncable state without
+closing the realm or missing notifications.
+
+This strategy comes with a downside: it permanently deletes all
```
cmelchior, line 50: This isn't entirely true. We do trigger a callback after the client reset where the old data is still available. So you can still move any old data if you want. But not sure if that nuance fits here.

source/sdk/android/examples/reset-a-client-realm.txt
cmelchior: It is used to transfer data, but you are correct, the data gets lost if your app crashes while the method is running.

PR #1635, opened 2022-02-10
Update language to make it clear that all clients client reset after …

nathan-contino-mongo:
…a destructive change

### Jira

- slack conversation

- [Reset a Client Realm - Android](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/client-reset-destructive-change-clarification/sdk/android/examples/reset-a-client-realm/#discard-unsynced-changes-after-destructive-schema-changes)
- [Client Resets](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/client-reset-destructive-change-clarification/sync/error-handling/client-resets/)

### Review Guidelines

### DIFF
--- a/source/includes/destructive-schema-change-app-update.rst
+++ b/source/includes/destructive-schema-change-app-update.rst
@@ -1,6 +1,9 @@
 .. note:: Destructive Schema Changes Require an App Schema Update
 
-   After a destructive schema change, clients must update their local
-   object models before they can resume syncing. Only clients that use
-   models affected by that schema change need to update. All clients
-   must perform a client reset.
+   After a destructive schema change:
+
+   - All clients must perform a client reset.
+   - Clients that use models destructively affected by the schema
+     change must update local object models.
+
+
--- a/source/includes/destructive-schema-change-app-update.rst
+++ b/source/includes/destructive-schema-change-app-update.rst
@@ -2,5 +2,5 @@
 
    After a destructive schema change, clients must update their local
    object models before they can resume syncing. Only clients that use
-   models affected by that schema change need to update and perform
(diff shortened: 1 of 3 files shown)


source/includes/destructive-schema-change-app-update.rst
```diff
@@ -2,5 +2,5 @@
 
    After a destructive schema change, clients must update their local
    object models before they can resume syncing. Only clients that use
-   models affected by that schema change need to update and perform
-   a client reset.
\ No newline at end of file
+   models affected by that schema change need to update. All clients
```
dacharyc, line 5: Nit: we might want to lead with the thing that applies to everyone, and then mention the thing that applies to a subset of users? Maybe something like:

> After a destructive schema change:
> - All clients must perform a client reset
> - Clients that use models affected by the schema change must update local object models

dacharyc (approved):
Small nit, but LGTM overall!

source/includes/destructive-schema-change-app-update.rst
dacharyc: I like it! But I wonder if we need:

"models destructively affected by the schema change" since the first line is "After a destructive schema change" - I know we're trying to be explicit about destructive vs. additive, but I think that's implied by the first line? (i.e. maybe we can remove "destructively" and just have "models affected by the schema change"

dacharyc: I don't think we need more information; I just think the "destructive" in the heading implies the "destructive" in the second bullet. But if folks are getting tripped up on that, then it makes sense to keep it!

PR #1642, opened 2022-02-14
(DOCSP-19777): Add size limit guidance to custom user data

cbush:
## Pull Request Info

Adds guidance around size limitations of custom user data since the data is stored in the access token.

However, not sure if this is actually a good change since any limitations/the meaning of "small" would depend on the other infrastructure of the user's site.

Is this a product issue? Should the custom user data be fetched in the access token at all?

Opening for discussion.

### Jira

- https://jira.mongodb.org/browse/DOCSP-19777

- https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/user-data-size/users/enable-custom-user-data/

### Review Guidelines

### DIFF
--- a/source/users/enable-custom-user-data.txt
+++ b/source/users/enable-custom-user-data.txt
@@ -29,10 +29,11 @@ and their client application access token.
 
 .. note::
 
-   Because {+service+} stores the custom user data in the access token,
-   you should keep the custom user data payload small (less than 16KB).
-   Otherwise, the payload might exceed header size limits, depending on
-   your infrastructure.
+   Because {+service+} stores the custom user data in the access token
+   in the HTTP header, you should keep the custom user data payload
+   small (say, less than 16KB). Other services might limit the HTTP
+   header size, which means that large custom user data objects can
+   cause integration issues.
 
 {+service+} does not manage custom user documents so you are responsible for
 creating and deleting them. The underlying data is a regular MongoDB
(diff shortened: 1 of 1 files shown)


MongoCaleb:
I agree this is a problem, and I think we _could_ document it as a band-aid, but really this info shouldn't be int he jwt -- seems like a significant security issue (potentially).

cbush:
@MongoCaleb how is it a security issue? The jwt will not be visible over HTTPS

source/users/enable-custom-user-data.txt
```diff
@@ -27,6 +27,13 @@ in their access token when they log in. You can access the data in the
 <context-get-user-information>`, the :json-expansion:`%%user` expansion,
 and their client application access token.
 
+.. note::
+
+   Because {+service+} stores the custom user data in the access token,
+   you should keep the custom user data payload small (less than 16KB).
+   Otherwise, the payload might exceed header size limits, depending on
+   your infrastructure.
```
nlarew, line 35: To be clear - this isn't actually a limitation of Realm itself, right? i.e. Realm would dutifully include 16+KB of custom user data in the JWT without issue but other systems may have issues with that JWT.

Also "header" here refers to HTTP headers right?

Maybe we could state that some external services limit HTTP header sizes, which means that large custom user data objects will be too big and cause integration issue.

nlarew:
I think part of the issue is that custom user data in the context of an application could be large (especially with our PBS function rules approach) but for a specific service external service/integration only a subset of data in the document is relevant.

This ultimately feels like a product limitation and is something we should improve. That said it's the way things are for the foreseeable future so I think it's worth it to document here.

nlarew (approved):
LGTM on the addition + a thought on how we might clarify the "depends on external system" point

PR #1677, opened 2022-02-23
(DOCSP-19947): Kotlin SDK CRUD Operations

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-19947

Given the amount of content, it's probably best to just browse the entire SDK realm database/sync sections in the staging site, and then comment in the "files changed" tab only when you identify things that need changes. Thanks!

- [Kotlin SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/actuallyKotlinCRUD/sdk/kotlin/) -- entire "Realm Database" and "Sync" sections

### Review Guidelines


source/includes/kotlin-install-prerequisites.rst
```diff
@@ -0,0 +1,42 @@
...
+  for Android Studio, version 0.2.6 or higher.
+- A KMM App created using the "KMM Application" template in Android
+  Studio. Follow the instructions in the `KMM documentation
+  <https://kotlinlang.org/docs/mobile/create-first-app.html>`__.
+- An Android Virtual Device (AVD) using either the :file:`x86_64` or
+  :file:`arm64` architecture.
+
+  .. important:: x86 Devices are not Compatible with the SDK
+
+     The Kotlin SDK does not support :file:`x86` devices. Since :file:`x86`
```
cmelchior, line 18: This is true for now... but we will have x86 support in the next release.

source/sdk/kotlin/install/android.txt
```diff
@@ -0,0 +1,137 @@
...
+      - Add :file:`io.realm.kotlin:library-base` to the list of
+        dependencies.
+
+      .. code-block:: kotlin
+         :emphasize-lines: 35
+         :copyable: false
+
+         plugins {
+             id 'com.android.application'
+             id 'org.jetbrains.kotlin.android'
```
cmelchior, line 32: Looks like you are missing the realm plugin here?

source/sdk/kotlin/install/android.txt
```diff
@@ -0,0 +1,137 @@
...
+      - Add :file:`io.realm.kotlin:library-sync` to the list of
+        dependencies.
+
+      .. code-block:: kotlin
+         :emphasize-lines: 35
+         :copyable: false
+
+         plugins {
+             id 'com.android.application'
+             id 'org.jetbrains.kotlin.android'
```
cmelchior, line 90: Same

source/sdk/kotlin/realm-database/schemas/supported-types.txt
```diff
@@ -0,0 +1,48 @@
...
+- :file:`Integer` or :file:`int`
+- :file:`Short` or :file:`short`
+- :file:`Long` or :file:`long`
+- :file:`Byte` or :file:`byte[]`
+- :file:`Double` or :file:`double`
+- :file:`Float` or :file:`float`
+- :file:`String`
+- :file:`Date`
+- :file:`Decimal128` from :file:`org.bson.types`
+- :file:`ObjectId` from :file:`org.bson.types`
```
cmelchior, line 27: These are not supported yet.
Also `ByteArray` support is also not supported, but is coming.
We also support `Char`

source/sdk/kotlin/realm-database/schemas/supported-types.txt
```diff
@@ -0,0 +1,48 @@
...
+The :file:`Byte`, :file:`Short`, :file:`Integer`, and :file:`Long` types and their
+lowercase primitive alternatives are all stored as :file:`Long` values
+within Realm Database. Similarly, Realm Database stores objects
+of the :file:`Float` and :file:`float` types as type :file:`Double`.
+
+Realm Database does not support fields with modifiers :file:`final` and
+:file:`volatile`, though you can use fields with those modifiers if you
+:kotlin-sdk:`ignore <{+kotlin-local-prefix+}io.realm.annotations/-ignore/index.html>`
+them. If you choose to provide custom constructors, you must declare a
+public constructor with no arguments.
```
cmelchior, line 40: Maybe an example for this would be good:

```
class Person(var name: String): RealmObject {
  constructor(): this("no-name")
  // ... other properties
} 
```

``

source/sdk/kotlin/realm-database/write-transactions.txt
```diff
@@ -0,0 +1,73 @@
+.. _kotlin-write-transactions:
+
+=============================
+Write Transactions - Java SDK
```
cmelchior, line 4:
```suggestion
Write Transactions - Kotlin SDK
```

cmelchior (commented):
Nice...I added a few comments to some of the code samples where we might consider improving them to avoid leading people towards the wrong patterns.

source/sdk/kotlin/install/android.txt
nathan-contino-mongo: I tested this, and it seemed all I needed to pull the dependency down was what's shown here. Do you know why that might be? I'm hesitant to add an extra line to the doc when the SDK seems to work fine without it. But if it's likely to break in the future, or if it will cause problems I don't know about yet, happy to add it.

source/examples/generated/kotlin/CRUDTest.codeblock.delete-all-objects-of-a-type.kt
```diff
@@ -0,0 +1,6 @@
+realm.writeBlocking {
```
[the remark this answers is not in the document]
nathan-contino-mongo: Interesting point. I'm using `writeBlocking` now because it makes it easy to guarantee that the logic in the test runs before the text exits. The idea is that you'd be running these things on a background thread anyway, and it's easier to just do things synchronously in that environment. But maybe it's better to demonstrate this in a "ui-thread" friendly way. I'll see if I can switch these and use some manual workarounds to ensure that we delay test execution until they complete.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/CRUDTest.kt
```diff
@@ -0,0 +1,341 @@
...
+                val frog = this.copyToRealm(Frog().apply {
+                    _id = Random.nextLong(1000000)
+                    name = "Wirt"
+                    age = 45
+                    species = "Green"
+                    owner = "Jim"
+                })
+            }
+            // :code-block-start: upsert-an-object
+            realm.writeBlocking {
```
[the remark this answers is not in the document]
nathan-contino-mongo: I figured! Let me know when that drops and I can update this example. I left this in because it's a common enough use case and we should probably have *something* for it, even if it's suboptimal at the moment.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/CRUDTest.kt
```diff
@@ -0,0 +1,341 @@
...
+            // :code-block-start: sort
+            // sort in descending order, frogs with distinct owners, only the first 5
+            val frogs: RealmResults<Frog> =
+                realm.query<Frog>("name = 'George Washington' SORT(age DESC) DISTINCT(owner) LIMIT(5)").find()
+            // :code-block-end:
+            realm.close()
+        }
+    }
+
+    //@Test TODO: investigate why this flow collect() call hangs
```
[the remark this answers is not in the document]
nathan-contino-mongo: Thanks, this makes perfect sense. Updated the example and fixed the test.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/CRUDTest.kt
```diff
@@ -0,0 +1,341 @@
...
+                // specify name so realm doesn't just use the "default.realm" file for this user
+                .name(REALM_NAME)
+                .path(randomTmpRealmPath()) // :hide: // default location for jvm is... in the project root
+                .build()
+            val realm = Realm.open(config)
+            Log.v("Successfully opened realm: ${realm.configuration.name}")
+            // :code-block-start: sort
+            // sort in descending order, frogs with distinct owners, only the first 5
+            val frogs: RealmResults<Frog> =
+                realm.query<Frog>("name = 'George Washington' SORT(age DESC) DISTINCT(owner) LIMIT(5)").find()
```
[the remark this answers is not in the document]
nathan-contino-mongo: TIL, thanks for pointing this out. I'll demonstrate both so users know both ways.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/CRUDTest.kt
```diff
@@ -0,0 +1,341 @@
...
+                    age = 45
+                    species = "Green"
+                    owner = "Jim"
+                })
+
+                // update the frog
+                frog.age = frog.age + 1
+
+                // get all frogs that belong to Jim
+                val jimsFrogs = this.query<Frog>("owner == 'Jim'").find()
```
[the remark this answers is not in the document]
nathan-contino-mongo: Gotcha. I'll update the "iteration" section to include this information, and add a warning/important block to each of the read pages where users might get the wrong idea (I take it that `find()` is fine when paired with `first()` since it's only fetching one element? Or is there a footgun there as well?

source/sdk/kotlin/realm-database/query-language.txt
```diff
@@ -0,0 +1,13 @@
+.. _kotlin-query-language:
```
mongodben, line 1: i think this page should be under the `realm-database/read` as RQL is just used in read operations

nathan-contino-mongo: Ah, but RQL also applies to deletes and updates (via the reads you need to perform to get objects for those operations), hence my decision to put it... here. Though those are still just reads, so maybe it does belong in the read section? I mostly wanted to keep it easier to find for users, and figured it didn't quite fit in with the explanation/example pattern of the other CRUD pages. But I'll defer to your opinion on this, @mongodben

source/sdk/kotlin/realm-database/schemas/supported-types.txt
nathan-contino-mongo: We'll have more object schema examples coming soon in this section -- I'll save this example in my TODO collection :)

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/CRUDTest.kt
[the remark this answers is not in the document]
cmelchior: Queries can still be expensive to run, even if they only find one element, so even `first()` queries can be dangerous.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/CRUDTest.kt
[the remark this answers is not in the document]
cmelchior: Yup, this is super common, so fine to have to have this.

cmelchior (approved):

source/sdk/kotlin/realm-database/overview.txt
```diff
@@ -0,0 +1,15 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. include:: /includes/realm-database.rst
```
mongodben, line 15: i think you might have to edit the realm-database.rst include to only cover the kotlin specific features. for example, i didn't see any reference ot persistent/in memory realms in the docs or encryption. these sections should be removed.

nathan-contino-mongo: Good point -- because snooty doesn't allow us to selectively display only parts of included content, I'm just going to add a disclaimer at the top about how "this alpha SDK is a work in progress, may not support all of these features yet". Seems like the only way to avoid duplicating a ton of content.


... or maybe I'll just duplicate the relevant content, with the intent of restoring the include once we're at feature parity. Hmmmm.

source/sdk/kotlin/realm-database/schemas.txt
```diff
@@ -0,0 +1,16 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. toctree::
+   :titlesonly:
+   
+   Supported Types </sdk/kotlin/realm-database/schemas/supported-types>
```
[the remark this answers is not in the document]
nathan-contino-mongo: Yep, examples of various models with field constraints are coming -- unfortunately we have very limited annotations right now, basically just primary key and ignore, and no support for realmdictionary/realmset/realmany, so there isn't as much to build on. But I would expect additional content in this section very soon.

source/sdk/kotlin/realm-database/open-a-realm.txt
```diff
@@ -0,0 +1,34 @@
+.. _kotlin-open-a-realm:
+
+=========================
+Open a Realm - Kotlin SDK
```
mongodben, line 4: why not rename to `Open & Close a Realm` like it is for other SDKs?

source/sdk/kotlin/realm-database/open-a-realm.txt
```diff
@@ -0,0 +1,34 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+To open a {+realm+}, create a
```
mongodben, line 15: if page renamed to `Open & Close a Realm`, add another header here for `Open a Realm`

source/reference/realm-query-language.txt
```diff
@@ -12,436 +12,11 @@ Realm Query Language
...
-.. example::
-
-   We use the query engine's string operators to find:
-
-   - Projects with a name starting with the letter 'e'
-   - Projects with names that contain 'ie'
-
-   .. literalinclude:: /examples/generated/realm-query-language/realm-query-language.codeblock.string-operators.js
-      :language: javascript
+.. include:: /includes/rql.rst
```
mongodben, line 15: i don't think you can just include this as there's some content that doesn't make sense for Kotlin SDK: 

for example, second paragraph refers to swift: 
> You can use Realm Query Language in most Realm SDKs with your SDK's filter or query methods. The **Swift SDK** is the exception, as it uses the NSPredicate query API.

also the 'About the Examples on This Page' note refers to some models that i don't think are in kotlin:

```
class Task {
    ObjectId id;
    string name;
    bool isComplete;
    string? assignee;
    int priority;
    int progressMinutes;
}

class Project {
    ObjectId id;
    string name;
    Task[] tasks;
    int? quota;
}
```

mongodben: would also be good to have some reference to the kotlin sdk methods that actually use RQL

source/sdk/kotlin/realm-database/read/sort-queries.txt
```diff
@@ -0,0 +1,42 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+To sort results, specify a sort in the query passed to
+:kotlin-sdk:`realm.query()
+<{+kotlin-local-prefix+}io.realm.query/-realm-query/query.html>`
+with the :file:`SORT` key word. Similarly, use the :file:`DISTINCT`
+key word to constrain results to unique values of a field, and
+the :file:`LIMIT` key word to cap the number of results. The SDK
```
mongodben, line 20: should it be "keyword" not "key word"?

https://forum.wordreference.com/threads/key-word-vs-keyword.3088253/#:~:text=Key%20word%20is%20a%20word,you%20emphasized%20context%20or%20usage.

source/sdk/kotlin/realm-database/read/sort-queries.txt
```diff
@@ -0,0 +1,42 @@
...
+provides convenience methods on :file:`RealmQuery` so you don't have
+to manually write the keywords:
+
+- :kotlin-sdk:`realmQuery.sort() <{+kotlin-local-prefix+}io.realm.query/-realm-query/sort.html>`
+- :kotlin-sdk:`realmQuery.distinct() <{+kotlin-local-prefix+}io.realm.query/-realm-query/distinct.html>`
+- :kotlin-sdk:`realmQuery.limit() <{+kotlin-local-prefix+}io.realm.query/-realm-query/limit.html>`
```
mongodben, line 26: move this content below the example. confusing to see that there are these convenience methods, but they aren't being used. consider also including a new heading on top of this section, smthn like 'Query Convenience Methods'

source/sdk/kotlin/realm-database/read/iteration.txt
```diff
@@ -0,0 +1,35 @@
...
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can iterate through results using `Flows
+<https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/>`__.
```
mongodben, line 16: i think it'd be worthwhile to expand on what flows are in a sentence for the benefit of readers unfamiliar w the concept (including myself)

source/sdk/kotlin/realm-database/read/iteration.txt
```diff
@@ -0,0 +1,35 @@
...
+   :copyable: false
+
+.. note:: :file:`find()` is Synchronous
+
+   You can also iterate through results returned via :file:`find()`.
+   :file:`find()` runs a synchronous query on the thread it is called from.
+   As a result, avoid using :file:`find()` on the UI thread or in
+   logic that could delay the UI thread. Prefer :kotlin-sdk:`asFlow()
+   <{+kotlin-local-prefix+}io.realm/-realm-results/as-flow.html>` in
+   time sensitive environments.
```
mongodben, line 35: can you provide a link to the relevant documentation on find somewhere in here?

source/examples/generated/kotlin/CRUDTest.codeblock.delete-an-object.kt
```diff
@@ -0,0 +1,7 @@
+realm.writeBlocking {
+    // fetch the frog by primary key value, passed in as argument number 0
+    val frogs: RealmResults<Frog> =
+        this.query<Frog>("_id == $0", PRIMARY_KEY_VALUE).find()
+    // call delete on the results of a query to delete the object permanently
+    frogs.delete()
```
mongodben, line 6: would be clearer if the variable were named `frog` instead of `frogs`

source/sdk/kotlin/sync/overview.txt
```diff
@@ -0,0 +1,51 @@
...
+a :ref:`MongoDB Realm backend application <realm-cloud>`. When a client 
+device is online, {+sync-short+} asynchronously synchronizes data in a 
+background thread between the device and your backend {+app+}. 
+
+When you use Sync in your client application, your implementation must match 
+the Sync Mode you select in your backend {+app+} configuration. The Sync Mode
+options are:
+
+- Partition-Based Sync
+- Flexible Sync
```
mongodben, line 30: there should be a flexible sync overview section on this page if there's a partition based sync one.

mongodben: or if flexible sync doesn't yet exist for kotlin sdk, then make reference to it less prominent and maybe include in a note that it's not yet available.

source/sdk/kotlin/sync/open-a-synced-realm.txt
```diff
@@ -1,8 +1,8 @@
-.. _kotlin-sync-changes-between-devices:
+.. _kotlin-open-a-synced-realm:
 
-=========================================
-Sync Changes Between Devices - Kotlin SDK
-=========================================
+================================
+Open a Synced Realm - Kotlin SDK
+================================
```
mongodben, line 5: would be good to make it clear on this page that it's only for partition-based sync.

source/sdk/kotlin/realm-database/overview.txt
mongodben: def duplicate imo. even if more verbose on our end, it's more accurate and clearer for reader.

source/reference/realm-query-language.txt
```diff
@@ -12,436 +12,11 @@ Realm Query Language
    :depth: 3
    :class: singlecol
```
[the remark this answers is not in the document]
nathan-contino-mongo: I do see it here -- might be that your window size is too narrow?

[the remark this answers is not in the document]
nathan-contino-mongo: Could also be a snooty bug, but I cannot reproduce.

source/reference/realm-query-language.txt
nathan-contino-mongo: I think that Swift SDK callout is actually one we should remove -- JS, Swift, and Kotlin all use NSPredicate. And Java provides a "raw query" NSPredicate interface that works just like the Kotlin query interface.

Creating a ticket to create "about the examples on this page" examples for each SDK, though -- and changing this to swift/kotlin tabs initially. This page *should* be universal to all SDKs, so I think these changes make sense.

source/sdk/kotlin/realm-database/create/create-a-new-object.txt
```diff
@@ -0,0 +1,46 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Instantiate Realm objects as you would any other object. In a
```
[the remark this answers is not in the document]
nathan-contino-mongo: Realm object should be capitalized, realm the database object shouldn't be capitalized. Confusing for sure. We might slacken this up if some naming conventions change in the future tho.

source/sdk/kotlin/realm-database/read/sort-queries.txt
nathan-contino-mongo: Thanks for catching this -- somehow I fixed the last occurrence, but not the first three. Removed a couple usages because when i ctrl+f it was far too many repeats of the same word in one paragraph.

source/sdk/kotlin/realm-database/read/sort-queries.txt
nathan-contino-mongo: Looks like this was displaying an outdated version of the snippet (snooty error, maybe? the bluehawk generated file looks fine in this commit) -- the latest version of the snippet shows *both* methods. Since NSPredicate-based RQL is the primary query interface, I figure it'll be a tossup as to which style users prefer.

source/sdk/kotlin/realm-database/read/iteration.txt
nathan-contino-mongo: Flows are unfortunately a very broad and fundamental Kotlin concept that I'm not sure we can adequately sum up in a sentence. I'll see if I can add some clarity here though.

source/sdk/kotlin/realm-database/read/iteration.txt
nathan-contino-mongo: Added. Also added to the other read pages.

source/examples/generated/kotlin/CRUDTest.codeblock.delete-an-object.kt
nathan-contino-mongo: There was a little API snafu with the object delete kotlin extension, but I think we're good now. TY!

source/sdk/kotlin/sync/overview.txt
nathan-contino-mongo: Leaving the mention of two sync modes, but pointing out clearly that the Kotlin SDK currently ONLY supports PBS. Good call.

source/sdk/kotlin/sync/open-a-synced-realm.txt
nathan-contino-mongo: This page will eventually include partition *and* flexible sync content -- but as far as the Kotlin SDK is concerned, only PBS exists right now. Adding a note to make it clear that Kotlin doesn't support flex yet.

source/sdk/kotlin/realm-database/read/sort-queries.txt
mongodben: ahh ok, yes snooty is still just sowing only the NSPredicate version.

source/reference/realm-query-language.txt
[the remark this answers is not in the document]
mongodben: def not a screensize thing, validated on my 24" monitor. bizarre that you see and i don't. can you try to hard reset this page and lmk if you see?

this definitely isn't something that should block release, but worth raising issue w Docs Platform

mongodben (approved):
lgtm

https://tenor.com/brIC3.gif

source/reference/realm-query-language.txt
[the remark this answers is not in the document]
nathan-contino-mongo: This page, right?

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/actuallyKotlinCRUD/reference/realm-query-language/

Screen Shot 2022-02-28 at 1 31 04 PM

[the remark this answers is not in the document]
nathan-contino-mongo: Fixed! Thanks for clearing this up. We were missing the "on this page" in the kotlin page version.

PR #1690, opened 2022-02-28
(DOCSP-21048): Java SDK -> Kotlin SDK Migration Guide, testing tables

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-21048

- [Migrate from Java SDK to Kotlin SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21048/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk/)

### Review Guidelines


examples/java/local/app/src/androidTest/java/com/mongodb/realm/examples/java/MigrateFromJavaToKotlinSDKTest.java
```diff
@@ -0,0 +1,432 @@
...
+                    new RealmConfiguration.Builder()
+                    .build();
+
+            Realm realm;
+            try {
+                realm = Realm.getInstance(config);
+                Log.v("EXAMPLE",
+                        "Successfully opened a realm: "
+                                + realm.getPath());
+            } catch (RealmFileException ex) {
```
cmelchior, line 45: We are not doing this in the Kotlin samples, which makes it a bit unfaithful... But I would also remove this as it distracts from the primary change. The API for opening the Realm.

source/sdk/kotlin/realm-database/read/iteration.txt
```diff
@@ -19,8 +19,8 @@ You can iterate through results using :file:`Flows`.
    To learn more about Kotlin Flows, check out `the kotlinx.coroutines documentation on Flows
    <https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/>`__.
 
-To convert your results into a :file:`Flow`, call :kotlin-sdk:`results.asFlow()
-<{+kotlin-local-prefix+}io.realm.query/-realm-element-query/as-flow.html>`.
+To convert your results into a :file:`Flow`, call :kotlin-sdk:`realmResults.asFlow()
```
cmelchior, line 22: in must cases it is probably better to call `RealmQuery.asFlow()` rather than `RealmResults.asFlow()` since a RealmResults only comes from running a query (and you don't want that on the UI thread).

examples/java/local/app/src/androidTest/java/com/mongodb/realm/examples/java/MigrateFromJavaToKotlinSDKTest.java
```diff
@@ -0,0 +1,432 @@
...
+import io.realm.Realm;
+import io.realm.RealmConfiguration;
+import io.realm.RealmMigration;
+import io.realm.RealmObjectSchema;
+import io.realm.RealmResults;
+import io.realm.RealmSchema;
+import io.realm.Sort;
+import io.realm.exceptions.RealmFileException;
+
+public class MigrateFromJavaToKotlinSDKTest extends RealmTest {
```
cmelchior, line 29: Do we really want to document the difference between Java and Kotlin as well? Doing a Realm Java (Kotlin) to Realm Kotlin would probably match better what most people are doing?

examples/java/local/app/src/main/java/com/mongodb/realm/examples/model/java/Sample.java
```diff
@@ -0,0 +1,26 @@
+package com.mongodb.realm.examples.model.java;
+
+import java.util.Date;
+
+import io.realm.RealmObject;
+import io.realm.annotations.Index;
+import io.realm.annotations.PrimaryKey;
+
+// :code-block-start: realm-object-model
+public class Sample extends RealmObject {
```
cmelchior, line 10: I would convert this to Kotlin syntax. It really displays the benefits of the Kotlin SDK better, i.e. no need to make the  class open

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/MigrateFromJavaToKotlinSDKTest.kt
```diff
@@ -0,0 +1,392 @@
...
+// :code-block-end:
+
+// :code-block-start: one-to-one-relationship
+class Child : RealmObject {
+    var frog: Frog? = null
+}
+// :code-block-end:
+
+// :code-block-start: one-to-many-relationship
+class Kid : RealmObject {
```
cmelchior, line 59: One of the primary benefits of the Kotlin SDK is also that the type system is now much better. This is e.g. visible when having lists of primitives where you no longer need the `@Required` annotation, but we fully understand the Kotlin type system, .e.g. `@Required list: RealmList<String>` is now just `list: RealmList<String>` (if required) or ` list: RealmList<String?>` if nullable values are allowed.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+you specify otherwise. The Kotlin SDK requires you to manually specify
+the Realm Object Models to use in your realm schema. Additionally:
+
+- The Java SDK used :file:`RealmConfiguration.Builder().build()` to
+  generate instances of :file:`RealmConfiguration`. With the Kotlin SDK,
+  use the :kotlin-sdk:`RealmConfiguration.with()
+  <{+kotlin-local-prefix+}io.realm/-realm-configuration/-companion/with.html>`
+  companion method :file:`RealmConfiguration` instead.
+
+- The Java SDK used the static :file:`Realm.getInstance()` method to
```
cmelchior, line 35: Also Kotlin SDK doesn't have the concept of `default configuration` so e.g. `Realm.getDefaultInstance()` and Realm.setDefaultConfiguration()` are no longer available. Basically, they are not required as you Realm is now a global singleton.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+The Java SDK and the Kotlin SDK differ in many ways. On this page, you'll
+find a high-level comparison of most of the ways the SDKs differ.
+
+Opening Realms
```
cmelchior, line 21: I would add a section first about the architectural differences since they drive a lot of changes in the API and address the biggest complaints we had with the Java SDK.

* Frozen-by-default: All objects are now frozen unless fetched from inside a write transaction.
* All Realm instances, objects, query results and lists can now be transferred across threads.
* You now only need 1 Realm instance, no need to open and close Realms in individual threads.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+
+.. seealso::
+
+   :ref:`Open & Close a Realm <kotlin-open-and-close-a-realm>`
+
+Realm Object Models
+-------------------
+
+Both the Java and Kotlin SDKs declare Realm objects the same way: extend
+:file:`RealmObject`, and annotate fields with properties like primary keys
```
cmelchior, line 70: This is not strictly true. In the Kotlin SDK, we no longer have a base class you need to extend. This was a common complaint in Java...now there is just an interface.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+     - Kotlin SDK
+
+   * - .. literalinclude:: /examples/generated/java/local/Child.codeblock.one-to-one-relationship.java
+          :language: java
+          :copyable: false
+     - .. literalinclude:: /examples/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.one-to-one-relationship.kt
+          :language: kotlin
+          :copyable: false
+
+One-to-Many
```
cmelchior, line 114: I would also mention lists of primitives here since this got a major upgrade as we now understand the Kotlin typesystem...e.g. `RealmList<String>` and `RealmList<String?>` is understood instead of needing to use the `@Required` annotation.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+------
+
+The Kotlin SDK introduces new names for the methods that write to realms.
+
+Async
+~~~~~
+
+With the Java SDK, you could write asynchronously to a realm with
+:file:`realm.executeTransactionAsync()`. The Kotlin SDK uses
+:kotlin-sdk:`realm.write()
```
cmelchior, line 147: Maybe mention that it is a suspend function?

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+
+There are several differences between queries in the Java SDK and queries
+in the Kotlin SDK:
+
+- With the Java SDK, you could query objects in realms using a fluent
+  interface or :ref:`Realm Query Language <kotlin-query-language>` (RQL).
+  The Kotlin SDK only uses RQL.
+
+- The Java SDK uses :file:`realm.where()`
+  to query realms, whereas the Kotlin SDK uses :kotlin-sdk:`realmQuery.query()
```
cmelchior, line 203:
```suggestion
  to query realms, whereas the Kotlin SDK uses :kotlin-sdk:`realm.query()
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+  <{+kotlin-local-prefix+>io.realm/query.html>`.
+
+- With the Java SDK, you could query asynchronously with
+  :file:`realmQuery.findAllAsync()` and :file:`realmQuery.findOneAsync()`.
```
cmelchior, line 207:
```suggestion
  :file:`realmQuery.findAllAsync()` and :file:`realmQuery.findFirstAsync()`.
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+  In the Kotlin SDK, query asynchronously with
+  :kotlin-sdk:`realm.asFlow()
+  <{+kotlin-local-prefix+}io.realm.query/-realm-element-query/as-flow.html>`.
+  Once you have a flow of results, you can `collect
+  <https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/collect.html>`__
+  the results.
+
+- With the Java SDK, you could query synchronously with
+  :file:`realmQuery.findAll()` and :file:`realmQuery.findOne()`.
```
cmelchior, line 216:
```suggestion
  :file:`realmQuery.findAll()` and :file:`realmQuery.findFirst()`.
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+
+.. seealso::
+
+   :ref:`Realm Query Language <kotlin-query-language>`
+
+Deletes
+-------
+
+In both SDKs, you can only delete live objects. The Kotlin SDK provides
+:file:`realm.findLatest()` to make it easy to access a live version of
```
cmelchior, line 262:
```suggestion
:file:`mutableRealm.findLatest()` to make it easy to access a live version of
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+   - :ref:`Delete all Objects of a Type <kotlin-delete-all-objects-of-a-type>`
+   - :ref:`Delete Multiple Objects <kotlin-delete-multiple-objects>`
+   - :ref:`Delete an Object <kotlin-delete-an-object>`
+
+Notifications
+-------------
+
+In both SDKs, you can subscribe to change to collections of results.
+With the Java SDK, you could use :file:`realmResults.addChangeListener()`
+to receive notifications whenever an element of a result set changes.
```
cmelchior, line 290: The Java SDK also had both RxJava support through `asFlowable()` and kotlin extensions for converting to flows using `toFlow()`. Most people used these rather than change listeners directly.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+     - Kotlin SDK
+
+   * - .. literalinclude:: /examples/generated/java/local/MigrateFromJavaToKotlinSDKTest.codeblock.notifications.java
+          :language: java
+          :copyable: false
+     - .. literalinclude:: /examples/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.notifications.kt
+          :language: kotlin
+          :copyable: false
+
+Threading
```
cmelchior, line 312: Maybe reference this from the intro section i proposed?

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,359 @@
...
+
+   :ref:`Frozen Architecture <kotlin-frozen-architecture>`
+
+Migrations
+----------
+
+With the Java SDK, migrations were a manual process. The Kotlin SDK
+automates migrations, but also gives you the ability to override those
+automatic migrations when necessary.
+
```
cmelchior, line 346: Not really...the schema migration is automatic, similar to Swift, JS and .NET. We just allow you migrate data just like they do.

cmelchior (commented):
Great start, but I left comments for some things could use improvements.

* The biggest bullet point is that having the old samples in Java is probably counterproductive. I don't think anyone migrating will have their old code in Java, and it makes it easier to compare what actually changed if you don't have to factor in syntax as well.

Also, I think it would be really helpful to have a "Design/Architecture" section at the top describing the overall architectural changes since they address a lof the complaints people had with the old API's.

examples/java/local/app/src/androidTest/java/com/mongodb/realm/examples/java/MigrateFromJavaToKotlinSDKTest.java
```diff
@@ -0,0 +1,432 @@
...
+            }
+            realm = Realm.getInstance(config);
+            realm.close();
+            expectation.fulfill();
+        });
+        expectation.await();
+    }
+
+    @Test
+    public void testQueryAggregates() {
```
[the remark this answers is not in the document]
nathan-contino-mongo: Good point, these aren't even given a category in the RQL docs: https://docs.mongodb.com/realm/reference/realm-query-language/#sort--distinct--limit

I just followed Nabil's outline's lead on this section title because I don't think there's a good general label for this concept. I'll just match the RQL section title instead.

examples/java/local/app/src/androidTest/java/com/mongodb/realm/examples/java/MigrateFromJavaToKotlinSDKTest.java
nathan-contino-mongo: Do we have any user research to indicate that folks aren't currently using the Java SDK with Java? I'll add a tab for kotlin examples in the Java SDK, but as long as the Kotlin SDK looks like the way forward for all Java SDK users, it seems like we're neglecting a portion of users by omitting java examples.

examples/java/local/app/src/main/java/com/mongodb/realm/examples/model/java/Sample.java
nathan-contino-mongo: Same as above.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/MigrateFromJavaToKotlinSDKTest.kt
nathan-contino-mongo: OK, I'll add a separate section on the type system. I omitted the nullable vs. non-nullable example from Nabil's outline because it wasn't actually an example of relationships (just a collection within an object) but this sounds meaty enough to justify a section and an example of its own. Thanks!

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Thanks, added a section on this.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Covered in a change in response to an above comment.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Never mind -- for some reason, I was having trouble bringing it up with search. But we already cover it in the frozen architecture documentation. I'll copy this example over to that section to give a better example of the method, though.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Added these examples as well, and transitioned into the consolidation of all of these notification interfaces into `asFlow()`.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Rephrased this to make it clear that you'll never really "override" the automatic migrations, just supplement them.

examples/java/local/app/src/androidTest/java/com/mongodb/realm/examples/java/MigrateFromJavaToKotlinSDKTest.java
cmelchior: We do have that data. We have 5x number of users of Kotlin vs. Java. Also 80% of top 1000 apps are using Kotlin by now according to Google and it is the official recommended language for all apps.

So we do have Java users, but many of them are not going to upgrade either way and there is also additional benefit of making it easier to compare the two SDK's.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Also curious how this differs from the existing `RealmModel` implementation option in the Java SDK.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+The Java SDK and the Kotlin SDK differ in many ways. On this page, you'll
+find a high-level comparison of most of the ways the SDKs differ.
+
```
mongodben, line 20: i think would be good to include a paragraph why a developer should consider migration.

mongodben: even if the answer is obvious to a java developer, it's not obvious to me :)

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+Architecture
+------------
+
+The Java SDK provided live objects, queries, and realms that
+automatically update when underlying data changes. The Kotlin SDK still
+provides this live interface in write transactions, but otherwise relies
+on a new frozen architecture that makes Realm objects easier to work
+with. With this new architecture, you'll notice a few differences:
```
mongodben, line 28:
```suggestion
with. With the Kotlin SDK architecture, you'll notice a few differences:
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+   :class: singlecol
+
+Overview
+--------
+
+The Java SDK and the Kotlin SDK differ in many ways. On this page, you'll
+find a high-level comparison of most of the ways the SDKs differ.
+
+Architecture
+------------
```
mongodben, line 22:
```suggestion
Kotlin SDK Architecture
-----------------------
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
mongodben: i think it'd also be beneficial to say what it means for an object to be frozen in a sentence before "With this new architecture..."

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Any suggestions here? I want to avoid explicitly saying "you should move to the kotlin SDK, java SDK SUCKS" because the Java SDK is still more feature complete and the Kotlin SDK is still in alpha (soon to be beta). Maybe something that boils down to "the Kotlin SDK is the future" without sounding too marketing-y?

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+
+The Java SDK provided live objects, queries, and realms that
+automatically update when underlying data changes. The Kotlin SDK still
+provides this live interface in write transactions, but otherwise relies
+on a new frozen architecture that makes Realm objects easier to work
+with. With this new architecture, you'll notice a few differences:
+
+- **Frozen by default**: All objects are now frozen. You can still obtain
```
mongodben, line 30: word choice. what does obtain mean here? 

maybe access? 
```suggestion
- **Frozen by default**: All objects are now frozen. You can still access
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+- **Thread-safety**: All realm instances, objects, query results, and
+  collections can now be transferred across threads.
+
+- **Singleton**: You now only need one instance of each realm.
+  :ref:`No need to open and close realms on individual threads.
+  <kotlin-migration-threading>`
+
+.. seealso::
+
+   :ref:`Frozen Architecture <kotlin-frozen-architecture>`
```
mongodben, line 43:
```suggestion
   :ref:`Kotlin SDK Frozen Architecture <kotlin-frozen-architecture>`
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
```
mongodben, line 16: worth mentioning in the overview section that the new Kotlin SDK differs from the Java SDK (Kotlin). i think a note would do.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+to customize your configuration even more:
+
+.. literalinclude:: /examples/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.open-a-realm-advanced.kt
+   :language: kotlin
+   :copyable: false
+   :caption: Kotlin SDK
+
+.. seealso::
+
+   :ref:`Open & Close a Realm <kotlin-open-and-close-a-realm>`
```
mongodben, line 109:
```suggestion
   :ref:`Open & Close a Realm (Kotlin SDK) <kotlin-open-and-close-a-realm>`
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+          .. tab:: Kotlin SDK
+             :tabid: kotlin-sdk
+
+             .. literalinclude:: /examples/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.realm-object-model.kt
+                :language: kotlin
+                :copyable: false
+
+.. seealso::
+
+   :ref:`Schemas: Supported Types <kotlin-supported-types>`
```
mongodben, line 148:
```suggestion
   :ref:`Supported Schemas Types (Kotlin SDK) <kotlin-supported-types>`
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+
+             .. literalinclude:: /examples/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.write-sync.kt
+                :language: kotlin
+                :copyable: false
+
+.. seealso::
+
+   - :ref:`Create a New Object <kotlin-create-a-new-object>`
+   - :ref:`Modify an Object <kotlin-modify-an-object>`
+   - :ref:`Delete an Object <kotlin-delete-an-object>`
```
mongodben, line 341:
```suggestion
   - :ref:`Create a New Object (Kotlin SDK) <kotlin-create-a-new-object>`
   - :ref:`Modify an Object (Kotlin SDK) <kotlin-modify-an-object>`
   - :ref:`Delete an Object (Kotlin SDK) <kotlin-delete-an-object>`
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+          .. tab:: Kotlin SDK
+             :tabid: kotlin-sdk
+
+             .. literalinclude:: /examples/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.query-sort-limit-distinct.kt
+                :language: kotlin
+                :copyable: false
+
+.. seealso::
+
+   :ref:`Realm Query Language <kotlin-query-language>`
```
mongodben, line 435:
```suggestion
   :ref:`Realm Query Language (Kotlin SDK) <kotlin-query-language>`
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+
+             .. literalinclude:: /examples/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.deletes.kt
+                :language: kotlin
+                :copyable: false
+
+.. seealso::
+
+   - :ref:`Delete all Objects of a Type <kotlin-delete-all-objects-of-a-type>`
+   - :ref:`Delete Multiple Objects <kotlin-delete-multiple-objects>`
+   - :ref:`Delete an Object <kotlin-delete-an-object>`
```
mongodben, line 477:
```suggestion
   - :ref:`Delete all Objects of a Type (Kotlin SDK) <kotlin-delete-all-objects-of-a-type>`
   - :ref:`Delete Multiple Objects (Kotlin SDK) <kotlin-delete-multiple-objects>`
   - :ref:`Delete an Object (Kotlin SDK) <kotlin-delete-an-object>`
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Good point. I just added a note above; take a look when you get the chance.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+.. _kotlin-migration-threading:
+
+Threading
+---------
+
+With the Java SDK, realms, Realm objects, and results cannot be passed
+between threads. The Kotlin SDK freezes these objects by default, making
+them thread-safe. Unlike the live objects used by the Java SDK, the
+frozen objects found in the Kotlin SDK do not automatically update when
+underlying data changes. You must use notifications to subscribe to
```
mongodben, line 544:
```suggestion
underlying data changes. With the Kotlin SDK, you must use notifications to subscribe to
```

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
```diff
@@ -0,0 +1,600 @@
...
+          .. tab:: Kotlin SDK
+             :tabid: kotlin-sdk
+
+             .. literalinclude:: /examples/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.threading.kt
+                :language: kotlin
+                :copyable: false
+
+.. seealso::
+
+   :ref:`Frozen Architecture <kotlin-frozen-architecture>`
```
mongodben, line 577:
```suggestion
   :ref:`Frozen Architecture (Kotlin SDK) <kotlin-frozen-architecture>`
```

mongodben:
great work! i think there could be some additional contextualization near the top as noted in the comments.

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
mongodben: i'd add a link to the kotlin SDK docs homepage toward the bottom of the note. otherwise seems good to me

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
mongodben: hmm idk. why don't you ask ian/product for their thoughts there so we don't have to make marketing-y decisions. 

that can be a fast (or not so fast follow) to this PR anyways.

nathan-contino-mongo: This is now covered in the intro note.

mongodben (approved):
lgtm!

see comments abt adding a link to note and how to contextualize the migration

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
nathan-contino-mongo: Added! Especially since this page is where many folks will first see the Kotlin docs, I think a link to the landing page is a good way to go. Also adding a link at the bottom about "next steps".

source/sdk/kotlin/migrate-from-java-sdk-to-kotlin-sdk.txt
cmelchior: I guess there is a bit confusion between extension and inheritance...they look similar in Kotlin

```
// Inheritence
class Person : RealmObject { }

// Extending base class
class Person : RealmObject() { }
```

I think people mostly complained about this out of principle (rather than a practical concern)... which was why we introduced the `RealmModel` interface in Java. 

For Kotlin we are taking advantage of "default methods" in interfaces which means that we can provide functionality even though you just implement an interface.

The primary concern people have is that by extending a base class you are tying the class to the library, but you are doing that just by using the interface and by using our annotations...so the argument is a pretty flawed...but people were pretty vocal about it. 

This is also why we are probably going to provide projections at some point that doesn't have any Realm code at all, so it is possible to completely separate classes from Realm...but that is a future project.

Also, base classes prevent you from having your own base class...which we don't support right now...but is something we also want to do.


But to my comment... I guess I was mostly objecting to use the word `extend` which indicates a base class. You would use `X implements Y` if it is an interface.

nathan-contino-mongo: This makes perfect sense, thanks for all of the explanation. I've got a fix coming shortly that adds a *little* bit of this and frames it as an advantage.

PR #1708, opened 2022-03-04
Kotlin prereq

nathan-contino-mongo:
### Jira

- Miscellaneous Eng-inspired fixes

- [Kotlin SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/kotlin-prereq/sdk/kotlin/)

### Review Guidelines

### DIFF
--- a/docusaurus/docs/generated/kotlin/CRUDTest.codeblock.delete-all-objects-of-a-type.kt
+++ b/docusaurus/docs/generated/kotlin/CRUDTest.codeblock.delete-all-objects-of-a-type.kt
@@ -2,5 +2,5 @@ realm.write {
     // fetch all frogs from the realm
     val frogs: RealmResults<Frog> = this.query<Frog>().find()
     // call delete on the results of a query to delete those objects permanently
-    frogs.delete()
+    frogs.also { delete(it) }
 }
--- a/docusaurus/docs/generated/kotlin/CRUDTest.codeblock.delete-all-objects-of-a-type.kt
+++ b/docusaurus/docs/generated/kotlin/CRUDTest.codeblock.delete-all-objects-of-a-type.kt
@@ -2,5 +2,5 @@ realm.write {
     // fetch all frogs from the realm
     val frogs: RealmResults<Frog> = this.query<Frog>().find()
     // call delete on the results of a query to delete those objects permanently
-    frogs.also { delete(it) }
+    delete(frogs)
 }
--- a/docusaurus/docs/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.deletes.kt
+++ b/docusaurus/docs/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.deletes.kt
@@ -6,7 +6,7 @@ val sample: Sample? =
 realm.writeBlocking {
     if (sample != null) {
         findLatest(sample)
-            .also { delete(it!!) }
+            ?.also { delete(it) }
     }
 }
 
@@ -16,6 +16,6 @@ GlobalScope.launch {
         query<Sample>()
             .first()
             .find()
-            .also { delete(it!!) }
+            ?.also { delete(it) }
     }
 }
--- a/docusaurus/docs/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.deletes.kt
+++ b/docusaurus/docs/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.deletes.kt
@@ -4,9 +4,10 @@ val sample: Sample? =
 
 // delete one object synchronously
 realm.writeBlocking {
-    val liveSample: Sample? =
-        this.findLatest(sample!!)
-    liveSample?.delete()
+    if (sample != null) {
+        findLatest(sample)
+            .also { delete(it!!) }
+    }
 }
 
 // delete a query result asynchronously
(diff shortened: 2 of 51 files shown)


docusaurus/docs/generated/kotlin/CRUDTest.codeblock.delete-all-objects-of-a-type.kt
```diff
@@ -2,5 +2,5 @@ realm.write {
     // fetch all frogs from the realm
     val frogs: RealmResults<Frog> = this.query<Frog>().find()
     // call delete on the results of a query to delete those objects permanently
-    frogs.delete()
+    frogs.also { delete(it) }
```
cmelchior, line 5:
```suggestion
    delete(frogs)
```

docusaurus/docs/generated/kotlin/MigrateFromJavaToKotlinSDKTest.codeblock.deletes.kt
```diff
@@ -4,9 +4,10 @@ val sample: Sample? =
 
 // delete one object synchronously
 realm.writeBlocking {
-    val liveSample: Sample? =
-        this.findLatest(sample!!)
-    liveSample?.delete()
+    if (sample != null) {
+        findLatest(sample)
+            .also { delete(it!!) }
```
cmelchior, line 9:
```suggestion
        findLatest(sample)?.also { delete(it) }
```

docusaurus/docs/generated/kotlin/QuickStartTest.codeblock.change-listeners.kt
```diff
@@ -1,5 +1,4 @@
-val config = RealmConfiguration.Builder()
-    .schema(setOf(Task::class))
+val config = RealmConfiguration.Builder(setOf(Task::class))
```
cmelchior, line 1: For this...the easy way is `RealmConfiguration.with(schema = setOf(Task::class))`

docusaurus/docs/generated/kotlin/SyncTest.codeblock.configure-a-synced-realm.kt
```diff
@@ -1,10 +1,9 @@
 val app = App.create(YOUR_APP_ID)
-runBlocking() {
+runBlocking {
     val user = app.login(Credentials.anonymous())
-    val config = SyncConfiguration.Builder(user, PARTITION)
+    val config = SyncConfiguration.Builder(user, PARTITION, setOf(/*realm object models here*/))
         // specify name so realm doesn't just use the "default.realm" file for this user
         .name(PARTITION)
```
cmelchior, line 6: This is done automatically, so only needed if you want another Realm name than one based on the partition value

cmelchior (approved):

docusaurus/docs/generated/kotlin/QuickStartTest.codeblock.change-listeners.kt
nathan-contino-mongo: It seems that in the JVM environment in Android studio testing, the SDK likes to dump all realm files right in the middle of my source code by default. So I have a hidden line that sets the directory/name of each file manually for now. When/if that behavior changes, i can move this -- but I'd rather have the real codeblock be tested, rather than replace the entire instantiation with a different piece of logic.

docusaurus/docs/generated/kotlin/SyncTest.codeblock.configure-a-synced-realm.kt
nathan-contino-mongo: In my experience, synced realms all use "default.realm" as the default name, rather than the partition key.

cmelchior: That is true for Flexible Sync, but shouldn't be the case for partition-based sync?

cmelchior: But no matter what the Realms should be put in a unique location `<appId>/<userId>/<partition>.realm` so there is very few reasons to want to to override the name.

nathan-contino-mongo: issue i experienced was that if i opened one realm with a user, then another, they would both try to use the same default.realm file in that user's directory. So I'm pretty sure it's a bug with the current partition-based implementation in the SDK.

nathan-contino-mongo: And I still think flexible sync needs to generate some kind of unique ID based on the query for each query to avoid this same issue!

cmelchior: Please report issues like this if you see them...this is the first time I'm hearing it is a potential problem and it isn't supposed to work that way, so if it does, it is a bug. I created https://github.com/realm/realm-java/issues/7653 so we can investigate. If you can add some more details there, it would be great.

Flexible Sync Realms have all queries in the same Realm...you effectively just have 1 Realm pr. user, so there using `default.realm` should be fine.

cmelchior: In either case...it isn't technically wrong to have `name()` here...It just isn't a pattern that is strictly needed.

MongoCaleb (approved):

PR #1710, opened 2022-03-04
Get files in subdirectories

mongodben:
Make Docusaurus staging action get files in subdirectories.

right now it only gets files in `docusaurus`, not `docusaurus/path/to/myfile.md`.

this should fix per https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet


nathan-contino-mongo (approved):
LGTM

nathan-contino-mongo:
Good catch, I should have noticed this on the other PR.

PR #1715, opened 2022-03-07
Upgrade actions/checkout to v3

cbush:
Checking if v3 supports the pull_request_target event to actually check out the PR branch by default


nathan-contino-mongo (approved):
LGTM

cbush:
Boo, turns out my hypothesis was incorrect! Going with solution here: https://github.com/actions/checkout/issues/518#issuecomment-890401887

Since we want the same "default" behavior of checkout on the pull_request event, we're checking out the merge commit.

PR #1736, opened 2022-03-11
Versioned-supported sites for Java & Kotlin SDKs

mongodben:
## Pull Request Info

Add versioned sites for Java and Kotlin SDKs. Didn't create any versions. This is necessary set up for creating SDK version selector. 

Note that this change removes the description text from the versions on the site homepage. I tried to add the descriptions back, but wasn't able to come up with an easy solution. I think it's fine for our current POC.


nathan-contino-mongo (approved):
LGTM, with one nit -- it looks like we've lost the nice descriptions on http://realm-io-staging.s3-website-us-east-1.amazonaws.com/mongodben/docs-realm/versioned_sites_java_kotlin/index.html for the SDK cards. It *should* inherit the titles from the 'description' front matter field -- can you look into this a little bit before merging? Not a huge deal, but would be nice.

mongodben:
yep that was noted in the PR. i believe that's lost b/c now the homepage doesn't have access to the docs themselves, it's only links to them. this is b/c these docs are now living in separate sites.

docusaurus/sidebars.js
```diff
@@ -27,16 +27,27 @@ const sidebars = {
           type: "link",
           label: "Java SDK",
           href: "/sdk/java/intro",
+          customProps: {
+            description:
+              "Build Android applications with Realm in Java or Kotlin.",
```
nathan-contino-mongo, line 32: beautiful

PR #1743, opened 2022-03-14
Prototype Docusaurus version changer

mongodben:
## Pull Request Info

Create prototype of the SDK version changer. 

note that this is just a prototype and the code will need significant refactoring for the production site. it should also be noted that the version selector does not currently work on the mobile site. this functionality should also be added. 

### Review Guidelines

### DIFF
--- a/docusaurus/src/components/VersionChanger.tsx
+++ b/docusaurus/src/components/VersionChanger.tsx
@@ -1,8 +1,7 @@
 import { useHistory } from "react-router-dom";
 import React, { useState, useEffect, Fragment, memo } from "react";
 import { useAllPluginInstancesData } from "@docusaurus/useGlobalData";
-import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
-import { versions } from "process";
+import Select, { StylesConfig } from "react-select";
 
 const CONTEXTS = [
   {
@@ -18,6 +17,17 @@ const CONTEXTS = [
     name: "Flutter",
   },
 ];
+
+const colorStyles: StylesConfig = {
+  control: (styles) => ({ ...styles, backgroundColor: "white" }),
+  option: (styles, { data, isDisabled, isFocused, isSelected }) => {
+    return {
+      ...styles,
+      backgroundColor: isSelected ? "#f8aaca" : undefined,
+      color: isSelected ? "white" : "#f15192",
+    };
+  },
+};
 
 const getContext = (id) => CONTEXTS.find((context) => context.id === id);
 
@@ -36,7 +46,6 @@ export const getCurrentSiteInstance = (history) => {
     const sdkWordIdx = splitPath.findIndex((value) => value === "sdk");
     const sdkInstance = splitPath[sdkWordIdx + 1];
     const sdkPage = splitPath.slice(sdkWordIdx + 2).join("/");
-    console.log(sdkPage);
     const sdkVersion = isNaN(splitPath[sdkWordIdx + 2][0])
       ? "current"
       : splitPath[sdkWordIdx + 2];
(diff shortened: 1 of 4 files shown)


docusaurus/src/components/VersionChanger.tsx
```diff
@@ -0,0 +1,165 @@
...
+  option: (styles, { data, isDisabled, isFocused, isSelected }) => {
+    return {
+      ...styles,
+      backgroundColor: isSelected ? "#f8aaca" : undefined,
+      color: isSelected ? "white" : "#f15192",
+    };
+  },
+};
+
+const getContext = (id) => CONTEXTS.find((context) => context.id === id);
```
cbush, line 32: Why not make CONTEXTS a `Record<string /* id */, SdkContext>` (or Map)?

docusaurus/src/components/VersionChanger.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const {
+    location: { pathname },
+  } = history;
+  const splitPath = pathname.split("/").slice(1);
+  if (!splitPath.includes("sdk")) {
+    return {
+      sdkInstance: "default",
+      sdkPage: null,
+    };
+  } else {
```
cbush, line 44: No need for else after early return

docusaurus/src/components/VersionChanger.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const sdkWordIdx = splitPath.findIndex((value) => value === "sdk");
+    const sdkInstance = splitPath[sdkWordIdx + 1];
+    const sdkPage = splitPath.slice(sdkWordIdx + 2).join("/");
+    const sdkVersion = isNaN(splitPath[sdkWordIdx + 2][0])
+      ? "current"
+      : splitPath[sdkWordIdx + 2];
+    return {
```
cbush, line 51: A programmer has a problem. "I know," they say, "I'll use a regex." They now have two problems. (I would have used a regex/match() for all of this)

docusaurus/src/components/VersionChanger.tsx
```diff
@@ -0,0 +1,165 @@
...
+      sdkVersion,
+    };
+  }
+};
+
+const pathExists = (path, data) => {
+  return data.docs.some((doc) => doc.path === path);
+};
+
+function SDKSelector({ sdkSitesMetaData, activeSDK }) {
```
cbush, line 63: Nitpick but `SdkSelector` would have been my style choice here. Then you don't see "SDKS Elector".

docusaurus/src/components/VersionChanger.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const history = useHistory();
+  function changeHandler({ value: sdkName }) {
+    const sdkMetadata = sdkSitesMetaData[sdkName];
+    const sdkBasePage = sdkMetadata.path + "intro"; //TODO: refactor into smthn better
+    history.push(sdkBasePage);
+  }
+  delete sdkSitesMetaData.default;
```
cbush, line 70: Why? Should have a comment.

This actually modifies the input instance which seems questionable.

To copy everything but one property, you could do something like:

```js
const {default, ...sdkSitesMetaDataCopy} = sdkSitesMetaData;
```

docusaurus/src/components/VersionChanger.tsx
```diff
@@ -0,0 +1,165 @@
...
+  useEffect(() => {
+    const { sdkPage } = getCurrentSiteInstance(history);
+
+    const currContext = getContext(sdkPage);
+    if (currContext && currContext.id !== context.id) {
+      setContext(currContext);
+    }
+    const { sdkInstance: sdkName } = getCurrentSiteInstance(history);
+    setVersions(data[sdkName]?.versions);
+  }, []);
```
cbush, line 140: Shouldn't "history" be in the deps list?
```suggestion
  }, [history]);
```

cbush (approved):
Looks great for a POC!

Your TypeScript config doesn't seem to be as strict as I would like. It should disallow any, require return types declared on exported functions, etc. etc. Fine for a POC but will need to update later, and I don't want to have things that "work" suddenly "break" when we do make the compiler point out some bugs.

mongodben:
>Your TypeScript config doesn't seem to be as strict as I would like. It should disallow any, require return types declared on exported functions, etc. etc. Fine for a POC but will need to update later, and I don't want to have things that "work" suddenly "break" when we do make the compiler point out some bugs.

makes sense to me. i didn't add any TS config myself. i would assume it's either inherited from docusaurus or something in my dev env. 

before i merge, in addition to implementing the changes in your comments, going to add a note to the top of the component file about how this is just a prototype and refer back to this PR for more details/conversation.

docusaurus/src/components/VersionChanger.tsx
mongodben: going to keep this as is for now, b/c the final implementation should be based on some existing docusaurus hooks to get site instance data.

in particular, i think we should use the `useAllPluginInstancesData("docusaurus-plugin-content-docs")` to generate the list of context.

docusaurus/src/components/VersionChanger.tsx
mongodben: going to keep this as is for now, b/c the final implementation should be based on some existing docusaurus hooks to get site instance data.

in particular, i think we should use the hook `useActiveDocContext(SDK_SITE_ID)` to get this information.

cbush: ???

cbush: I was referring to the split()/indexing and all that. A regex could match the /sdk/ and then extract the sdk name and version number.

That said, per my joke above, I wouldn't bother. ⛳️

mongodben: > ???
i think we can get the active doc from `useActiveDocContext` rather than parsing the path. w/ that being said, i agree regex would be a better way of parsing path, per your suggestion, just not right now since the whole path-parsing implementation will become null w the hook implementation.

mongodben: i've become unfortunately well acquainted w regex recently - https://github.com/mongodben/jurassic-park/blob/main/src/lib/parsers/snooty-to-pandoc-rst/replace-link-roles-with-links.ts#L1

PR #1753, opened 2022-03-16
(DOCSP-21333): Add tip about changing property name without triggering migration

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-21333

- [Update Your Schema](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21333/sync/data-model/update-schema/#change-a-property-name): New tip in the section about how to rename a property

### Review Guidelines

### DIFF
--- a/source/sync/data-model/update-schema.txt
+++ b/source/sync/data-model/update-schema.txt
@@ -119,6 +119,12 @@ the client-side object model. For more details, see relevant sections below.
      - Breaking
      - Breaking
 
+.. tip:: Rename a property
+
+   While renaming a property is a breaking change, some of the Realm SDKs
+   offer a workaround to remap the property name. See 
+   :ref:`schema-update-change-property-name` for more details.
+
 .. _schema-update-add-object-type:
 
 Add an Object Type
--- a/source/sync/data-model/update-schema.txt
+++ b/source/sync/data-model/update-schema.txt
@@ -107,8 +107,8 @@ the client-side object model. For more details, see relevant sections below.
      - Breaking
      - Breaking
    * - :ref:`Change a property name <schema-update-change-property-name>`
-     - Breaking
-     - Breaking
+     - Breaking*
+     - Breaking*
    * - :ref:`Change property type but keep the name <schema-update-change-property-type-but-keep-name>`
      - Breaking
      - Breaking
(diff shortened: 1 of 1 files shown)


source/sync/data-model/update-schema.txt
```diff
@@ -107,8 +107,8 @@ the client-side object model. For more details, see relevant sections below.
      - Breaking
      - Breaking
    * - :ref:`Change a property name <schema-update-change-property-name>`
-     - Breaking
-     - Breaking
+     - Breaking*
```
MongoCaleb, line 110: Why these added *s? I was expecting a footnote below the table, or at the very least an explanation of why the * in the linked section.

source/sync/data-model/update-schema.txt
```diff
@@ -414,6 +414,20 @@ to both the server-side schema and the client-side object model.
 How to Change a Property Name
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. tip:: Change a property name
```
MongoCaleb, line 417: Ah, I see. This answers the above question. I guess if I overlooked it, others might as well. Would it make sense to change the table to "Breaking in some SDKs" or similar? Perhaps a footnote under the table?

source/sync/data-model/update-schema.txt
```diff
@@ -414,6 +414,20 @@ to both the server-side schema and the client-side object model.
...
+
+   While changing a property name is a breaking change in both the server-side
```
MongoCaleb, line 419: Just toying here with a little rewrite; take it or leave it as you wish. 

Changing a property name is a breaking change except in the following SDKs,
which provide an API to map the name defined in a model class to a different internal name. 
This API allows you to avoid renaming a property and triggering migrations. SDKs that provide this additional API are:

MongoCaleb (approved):
Approved with a few sugestions.

source/sync/data-model/update-schema.txt
dacharyc: It's a breaking change in all SDKs, but this is a workaround available to some. But yeah, I think a footnote under the table is the good idea. I originally had this note there, and then moved it down and forgot to deal with the * in the table.

source/sync/data-model/update-schema.txt
dacharyc: I think I see what you're going for, but I don't consider this a "breaking change except in these SDKs" thing so much as a "this is a breaking change, but here's a workaround you might prefer."

PR #1765, opened 2022-03-17
(DOCSP-21382): New section with embedded object relationship examples

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-21382
- https://jira.mongodb.org/browse/DOCSP-21383

- [Relationships](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21382/schemas/relationships/#embedded-object-relationships-with-foreign-collections): New section for Embedded Object Relationships

### Review Guidelines

### DIFF
--- a/source/schemas/relationships.txt
+++ b/source/schemas/relationships.txt
@@ -322,6 +322,11 @@ you can specify the foreign collection details and foreign key field.
      }
    }
 
+.. tip:: Use the same syntax with dictionaries and sets
+
+   You can use this same ``field1.[].field2`` syntax when creating relationships
+   within dictionaries and sets.
+
 .. _define-a-relationship:
 
 Define a Relationship
--- a/source/schemas/relationships.txt
+++ b/source/schemas/relationships.txt
@@ -312,6 +312,11 @@ To access a embedded object property contained in a list, use:
 ``field1.[].field2``, e.g. ``pets.[].favoriteToyBrand``. From there, 
 you can specify the foreign collection details and foreign key field.
 
+.. tip:: Use the same syntax with dictionaries and sets
+
+   You can use this same ``field1.[].field2`` syntax when creating relationships
+   within dictionaries and sets.
+
 .. code-block:: json
 
    {
@@ -322,10 +327,12 @@ you can specify the foreign collection details and foreign key field.
      }
    }
 
-.. tip:: Use the same syntax with dictionaries and sets
+.. note:: Primitives versus lists, dictionaries, and sets in relationships
 
-   You can use this same ``field1.[].field2`` syntax when creating relationships
-   within dictionaries and sets.
+   In the example above, ``is_list`` is set to false. The field at the end
+   of the relationship string here is a primitive, not a list. The embedded 
+   object is *contained* in a list, but the ``favoriteToyBrand`` property 
+   itself is not a dictionary, set or list.
 
 .. _define-a-relationship:
 
--- a/source/schemas/relationships.txt
+++ b/source/schemas/relationships.txt
@@ -242,8 +242,7 @@ Embedded Object Relationships
 -----------------------------
 
 Embedded Objects can have relationships with foreign collections. Use 
-dot notation to access properties in embedded objects when creating
-relationships with foreign collections.
+dot notation to access properties in embedded objects.
 
 .. _relationships-from-embedded-object-to-another-collection:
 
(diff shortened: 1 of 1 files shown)


source/schemas/relationships.txt
```diff
@@ -238,6 +238,91 @@ or a null value in resolved GraphQL types and SDK models:
            }
          }
 
+Embedded Object Relationships
+-----------------------------
+
+Embedded Objects can have relationships with foreign collections. Use 
+dot notation to access properties in embedded objects when creating
+relationships with foreign collections.
```
nathan-contino-mongo, line 246:
```suggestion
dot notation to access properties in embedded objects.
```

Maybe I'm missing something, but the extra clause on this sentence:
- doesn't seem to add anything (don't you *always* use dot notation when referencing embedded objects?)
- makes the sentence less readable

So I suggest removing it. If it does add something, maybe we can break the sentence up?

source/schemas/relationships.txt
```diff
@@ -238,6 +238,91 @@ or a null value in resolved GraphQL types and SDK models:
...
+
+.. _relationships-from-embedded-object-to-another-collection:
+
+Embedded Object to Another Collection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An embedded object can have a relationship with an object in a foreign 
+collection.
```
nathan-contino-mongo, line 254:
```suggestion
collection:
```

Maybe other examples on this page don't use colons for the transition, but it feels very natural here. Could always update the other examples, too.

source/schemas/relationships.txt
```diff
@@ -238,6 +238,91 @@ or a null value in resolved GraphQL types and SDK models:
...
+             "favoriteToyBrand": { "bsonType": "objectId" }
+           }
+         }
+       }
+      // ...additional model properties
+     }
+   }
+
+The dot notation to access a embedded object property contained in a list 
+is: ``pets.[].favoriteToyBrand``. From there, you can specify the
```
nathan-contino-mongo, line 313:
```suggestion
To access an embedded object property contained in a list, use ``.[].``, e.g.
``pets.[].favoriteToyBrand``. From there, you can specify the 
```
Passive -> active reword suggestion. Take it or leave it or change it, up to you.
"a" -> "an", should probably take it.

nathan-contino-mongo (approved):
LGTM with some smol nits. Nice work!

source/schemas/relationships.txt
```diff
@@ -238,6 +238,90 @@ or a null value in resolved GraphQL types and SDK models:
...
+         "foreign_key": "_id", 
+         "is_list": false }
+   }
+
+.. _relationships-to-embedded-object-within-list:
+
+Embedded Object Within a List
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An embedded object that is within a list property can have a relationship
```
tkaye407, line 289: Looks good to me. Dont think you need another example but you can add a note that dictionaries and sets use the same [] syntax

tkaye407 (approved):

source/schemas/relationships.txt
```diff
@@ -238,6 +238,90 @@ or a null value in resolved GraphQL types and SDK models:
...
+``field1.[].field2``, e.g. ``pets.[].favoriteToyBrand``. From there, 
+you can specify the foreign collection details and foreign key field.
+
+.. code-block:: json
+
+   {
+     "pets.[].favoriteToyBrand": {
+       "ref": "#/relationship/mongodb-atlas/example/ToyBrand",
+       "foreign_key": "_id",
+       "is_list": false
```
kmorkos, line 321: @tkaye407 is this supposed to be `"is_list": true`?

source/schemas/relationships.txt
dacharyc: Good note, I'll add that - thanks!

source/schemas/relationships.txt
dacharyc: I had assumed in this example we're referring to the foreign_key field as a list (or not) - in which case the ToyBrand `_id` is unlikely to be a list. But it's very possible my understanding is backwards!

tkaye407: Nope. Is_List refers to if the field at the end of the relationship string is a list. So if favoriteToyBranc was itself a list / set / dictionary, then is_list would need to be true. Here, the property being pointed to: "pets.[].favoriteToyBrand" is just a primitive

tkaye407: Perhaps its worth clarifying (a) that is_list is also true for sets / dictionaries and (b) that it refers to whether the property itself is a list / set / dictionary, not if it is withing a list of embedded objects

kmorkos: ah i see the distinction - thanks for confirming!

PR #1818, opened 2022-04-01
(DOCSP-13022): Encrypted realms cannot be opened across processes

cbush:
### Jira

- https://jira.mongodb.org/browse/DOCSP-13022

- https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/encrypt-caveat/

### Review Guidelines


source/sdk/java/advanced-guides/encryption.txt
```diff
@@ -143,3 +143,10 @@ method:
          :language: java
          :copyable: false
 
+Accessing an Encrypted Realm from Multiple Process
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Realm forbids opening the same encrypted realm from multiple processes.
+Attempting to do so will throw the error: "Encrypted interprocess sharing is
+currently unsupported."
```
nathan-contino-mongo, line 151: Not the wording of this error on the Java SDK, since I've seen this error before. Skeptical if we should claim a specific error message (for any SDK!) because these are subject to change with any update.

nathan-contino-mongo (commented):
See comment -- I'm not sure if we should claim a specific error message when that message doesn't seem to be in any of the SDKs. Funnily enough, I *can* find that exact error message in Java... but only in a changelog entry from 2016, not in an actual exception class anywhere. I wonder if this information is better suited to a "warning" admonition on the fundamentals Realm page discussion of encrypted realms.

source/sdk/java/advanced-guides/encryption.txt
cbush: I think it comes from the core. We could also put something like "something like"? The reason for including it at all would be for searchability when one does encounter this error.

nathan-contino-mongo: Ah, strange. I don't see it there either: https://github.com/realm/realm-core/search?q=interprocess

But maybe it's just not showing up in github's search?

cbush: This seems to be where it's coming from: https://github.com/realm/realm-core-private/blob/812f88ae6c1cf0929861e05739153b5196ba2830/src/realm/db.cpp#L1184-L1190

nathan-contino-mongo: LOL github search fail. In that case, approving now!

nathan-contino-mongo (approved):
LGTM. Though I do still think this is suited well to an admonition on the realm database page (bonus: that's shared across SDKs, so you only have to write it once)

source/sdk/java/advanced-guides/encryption.txt
cbush: To be fair this is in the realm-core-private repo, not realm-core 😵‍💫

PR #1832, opened 2022-04-06
(DOCSP-21708): Use endpoints `response` object with  function returns as body

mongodben:
## Pull Request Info

update docs to reflect the fact that users can use the endpoint function return value along with the function's `response` object methods. 

(previously it'd been only the response object or the endpoint function return value, not combo of both)

### Jira

- https://jira.mongodb.org/browse/DOCSP-21708

- [Create HTTPS Endpoint](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21708/endpoints/configure/#configure-the-endpoint-response)
- [Convert Webhooks to Endpoints](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21708/services/convert-webhooks-to-endpoints/)

### Review Guidelines

### DIFF
--- a/source/includes/steps-create-an-endpoint-cli.yaml
+++ b/source/includes/steps-create-an-endpoint-cli.yaml
@@ -21,7 +21,7 @@ content: |
   To create a new endpoint, add an :ref:`endpoint configuration
   <appconfig-endpoints>` object to the array in
   :file:`/http_endpoints/config.json`.
-  
+
   The configuration file has the following form:
 
   .. code-block:: json
@@ -47,7 +47,7 @@ content: |
   An endpoint's route determines the URL used to call the endpoint. You
   can define at most one handler for a given route and HTTP method
   combination in your app.
-  
+
   1. Enter a name for the endpoint in ``route`` field. Route
      names must begin with a forward slash (``/``) and may contain
      additional forward slashes to indicate a nested path.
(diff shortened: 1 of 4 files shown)


source/includes/steps-create-an-endpoint-cli.yaml
```diff
@@ -85,14 +85,15 @@ ref: configure-the-endpoint-response
...
   :mdn:`HTTP 200 <Web/HTTP/Status/200>` response that includes the
   endpoint function return value as its ``body`` field.
-  
+
   You can configure a custom HTTP response from within the endpoint
-  function using the ``response`` object. For more information, see
-  :ref:`endpoint-response-object`.
+  function using the ``response`` object. If you use ``response.setBody(value)``,
+  the endpoint returns that value, not the function return value.
+  For more information, see :ref:`endpoint-response-object`.
```
jo-tse-mongo, line 96: I think to kush’s point, maybe adding a reference to the new `Differing Behavior` section in the `endpoint-response-object` page would help also 
If I were just looking at this page, I'd think to go to the `endpoint-response-object` doc page for more information on the `response.SetBody` behavior and why it works that way

mongodben: good point. will add

source/services/convert-webhooks-to-endpoints.txt
```diff
@@ -186,3 +186,13 @@ To migrate to the converted HTTPS endpoint URLs:
...
+
+.. _differences-webhooks-converted-endpoints: 
+
+Differences between Webhooks and Their Converted Endpoints
+-----------------------------------------------------------------
+
+If you configured your converted HTTPS endpoint to use :ref:`Respond with Result <configure-endpoints>`
+and the webhook function it's based on returns a value but does not invoke ``response.setBody()``,
+the generated endpoint includes the function's return value as its response body.
+The webhook, in contrast, only returns a body specified by ``response.setBody()``.
```
kpatel71716, line 198: This looks good, but might be worth clarifying that if a webhook function calls setHeader or setStatusCode and returns a value instead of using response.SetBody, the response value will not be sent. If the webhook is converted, then the returned value of the function will be sent as the response body. It's a minor edge case where a user might just see data they didn't for a webhook. Let me know if this doesn't make sense and I can clarify

E.x:
```exports = function({ query, headers, body}, response){
  response.setStatusCode(200)
  return "HI"
};```
will return no response body for a webhook, but the converted endpoint response body will be `"HI"`

mongodben: added a brief example per your comment. please take a look and lmk what you think!

kpatel71716: This looks great to me! @jo-tse-mongo let us know if this looks good to you as well, thanks Ben!

nathan-contino-mongo (approved):
LGTM

PR #1847, opened 2022-04-14
Update swift examples README for SPM + other hints

cbush:


examples/ios/README.md
```diff
@@ -140,7 +149,7 @@ The `QuickStartSwiftUI` directory -- `QuickStartSwiftUI/QuickStart.swift` in
 particular -- contains the code for the SwiftUI and Combine quick start. To run,
 open the workspace in Xcode and select the "QuickStartSwiftUI" target.
 
->⚠️ Note: you will need an iOS 14.0+ simulator or device to run this target.
+> ⚠️ Note: you will need an iOS 14.0+ simulator or device to run this target.
```
dacharyc, line 152: Have you verified it builds with 14.0? There was a dependency on 15.x versions which might have been removed now - I am still building with 15.0 so am not sure what is actually required at this point.

dacharyc (approved):
Overall LGTM, just one question about what iOS target version we actually need to use when running the tests.

PR #1852, opened 2022-04-14
remove alt apostrophes

mongodben:
### Jira

n/a

n/a
### Review Guidelines

### DIFF
--- a/source/sdk/swift/examples/read-and-write-data.txt
+++ b/source/sdk/swift/examples/read-and-write-data.txt
@@ -322,7 +322,7 @@ query.
 .. example::
 
    To get a result set for tan dogs, and tan dogs whose names start with
-   ‘B’, chain two queries like this:
+   ‘B', chain two queries like this:
 
    .. tabs::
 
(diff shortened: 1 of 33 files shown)


source/sdk/swift/examples/read-and-write-data.txt
```diff
@@ -322,7 +322,7 @@ query.
 .. example::
 
    To get a result set for tan dogs, and tan dogs whose names start with
-   ‘B’, chain two queries like this:
+   ‘B', chain two queries like this:
```
dacharyc, line 325:
```suggestion
   'B', chain two queries like this:
```
Weird that find and replace caught one of those but not the other!

dacharyc (approved):
Awesome! Just one spot where find-and-replace missed one but I'm so happy you did this ✋

source/sdk/swift/examples/read-and-write-data.txt
mongodben: nice catch! b/c this is a different character! `‘` vs `’`

mongodben: did a global search in repo for `‘` and this was actually the only occurrence.

PR #1861, opened 2022-04-20
(DOCSP-22153): Fix Broken JS/TS tests

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-22153

N/A

### Review Guidelines


examples/node/.gitignore
```diff
@@ -1,11 +1,12 @@
 .DS_Store/
 node_modules/
 mongodb-realm/
-.env
-/default.realm
-/default.realm.lock
-/default.realm.management/
-/mongodb-realm.lock
+/data-type-realm
```
nlarew, line 4:
```suggestion
# Ignore .realm files
/data-type-realm
```

Maybe we can indicate that these should each be a `path` we use in a test?

examples/node/Examples/data-types.js
```diff
@@ -53,11 +53,6 @@ const PetOwnerSchema = {
 // :code-block-end:
 
 describe("Node.js Data Types", () => {
-  afterEach(() => {
-    if (realm && !realm.isClosed) {
-      realm.close();
-    }
-  });
```
nlarew, line 60: Since we're no longer cleaning up a shared reference to the `realm` variable after each test, can we delete the shared variable (i.e. `let realm;` at the top of the file) and declare a new block-level `realm` variable for each test?

i.e. instead of
```js
realm = await Realm.open({
    path: "data-type-realm",
    schema: [PersonSchema],
});
```
we would add a const declaration
```js
const realm = await Realm.open({
    path: "data-type-realm",
    schema: [PersonSchema],
});
```
for every test on this page

examples/node/.gitignore
nlarew: Alternatively, can we change all the realm `path` strings to used a single nested directory and then we just gitignore that? It'll keep our project roots clean & means we don't have to update .gitignore every time we open a new path - just need to enforce the directory convention.

e.g. instead of 
```js
const realm = await Realm.open({
    path: "data-type-realm",
    schema: [PersonSchema],
});
```
we'd do
```js
const realm = await Realm.open({
    path: "realm-files/data-type-realm",
    schema: [PersonSchema],
});
```
for every test

nlarew (approved):
Ran through the tests a few times and everything seems to work! Left a few suggestions but otherwise LGTM.

Also, I noticed that we're still skipping a few tests (3 suites in `test:js` and 1 in `test:ts`) - are we tracking these already / are we able to flip any of them back on too?

examples/node/.gitignore
mongomoe: great idea

mongomoe:
@nlarew regarding the skipped tests, I still need to fix those separately, since they mostly relate to sync and we can now unskip them (after some changes)

PR #1863, opened 2022-04-20
(DOCSP-21149): Refactor to improve readability from grade C to B, or grade 12 to 8

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-21149

- [Reset a Client Realm - Java SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21149/sdk/java/examples/reset-a-client-realm/): Refactor to improve readability from grade C/73% reach to grade B/93% reach in Readable, or grade 12 to 8 in Hemingway. Screenshots in attached Jira ticket.

### Review Guidelines

### DIFF
--- a/source/includes/destructive-schema-change-app-update.rst
+++ b/source/includes/destructive-schema-change-app-update.rst
@@ -3,6 +3,6 @@
    After a breaking schema change:
 
    - All clients must perform a client reset.
-   - Clients affected by the schema change must update their local object models.
+   - You must update client models affected by the breaking schema change.
 
 
--- a/source/includes/destructive-schema-change-app-update.rst
+++ b/source/includes/destructive-schema-change-app-update.rst
@@ -3,7 +3,6 @@
    After a breaking schema change:
 
    - All clients must perform a client reset.
-   - Clients that use models destructively affected by the schema
-     change must update local object models.
+   - Clients affected by the schema change must update their local object models.
 
 
--- a/source/sdk/java/examples/reset-a-client-realm.txt
+++ b/source/sdk/java/examples/reset-a-client-realm.txt
@@ -26,7 +26,7 @@ This error means you must reset the realm file in the client application.
 Clients in this state may continue to run and save data locally. Until you 
 perform the client reset, the realm does not sync with the backend.
 
-Choose a **client reset strategies** to handle client reset errors. These 
+Choose a **client reset strategy** to handle client reset errors. These 
 strategies restore realm to a syncable state, but have tradeoffs:
 
 - :ref:`Discard Unsynced Changes <java-discard-unsynced-changes>`. Restore 
--- a/source/sdk/java/examples/reset-a-client-realm.txt
+++ b/source/sdk/java/examples/reset-a-client-realm.txt
@@ -17,7 +17,7 @@ Reset a Client Realm - Java SDK
    To learn about the causes of and strategies for handling client
    resets, check out the Sync :ref:`Client Resets <client-resets>` page.
 
-The realm file in your client handles reads and writes on the device. When 
+The SDK reads and writes to a realm file on the device. When 
 you use Realm Sync, this local realm syncs with the application backend. 
 Some conditions can cause the realm to be unable to sync with the backend. 
 When this occurs, you get a **client reset error**. 
@@ -55,12 +55,9 @@ This strategy requires minimal code. This strategy performs a reset without
 closing the realm or missing notifications.
 
 It *does* delete all local changes made since the last successful sync. 
-Do not use this strategy if your application cannot lose unsynced data. 
-
-This strategy provides a frozen copy of the realm from before reset. You 
-can attempt to recover changes from this frozen realm. However, the code 
-required for a manual recovery is quite complex. This is not recommended 
-for most applications.
+This includes any data already written to the realm but not yet synced to
+the application backend. Do not use this strategy if your application 
+cannot lose unsynced data. 
 
 Discard unsynced changes cannot handle :ref:`breaking or destructive schema 
 changes <destructive-changes-synced-schema>`. When breaking changes occur, 
@@ -81,7 +78,9 @@ methods:
 - ``onAfterReset()``. The SDK calls this block after successfully executing 
   this strategy. This block provides a frozen copy of the original realm. 
   It also returns a live instance of the realm in a syncable state.
-- ``onError()``: Mimics the "manually recover unsynced changes" strategy. 
+- ``onError()``: Behaves similar to 
+  :java-sdk:`defaultClientResetStrategy() 
+  <io/realm/mongodb/AppConfiguration.Builder.html#defaultSyncClientResetStrategy(io.realm.mongodb.sync.ManuallyRecoverUnsyncedChangesStrategy)>`. 
   The SDK calls this method during a breaking schema change.  
   
   .. seealso::
--- a/source/sdk/java/examples/reset-a-client-realm.txt
+++ b/source/sdk/java/examples/reset-a-client-realm.txt
@@ -17,23 +17,31 @@ Reset a Client Realm - Java SDK
    To learn about the causes of and strategies for handling client
    resets, check out the Sync :ref:`Client Resets <client-resets>` page.
 
-A **client reset error** is a scenario where a client realm cannot sync
-data with the application backend. Clients in this state may continue to
-run and save data locally but cannot send or receive sync changesets
-until they perform a client reset strategy.
+The realm file in your client handles reads and writes on the device. When 
+you use Realm Sync, this local realm syncs with the application backend. 
+Some conditions can cause the realm to be unable to sync with the backend. 
+When this occurs, you get a **client reset error**. 
 
-You can handle client resets in your application using one of the
-available **client reset strategies**:
+This error means you must reset the realm file in the client application. 
+Clients in this state may continue to run and save data locally. Until you 
+perform the client reset, the realm does not sync with the backend.
 
-- :ref:`Discard Unsynced Changes <java-discard-unsynced-changes>`:
-  Restores the realm to a syncable state by discarding changes made
-  since the last sync.
+Choose a **client reset strategies** to handle client reset errors. These 
+strategies restore realm to a syncable state, but have tradeoffs:
 
-- :ref:`Manually Recover Unsynced Changes <java-manually-recover-unsynced-changes>`:
-  Downloads a new copy of the realm, and moves the unsyncable realm
-  to a backup. Migrate unsynced data from the backup copy of the
-  realm to the new syncable copy.
+- :ref:`Discard Unsynced Changes <java-discard-unsynced-changes>`. Restore 
+  Sync by discarding local changes since the last sync. Maintains change 
+  listeners.
+- :ref:`Manually Recover Unsynced Changes 
+  <java-manually-recover-unsynced-changes>`:. Move the unsyncable realm and 
+  download a new copy. Invalidates change listeners.
 
+Both options let you write custom logic to recover local changes. Neither 
+option can recover local changes for you. 
+
+Discard unsynced changes is an improved version of manual recovery. However, 
+this strategy cannot handle every client reset error. You must maintain a 
+manual client reset handler as a fallback.
 
 .. _java-discard-unsynced-changes:
 
@@ -42,26 +50,24 @@ Discard Unsynced Changes
 
 .. versionadded:: 10.10.0
 
-The **discard unsynced changes** client reset strategy helps you perform
-a client reset with minimal code and minimal disruption to your application
-workflow. This strategy restores your local realm to a syncable state without
-closing the realm or missing notifications.
+The **discard unsynced changes** performs client resets "seamlessly." This 
+strategy requires minimal code and minimal disruption. This strategy 
+does not close the realm or miss notifications.
 
-This strategy comes with a downside: it permanently deletes all
-local unsynced changes made since the last successful sync of the realm.
-Do not use the "discard unsynced changes" strategy if your application
-cannot lose data already written to the client realm file but not yet
-synced to the backend.
+It *does* delete all local changes made since the last successful sync. 
+Do not use this strategy if your application cannot lose unsynced data. 
 
-The "discard unsynced changes" strategy can handle every kind of client
-reset error *except* for client resets triggered by
-:ref:`breaking or destructive schema changes <destructive-changes-synced-schema>`.
-If your application experiences a breaking schema change, this strategy
-falls back to a mode that mimics the "manually recover unsynced changes"
-strategy.
+This strategy provides a frozen copy of the realm from before reset. You 
+can attempt to recover changes from this frozen realm. However, the code 
+required for a manual recovery is quite complex. This is not recommended 
+for most applications.
 
-To handle client resets with the "discard unsynced changes" strategy, 
-pass an instance of :java-sdk:`DiscardUnsyncedChangesStrategy
+Discard unsynced changes cannot handle :ref:`breaking or destructive schema 
+changes <destructive-changes-synced-schema>`. With breaking changes, this 
+strategy falls back to manual recovery mode.
+
+To use this strategy, pass an instance of 
+:java-sdk:`DiscardUnsyncedChangesStrategy
 <io/realm/mongodb/sync/DiscardUnsyncedChangesStrategy.html>` to the
 :java-sdk:`defaultSyncClientResetStrategy()
 <io/realm/mongodb/AppConfiguration.Builder.html#defaultSyncClientResetStrategy(io.realm.mongodb.sync.DiscardUnsyncedChangesStrategy)>`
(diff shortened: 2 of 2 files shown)


source/includes/destructive-schema-change-app-update.rst
```diff
@@ -3,7 +3,6 @@
    After a breaking schema change:
 
    - All clients must perform a client reset.
-   - Clients that use models destructively affected by the schema
-     change must update local object models.
+   - Clients affected by the schema change must update their local object models.
```
nathan-contino-mongo, line 6: The "destructive" part here is critical -- a schema change could include many different changes, but only if the local schema was *destructively effected* does a client reset occur. So I'd propose we restore this to the original wording.

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -17,23 +17,31 @@ Reset a Client Realm - Java SDK
    To learn about the causes of and strategies for handling client
    resets, check out the Sync :ref:`Client Resets <client-resets>` page.
 
-A **client reset error** is a scenario where a client realm cannot sync
-data with the application backend. Clients in this state may continue to
-run and save data locally but cannot send or receive sync changesets
-until they perform a client reset strategy.
+The realm file in your client handles reads and writes on the device. When
```
nathan-contino-mongo, line 20: Technically the SDK handles reads and writes (is it ever correct to say that a file "handles reads and writes"? I'm not sure this sentence means anything in its current state. Add in the passive wording beneath and I suspect this paragraph was better in its original form. Maybe we could just break up the original second sentence into two sentences instead?

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -17,23 +17,31 @@ Reset a Client Realm - Java SDK
...
+The realm file in your client handles reads and writes on the device. When 
+you use Realm Sync, this local realm syncs with the application backend. 
+Some conditions can cause the realm to be unable to sync with the backend. 
+When this occurs, you get a **client reset error**. 
 
-You can handle client resets in your application using one of the
-available **client reset strategies**:
+This error means you must reset the realm file in the client application. 
+Clients in this state may continue to run and save data locally. Until you 
+perform the client reset, the realm does not sync with the backend.
```
nathan-contino-mongo, line 27: I like this rewording, definitely an improvement from the original wording. Probably a good idea to split this into two paragraphs anyway, where one talks about what a client reset is and the other discusses the impact a client reset has on your application.

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -17,23 +17,31 @@ Reset a Client Realm - Java SDK
...
-  since the last sync.
+Choose a **client reset strategy** to handle client reset errors. These 
+strategies restore realm to a syncable state, but have tradeoffs:
 
-- :ref:`Manually Recover Unsynced Changes <java-manually-recover-unsynced-changes>`:
-  Downloads a new copy of the realm, and moves the unsyncable realm
-  to a backup. Migrate unsynced data from the backup copy of the
-  realm to the new syncable copy.
+- :ref:`Discard Unsynced Changes <java-discard-unsynced-changes>`. Restore 
+  Sync by discarding local changes since the last sync. Maintains change
```
nathan-contino-mongo, line 33: This wording is great, huge improvement!

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -17,23 +17,31 @@ Reset a Client Realm - Java SDK
...
+  Sync by discarding local changes since the last sync. Maintains change 
+  listeners.
+- :ref:`Manually Recover Unsynced Changes 
+  <java-manually-recover-unsynced-changes>`:. Move the unsyncable realm and 
+  download a new copy. Invalidates change listeners.
 
+Both options let you write custom logic to recover local changes. Neither 
+option can recover local changes for you. 
+
+Discard unsynced changes is an improved version of manual recovery. However,
```
nathan-contino-mongo, line 42: "Improved" is a little subjective here -- maybe, "easier to implement" would be more accurate? (and a reword to support that phrasing)

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -42,44 +50,43 @@ Discard Unsynced Changes
...
-:ref:`breaking or destructive schema changes <destructive-changes-synced-schema>`.
-If your application experiences a breaking schema change, this strategy
-falls back to a mode that mimics the "manually recover unsynced changes"
-strategy.
-
-To handle client resets with the "discard unsynced changes" strategy, 
-pass an instance of :java-sdk:`DiscardUnsyncedChangesStrategy
+The **discard unsynced changes** performs client resets "seamlessly." This 
+strategy requires minimal code and minimal disruption. This strategy 
+does not close the realm or miss notifications.
```
nathan-contino-mongo, line 55: I think I prefer the original first sentence to the new first two sentences, the new third sentence is a big improvement though. Personally I feel that the new first sentence has no real meaning to the reader and we should just trim it.

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -42,44 +50,43 @@ Discard Unsynced Changes
...
+
+It *does* delete all local changes made since the last successful sync. 
+Do not use this strategy if your application cannot lose unsynced data.
```
nathan-contino-mongo, line 58: The SDK team specifically asked for the "data already written to the client realm file but not yet synced to the backend" clarification, IIRC -- "unsynced data" is not necessarily a clear concept for readers. I would probably add a sentence that clarifies that "unsynced data == local changes made since the last successful sync".

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -42,44 +50,43 @@ Discard Unsynced Changes
...
+strategy requires minimal code and minimal disruption. This strategy 
+does not close the realm or miss notifications.
+
+It *does* delete all local changes made since the last successful sync. 
+Do not use this strategy if your application cannot lose unsynced data. 
+
+This strategy provides a frozen copy of the realm from before reset. You 
+can attempt to recover changes from this frozen realm. However, the code 
+required for a manual recovery is quite complex. This is not recommended 
+for most applications.
```
nathan-contino-mongo, line 63: I believe I omitted this detail originally because the "frozen copy" is impermanent and not reliable for a real recovery operation, so users can't really rely on it for anything. As a result it's easier and simpler to leave it out of the explanation since it just confuses things (IMO)

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -42,44 +50,43 @@ Discard Unsynced Changes
...
+
+Discard unsynced changes cannot handle :ref:`breaking or destructive schema 
+changes <destructive-changes-synced-schema>`. With breaking changes, this
```
nathan-contino-mongo, line 66:
```suggestion
changes <destructive-changes-synced-schema>`. When breaking changes occur, this 
```
I don't think there's any actual grammatical issue here, but for some reason "With breaking changes" just sounds... strange? to me. Take it or leave it.

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -42,44 +50,43 @@ Discard Unsynced Changes
...
-- ``onAfterReset()``: called after the SDK successfully executes this
-  strategy. Provides a frozen copy of the original realm, as well as
-  a live instance of the realm in a syncable state.
-- ``onError()``: mimics the "manually recover unsynced changes"
-  strategy. The SDK only calls this method when your application
-  experiences a breaking schema change. For more information on
-  how to handle this scenario, see
-  :ref:`Discard Unsynced Changes after Breaking Schema Changes
-  <java-discard-unsynced-changes-after-destructive-schema-changes>`.
+- ``onBeforeReset()``. Realm calls this block when it receives a client
```
nathan-contino-mongo, line 78:
```suggestion
- ``onBeforeReset()``. The SDK calls this block when it receives a client 
```
I try to avoid using "Realm" as a subject for active verbs since realm (the file), Realm (the backend), Realm (the company), etc. are really easy to confuse. I'd suggest "Realm Database" here instead, but that gets kind of repetitive in two bullet points in a row.

nathan-contino-mongo: Additionally, I think it's useful to clarify that this method call happens *BEFORE* any recovery logic runs. Maybe we can add that in somewhere.

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -42,44 +50,43 @@ Discard Unsynced Changes
...
-  strategy. The SDK only calls this method when your application
-  experiences a breaking schema change. For more information on
-  how to handle this scenario, see
-  :ref:`Discard Unsynced Changes after Breaking Schema Changes
-  <java-discard-unsynced-changes-after-destructive-schema-changes>`.
+- ``onBeforeReset()``. Realm calls this block when it receives a client 
+  reset error from the backend.
+- ``onAfterReset()``. Realm calls this block after successfully executing 
+  this strategy. This block provides a frozen copy of the original realm. 
+  It also returns a live instance of the realm in a syncable state.
```
nathan-contino-mongo, line 82: I like this, aside from the Realm active subjectification I mentioned in the previous comment.

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -42,44 +50,43 @@ Discard Unsynced Changes
...
+- ``onError()``: Mimics the "manually recover unsynced changes" strategy.
```
nathan-contino-mongo, line 83: Perhaps instead of "mimics" we can point to the specific method in "manually recover unsynced changes" that this behaves just like.

source/sdk/java/examples/reset-a-client-realm.txt
```diff
@@ -106,12 +113,9 @@ Discard Unsynced Changes after Breaking Schema Changes
...
 
-If your application experiences a breaking schema change, the "discard
-unsynced changes" strategy cannot handle the resulting client reset
-automatically. Instead, you must provide a manual client reset
-implementation in the ``onError()`` method. The following
-example demonstrates how you can manually handle this error case by
-discarding all unsynced changes:
+The discard unsynced changes strategy cannot handle breaking changes. You 
+must manually handle the client reset in the ``onError()`` method. This 
+example discards unsynced changes to handle the client reset:
```
nathan-contino-mongo, line 118:
```suggestion
example manually discards unsynced changes to handle the client reset:
```

I'd specifically call out that this is a *manual* process, as opposed to the automatically discard strategy this happens inside of. Otherwise nice work streamlining this!

nathan-contino-mongo (approved):
LGTM -- I nitpicked a ton of stuff here since I wrote the original page here, but I only hope you incorporate the technical inaccuracies I noticed on the new version of the page. I trust that you'll make any technical changes before merging if they really matter.

source/includes/destructive-schema-change-app-update.rst
dacharyc: The reason I took the "destructive" out here was because the start of this bullet list establishes "After a breaking schema change" and the bullet itself says "Clients affected by the schema change."

I _think_ that's logically sound enough to establish that we're talking about a breaking change here. But if you think it's important to repeat in the bullet, I'll reword it - I'd just advise we say "Clients affected by the breaking schema change" instead of "destructively effected" because that clause reads oddly/is difficult to parse.

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: Your point is fair. But I think as-written, we're missing an opportunity to better set up this problem to make it easier or more intuitive for devs to understand the concept of a local realm having changes that don't exist in the Cloud. I'll have a think about this intro.

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: What about:
> Discard unsynced changes is a less complex alternative to manual recovery.

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: The original first sentence is _long_ and difficult to parse. I'm not tied to this rewording specifically, but I do think we need to break that sentence up and make it easier to parse. If you want to wordsmith it a bit with Hemingway or Readable, I'm happy to go with a different wording that you prefer. Or I can take another shot at this and try to find an alternative wording you'll like better.

(That "seamless" thing was a nod to our product positioning for this feature, but yeah, it doesn't actually communicate effectively to the developer.)

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: That's fair - the wording was just awkward and difficult to parse. This is part of what I was trying to solve with the intro rewording - trying to establish that the SDK persists changes to the local realm, and then Sync attempts to apply these changes to the data in the cloud, which I think makes the concept of unsynced data clearer. I'll have another think about how we might communicate this more clearly.

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: When I was working on the Swift SDK version of this doc, the team thought it was important to call this out as it gives devs an _option_ to do something with the data if they want it, whereas they can't do that if they don't know it exists. Can this frozen realm not be copied somewhere more permanent and used for a recovery operation? That was the suggestion in the Swift SDK.

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: That's fair. I did try that initially but it made the sentence "too long." What about:

> When breaking changes occur, the SDK falls back to manual recovery mode.

source/includes/destructive-schema-change-app-update.rst
nathan-contino-mongo: Yeah, the big distinction to make here is that clients may only contain a subset of realm object classes, and if that subset doesn't include the breaking change... no client reset. As long as we convey that, I'm happy -- I just don't want to lose that fact from this page!

source/sdk/java/examples/reset-a-client-realm.txt
nathan-contino-mongo: LGTM

source/sdk/java/examples/reset-a-client-realm.txt
nathan-contino-mongo, line 55: How about:

```suggestion
**Discard unsynced changes** is a client reset strategy provided by the SDK.
This strategy requires minimal code. This strategy does not close the
realm or miss notifications.
```

nathan-contino-mongo: Added a suggested wording, I'm not 100% happy with it but maybe you can take some inspiration.

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: Is that this: [defaultClientResetStrategy](https://www.mongodb.com/docs/realm-sdks/java/latest/io/realm/mongodb/AppConfiguration.Builder.html?_ga=2.12531955.691438089.1650290917-78975312.1612214951#defaultSyncClientResetStrategy(io.realm.mongodb.sync.ManuallyRecoverUnsyncedChangesStrategy))?

source/sdk/java/examples/reset-a-client-realm.txt
nathan-contino-mongo: Interesting, as far as I know you can't copy the "frozen" instance of the realm because it only exists as a diff state in the realm file itself. I say it's impermanent because if your application crashes, or gets closed by the user, or the device battery dies, or whatever... you'll never be able to access that frozen instance again. So there's no guarantee you can do anything with it. Probably not a big deal to mention it here, I didn't think it had much value for developers though.

source/sdk/java/examples/reset-a-client-realm.txt
nathan-contino-mongo: Yep!

source/includes/destructive-schema-change-app-update.rst
dacharyc: Ahh, I see the distinction - thank you for clarifying! What if I change this second bullet to:
> You must update client models affected by the breaking schema change.

nathan-contino-mongo: Directly actionable, perfect!

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: Hmm. Removing this paragraph, then. If your SDK team didn't feel it was important to mention, and you don't think it provides much value to developers, then I suppose there's no need to mention it - better to not give developers the wrong impression about what they can do with this client reset stuff.

source/sdk/java/examples/reset-a-client-realm.txt
dacharyc: What if I change the first line in this paragraph to:
> The SDK reads and writes to a realm file on the device.

Then the opening paragraph becomes:
> The SDK reads and writes to a realm file on the device. When 
you use Realm Sync, this local realm syncs with the application backend. 
Some conditions can cause the realm to be unable to sync with the backend. 
When this occurs, you get a **client reset error**.

PR #1871, opened 2022-04-25
(DOCSP-20436): Add Null Support for Optional Types in Realm Schema

mongodben:
## Pull Request Info

Document how to add null support for optional types in a realm schema

### Jira

- https://jira.mongodb.org/browse/DOCSP-20436

- [Enforce a Schema page > Validate Null Types section](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-20436/schemas/enforce-a-schema#validate-null-types)

### Review Guidelines

### DIFF
--- a/source/schemas/enforce-a-schema.txt
+++ b/source/schemas/enforce-a-schema.txt
@@ -43,3 +43,21 @@ Procedure
       :tabid: cli
       
       .. include:: /includes/steps/enforce-a-schema-cli.rst
+
+.. _validate-null-types:
+
+Validate Null Types
+-------------------
+
+You can configure {+service-short+} to pass schema validation for any ``null`` values
+when their field types are set as optional in the schema. 
+
+If you do not enable null type schema validation, {+service-short+} rejects ``null``
+values passed to optional fields. 
+
+To enable null type schema validation: 
+
+#. In the left navigation menu, select :guilabel:`App Settings` beneath :guilabel:`Manage`.
+#. On the :guilabel:`General` tab, navigate to the :guilabel:`Null Type Schema Validation`
+   section. Toggle the switch to be :guilabel:`ON`.
+#. Click the :guilabel:`Save` button. 
--- a/source/schemas/enforce-a-schema.txt
+++ b/source/schemas/enforce-a-schema.txt
@@ -52,13 +52,13 @@ Validate Null Types
 You can configure {+service-short+} to pass schema validation for any ``null`` values
 when their field types are set as optional in the schema.
 Enabling null type validation allows the value for a field to be persisted as
-the type in the schema or :manual:`BSON null </reference/bson-types/>`.
+the type in the schema or the :manual:`BSON null </reference/bson-types/>` type.
 If you do not enable null type schema validation, {+service-short+} rejects ``null``
 values passed to optional fields. 
 
 .. note:: 
 
-   Schema fields are not nullable by default because null is a unique BSON type.
+   Schema fields are not nullable by default because ``null`` is a unique BSON type.
    The default schema behavior is to accept only a single type for a field.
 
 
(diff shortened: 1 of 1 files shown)


source/schemas/enforce-a-schema.txt
```diff
@@ -43,3 +43,20 @@ Procedure
...
       
       .. include:: /includes/steps/enforce-a-schema-cli.rst
+
+.. _validate-null-types:
+
+Validate Null Types
+-------------------
+
+You can configure {+service-short+} to pass schema validation for any ``null`` values
+when their field types are set as optional in the schema.
```
sumedhamehta, line 53: Can we write something about why this isn't the default behavior?

maybe "_Even though null is a separate type in MongoDB's bson types <link to bson types>"_

mongodben: this could be a note perhaps. will add something

sumedhamehta (approved):

source/schemas/enforce-a-schema.txt
```diff
@@ -43,3 +43,28 @@ Procedure
...
+Enabling null type validation allows the value for a field to be persisted as
+the type in the schema or the :manual:`BSON null </reference/bson-types/>` type.
+If you do not enable null type schema validation, {+service-short+} rejects ``null``
+values passed to optional fields. 
+
+.. note:: 
+
+   Schema fields are not nullable by default because ``null`` is a unique BSON type.
```
MongoCaleb, line 61: This note feels like the lede for the section. Would you consider the following re-arrangment?

- Default behavior is to accept only a single type of field; "null" is a defined BSON type
- Schema fields are not nullable by default, so schema validation fails on nullable values.
- However, you can configure Realm to ignore null values in optional field types.

cbush (approved):

source/schemas/enforce-a-schema.txt
```diff
@@ -43,3 +43,28 @@ Procedure
...
+   The default schema behavior is to accept only a single type for a field.
+
+
+To enable null type schema validation: 
+
+#. In the left navigation menu, select :guilabel:`App Settings` below :guilabel:`Manage`.
```
MongoCaleb, line 67: I generally try to put UI steps in the order that the user clicks. So this first step might be:

> In the left navigation menu, under Manage, select App Settings.

MongoCaleb (commented):
Suggestion to reorganize a bit to put the problem statement at the beginning of the section

source/schemas/enforce-a-schema.txt
mongodben: they don't actually click 'manage' which is why it's like this. manage is just a UI label

source/schemas/enforce-a-schema.txt
mongodben: good point. will refactor.

source/schemas/enforce-a-schema.txt
MongoCaleb: I still think it's useful to describe the location before the action. "In the backyard, find the $20 bill....under the rock" is less clear (to me) as "In the backyard, under the rock, find the $20 bill." I'm willing to let it go, though, if you disagree.

mongodben: i'm not willing to disagree. implemented your suggested changes :)

source/schemas/enforce-a-schema.txt
```diff
@@ -43,3 +43,28 @@ Procedure
...
+
+Validate Null Types
+-------------------
+
+{+service+}'s default behavior is to only accept a single type for each field. 
+Schema fields are not nullable by default because ``null`` is a unique 
+:manual:`BSON type </reference/bson-types/>`.
+
+You can configure fields to pass schema validation for ``null`` values
+when their types are set as optional in the schema.
```
MongoCaleb, line 57: This _could_ be read as: "you can configure a field to pass schema validation if you set its type to optional." Perhaps:
You can configure any **optional** field to pass schema validation for ``null`` values.

MongoCaleb (approved):
LGTM with one nit

source/schemas/enforce-a-schema.txt
mongodben: nice, reduces wordiness too

PR #1872, opened 2022-04-25
Add Offline Login to .NET docs

MongoCaleb:
### Jira
https://jira.mongodb.org/browse/DOCSP-22071

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/offline_login/sdk/dotnet/examples/authenticate/#offline-login

### DIFF
--- a/source/sdk/dotnet/examples/authenticate.txt
+++ b/source/sdk/dotnet/examples/authenticate.txt
@@ -134,6 +134,9 @@ Offline Login
 
 .. include:: /includes/offline-login.rst
 
+The following example checks if there is a cached User object. If not, it logs 
+the user in. Otherwise, it uses the cached credentials:
+
 .. literalinclude:: /examples/generated/dotnet/OpenARealmExamples.codeblock.check-if-offline.cs
    :language: csharp
 
--- a/source/sdk/dotnet/examples/authenticate.txt
+++ b/source/sdk/dotnet/examples/authenticate.txt
@@ -128,6 +128,15 @@ you can log in using the following code:
 
 .. include:: /includes/authorization-appleidcredential-string.rst
 
+
+Offline Login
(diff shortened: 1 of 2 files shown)


source/sdk/dotnet/examples/authenticate.txt
```diff
@@ -128,6 +128,18 @@ you can log in using the following code:
 
 .. include:: /includes/authorization-appleidcredential-string.rst
 
+
+Offline Login
+-------------
+
+.. include:: /includes/offline-login.rst
```
nlarew, line 135: Can we tweak the title of the shared note in this include?

Current:
A user must have a network connection for first login on client

Proposed:
- A user must have a network connection for their first login on a client
- A network connection is required for a user's first login on a client

nlarew (approved):
LGTM w/ one proposed change to the shared note

source/sdk/dotnet/examples/authenticate.txt
MongoCaleb: @nlarew - the proposed change puts it in the passive voice. Is your objection to the first sentence about the user having a network connection? If so, me too. Since this is a "title" of sorts, how do you like:

> An initial login requires a network connection

dacharyc: What about just:
> Initial login requires a network connection

I don't have strong feelings about this as long as we communicate that a user has to have a network connection the first time they login, but "An" just reads weird to me here.

nlarew: "A user must have a network connection for their first login on a client" is active voice - the second bullet point is passive but I included it because it's just a little bit shorter.

No strong opinions on the updated "initial login" title - either one would also work

PR #1873, opened 2022-04-25
(DOCSP-18418): GraphQL Overview 'Code-First Approach' Clarifications

mongodben:
## Pull Request Info

Add clarifications about how Realm GraphQL uses a code-first approach.

also updated `:doc:` links to use `:ref:` throughout the page

### Jira

- https://jira.mongodb.org/browse/DOCSP-18418

- [GraphQL overview](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-18418/graphql)

### Review Guidelines

### DIFF
--- a/source/graphql.txt
+++ b/source/graphql.txt
@@ -38,20 +38,14 @@ GraphQL API
 Overview
 --------
 
-.. note:: Version Requirement
-
-   GraphQL requires MongoDB version 4.0 or greater.
-
 The {+service+} :graphql:`GraphQL <>` API allows client applications to access
 data stored in a linked {+atlas+} cluster using any standard GraphQL client.
 
 {+service-short+} automatically creates GraphQL types for every linked
 collection that has a defined :ref:`schema <schemas>` and evaluates
-:ref:`role-based permissions <define-roles-and-permissions>` for all
-GraphQL requests. 
-
-To learn how to make data available through the GraphQL API, see
-:ref:`Expose Data in a Collection <graphql-expose-data>`.
+:ref:`role-based permissions <define-roles-and-permissions>` for all GraphQL requests.
+To learn how to make data available through the GraphQL API,
+see :ref:`Expose Data in a Collection <graphql-expose-data>`.
 
 To learn about the generated types and operations that you can use with the {+service-short+}
 GraphQL API, see :ref:`GraphQL Types & Resolvers <graphql-types-and-resolvers>`.
@@ -83,13 +77,11 @@ circumvents the need for multiple costly round trips to the server.
 To learn more about GraphQL, check out the :graphql:`official GraphQL
 tutorial <learn>`.
 
-Approach to Schema Definition
------------------------------
+How {+service-short+} Creates GraphQL Schemas
+---------------------------------------------
 
-{+service+} takes a unique approach to schema definition. You
-generates the GraphQL schema and resolvers from JSON schemas
-for MongoDB collections. This differs from the traditional `code-first and schema-first approaches
-<https://blog.logrocket.com/code-first-vs-schema-first-development-graphql/>`__
+Using {+service+}, you generate the GraphQL schema and resolvers from JSON schemas
+for MongoDB collections. This differs from the traditional code-first and schema-first approaches
 to GraphQL schema development.
 
 To define your GraphQL schema with {+service-short+}:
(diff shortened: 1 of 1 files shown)


source/graphql.txt
```diff
@@ -43,34 +43,37 @@ Overview
    GraphQL requires MongoDB version 4.0 or greater.
 
 The {+service+} :graphql:`GraphQL <>` API allows client applications to access
-data stored in a linked MongoDB cluster using any standard GraphQL client.
+data stored in a linked {+atlas+} cluster using any standard GraphQL client.
+
+{+service+} takes a code-first approach to GraphQL schema generation.
```
nlarew, line 48: > MongoDB Realm takes a code-first approach to GraphQL schema generation.

So this is only kind of true - it depends on your perspective. I'm a bit hesitant to just say it outright and think it might be better to focus on guiding devs to the right mental model.

_From https://blog.graphqleditor.com/graphql-schemafirst-codefirst_
> **Code-first** (often called resolver-first) is a process where the _schema is defined and implemented programmatically_. The design process begins with coding the resolvers and the SDL version of the GraphQL schema is a generated artifact.
> 
> **Schema-first** is an approach that puts _schema as your source of truth_ and forces your code to follow the definitions stored in your schema.

What Realm _actually does_ is "code-first" in a way - it generates a GraphQL schema based on external code. Typically, though, code-first implies that you're generating the types in your GraphQL schema from your resolver functions, which is not the case here!

Instead, the external code is user-defined JSON schema definitions which get translated into a GraphQL schema. The actual resolvers (so far) are hidden because we can create them automatically. Sounds a lot like "schema first" to me even though it's not technically defining the GraphQL schema directly...

This feels especially weird when you consider custom resolvers. You write the resolver function & specify its input/return type with JSON schema definitions. Realm translates these into your GraphQL schema just like it translates collection JSON schemas - the only difference is now you write the function instead of us auto-generating it.

Either way, a (JSON) schema is the first thing:

1. Define a JSON schema (for a collection or resolver input/return types)
2. Write (or generate) a resolver that conforms to your schemas (or the standard generated schemas)
3. Translate the JSON schemas to a GraphQL schema

mongodben: thanks for the very thorough response nick. i see what you mean. what we're doing is really a combination of the 2 approaches. 

regarding the documentation here, i think we can either:
1. remove the line `{+service+} takes a code-first approach to GraphQL schema generation.` and keep rest of PR as is (maybe w minor tweaks)
2. remove the line `{+service+} takes a code-first approach to GraphQL schema generation.` and build out new section on this page about 'realm graphql approach' noting what it is and how it differs from traditional code/schema first approaches (refinement of content in your PR)

second option is preferable, since i think this is a fairly important conceptual description. curious your thoughts here @nlarew

source/graphql.txt
```diff
@@ -82,6 +83,25 @@ circumvents the need for multiple costly round trips to the server.
 To learn more about GraphQL, check out the :graphql:`official GraphQL
 tutorial <learn>`.
 
+Approach to Schema Definition
+-----------------------------
+
+{+service+} takes a unique approach to schema definition. You
```
nlarew, line 89: I'd omit "unique" here - it's not really that crazy of an idea to generate GraphQL schemas from JSON schemas

source/graphql.txt
```diff
@@ -82,6 +83,25 @@ circumvents the need for multiple costly round trips to the server.
...
+generates the GraphQL schema and resolvers from JSON schemas
```
nlarew, line 90: - Check the grammar here - "You generates..."
- Let's link to the [Schemas](https://www.mongodb.com/docs/realm/schemas/) page here

source/graphql.txt
```diff
@@ -82,6 +83,25 @@ circumvents the need for multiple costly round trips to the server.
...
+for MongoDB collections. This differs from the traditional `code-first and schema-first approaches
+<https://blog.logrocket.com/code-first-vs-schema-first-development-graphql/>`__
+to GraphQL schema development.
```
nlarew, line 93: Let's remove the link to the logrocket blog - I'd prefer to have a more definitive link but I also don't think no link is too big of a deal here. 

We can let the process speak for itself, as long as we're clear that we're "schema first" but not "GraphQL SDL first"

source/graphql.txt
```diff
@@ -82,6 +83,25 @@ circumvents the need for multiple costly round trips to the server.
...
+
+To define your GraphQL schema with {+service-short+}:
+
+#. :ref:`Define a JSON schema <schemas>` for a MongoDB collection in your {+atlas+} cluster.
+   You can :ref:`enforce the shape of the collection schema <enforce-a-schema>`
+   based on custom definitions or use a generated schema based on the documents in the collection.
+#. Configure {+service-short+} to :ref:`generate the GraphQL schema and resolvers
```
nlarew, line 100: Technically there's nothing to configure here - once you've defined a JSON schema you also get a corresponding GraphQL schema/resolvers automatically

source/graphql.txt
```diff
@@ -82,6 +83,25 @@ circumvents the need for multiple costly round trips to the server.
...
+   <graphql-expose-data>` based on your collection JSON schema. 
+#. Optionally extend the functionality of your generated GraphQL schema with
+   :ref:`custom resolvers <graphql-custom-resolvers>`.
+
 Limitations
```
nlarew, line 105: Can we bump the **Limitations** section to the bottom of the page?

source/graphql.txt
```diff
@@ -82,6 +83,25 @@ circumvents the need for multiple costly round trips to the server.
 To learn more about GraphQL, check out the :graphql:`official GraphQL
 tutorial <learn>`.
 
+Approach to Schema Definition
```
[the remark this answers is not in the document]
mongodben: sure. i like "How Realm Generates GraphQL Schemas".  maybe  change to "How Realm Creates GraphQL Schemas" since it's a mix of user action and generation

source/graphql.txt
mongodben: sounds good to me. i was hesitant to include for same reason but couldn't find anything more authoritative. 

but no link should be fine. if someone is curious they can google "code-first and schema-first approaches to GraphQL" and find that article as the first link (as I did)

source/graphql.txt
mongodben: good point. removing "Configure {+service-short+} to "

source/graphql.txt
mongodben: removing "takes a unique approach to schema definition. You" per this and prev comment. changing to "{+service+} generates..."

mongodben: not linking to the schemas page b/c i do this below

nlarew (approved):
LGTM

PR #1885, opened 2022-04-28
(DOCSP-21902): External Dependencies

nlarew:
## Pull Request Info

- Combines [Add External Dependencies](https://www.mongodb.com/docs/realm/functions/add-external-dependencies/) and [Import External Dependencies](https://www.mongodb.com/docs/realm/functions/import-external-dependencies/) in a new page

### Jira

- (DOCSP-21902): External Dependencies

- [Functions > External Dependencies](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/fns-dependencies/functions/dependencies/)

### Review Guidelines


source/functions/dependencies.txt
```diff
@@ -0,0 +1,128 @@
...
+=====================
+External Dependencies
+=====================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 3
```
dacharyc, line 12: The "On This Page" for the "Add an External Package" section is quite long, because of the tabs/multiple steps in each tab. We should probably reduce the depth to make the "On This Page" more useful for folks.

source/functions/dependencies.txt
```diff
@@ -0,0 +1,128 @@
...
+   :class: singlecol
+
+An **external dependency** is a library that includes code that you
+can't or don't want to implement yourself. For example, you might use an
+official library for an external services or a custom implementation of
```
dacharyc, line 17:
```suggestion
official library for an external service or a custom implementation of
```

source/functions/dependencies.txt
```diff
@@ -0,0 +1,128 @@
...
+
+To import and use an external dependency, you first need to add the
+dependency to your application. You can either :ref:`add packages by
+name <dependencies-add-by-name>` or :ref:`upload a directory of
+dependencies <dependencies-upload-zip>`.
+
+.. important:: Override Existing Dependencies
+   
+   You can only use one method at a time to specify the external packages
+   your app can use. For example, a package added by name through the UI
```
dacharyc, line 45: Interesting! Is there an order in which dependencies override one another? I.e. UI > CLI > GitHub? Might be useful for devs to know which one "wins."

source/functions/dependencies.txt
```diff
@@ -0,0 +1,128 @@
...
+
+   .. tab:: CLI
+      :tabid: cli
+
+      .. include:: /includes/steps/add-external-dependencies-cli.rst
+   
+   .. tab:: Github Auto Deploy
+      :tabid: github
+      
+      .. include:: /includes/steps/add-external-dependencies-github.rst
```
dacharyc, line 73: I see that this include is outside the scope of this PR, but in looking at the staged page, I see step 5 of the "GitHub Auto Deploy" tab talks about checking for operation success:

> Once it finishes adding the dependencies, the CLI tells you if all packages were successfully added.

But this is _after_ Step 4, which has you pushing dependencies to GitHub using Git Push - not the CLI. So I think Step 5 should be updated to let folks know how to check for success when they're using deploy with GitHub. I don't know if you want to do this as part of this PR, or if we should make a separate Jira ticket for it?

dacharyc (approved):
I spotted a handful of small things I have Qs about, including a typo to fix, but otherwise LGTM!

source/functions/dependencies.txt
nlarew: Good call - the winner is whatever method you used most recently. Updated!

source/functions/dependencies.txt
nlarew: Good catch! Updated the step to not use the CLI

PR #1888, opened 2022-04-29
(DOCSP-21899): Query > Write

nlarew:
### Jira

- (DOCSP-21899): Query > Write

- [Functions > Query MongoDB > Write](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/fns-mql-write/functions/mongodb/write/)

### Review Guidelines


source/functions/mongodb/write.txt
```diff
@@ -13,66 +15,99 @@ Update Documents
...
+which documents to update and :manual:`update operators
+</reference/operator/update>` to describe the changes to apply to
+matching documents.
+
+.. note::
+   
+   While running update operations, {+backend+} temporarily adds a
+   reserved field, ``_id_baas_transaction``, to documents. If you modify
+   data used by your app outside of {+service-short+}, you may need to
+   consider this field. For more information, see :ref:`Transactional
```
MongoCaleb, line 79: Do we want to be more prescriptive here and say "don't touch it, yo!"?

source/functions/mongodb/write.txt
```diff
@@ -271,3 +310,88 @@ elements in an array field based on an array filter:
...
+
+MongoDB supports :manual:`multi-document transactions
+</core/transactions/>` that let you read and write multiple documents
+atomically, even across collections.
+
+To perform a transaction:
+
+1. Obtain and start a client session with ``client.startSession()``.
+
+#. (Optional) Define a read and write settings object for the
```
MongoCaleb, line 373: This one confused me a bit until I looked through the code snippet below. Could you consider something like the following?
> Define options for the transaction, such as readPreference and writeConcern, and then pass this object to the ``session.withTransaciton()`` method described in the next step.

MongoCaleb: And, actually, do we describe this object anywhere else? What properties are valid, etc.? If so, we need a link to it. If not....we need it. :)

source/functions/mongodb/write.txt
```diff
@@ -271,3 +310,88 @@ elements in an array field based on an array filter:
...
+   transaction. Pass this object to the method called in the next step.
+
+#. Use the ``session.withTransaction()`` method to start a transaction.
```
MongoCaleb, line 376: Consider showing ``session.withTransaction(options)``

MongoCaleb (commented):
LGTM but i think there's one place that could use a little bit more clarification.

source/functions/mongodb/write.txt
nlarew: Yes but you actually DO want to touch it! IIRC if you modify a document in a baas_transaction then the field doesn't get cleaned up automatically - you have to make sure to `$unset` it.

MongoCaleb (approved):
LGTM!

PR #1905, opened 2022-05-04
(DOCSP-21890): Data Sources

nlarew:
## Pull Request Info

- Updates **MongoDB Data Access** -> **MongoDB Data Sources** and refreshes the content on `/mongodb`
- Link a Data Source is no longer a separate page. Rather, it's a section on `/mongodb`
- Shuffles the `/mongodb` toctree to match our desired final state

### Jira

- (DOCSP-21890): Data Sources

- [MongoDB Data Sources](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/data-sources/mongodb/)

### Review Guidelines


source/mongodb.txt
```diff
@@ -16,86 +15,172 @@ MongoDB Atlas Overview
...
+
+To learn how to work with a data source from a Realm SDK, refer to the
+docs for the SDK:
+
+- :ref:`Query MongoDB - Swift SDK <ios-mongodb-data-access>`
+- :ref:`Query MongoDB - .NET SDK <dotnet-mongodb-data-access>`
+- :ref:`Query MongoDB - Java SDK <java-mongodb-data-access>`
+- :ref:`Query MongoDB - Node.js SDK <node-mongodb-data-access>`
+- :ref:`Query MongoDB - React Native SDK <react-native-mongodb-data-access>`
+- :ref:`Query MongoDB - Web SDK <web-mongodb-data-access>`
```
mongodben, line 70: seems like a good place for language specific tabs

source/mongodb.txt
```diff
@@ -1,9 +1,8 @@
-.. _mongodb-service:
-.. _mongodb-data-access:
```
mongodben, line 2: i recall that in comments on the outline we'd discussed having a folder on top of the pages in this section called "Query MongoDB Atlas". would you consider adding that back in? 

i think it'd be beneficial to have. personally, i don't like when a page is also a folder, generally speaking, as it is here. esp when there's very important but easy to miss information like the "link a data source" section. 

also, generally not sure how i feel about removing a separate page for 'link a data source' since that's a clear task that every user will be doing and if we're moving in a task-based direction i think that'd be worth having on a distinct page.

source/mongodb/crud-and-aggregation-apis.txt
```diff
@@ -17,27 +17,31 @@ CRUD & Aggregation APIs
...
-you connect to {+atlas+} through {+backend+}. For a high-level overview of
-MongoDB feature availability in {+service+}, see
-:ref:`Service Limitations <mongodb-service-limitations>`.
+Linked data sources can run most CRUD and Aggregation operations;
+however, {+service+} does not support all operations available in
+standard tools and clients. This page lists support for specific MongoDB
+operations, operators, options, and stages.
+
+For information on how to read and write data in MongoDB from a
+function, see :ref:`Query MongoDB Atlas <functions-mongodb>`.
```
mongodben, line 26: this link leads me to a blank page - https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/data-sources/functions/mongodb/#std-label-functions-mongodb

source/rules/expressions.txt
```diff
@@ -387,29 +387,29 @@ MongoDB Document Expansions
 .. json-expansion:: %%this
    
    | **Type:** ``any``
-   | **Usable In:** :ref:`MongoDB Data Sources <mongodb-service>`
+   | **Usable In:** :ref:`MongoDB Data Sources <data-sources>`
```
mongodben, line 390: thought: rather than saying 'MongoDB data source' should we say 'MongoDB *Atlas* Data Source'?

source/mongodb/crud-and-aggregation-apis.txt
nlarew: Ah yeah I knew I was missing something - still need to build out an index page here

source/mongodb.txt
nlarew: > we'd discussed having a folder on top of the pages in this section called "Query MongoDB Atlas"

That folder still exists in the Functions section - see `/functions/mongodb/` (the blank page you mentioned in another comment).

> i don't like when a page is also a folder

Perhaps we should chat as a team about this one. I land on the opposite side where I think page-folders are useful for reducing TOC bloat. Especially given our current tools - Sphinx/rst was designed to do page-folders - where non-page folders are still implemented as pages, which is why `/functions/mongodb/` is blank atm.

> also, generally not sure how i feel about removing a separate page for 'link a data source' since that's a clear task that every user will be doing and if we're moving in a task-based direction i think that'd be worth having on a distinct page

I've had a bit of a change of heart in how I think we should approach "task-oriented" content. I used to push for smaller, verb-phrase pages that focus on specific tasks. I think that approach leads to a more bloated TOC and adds unnecessary searching/clicking for the reader.

In recents projects (on the cloud side) I've been aiming to consolidate pages into topics that mix content types (e.g. a page with a conceptual overview _and_ a procedure on how to set it up. That way you only need to navigate to a topic you're interested in and then it can get task-oriented on the page itself. For example the "Link a Data Source" page on its own isn't that helpful for a new user - they should also get some context on _what_ a data source is and how they'll use it. I think combining onto one page accomplishes that better.

mongodben: let's discuss IRL

source/rules/expressions.txt
mongodben: maybe on first reference per page? 

it is more descriptive + some branding

source/mongodb/crud-and-aggregation-apis.txt
mongodben: can you make it reachable from the sidenav? it's confusing to click the link, see content, and then be dissociated from the side nav?

nlarew: Sure - added

mongodben (approved):
lgtm

PR #1911, opened 2022-05-06
Add step by step guide for Read/Write Own Data

cbush:
Note: merging to feature branch

See also: https://github.com/cbush/realm-template-apps/tree/flex-sync-permissions

### Jira

- [DOCSP-22227](https://jira.mongodb.org/browse/DOCSP-22227)

- Staging builds are not working right now

### Review Guidelines


source/sync/data-access-patterns/flexible-sync-permissions-guide.txt
```diff
@@ -0,0 +1,161 @@
...
+To set this up, log in to the Realm UI. Click :guilabel:`Sync` on the left hand
+panel and fill out the following information:
+
+- :guilabel:`Sync Type`: choose :guilabel:`Flexible`
+- Press the toggle to enable Development Mode.
+- Select the cluster you wish to sync.
+- :guilabel:`Define a Database Name`: select :guilabel:`+Add a new
+  database` and type a name for the database Realm will use to store your synced
+  objects. You can name it anything you want. A common strategy would be to name
+  the database after the app you're making.
```
nathan-contino-mongo, line 42: It looks like you're going to put a lot of different "cases" of permissions on this page. I suspect we should either refactor the first steps listed here into an include (to avoid redundancy) or pull them out into some sort of "get started with permissions" section.

Additionally... should we create a bulleted list/table of contents so users can quickly scan for the use case they're looking for (the "on this page" on the right is easy to ignore or miss)? Or should we consider breaking these out into individual pages for each case, so users can scan the different cases through the TOC?

cbush: Completely agree with the overview list.

For the steps, I agree it's repetitive, but on the other hand it is nice to have all of the steps listed in one place. Maybe I can use an include to repeat the repeated stuff?

cbush:
Merged to feature branch

PR #1916, opened 2022-05-10
Permissions guide: Admin Privileges

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-22230

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/premissions-guide-admin/sync/data-access-patterns/flexible-sync-permissions-guide/#administrator-privileges

### DIFF
--- a/source/sync/data-access-patterns/flexible-sync-permissions-guide.txt
+++ b/source/sync/data-access-patterns/flexible-sync-permissions-guide.txt
@@ -136,8 +136,8 @@ panel and fill out the following information:
   may read all data, but write only if the document's ``owner_id`` field is equal 
   to the user's ID.
 
-- You can skip :guilabel:`Advanced Configuration` for now.
-- Click :guilabel:`Save Changes` to enable Sync with these permissions.
+- Skip :guilabel:`Advanced Configuration`, and then click :guilabel:`Save Changes` 
+  to enable Sync with these permissions.
 
 .. _admin-privileges:
 
--- a/source/sync/data-access-patterns/flexible-sync-permissions-guide.txt
+++ b/source/sync/data-access-patterns/flexible-sync-permissions-guide.txt
@@ -29,7 +29,6 @@ These assume you have already created a Realm app.
    :ref:`create-realm-apps`
    :ref:`flexible-sync-rules-and-permissions`
 
-
 .. _read-write-own-data:
 
 Read & Write Own Data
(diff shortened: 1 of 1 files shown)


source/sync/data-access-patterns/flexible-sync-permissions-guide.txt
```diff
@@ -93,35 +88,141 @@ panel and fill out the following information:
...
+       }
+     ]  
+   }
+
+.. note:: Change the default settings
+
+   The generated expression has two objects in the "defaultRoles" 
+   array. The first defines the permissions for an administrator. Note that the 
+   auto-generated expression assumes there is a boolean field in the custom user 
+   data document named ``isGlobalAdmin``. Depending on how you defined your custom
```
cbush, line 212: I wonder if we would be better off telling people to use "isGlobalAdmin" above so that it "just works™", but then we can keep the info in this note. Otherwise, it seems like we're telling people to do one thing just to change it later. Thoughts?

source/sync/data-access-patterns/flexible-sync-permissions-guide.txt
```diff
@@ -93,35 +88,141 @@ panel and fill out the following information:
...
+
+In this permission strategy, users with a specific "administrator" role can 
+read and write any document. Users who do not have the specified role can only 
+read and write their own data.
 
-In this permission strategy, users with the ``admin`` role can read and 
-write any document. Users who do not have the ``admin`` role can read any
-document, but can only write their own data.
+To make this strategy work, you first need to :ref:`enable customer user data
+<custom-user-data>`. In the customer use data document that you define, define
```
cbush, line 151:
```suggestion
<custom-user-data>`. In the custom user data document that you define, define 
```

cbush (approved):
This is a great start on this section and should be merged to feature branch (with comments) but I'm thinking about a few things:
- Should we include how to set up the custom user data right on this page?
- Should we demonstrate how to actually make a user an admin?
- How is write access to the custom user data object/collection protected? (i.e. clients are not supposed to be able to set their own custom user data, but is that actually enforced?)

source/sync/data-access-patterns/flexible-sync-permissions-guide.txt
MongoCaleb: I was pondering the same. Since you saw it, too, I'll change it.

PR #1931, opened 2022-05-18
Test autobuilder check

cbush:
This updates the action to use https://github.com/cbush/snooty-autobuilder-check


mongodben (approved):
generally looks like it should work. let's give it a try. 

one question though, why did you do `runs.using: composite`? https://github.com/cbush/snooty-autobuilder-check/blob/main/action.yml#L8-L9

cbush:
I used composite because it still has to install and build. I didn't see a way to use the "javascript" mode and also run a package.json script, but let me know if there is indeed a way.

PR #1932, opened 2022-05-19
Remove cloud

cbush:
Staged: https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/remove-cloud/


dacharyc (approved):
Whoa, that's a lot of files. LGTM, probably?

snooty.toml
```diff
@@ -2,35 +2,13 @@ name = "realm"
 title = "MongoDB Realm"
 
 intersphinx = [
-  "https://www.mongodb.com/docs/manual/objects.inv",
-  "https://docs.atlas.mongodb.com/objects.inv"
```
dacharyc, line 6: Why are we removing these things? Are the intersphinx references to manual/atlas only in the Cloud side?

cbush: Correct. And we only use one anyway (billing). Really not needed!

PR #1948, opened 2022-05-27
(DOCSP-22452): Authentication Provider Examples, Kotlin SDK

nathan-contino-mongo:
### Jira

- https://jira.mongodb.org/browse/DOCSP-22452

- [Kotlin SDK - Authenticate a User](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-22452/sdk/kotlin/app-services/authenticate-users/)

### Review Guidelines


source/sdk/kotlin/app-services/authenticate-users.txt
```diff
@@ -39,42 +39,225 @@ In the event of a failure, :file:`app.login()` throws an
...
 
-The :ref:`Email/Password authentication provider
-<email-password-authentication>` enables users to log in to your application
-with an email username and a password. To log in with email/password
-authentication, create an email/password credential by calling
-:kotlin-sdk:`AuthenticationProvider.EMAIL_PROVIDER()
-<{+kotlin-sync-prefix+}io.realm.mongodb/-authentication-provider/-e-m-a-i-l_-p-a-s-s-w-o-r-d/index.html>`
-with the user's email and password. Then pass the generated credential to
-:file:`app.login()`.
+The :doc:`Email/Password authentication provider </authentication/email-password>`
```
mongodben, line 65: i think you wanna keep this as a ref so intersphinx can point to the right place, as opposed to this which goes to a deleted page

source/sdk/kotlin/app-services/authenticate-users.txt
```diff
@@ -39,42 +39,225 @@ In the event of a failure, :file:`app.login()` throws an
...
 .. literalinclude:: /examples/generated/kotlin/AuthenticationTest.snippet.email-password-authentication.kt
    :language: kotlin
    :copyable: false
 
+.. _kotlin-login-jwt:
+
+JWT
+~~~
+
+TThe :doc:`Custom JWT authentication provider </authentication/custom-jwt>`
```
mongodben, line 83:
```suggestion
The :doc:`Custom JWT authentication provider </authentication/custom-jwt>`
```

mongodben: Also ditto prev comment about not using the `:doc:` role here

source/sdk/kotlin/app-services/authenticate-users.txt
```diff
@@ -39,42 +39,225 @@ In the event of a failure, :file:`app.login()` throws an
...
+.. literalinclude:: /examples/generated/kotlin/AuthenticationTest.snippet.jwt-authentication.kt
+   :language: kotlin
+   :copyable: false
+
+.. _kotlin-login-api-key:
+
+API Key
+~~~~~~~
+
+The :doc:`API Key authentication provider </authentication/api-key>`
```
mongodben, line 101: ditto abt `ref` instead of `doc`

source/sdk/kotlin/app-services/authenticate-users.txt
```diff
@@ -39,42 +39,225 @@ In the event of a failure, :file:`app.login()` throws an
...
+1. In the `Google Cloud Platform console
+   <https://console.cloud.google.com/apis/credentials>`__, create an
+   OAuth 2.0 client ID of type "Web application".
+
+#. Configure your backend {+app+} to use that client ID and the associated
+   client secret.
+
+#. Enable OpenID Connect on the backend.
+
+Use Google's official :google:`Sign-In for Android <identity/sign-in/android>`
```
mongodben, line 130: do you only do this if doing for android? what about KMM?

source/sdk/kotlin/app-services/authenticate-users.txt
```diff
@@ -39,42 +39,225 @@ In the event of a failure, :file:`app.login()` throws an
...
+
+#. Create a :google:`GoogleSignInOptions
+   <android/reference/com/google/android/gms/auth/api/signin/GoogleSignInOptions>`
+   with the following builder options:
+
+   - :google:`default sign in
+     <android/reference/com/google/android/gms/auth/api/signin/GoogleSignInOptions.Builder#public-builder-googlesigninoptions-googlesigninoptions>`
+   - an :google:`ID token request
+     <android/reference/com/google/android/gms/auth/api/signin/GoogleSignInOptions.Builder#public-googlesigninoptions.builder-requestidtoken-string-serverclientid>`
+     -- pass your OAuth 2.0 client ID as the ``serverClientId``.
```
mongodben, line 155: i think there's something funky here with the "--"

source/sdk/kotlin/app-services/authenticate-users.txt
```diff
@@ -39,42 +39,225 @@ In the event of a failure, :file:`app.login()` throws an
...
+
+Use Google's official :google:`Sign-In for Android <identity/sign-in/android>`
+to authenticate Google users in your Android application:
+
+.. note:: Code Example Below
+
+   For an implementation of these instructions, check out the code block
+   below.
+
+1. Add the Google Sign-In for Android dependency to the ``dependencies``
```
mongodben, line 138:
```suggestion
#. Add the Google Sign-In for Android dependency to the ``dependencies``
```

mongodben, line 138: im not sure if it's prudent to have a list of all the method calls in the code here. would probably just be clearer to have the code example, as you do below, and heavily annotate it with comments explaining what's going on

source/sdk/kotlin/app-services/authenticate-users.txt
```diff
@@ -39,42 +39,225 @@ In the event of a failure, :file:`app.login()` throws an
...
+``loginWithGoogle()``:
+
+The :doc:`Email/Password authentication provider </authentication/email-password>`
+enables users to log in to your application with an email username and
+a password. To log in with email/password authentication, create an
+email/password credential by calling :kotlin-sdk:`Credentials.google()
+<{+kotlin-sync-prefix+}io.realm.mongodb/-credentials/-companion/google.html>`
+with the user's email and password. Then pass the generated credential
+to :kotlin-sdk:`app.login()
+<{+kotlin-sync-prefix+}io.realm.mongodb/-app/login.html>`.
```
mongodben, line 205: is this paragraph supposed to be here? confuses me. think there might be an error

source/sdk/kotlin/app-services/authenticate-users.txt
```diff
@@ -39,42 +39,225 @@ In the event of a failure, :file:`app.login()` throws an
...
+Facebook
+~~~~~~~~
+
+The :doc:`Facebook </authentication/facebook>` authentication provider allows
+you to authenticate users through a Facebook app using their existing Facebook
+account.
+
+.. include:: /includes/note-facebook-profile-picture-url.rst
+
+Follow the official :facebook:`Facebook Login for Android Quickstart
```
mongodben, line 228: ditto a prev question..what if using KMM and not only android?

source/sdk/kotlin/app-services/authenticate-users.txt
nathan-contino-mongo: Investigating a bit about the best way to approach this. We're really just targeting Android docs parity at the moment, but this raises a good point: certain content should probably cover the iOS use case as well (and other platforms, if they differ significantly). This is a large can of worms, hence our firm "talk about it later" stance so far.

At the very least though, I'll add a disclaimer here that the iOS use case will differ, though. I might be able to point to the Swift SDK examples for most of the "get token on iOS" side of things.

source/sdk/kotlin/app-services/authenticate-users.txt
nathan-contino-mongo: Definitely an error, thanks!

source/sdk/kotlin/app-services/authenticate-users.txt
nathan-contino-mongo: SAA

source/sdk/kotlin/app-services/authenticate-users.txt
nathan-contino-mongo: Interesting thought. I'll see if I can trim this down. It's unfortunately a lengthy process that Google really hasn't adequately documented... but at some point too much doc is too much doc.

source/sdk/kotlin/app-services/authenticate-users.txt
nathan-contino-mongo: I've created [a ticket](https://jira.mongodb.org/browse/DOCSP-22992) to investigate iOS/Android (and other) platform differences for auth with the SDK. This isn't a top priority concern for GA release at World, but I'll get to it as soon as possible because there is a lot of groundwork to lay with tests, CI, etc. for each platform we want to support directly in the docs.

mongodben (approved):
lgtm!

examples/kotlin/shared/src/androidMain/kotlin/com/mongodb/realm/realmkmmapp/MainActivity.kt
```diff
@@ -1,9 +1,95 @@
...
+        resultLauncher.launch(signInIntent)
+    }
+
+    fun handleSignInResult(completedTask: Task<GoogleSignInAccount>) {
+        try {
+            if (completedTask.isSuccessful) {
+                val account: GoogleSignInAccount? = completedTask.getResult(ApiException::class.java)
+                val token: String = account?.idToken!!
+                val app: App = App.create(YOUR_APP_ID) // Replace this with your App ID
+                runBlocking {
```
cmelchior, line 53: Doing this is a bit dangerous. For ease of use, this is fine, but it will block the UI thread, so if this leaks into the documentation you might want to put a warning here

examples/kotlin/shared/src/androidMain/kotlin/com/mongodb/realm/realmkmmapp/MainActivity.kt
```diff
@@ -1,9 +1,95 @@
...
+        FacebookSdk.setApplicationId(YOUR_FACEBOOK_SDK_APP_ID)
+        FacebookSdk.sdkInitialize(activity)
+        val callbackManager = CallbackManager.Factory.create()
+        LoginManager.getInstance().registerCallback(
+            callbackManager,
+            object : FacebookCallback<LoginResult> {
+                override fun onSuccess(loginResult: LoginResult) {
+                    // Signed in successfully, forward credentials to MongoDB Realm.
+                    val accessToken = loginResult.accessToken
+                    runBlocking {
```
cmelchior, line 78: Same, a bit dangerous.

source/sdk/kotlin/app-services/authenticate-users.txt
cmelchior: Yup, this wouldn't be supported in KMM right now, but would depend on platform specific libraries on Android and iOS. I would assume the Swift SDK has the same feature documented. You could just refer to that?

cmelchior (approved):
Looks good to me. I have never used the Google and Facebook login methods, so I cannot speak to that code, but it looks sane.

Only concern I have is the use of `runBlocking` which is a bit dangerous especially if the the thing you run inside is a network operation like login.

For simplicity it is probably fine, but there should probably be a warning about not doing it in a production app.

PR #1949, opened 2022-05-27
(DOCSP-21678): Add Flex Sync property wrappers to docs and SwiftUI Quick Start

dacharyc:
## Pull Request Info

In addition to adding the Flexible Sync versions of the property wrappers, this PR also updates the name of the `Group` object in the SwiftUI QuickStart to `ItemGroup` to resolve a namespace issue raised in docs feedback.

For a reviewer: the CI is failing, but this is expected. This related Flex Sync docs PR fixes the CI issues: #1952 .

TODO Before Merging:
- [x] Update SPM to point at Swift SDK release that includes these updates
- [x] Make sure versionadded points to the right versions
- [x] Rebase from upstream/master to make sure this PR doesn't clobber any name changes

### Jira

- https://jira.mongodb.org/browse/DOCSP-22203

- [SwiftUI Quick Start](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21678/sdk/swift/swiftui/): Add Flex Sync code examples, change `Group` object name to `ItemGroup` in all three versions of the app (Flex Sync, Partition-Based Sync and Local-only)
- [SwiftUI Guide](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21678/sdk/swift/examples/swiftui-guide/#open-a-synced-realm): Update the "Open a Synced Realm" section with Flex Sync property wrappers

### Review Guidelines


mongomoe:
Can we get rid of the sentence: `A common Realm data modeling use case is to have "things" and "containers of things"`  
in https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-21678/sdk/swift/swiftui/#define-models 

or at least not use the word "things". A good alternative maybe:  `A common Realm data modeling use case is to have items and a set containing those items`?

examples/ios/QuickStartSwiftUI/QuickStart.swift
```diff
@@ -35,7 +35,7 @@ let randomNouns = [
     "cork", "mouse pad"
 ]
 
-/// An individual item. Part of a `Group`.
+/// An individual item. Part of a `ItemGroup`.
```
mongomoe, line 38: Might be a silly q, given my inexperience w/ SwiftUI, but why are comments three forward slashes (`///`) rather than two (`//`)

mongomoe:
Is the Feedback subsection in the SwiftUI Quickstart necessary? No other quickstarts specifically call out the feedback form, and the page is already long enough IMO.

mongomoe (commented):
Generally looks pretty good, though I found the QuickStart with Swift UI to be a bit long (though not necessarily a bad thing). Let me know if you have any questions. I think this is close to an LGTM, but I would consider finding ways to shorten that QuickStart (or make it more readable in 1 sitting)

examples/ios/SwiftUIFlexSyncExample/SwiftUIFlexSyncExampleApp.swift
```diff
@@ -0,0 +1,365 @@
...
+            // If there is no user logged in, show the login view.
+            LoginView()
+        }
+    }
+}
+// :snippet-end:
+// :snippet-start: open-realm-view-flex-sync
+// :snippet-start: subscription-state-enum
+enum SubscriptionState {
+    case initial
```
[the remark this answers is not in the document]
dacharyc: Fair point - this isn't actually a thing in the API, but is a thing we are doing programmatically here to defer using the realm until after the subscriptions have been added. There is a Swift SDK PR that manages this, but I'm not sure if it will get merged before MDBW. If it does, I'll remove this from the code. But in the meantime, I've added more explanation that hopefully removes the confusion.

source/sdk/swift/swiftui.txt
```diff
@@ -188,20 +210,41 @@ across devices. Before you can add sync to this app, make sure to:
 - :ref:`Enable anonymous authentication <anonymous-authentication-configuration>`.
 - :ref:`Enable {+sync+} <enable-sync>`.
 
-  1. Use :ref:`Partition-Based Sync <partition-based-sync>`
+  1. Choose :ref:`Partition-Based Sync <partition-based-sync>` or :ref:`Flexible Sync <flexible-sync>`
```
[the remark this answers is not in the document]
dacharyc: When we originally wrote this, we did not have this list of steps here - but a user in the forum had trouble setting it up. It's not enough to just link to "Enable Realm Sync" - we have to specify the configuration details that a user must use when they enable Sync. If we don't tell them what partition key/queryable field to set, or what permissions to use, they can't successfully set up the app to use this quick start.

examples/ios/QuickStartSwiftUI/QuickStart.swift
mongomoe: got it, thanks for the explanation!

mongomoe (approved):
LGTM 👍  great work

source/sdk/swift/swiftui.txt
```diff
@@ -235,31 +278,100 @@ This view has two possible states:
...
+      that includes the ``initialSubscriptions`` parameter. We can use this
+      parameter to :ref:`subscribe to queryable fields 
+      <ios-sync-subscribe-to-queryable-fields>`. These initial subscriptions
+      search for data that matches the queries, and syncs that data to the
+      realm. If no data matches the queries, the realm opens with an initial
+      empty state. 
+      
+      Your client application can only write objects that match the 
+      subscription query to a realm opened with a ``flexibleSyncConfiguration``. 
+      You cannot write objects that don't match the subscription query to the
```
mongomoe, line 301: Nitpick/not required for lgtm: 

The second sentence ("You cannot write objects that don't match the subscription query...) gives very similar info to the first ("Your client application can only write objects that match the 
      subscription query ...) would it be worth combining them into 1 sentence to reduce redundancy?

source/sdk/swift/examples/swiftui-guide.txt
```diff
@@ -97,8 +97,48 @@ for apps that require up-to-date information from the server, such as game
...
+      Realm Swift SDK version 10.27.0 adds :ref:`Flexible Sync <flexible-sync>` 
+      versions of the property wrappers to open Realm with SwiftUI. You can 
+      add subscription queries in ``.onAppear`` after opening the realm.
+
+      .. literalinclude:: /examples/generated/swiftui/flex-sync/SwiftUIFlexSyncExampleApp.snippet.fs-property-wrapper-sans-config-comment.swift
+         :language: swift
+
+      .. versionadded:: 10.28.0
+
+      If you're using Realm Swift SDK version 10.28.0, you can create a
```
mongomoe, line 116: Nitpick/Suggestion not required for LGTM:

These two sentences could be simplified quite a bit and accomplish the same purpose:
1. Remove the "If you're using Realm Swift SDK version 10.28.0" (since the version added is right above this, also presumably this would apply to version 10.29.0 and so on, whenever those release)
2. The first and second sentence could be combined.

"You can create a flexibleSyncConfiguration() with the initialSubscriptions parameter to subscribe to Flexible Sync queries in the configuration."

mongomoe (approved):
Re-LGTM w/ a couple suggestions 👍

examples/ios/SwiftUIFlexSyncExample/SwiftUIFlexSyncExampleApp.swift
```diff
@@ -0,0 +1,354 @@
...
+struct LoginView: View {
+    // Hold an error if one occurs so we can display it.
+    @State var error: Error?
+    
+    // Keep track of whether login is in progress.
+    @State var isLoggingIn = false
+
+    var body: some View {
+        VStack {
+            if isLoggingIn {
```
jsflax, line 230: Kind of surprised this works without it being an if/elseif/else, but if it does, great!

jsflax (approved):

PR #1967, opened 2022-06-07
[Swift SDK] Update Realm/App Services naming

dacharyc:

- [Swift SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/swift-sdk-naming/sdk/swift/)

### Review Guidelines


snooty.toml
```diff
@@ -60,7 +60,7 @@ backend-schema = "Realm Schema"
 frontend-schema = "Realm Object Model"
 cli-bin = ":binary:`realm-cli`" # binary -- DO NOT USE IN LINKS! Will break them.
 cli = "Realm CLI"
-ui = "Realm UI"
+ui = "Atlas App Services UI"
```
nlarew, line 63: Thoughts on changing this to "Admin UI"? It aligns with the Admin API and is a bit *snappier*.

Or maybe even just "UI"?

source/sdk/swift/advanced-guides/client-reset.txt
```diff
@@ -120,7 +120,7 @@ You can choose to write a handler to incorporate any local changes that exist
 in the old copied realm file into the newly downloaded realm.
 
 The SDK performs client resets automatically on startup when instructed 
-to do so by the server. When a client reset occurs, the SDK creates a 
+to do so by the Atlas App Services App. When a client reset occurs, the SDK creates a
```
nlarew, line 123: Do/Should we have a source constant for "Atlas App Services"?

nlarew (approved):
LGTM w/ a couple thoughts

snooty.toml
dacharyc: I think, with the renaming, I'd rather it be clearer what UI we're talking about. Admin UI isn't necessarily a term that Realm Swift SDK users will be able to grok from context, while Atlas App Services UI ought to be pretty easy to figure out from context for people who stumble across the term. But I _also_ didn't check and see how this was used everywhere - I'm just basing it off how I saw it used in the Swift SDK.

source/sdk/swift/advanced-guides/client-reset.txt
dacharyc: There is one, but we don't have one for "Atlas App Services App". I didn't necessarily want to lump this under "{+backend+} App" (or "{+backend+} {+app+}") because that just seems like it would make future renamings more complicated. As it is, a lot of these manual "Atlas App Services App" switches are because source constants were previously used incorrectly or inconsistently - I think I'm coming around to @cbush 's idea that they can be more hassle than they're worth.

cbush: In this case I would just call it "Device Sync backend" or just "backend". (In a lot of cases I preferred "App Services backend" over "App Services App")

dacharyc: If we are opposed to the "Atlas App Services App" I can always follow with another PR to update these references. But I'd rather get the naming updated before World and then retroactively tweak things than to leave this unmerged for now.

PR #1992, opened 2022-06-28
(DOCSP-23481): Incorporate docs feedback to clarify using configurations w/SwiftUI

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-23481

- [SwiftUI Guide/Open a Realm with a Configuration](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-23481/sdk/swift/examples/swiftui-guide/#open-a-realm-with-a-configuration)

### Review Guidelines

### DIFF
--- a/source/sdk/swift/examples/swiftui-guide.txt
+++ b/source/sdk/swift/examples/swiftui-guide.txt
@@ -665,7 +665,7 @@ environment value <inject-realm-as-environment-value>`:
 .. _swiftui-preview-use-in-memory-realm:
 
 Use an In-Memory Realm
-^^^^^^^^^^^^^^^^^^^^^^
+``````````````````````
 
 When possible, use an :ref:`in-memory realm <ios-open-an-in-memory-realm>` 
 to get around some of the state-related issues that can come from using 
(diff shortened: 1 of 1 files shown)


source/sdk/swift/examples/swiftui-guide.txt
```diff
@@ -65,8 +65,14 @@ objects or results.
 
 When you do not specify a configuration, these property wrappers use the 
 :swift-sdk:`defaultConfiguration <Structs/Realm/Configuration.html#/s:10RealmSwift0A0V13ConfigurationV07defaultC0AEvpZ>`. 
-However, you can :ref:`use environment injection to specify a different 
-configuration <ios-swiftui-inject-realm-configuration>`.
+
+You can alternately customize the configuration that the property wrappers use
```
nathan-contino-mongo, line 69:
```suggestion
You can also customize the configuration that the property wrappers use 
```

I think the right word here is alternatively, unless I'm misunderstanding the context. But I think we can cut down on wordaceousness even more.

nathan-contino-mongo (approved):
LGTM, one suggestion

source/sdk/swift/examples/swiftui-guide.txt
dacharyc: Hmm, yeah, you're probably right that alternatively is a better fit here. Normally I'm all in favor of smaller words, but "also" here reads like additive option, where alternatively reads like an either/or, and I think that's what we need here based on the docs feedback.

PR #2060, opened 2022-07-27
(DOCSP-23222): Change Set is cleared - .NET SDK

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-23222

- [React to Changes - .NET SDK # Check if the ChangeSet Has Been Cleared](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-23222-ChangeSet-IsCleared-dotnet/sdk/dotnet/examples/react-to-changes/#check-if-the-changeset-has-been-cleared)

### Review Guidelines

### DIFF
--- a/examples/dotnet/Examples/WorkWithRealm.cs
+++ b/examples/dotnet/Examples/WorkWithRealm.cs
@@ -226,6 +226,7 @@ namespace Examples
                 Id = ObjectId.GenerateNewId(),
                 Name = "Fido"
             };
+            
             realm.Write(() =>
             {
                 fido = realm.Add<Dog>(fido);
--- a/examples/dotnet/Examples/WorkWithRealm.cs
+++ b/examples/dotnet/Examples/WorkWithRealm.cs
@@ -251,12 +251,16 @@ namespace Examples
 
                 if (changes.IsCleared)
                 {
-                    // ... handle a CollectionChanged Event with action `Reset`
+                    // ... handle collection has been cleared ...
                     // :remove-start:
                     changesHaveBeenCleared = true;
                     // :remove-end:
                 }
             });
+            // :snippet-end:
+
+            // :snippet-start: call-handle-collection-changed
+            fido.Owners.AsRealmCollection().CollectionChanged += HandleCollectionChanged;
             // :snippet-end:
 
             // :snippet-start: clear-collection
@@ -271,6 +275,17 @@ namespace Examples
             Assert.IsTrue(changesHaveBeenCleared);
 
         }
+
+        // :snippet-start: define-handle-collection-changed
+        private void HandleCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
+        {
+            if(e.Action == NotifyCollectionChangedAction.Reset)
+            {
+                // ... handle a CollectionChanged Event with action `Reset`
+            }
+        }
+        // :snippet-end:
+
 
         class NotificationUnsub
         {
--- a/examples/dotnet/Examples/WorkWithRealm.cs
+++ b/examples/dotnet/Examples/WorkWithRealm.cs
@@ -209,6 +209,68 @@ namespace Examples
         // :replace-end:
         //:snippet-end:
 
+        [Test]
+        public void ChangeSetCleared()
+        {
+            myRealmAppId = Config.appid;
+            var app = App.Create(myRealmAppId);
+            var realm = Realm.GetInstance("");
+
+            realm.Write(() =>
+            {
+                realm.RemoveAll<Dog>();
+            });
+
+            var fido = new Dog()
+            {
+                Id = ObjectId.GenerateNewId(),
+                Name = "Fido"
+            };
+            realm.Write(() =>
+            {
+                fido = realm.Add<Dog>(fido);
+            });
+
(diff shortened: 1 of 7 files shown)


source/examples/generated/dotnet/WorkWithRealm.snippet.get-notification-if-collection-is-cleared.cs
```diff
@@ -0,0 +1,10 @@
+var token = fido.Owners.SubscribeForNotifications((sender, changes, error) =>
+{
+    if (error != null) return;
+    if (changes == null) return;
+
+    if (changes.IsCleared)
+    {
+        // ... handle a CollectionChanged Event with action `Reset`
```
MongoCaleb, line 8: Do we need to talk about why a Reset is needed here?

MongoCaleb (approved):
LGTM with a question about showing -- or explaining -- why we need to do a Reset.

source/examples/generated/dotnet/WorkWithRealm.snippet.get-notification-if-collection-is-cleared.cs
mongomoe: We don't necessarily need to, I could have made the comment more similar to the other comments in this page, such as "// ... handle ChangeSet cleared ...". I just wanted to be explicit with the info provided in the [release notes for 10.14.0](https://github.com/realm/realm-dotnet/releases/tag/10.14.0):

"Added boolean property ChangeSet.IsCleared that is true when the collection gets cleared. Also Realm collections now raise CollectionChanged event with action Reset instead of Remove when the collections is cleared. Please note that this will work only with collection properties, such as IList and ISet. (Issue https://github.com/realm/realm-dotnet/issues/2856)"

examples/dotnet/Examples/WorkWithRealm.cs
```diff
@@ -209,6 +209,68 @@ public void Notifications()
...
+            realm.Write(() =>
+            {
+                realm.RemoveAll<Dog>();
+            });
+
+            var fido = new Dog()
+            {
+                Id = ObjectId.GenerateNewId(),
+                Name = "Fido"
+            };
```
papafe, line 228: I would put an empty line after this

source/examples/generated/dotnet/WorkWithRealm.snippet.get-notification-if-collection-is-cleared.cs
papafe: I think that maybe this could cause some confusion. Because effectively there are two ways that one can get notifications in this case:
1) `fido.owners.SubscribeForNotifications(....)`. This is the default, Realm-specific listener for changes on collection. In this case `changes.IsCleared` will be true.
2) `fido.owners.AsRealmCollection().CollectionChanged += HandleCollectionChanged`.  This is because the realm collection implements `INotifyCollectionChanged` and is not realm-specific, but more general and is generally used with data binding. 
Inside `HandleCollectionChanged` there will be a parameter `NotifyCollectionChangedEventArgs args` with `args.Action == Reset.`

The two ways are actually independent but the text "handle a CollectionChanged Event with action Reset" make it sound like one is causing the other, but that's not the case. 

I think it would make sense to be more explicit in the page about the fact that there are two ways to listen to changes for a collection, not only related to  this specific `IsCleared` case, but in general. Also it's a little bit confusing that the `Unregister a Change Listener` talks about how to unregister if one uses `CollectionChanged`, but not if one uses `SubscribeForNotification`.

I don't know if my explanation was clear enough, but feel free to contact me if you need additional clarifications.

source/sdk/dotnet/examples/react-to-changes.txt
```diff
@@ -135,6 +135,62 @@ list or other view that represents the collection in the UI.
...
+
+With CollectionChanged  then NotifyCollectionChangedEventArgs args will have args.Action = NotifyCollectionChangedAction.Reset.
+
+
+ Realm Database raises a ``CollectionChanged`` event with the
+action `Reset
+<https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.notifycollectionchangedaction?view=net-6.0>`_.
+
+Register a change listener on a collection's ``IList`` or ``ISet`` property to
+check if that property has been cleared.
```
papafe, line 191: I think that all the necessary info are already contained in the previous text of this section, probably this could be removed

source/sdk/dotnet/examples/react-to-changes.txt
```diff
@@ -135,6 +135,62 @@ list or other view that represents the collection in the UI.
    Every Realm collection implements ``INotifyCollectionChanged``, which allows 
    you to use a collection directly in data-binding scenarios.
 
+.. _dotnet-change-set-cleared:
+
+Check if the ChangeSet Has Been Cleared
```
papafe, line 140: It should be "check if the collection has been cleared"

source/sdk/dotnet/examples/react-to-changes.txt
mongomoe: Whoops sorry, this text wasn't supposed to be in there - I was just using it to base the earlier paragraph on.

papafe (approved):

PR #2061, opened 2022-07-28
(DOCSP-22738): Flutter Quick Start with Sync

mongodben:
## Pull Request Info

- Update quick start with sync info. 
- Fixed some references to "MongoDB Realm"

Must be merged after https://github.com/mongodb/docs-realm/pull/2123 b/c references the tutorial, which is in that PR.

### Jira

- https://jira.mongodb.org/browse/DOCSP-22738

- [Quick Start - Flutter SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-22738/sdk/flutter/quick-start/)

### Review Guidelines


MongoCaleb:
General thoughts:
1. Looks great overall!
2. I would remove the section about the template app. As Dachary mentioned, it *seems* like this page is about the template app rather than "rolling your own". Instead, considering putting in a short note there: "For a complete end-to-end solution that shows all of this foo, check out the <link>Flutter Quick Start</link>." This leads me to #3...
3. I think the Quick Starts should not show a complete solution, because we have the complete solution in the template app and the tutorials based on it, and having it here just adds to the code maintenance / tech debt.

source/sdk/flutter/quick-start.txt
```diff
@@ -255,10 +255,124 @@ Once you've finished working with a realm, close it to prevent memory leaks.
...
+#. Create an Atlas App Services ``App`` instance. Realm uses the ``App``
+   to help manage the Device Sync session.
+#. Log a user into the ``App``.
+#. Open a synced Realm Database instance.
+#. Add a Device Sync subscription to synchronize data between the client app
+   and MongoDB Atlas.
+
+.. procedure::
+
+   .. step:: Configure App Services Backend
```
dacharyc, line 287: I think this makes more sense as a prerequisite than a step 1. The rest of the quick start steps are about using Realm/Device Sync in a client, and this is a whole bunch of backend stuff that you have to do before you can use Realm/Device Sync in a client. Maybe this could be moved into a callout box as a prerequisite similar to how Moe did the Kotlin quick start?

source/sdk/flutter/quick-start.txt
```diff
@@ -255,10 +255,124 @@ Once you've finished working with a realm, close it to prevent memory leaks.
...
+      .. tip:: Use Realm Flutter Template App
+
+         If you want a working Flutter app with Device Sync already set up in the client
+         and on that App Service backend, use the :ref:`Flutter Template App <flutter-template-app>`.
+
+   .. step:: Initialize App Services
+
+      To use App Services features such as authentication and sync, you must
+      access your App Services App using your App ID. You can find your
+      App ID in the App Services UI.
```
dacharyc, line 313: This might be a good place to link to the App Services page on finding your App ID.

source/sdk/flutter/quick-start.txt
```diff
@@ -255,10 +255,124 @@ Once you've finished working with a realm, close it to prevent memory leaks.
...
+#. Create an App Services App backend that you can use to sync data
+   from your client app to MongoDB Atlas.
+#. Create an Atlas App Services ``App`` instance. Realm uses the ``App``
+   to help manage the Device Sync session.
+#. Log a user into the ``App``.
+#. Open a synced Realm Database instance.
+#. Add a Device Sync subscription to synchronize data between the client app
+   and MongoDB Atlas.
+
+.. procedure::
```
dacharyc, line 285: This is totally subjective, but seeing this as a procedure - especially with all the "See also" callout boxes - makes this section feel really chonky and like it takes up a lot of space relative to the rest of the page. I think we should probably restructure it to a section with subheadings, because I think the procedure styling makes this look, well, more procedural than like a "quick start." It "feels" more like a tutorial than just "here's what you need to do to quickly get going."

Also, I think the see also callouts here are probably overkill. We're not using them consistently in every section in the page above, and I think with the new docs styling it would feel pretty overwhelming if we did. I'd probably just use a line instead of the callout box in each section to connect users to the related docs.

dacharyc (changes requested):
I think this is a good step toward what we discussed in lobby a few weeks ago, but I still think it leans more toward tutorial than quick start. I think we can make it leaner in text and styling to make it feel more "quick start"-ey.

dacharyc (approved):
Nice work - LGTM! 👍

PR #2103, opened 2022-08-16
(DOCSP-24486): Android with Kotlin Tutorial

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-24486

- [Android with Kotlin](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-24486-kotlin-sdk/tutorial/kotlin/)

### Review Guidelines


source/tutorial/kotlin.txt
```diff
@@ -0,0 +1,392 @@
...
+         feedback to tell the user they have successfully created a new todo item with
+         their summary and priority. Within the ``Toast.makeText()`` method alter the
+         existing text to the following:
+
+         .. code-block:: kotlin
+            :emphasize-lines: 3
+
+            Toast.makeText(
+                 this,
+                 "Item created: ${item.summary} with priority: ${item.priority}",
```
dacharyc, line 253: I don't know much about how Toasts work - seems to me that this line of text with the priority added could get quite long if the user enters a long todo name. But the next line has `Toast.LENGTH_SHORT`. Will this truncate or wrap the text if it's too long? Do we need to change this `Toast.LENGTH_SHORT` to something else?

source/tutorial/kotlin.txt
```diff
@@ -0,0 +1,392 @@
...
+      ``SyncConfiguration.Builder().initialSubscriptions()`` function that our
+      application calls on start. Within the ``add()`` method, you can see that
+      we are currently subscribing to all documents where the ``OwnerId``
+      property matches the authenticated user. We want to maintain that, but
+      *only* sync Items that are marked as High or Severe priority. 
+
+      As you may recall, the Priority field is of type ``int``, where the highest 
+      priority ("Severe") has a value of 0, and the lowest priority ("Low") has 
+      a value of 3. We can make direct comparisons between an integer and the 
+      Priority property. To do so, edit the LINQ statement to include documents
```
dacharyc, line 300: Isn't `LINQ statement` a .NET term? Should we call this a query statement or something more KMM-ey?

dacharyc (approved):
Overall, this looks good. I just had a couple of questions, and there's a missing hyphen and a spot where I think we need to change the text because it's using a .NET term.

source/tutorial/kotlin.txt
```diff
@@ -0,0 +1,392 @@
...
+               android:layout_marginLeft="20dp"
+               android:layout_marginRight="20dp"
+               android:minHeight="48dp"
+               android:paddingLeft="5px"
+               android:paddingRight="5px"
+               tools:ignore="SpeakableTextPresentCheck" />
+               
+      #. Next, we will populate the spinner with user choices. Since the choices
+         for priority levels are predetermined, we will provide them through the
+         :android:`string resources file <guide/topics/resources/string-resource>`.
```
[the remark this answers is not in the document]
mongomoe: I'm not sure of a quick solution for only having these priorities defined in one location. [Android's Spinner must be populated from a SpinnerAdapter, such as an ArrayAdapter](https://developer.android.com/guide/topics/ui/controls/spinner). The ArrayAdapter. createFromResource method takes a resource id. In my current code, I pass in the resource id of the "priorities" string-array. However, I can't pass in the Enum (since it's not a resource, and I can't use the enum in the string resource file since its an XML file).

Looks like the .NET SDK tutorial has a similar issue:

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-24486-kotlin-sdk/tutorial/dotnet/#add-a-new-property

source/tutorial/kotlin.txt
mongomoe: LENGTH_SHORT refers to the duration the message is displayed for. There's no options for wrapping the text unfortunately. For a tutorial app + a quick piece of feedback, I think it gets the message across. If users want to customize their app for a popup/feedback/modal with a more versatile set of options for styling, there's alternative Android UI features they can use (however, it wouldn't make sense to show those for a tutorial).

source/tutorial/kotlin.txt
mongomoe: Whoops sorry accidentally left that in when I copied and pasted from .NET, it should be RQL - will change that.

source/tutorial/kotlin.txt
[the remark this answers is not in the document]
cmelchior: You can do it using code instead of an XML resource, but it requires a bit of a refactor: https://developer.android.com/develop/ui/views/components/spinner

source/tutorial/kotlin.txt
```diff
@@ -0,0 +1,392 @@
...
+      shown here:
+
+      .. code-block:: kotlin
+         :emphasize-lines: 3-8
+
+         config = SyncConfiguration.Builder(user, setOf(Item::class))
+             .initialSubscriptions { realm ->
+                 add(
+                     realm.query<Item>(
+                         "owner_id == \$0 AND priority <= ${PriorityLevel.High.priority.toString()}",
```
cmelchior, line 311: It shouldn't be needed to escape the `$0` in this case. Also it looks a bit weird that a combination of `$0` and `${PriorityLevel.High.priority.toString()}` is used. I would probably either pass both in through `$0` and `$1` or create both in place like the enum value.

cmelchior (approved):

PR #2105, opened 2022-08-17
React Native Tutorial

nlarew:
### Jira

- (DOCSP-23975) Update React Native Tutorial to Flexible Sync

- [React Native Tutorial](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/rn-tut/tutorial/react-native/)

### Review Guidelines


docs-release-notes.md
```diff
@@ -1,3 +1,31 @@
+# August 12, 2022
```
mongomoe, line 1: Why are there release notes in this PR?

examples/java/local/README.md
```diff
@@ -30,11 +30,29 @@ by Android Studio, after waiting a short time (30 seconds to a minute) you
 should be able to run the "All Tests" target on a selected AVD using the green play button in the
 toolbar.
 
+#### Test via Android Studio
```
mongomoe, line 33: you may have pulled from a different branch so you have all these changes that are unrelated to the tutorial. Or maybe your branch (rn-tut) is ahead of 'upstream:tutorial-phase1' and then we need to pull from 'upstream:master' into 'upstream:tutorial-phase1'

nlarew: Yeah switching branches pulled in extra commits from master. I updated the commit history so this should be fixed.

nlarew:
@mongomoe Fixed the git history

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
 Overview
 --------
 
-Realm provides a Node SDK that allows you to create a Mobile Application with
-React Native. This tutorial illustrates the creation of a "Task Tracker" React Native
-application that allows users to: 
+You can use the Realm JavaScript SDK and the official Realm React hooks
```
mongomoe, line 18:
```suggestion
You can use the Realm JavaScript SDK and the official Realm React
```

I know we call it Realm Hooks sometimes internally, but iirc we're supposed to just refer to it as either the "Realm React" or "@realm/react" library.

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+library to build a mobile application with React Native. This tutorial
+walks you through how to build your own app that uses flexible sync.
 
-- Sign in to their account with their email and password and sign out later.
-- View a list of projects they are a member of.
-- View, create, modify, and delete tasks in projects.
-- View a list of team members in their project.
-- Add and remove team members to their project.
+For this tutorial, we'll start with a pre-built template application to
```
mongomoe, line 22: Suggestion (not required for LGTM): I think the other SDKs had a slightly more descriptive variation of a sentence that explains the tutorial:
```
This tutorial adds functionality to the Template App. You will add a new Priority field to the existing Item model and update the Flexible Sync Subscription to only show items within a range of priorities.
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+they finish it. Some basic functionality is already set up for you:
 
-- The React Native CLI. Detailed instructions for setting up your development 
-  environment, including the CLI tools, can be found in the 
-  `React Native docs <https://reactnative.dev/docs/environment-setup>`_.
+- Users can create email/password accounts and log in and out of the app.
 
-- :ref:`Set up the Atlas App Services backend <tutorial-task-tracker-create-realm-app>`.
+- Users can create, read, update, and delete ``Item`` objects that
+  represent individual to-do tasks.
```
mongomoe, line 59:
```suggestion
  represent individual items to do.
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+   :header-rows: 1
+   :widths: 20 80
+
+   * - File Name
+       
+     - Description
+     
+   * - ``ItemSchema.js``
+       
+     - This is where we define the ``Item`` class, including its object
```
mongomoe, line 221: Suggestion/Not required for LGTM: For this table, I don't think it's necessary to start each description sentence with "This is", it gets repetitive and doesn't add value to the sentence. 

For example, the first description (ItemSchema.js), the following might be a bit more concise and avoid the fragment 'this is':

```
The Item class, including its object schema. We import the class in RealmContext.js
to include it in the app's Realm schema.
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
-      will close the realm any time the user changes or the component is unmounted:
-
-      .. literalinclude:: /tutorial/generated/code/final/AuthProvider.snippet.user-realm-cleanup.js
-         :language: javascript
-
-
+   .. step:: Define the Priority Levels
+      
+      We want to allow a small number of named priority levels and we
+      want to easily be able sort the levels. In TypeScript, you could
```
mongomoe, line 274: Question: Do we really need to describe what to do in TS if the template app doesn't use typescript?

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
-
-
-   .. step:: Run and Test
-
-      Congratulations! Now that you have completed the code, you can run the app and
-      check functionality. Here are some things you can try in the app:
+         The priority levels in the enum are ordered from most important
+         to least. The corresponding index value for each level
+         increases from the most important, ``Priority[0]``, to the
+         least, ``Priority[3]``. This means that a "higher" priority
```
mongomoe, line 317:
```suggestion
         least important, ``Priority[3]``. This means that a "higher" priority
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
-      - Create another user with email *second@example.com*
-      - Navigate to *second@example.com*'s project
-      - Add, update, and remove some tasks
-      - Click "Manage Team"
-      - Add *first@example.com* to your team
-      - Log out and log in as *first@example.com*
-      - See two projects in the projects list
-      - Navigate to *second@example.com*'s project
-      - Collaborate by adding, updating, and removing some new tasks
+         At this point your React Native ``Item`` model and its
```
mongomoe, line 361: The other tutorials have a similar statement about why this doesn't break sync, do you just want to remove this paragraph and include that note, it reads:

```
Why Didn't This Break Sync?
Adding a property to a Realm object is not a breaking change and therefore does not 
require a client reset. The template app has Development Mode enabled, so changes
to the client Realm object are reflected in the server-side schema. For more information,
 see Enable or Disable Development Mode and Update Your Schema.

```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+           const [priority, setPriority] = useState(Priority.High);
+           
+           return (
+             <View style={styles.modalWrapper}>
+               <Text h4 style={styles.addItemTitle}>
+                 Add Item
+               </Text>
+               <Input placeholder="What do you want to do?" onChangeText={setSummary} />
+               <Picker
+                 style={{width: '80%'}}
```
mongomoe, line 417: Suggestion/not required for LGTM: This maybe a personal pet-peave of mine, but can we change the inline-styling to a styles.pickerWidth or something?

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+         useEffect(() => {
+           // initialize the subscriptions
+           const updateSubscriptions = async () => {
+             await realm.subscriptions.update(mutableSubs => {
+               // subscribe to all of the logged in user's to-do items
+               let ownItems = realm
+                 .objects("Item")
+                 .filtered(`owner_id == "${user.id}"`);
+               if(showImportantOnly) {
+                 // More important Items have lower priority values
```
mongomoe, line 598:
```suggestion
                 // Filter 'Severe' or 'High' priority level items
```
The sentence "More important Items have lower priority values" seems very clunky and is hard to understand (especially if the reader skimmed the explanation that priority value is inversely correlated to importance)

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+           }
+           
+           static schema = {
+             name: 'Item',
+             properties: {
+               _id: 'objectId',
+               isComplete: {type: 'bool', default: false},
+               summary: 'string',
+               owner_id: 'string',
+               priority: 'int' // Store the index value of the Priority enum rather than the name
```
mongomoe, line 353: Should this be?: 

```
priority: {type: 'int', default: Priority.High},
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+      .. code-block:: javascript
+         :caption: ItemSchema.js
+         :emphasize-lines: 6, 11, 21
+         
+         export class Item {
+           constructor({
+             _id = new BSON.ObjectId(),
+             isComplete = false,
+             owner_id,
+             priority = Priority.High, // Default to High priority if none is specified
```
mongomoe, line 338:
```suggestion
             priority = Priority["High"], // Default to High priority if none is specified
```

If it's Priority.High you will get the error: undefined is not an object

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+      - pass ``priority`` to the ``onSubmit()`` handler
+
+      .. code-block:: javascript
+         :caption: src/CreateToDoPrompt.js
+         :emphasize-lines: 4, 12-19, 23
+         
+         export function CreateToDoPrompt(props) {
+           const {onSubmit} = props;
+           const [summary, setSummary] = useState(null);
+           const [priority, setPriority] = useState(Priority.High);
```
mongomoe, line 408:
```suggestion
           const [priority, setPriority] = useState(Priority["High"]);
```
Priority.High will lead to an undefined is not an object error. See: https://github.com/mongodb/docs-realm/pull/2105/files#r952881610

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+      object. Let's update the app UI so that you can choose a priority
+      value when you add a new to-do to the list.
+      
+      First, we'll install an external library to implement the priority
+      picker component. Run the following in your terminal inside of
+      your project root:
+      
+      .. code-block:: shell
+         
+         npm install @react-native-picker/picker
```
mongomoe, line 382: Needed fix: The user needs to pod install after installing @react-native-picker/picker, otherwise they'll get the following error:

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+         :caption: ItemSchema.js
+         
+         function createEnum(arr) {
+           arr.forEach((p, i) => arr[p] = i);
+           return arr;
+         }
+         
+         // Priority.High === 1
+         // Priority[Priority.High] === "High"
+         const Priority = createEnum([
```
mongomoe, line 305: Needed fix: You need to export 'Priority' in order to import it into `CreateToDoPrompt.js` as shown in the step [3.Add a Priority Picker](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/rn-tut/tutorial/react-native/#add-a-priority-picker)

mongomoe, line 305:
```suggestion
         export const Priority = createEnum([
```

mongomoe: Needed fix: You need to export 'Priority' in order to import it into `CreateToDoPrompt.js` as shown in the step [3.Add a Priority Picker](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/rn-tut/tutorial/react-native/#add-a-priority-picker)

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+             {/* ... */}
+           </ListItem>
+         ))}
+         <Button
+           title={showImportantOnly ? 'Show All' : 'Show Important Only'}
+           buttonStyle={{
+             ...styles.addToDoButton,
+             backgroundColor: showImportantOnly ? '#00A35C' : '#FFC010',
+           }}
+           onPress={() => setShowImportantOnly(b => !b)}
```
mongomoe, line 570: Suggestion/not required for lgtm (but strongly suggest): Can you change it from 'b' to something more descriptive, maybe: ``` onPress={() => setShowImportantOnly(buttonType => !buttonType)}```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+                 Add Item
+               </Text>
+               <Input placeholder="What do you want to do?" onChangeText={setSummary} />
+               <Picker
+                 style={{width: '80%'}}
+                 selectedValue={priority}
+                 onValueChange={value => setPriority(value)}>
+                 {Priority.map(p => (
+                   <Picker.Item key={p} label={p} value={Priority[p]} />
+                 ))}
```
mongomoe, line 422: Suggestion/not required for LGTM (but strongly suggested): Can we choose something more descriptive than 'p', maybe:
```
                 {Priority.map((priorityValue, i) => (
                   <Picker.Item key={`picker-${priorityValue}-${i}`} label={priorityValue} value={Priority[priorityValue]} />
                 ))}
```

Note: I chose "`picker-${priorityValue}-${i}`" above because keys should be unique

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+     
+   * - ``CreateToDoPrompt.js``
+       
+     - This is a UI form that lets us enter data for new ``Item``
+       objects. The code that actually creates new objects is in
+       ``ItemListView.js``.
+     
+   * - ``LogoutButton.js``
+       
+     - This is a re-usable button that logs out an authenticated user.
```
mongomoe, line 257: Needed fix:
```suggestion
     - This is a reusable button that logs out an authenticated user.
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
-      finished with it. You can return a cleanup function from the effect, which
-      will close the realm any time the user changes or the component is unmounted:
-
-      .. literalinclude:: /tutorial/generated/code/final/AuthProvider.snippet.user-realm-cleanup.js
-         :language: javascript
-
-
+   .. step:: Define the Priority Levels
+      
+      We want to allow a small number of named priority levels and we
```
mongomoe, line 273:
```suggestion
      We want to allow a small number of named priority levels, and we
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+         okay for now.
+
+         The template app uses development mode by default. In this
+         mode, any data model changes that your client app syncs will
+         automatically update the backend schema. The next time you run
+         the React Native app, the schemas will sync.
+
+   .. step:: Add a Priority Picker
+
+      Your apps data model now includes a ``priority`` for each ``Item``
```
mongomoe, line 372:
```suggestion
      Your app's data model now includes a ``priority`` for each ``Item``
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+      .. important::
+
+         Don't forget to add ``showImportantOnly`` to the list of
+         dependencies in the second argument of ``useEffect``.
+
+   .. step:: Test the App
+
+      Your app is now set up to modify its sync subscription based on
+      the current mode. However, if you run the app and switch to
+      "important only" mode, you'll see that the list doesn't change and
```
mongomoe, line 617:
```suggestion
      "important only" mode, you'll see that the list doesn't change, and
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,669 @@ React Native Tutorial
...
+
+         npx mongodb-realm-cli push
+
+   .. step:: Run the Completed App
+      
+      Once your updated sync configuration is deployed, your app is
+      complete and fully functional. Congratulations!
+      
+      Rebuild and run the app to make sure everything works. You should
+      be able create, complete, and delete to-do items as well as switch
```
mongomoe, line 668:
```suggestion
      be able to create, complete, and delete to-do items as well as switch
```

mongomoe (commented):
Great job on this so far! There are a few fixes needed to get this to LGTM which I've labelled as required fix, and there's also various grammar/wording fixes needed. Most of the other comments are suggestions which you can choose whether or not to incorporate into the PR.

source/tutorial/react-native.txt
nlarew: I mainly wanted a handy reference to say "doing it this way gives you the TS enum behavior" - I'll update to focus less on TS while still keeping the pointer.

source/tutorial/react-native.txt
nlarew: Good point, updated

source/tutorial/react-native.txt
nlarew: I thought about doing this but it's an extra step for something that's not the point of the tutorial and is still a one-liner here. Not sure it justifies the extra step but if you think it's worth it we can change.

source/tutorial/react-native.txt
nlarew: Oops good catch! I forgot I did this 😅

source/tutorial/react-native.txt
nlarew: +1 for a more descriptive name. Keys only need to be unique within the map and the enum guarantees that each priority value is unique in the enum, so in this case just `priorityValue` will work.

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,689 @@ React Native Tutorial
...
+         }
+         
+         // Priority.High === 1
+         // Priority[Priority.High] === "High"
+         export const Priority = createEnum([
+           "Severe",
+           "High",
+           "Medium",
+           "Low",
+         ])
```
takameyer, line 295: The `createEnum` function is a bit overkill.  I would just define this as an object in descending order:
```
export const Priority = {
  Low: 0,
  Medium: 1,
  High: 2,
  Severe: 3
}
```
Then one can reference the priority as `Priority.Medium` which makes the code clearer when doing logic comparisons or assignments:
```
if( item.priority > Priority.High ) 
//...

realm.write(() => {
  item.priority = Priority.Low
})
```

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,689 @@ React Native Tutorial
...
-      named ``removeTeamMember``, that removes a team member:
-
-      .. literalinclude:: /tutorial/generated/code/final/ManageTeam.snippet.remove-team-member.js
-         :language: javascript
-
-
-   .. step:: Run and Test
+      .. tip::
+      
+         In a TypeScript app, you could define an `enum
```
takameyer, line 305: We should also have a TypeScript version of this template and tutorial.  When doing so, one should consult us about how to correctly do class based models (there are changes coming to this in v11)

takameyer (commented):
Since there are descrepencies between the source code snippets in this tutorial and the template that was downloaded, I am going to pause this review. 
I will continue reviewing when I have a template where the code matches the tutorial, so I don't have to pull my hair out fixing errors.

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,689 @@ React Native Tutorial
...
-      - Navigate to *second@example.com*'s project
-      - Add, update, and remove some tasks
-      - Click "Manage Team"
-      - Add *first@example.com* to your team
-      - Log out and log in as *first@example.com*
-      - See two projects in the projects list
-      - Navigate to *second@example.com*'s project
-      - Collaborate by adding, updating, and removing some new tasks
+         cd ios
+         pod install --repo-update
```
takameyer, line 397: This is only for `ios`.  It should be clear that they will have to rerun `npm run ios` or `npm run android` in order for the picker ot be useable.  They also might have to restart their realm bundler

takameyer (changes requested):
Really nice work here.  I was about to comment on the queryable field error, as I had tried to apply the filter right away and noticed it didn't work.  Fixed it myself and realized the next session was addressing said issue 😂 What was cool though is that the app updated without restarting as soon as I added the queryable field in the atlas ui.  Nice wow effect.

Feel free to get back to me when my feedback has been applied or if there are any questions on anything i've commented on.

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,689 @@ React Native Tutorial
...
-      If you plan to run the app on an iOS device or emulator, you need to install
-      additional dependencies with
-      `CocoaPods <https://guides.cocoapods.org/using/getting-started.html>`__. Change
-      directories to the ``ios`` directory and run ``pod install``, as shown here:
+      
+      To build and run the app on an iOS device or simulator, install
+      the additional iOS dependencies with `CocoaPods
+      <https://guides.cocoapods.org/using/getting-started.html>`__.
 
       .. code-block:: shell
```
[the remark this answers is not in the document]
nlarew: Neat! Good suggestion thanks

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,689 @@ React Native Tutorial
...
-            # use_flipper!
-            #post_install do |installer|
-            #  flipper_post_install(installer)
-            #end
+         .. tab:: Android
+            :tabid: android
+            
+            .. code-block:: shell
+               
+               npx react-native run-android
```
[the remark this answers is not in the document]
nlarew: Good catch, added

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,689 @@ React Native Tutorial
...
+         guide and ensure that you've followed all the steps for
+         your Development OS and Target OS.
+
+   .. step:: Test the App
+      
+      When the build completes, you should have a functional app running
+      on your simulator. In the app, register a new account and play
+      around with the app for a few minutes:
+
+      - Add a few to-do items to the list.
```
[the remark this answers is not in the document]
nlarew: Ah yeah I actually already installed the package just never added the import 😅

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,689 @@ React Native Tutorial
...
+----------------------------------
+
+Now that you have the template app running let's dive into the code to
+see what we're working with.
+
+Atlas App
+~~~~~~~~~
+
+The template app includes a fully configured Atlas App in the
+``backend`` directory. It has a unique ``app_id`` value in
```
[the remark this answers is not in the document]
nlarew: Yeah this is a side-effect of how we organize the template apps so not a simple change unfortunately. I can bring it up with the team.

source/tutorial/react-native.txt
nlarew: The code samples you have will work with the enum too. I chose to use the enum approach because it gives a two-way mapping between the level name and value. That way I can use the `Priority.High` syntax and _also_ map through the Priority enum to get the name and value together in the Picker component.

source/tutorial/react-native.txt
nlarew: Yeah agreed - I think we've held off on going full force into TS until the models stuff is fully baked but we're all about it.

nlarew:
@takameyer Thanks for the comments! Update the page, lemme know what you think.

source/tutorial/react-native.txt
takameyer: Ah, sure.  Makes sense that the Picker component needs an array instead of an object.

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,714 @@ React Native Tutorial
...
--------------
+Before You Begin
+----------------
 
-Ensure that you have the following:
+You must set up your local environment for React Native development
+before you can begin this tutorial. For detailed instructions, see
+`Setting up the development environment
+<https://reactnative.dev/docs/environment-setup>`_ in the React Native
+docs.
```
takameyer, line 44: We should add a note here about Node 17.  And perhaps guide the user to [`nvm`](https://github.com/nvm-sh/nvm) if they need to downgrade Node.

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,714 @@ React Native Tutorial
...
-      - See two projects in the projects list
-      - Navigate to *second@example.com*'s project
-      - Collaborate by adding, updating, and removing some new tasks
+      .. code-block:: shell
 
+         npx pod-install
+      
       .. tip::
+         
+         You may need to restart your project bundler after installing.
```
takameyer, line 423: They will also need to rebuild the app with `npm run ios` and `npm run android`.  I would recommend running these commands first, this way they can see the changes as they update their code and there won't be any errors.

takameyer (approved):
Just a couple more requests.  Otherwise LGTM :shipit:

source/tutorial/react-native.txt
```diff
@@ -15,418 +15,689 @@ React Native Tutorial
...
-         that enables Flipper and comment out those lines:
+         .. tab:: iOS
+            :tabid: ios
+            
+            .. code-block:: shell
 
-         .. code-block:: text
+               npx react-native run-ios
 
-            # Enables Flipper.
```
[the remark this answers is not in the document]
nlarew: Updated RN!

source/tutorial/react-native.txt
nlarew: Node 17 is supported now since we're on the latest version of RN

source/tutorial/react-native.txt
mongomoe: Good decision then :)

mongomoe (approved):
LGTM, fantastic job 👍

PR #2115, opened 2022-08-24
Data binding in Xamarin & MAUI

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-24750

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/data_binding/sdk/dotnet/fundamentals/data-binding/


source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+In Xamarin and MAUI applications, data binding updates the UI when an underlying 
+object changes, and vice versa.
```
papafe, line 16: I would say "allows updating the UI...". It's not strictly the data binding doing it. 
Also it should be "when an underlying object or collection changes"

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+In Xamarin and MAUI applications, data binding updates the UI when an underlying 
+object changes, and vice versa. 
+Realm objects are :ref:`live objects <dotnet-live-object>` that update automatically
```
papafe, line 17: This is maybe nitpicking, but maybe we should say "Realm objects and collections are live, and so update automatically to ...", so that we continue speaking about both objects and collections throughout.

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+   :class: singlecol
+
+In Xamarin and MAUI applications, data binding updates the UI when an underlying 
+object changes, and vice versa. 
+Realm objects are :ref:`live objects <dotnet-live-object>` that update automatically 
+to reflect data changes, making data binding the UI simple and straight-forward. 
+This is because Realm objects implement 
+`INotifyPropertyChanged <https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifypropertychanged>`__ 
+and Realm collections implement 
+`INotifyCollectionChanged <https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.inotifycollectionchanged>`__.
```
papafe, line 22: Maybe "Implement INotifyCollectionChanged underneath so all changes can be automatically reflected in the UI, without the need to do additional work"? Too much repetition?

papafe (changes requested):

papafe:
I would say that this page should contain snippets to show most of what people want to do with Realm and Xamarin/MAUI. I'm not sure if we need to have the code snippet coming necessarily from the template app, but I think we should show:

Let's say that we have a model like this:
```csharp
public class Person : RealmObject
{
    public string Name { get; set; }

    public int Age { get; set; }

    public IList<string> ChildNames { get; }

	public IList<Dog> Dogs { get; }
}

public class Dog : RealmObject
{
    public string Name { get; set; }

    public string Breed { get; set; }

   public int DogYears { get; set; }
}
```

--- Binding to an object ----
In code :
```csharp
public Person Person {get;}

...
Person = realm.All<Person>().First() //For example
```
In XAML then we can do:
```
            <Label Text="{Binding Person.Name}" />
            <Label Text="{Binding Person.Age}" />
```
Changing Name or Age in Person will update the UI. (if those were Editor/Entry or another input form, then changes in the UI are also saved in Realm, so the value of Age and Name will change)

--- Binding to a collection ---
You can bind to a query (this is what you've shown)

You can also bind to any collection (like Dogs or ChildNames in Person). For example:
In code :
```csharp
public IEnumerable<Dog> Dogs {get;}

...
...
Dogs = person.Dogs;
```
In XAML then we can do (using the same UI you have in the example with the list):
```
<ListView x:Name="listItems" ItemsSource="{Binding Dogs}">

...
... 


<Label Text="{Binding Breed}"/>


```

Finally, you can also filter collections of objects like it's done with queries by using AsRealmQueryable().
For example:
```csharp
OldDogs = person.Dogs.AsRealmQueryable().Where( d => d.DogYears > 55);
```

I understand it's a lot of different things, but those are probably the most common use cases that we get asked about, so it would be a good idea to have them shown somewhere

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+Realm objects are :ref:`live objects <dotnet-live-object>` that update automatically 
+to reflect data changes, making data binding the UI simple and straight-forward. 
+This is because Realm objects implement 
+`INotifyPropertyChanged <https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifypropertychanged>`__ 
+and Realm collections implement 
+`INotifyCollectionChanged <https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.inotifycollectionchanged>`__. 
+
+.. important:: Don't Use ToList()
+
+   Because Realm objects are live objects, you should not cast a Realm collection
```
papafe, line 26: Maybe "Because Realm collections are live"

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+to reflect data changes, making data binding the UI simple and straight-forward. 
+This is because Realm objects implement 
+`INotifyPropertyChanged <https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.inotifypropertychanged>`__ 
+and Realm collections implement 
+`INotifyCollectionChanged <https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.inotifycollectionchanged>`__. 
+
+.. important:: Don't Use ToList()
+
+   Because Realm objects are live objects, you should not cast a Realm collection 
+   using ``ToList()``. Casting the collection to a list makes the collection no
```
papafe, line 27: Calling toList() it's not exactly casting. You could say "This forces loading the collection objects from memory, and the resulting collection is also no more live."

papafe (changes requested):

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+In Xamarin and MAUI applications, data binding updates the UI when an underlying 
+object changes, and vice versa. 
+Realm objects are :ref:`live objects <dotnet-live-object>` that update automatically 
+to reflect data changes, making data binding the UI simple and straight-forward.
```
dacharyc, line 18: Nit: "simple" and "straight-forward" are value judgements that we should try to avoid, as not all developers may agree. We should probably reword to avoid using these terms. Building on papafe's note above, maybe something like:

> Realm objects and collections are live, and so update automatically to reflect data changes. When you bind these to your UI, your UI always reflects the latest data.

Or something more directly about the benefit, rather than a subjective value judgement.

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+
+.. important:: Don't Use ToList()
+
+   Because Realm objects are live objects, you should not cast a Realm collection 
+   using ``ToList()``. Casting the collection to a list makes the collection no 
+   longer a live object.
+
+Data Binding a Realm Collection
+-------------------------------
+In the following code example, we're creating a public ``IEnumerable`` of
```
dacharyc, line 32: Nit: indirect sentence structure is a bit less readable. Consider rewording to something like:

> In the following code example, we create a public...

dacharyc (approved):
Overall LGTM, just a couple of nits.

Also, a note about the page name. In the recent App Services & Realm renaming document, there's a note that says:

> When referring to the Realm SDKs avoid using terms like 'binding' or 'wrapper' that may add confusion or imply that the SDKs offer fewer features/power than they actually do

I wonder if there's another name we could use for this page that isn't so general - maybe something like "Bind Realm Collections to the UI"? I know that's a mouthful, but "Data Binding" is such a general term I would think using it as a page title falls within the spirit of this marketing note.

papafe:
@MongoCaleb Another thing I think it's definitely worth mentioning (probably in one of those warning boxes) is that, at the moment, Realm does not work with compiled bindings for writes.
This means that if you use Realm objects just to take values to show in the UI, then it's fine. But if you need to have a value of a Realm object to change when a UI field changes (like with an Editor/Entry, for instance), then this will give a runtime exception complaining that there is no transaction open. 
This limitation is something we're trying to solve, at least for MAUI

source/sdk/dotnet/fundamentals/data-binding.txt
MongoCaleb: > Too much repetition == true

source/sdk/dotnet/fundamentals/data-binding.txt
MongoCaleb: @papafe , this confuses me: 
> But if you need to have a value of a Realm object to change when a UI field changes (like with an Editor/Entry, for instance), then this will give a runtime exception complaining that there is no transaction open. 

Our current template app & tutorial work to update just fine. UI changes are reflected in the realm object. I must be missing a nuance in your comment.

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+                            <ColumnDefinition Width="1*"/>
+                            <ColumnDefinition Width="1*"/>
+                        </Grid.ColumnDefinitions>
+                        <Label Text="{Binding Summary}"/>
+                        <Switch x:Name="chkCompleted" 
+                           IsToggled="{Binding IsComplete}" 
+                           AutomationId="{Binding Id}"/>
+                        <StackLayout Grid.Column="2" VerticalOptions="Center">
+                            <Image WidthRequest="30" x:Name="btnDelete" 
+                                    Source="trash.png" AutomationId="{Binding Id}">
```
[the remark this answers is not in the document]
MongoCaleb: dunno why they were in there. Removed.

MongoCaleb:
> I understand it's a lot of different things, but those are probably the most common use cases that we get asked about, so it would be a good idea to have them shown somewhere

I'm not sure we're actually showing anything very different in these 3 examples. In each, we're binding the UI to a live object/collection. Showing an object and 2 different live collections doesn't make much sense to me, since it's the same binding technique in all 3, no? We show the filtering of a collection in other examples in the docs already. If you can help me understand how these 3 examples differ enough to help our users, that would be great.

source/sdk/dotnet/fundamentals/data-binding.txt
papafe: The problem is if compiled bindings are used. By default Xamarin does not use compiled bindings (while MAUI does it I think).

papafe:
> > I understand it's a lot of different things, but those are probably the most common use cases that we get asked about, so it would be a good idea to have them shown somewhere
> 
> I'm not sure we're actually showing anything very different in these 3 examples. In each, we're binding the UI to a live object/collection. Showing an object and 2 different live collections doesn't make much sense to me, since it's the same binding technique in all 3, no? We show the filtering of a collection in other examples in the docs already. If you can help me understand how these 3 examples differ enough to help our users, that would be great.

I would say this is more to be sure to cover the ground for all the most common cases that people encounter. A lot of new developers are confused about how data binding works, so having everything shown to them could help setting everything quicker. 
Binding to an object is definitely basic, but it shows that it's auto updating. The extra example about collection (person.Dogs) is to show that it's possible to bind also to collections inside objects, not only to queries (because it's not necessarily clear). Regarding the filtering, I understand it's shown in other parts, but maybe it's not evident that it's particularly useful in this case. 
We get a lot of questions/doubts/bug requests around bindings, and it would be nice to be able to just send a link to the docs with examples. Also, hopefully, this would reduce the number of such posts 😁

MongoCaleb:
Bluehawk error will exist until https://github.com/mongodb-university/Bluehawk/pull/129 is merged.

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.snippet.bind-list-xaml.xaml
+   :language: xml
+
+Binding to an Embedded Collection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+In this example, we bind to a collection of Items within the ``Items`` property 
+of a specific ``Employee``. 
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.xaml.snippet.bind-embedded-cs.cs
```
papafe, line 75: This example shows the wrong property to bind to.

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+Binding to a Realm Collection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+In this code example, we create a public ``IEnumerable`` of the ``Item``
+Realm object and then populate that collection with all 
+of the documents in the Items collection:
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.xaml.snippet.bind-list-cs.cs
+   :language: csharp
+
+In the associated XAML file, we bind a ``ListView`` to the **Items** ``IEnumerable``
```
papafe, line 64: It's not the listItems enumerable anymore that is being shown

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+Realm object and then populate that collection with all 
+of the documents in the Items collection:
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.xaml.snippet.bind-list-cs.cs
+   :language: csharp
+
+In the associated XAML file, we bind a ``ListView`` to the **Items** ``IEnumerable`` 
+and then use the Item's property values to display information in the ListView:
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.snippet.bind-list-xaml.xaml
```
papafe, line 67: The name of the ListView is "listItems", but this is a list of employees

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+
+Data Binding Examples
+---------------------
+The following code snippets show both the C# and XAML for 3 typical data 
+binding use cases. They all use the following Realm objects:
+
+.. literalinclude:: /examples/generated/dotnet/Models.snippet.binding-classes.cs
+   :language: csharp
+
+Binding to a Single Realm Object
```
papafe, line 41: Probably Single shouldn't be capitalized? (also object probably)

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,82 @@
...
+.. literalinclude:: /examples/generated/dotnet/MainPage.xaml.snippet.bind-list-cs.cs
+   :language: csharp
+
+In the associated XAML file, we bind a ``ListView`` to the **Items** ``IEnumerable`` 
+and then use the Item's property values to display information in the ListView:
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.snippet.bind-list-xaml.xaml
+   :language: xml
+
+Binding to an Embedded Collection
```
papafe, line 70: Maybe "embedded" collections isn't the right term here, it makes it sounds like it's related to embedded object. 
Either we call it "Binding to a Collection property", or this could also be part of the previous section (Binding to a Realm Collection), and we say something like "Please note that you can also bind to collection properties, like in the following example"). If we go this way then there are 2 cases and not 3 😁

source/sdk/dotnet/fundamentals/data-binding.txt
papafe: It also needs to use "AsRealmQueryable", otherwise it does not work. I would also add a note that we are using "AsRealmQueryable" because we're filtering the list, otherwise we can bind to the list directly, without the need to call it.

papafe (changes requested):
We also need a warning box that says something like:
" Please note that Realm is not yet compatible with compiled bindings"

source/sdk/dotnet/fundamentals/data-binding.txt
MongoCaleb: No, but removed anyway. :)

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,97 @@
...
+---------------------
+The following code snippets show both the C# and XAML for 3 typical data 
+binding use cases. They all use the following Realm objects:
+
+.. literalinclude:: /examples/generated/dotnet/Models.snippet.binding-classes.cs
+   :language: csharp
+
+.. note:: ToList()
+
+   Because Realm collections are live, you should not call ``ToList()`` when
```
papafe, line 43: Not sure what's happening, but this looks a little strange in the page

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,97 @@
...
+each employee's **EmployeeId** value:
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.snippet.bind-list-xaml.xaml
+   :language: xml
+
+We can also bind to a *collection property* of a Realm object. In this example, 
+we bind to the ``Items`` property of a specific Employee``. The ``Items`` 
+property on the ``Employee`` class is of type ``IList<Item>``.
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.xaml.snippet.bind-embedded-cs.cs
```
papafe, line 83: This example still shows the wrong property `public Employee Employee123 { get; }`, instead of `public IEnumerabel<Items> IncompleteItems { get; }`

papafe (approved):
Apart from those two small things it looks good! I've just looked at the staged page, and not at all the files that are in the PR. Not sure if that's necessary?

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,97 @@
...
+.. note:: Compiled Bindings
+
+	The Realm SDK does not support `Compiled Bindings 
+	<https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/data-binding/compiled-bindings>`__.
+
+
+Data Binding Examples
+---------------------
+The following code snippets show both the C# and XAML for 3 typical data 
+binding use cases. They all use the following Realm objects:
```
dacharyc, line 36: Nit: compound modifier here should be hyphenated.
```suggestion
The following code snippets show both the C# and XAML for 3 typical data-binding 
use cases. They all use the following Realm objects:
```

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,97 @@
...
+
+.. note:: Compiled Bindings
+
+	The Realm SDK does not support `Compiled Bindings 
+	<https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/data-binding/compiled-bindings>`__.
+
+
+Data Binding Examples
+---------------------
+The following code snippets show both the C# and XAML for 3 typical data
```
dacharyc, line 35: Our [style guide](https://www.mongodb.com/docs/meta/style-guide/style/numbers/) indicates we should spell out numbers from zero through nine:
```suggestion
The following code snippets show both the C# and XAML for three typical data 
```

source/sdk/dotnet/fundamentals/data-binding.txt
dacharyc: Agreed, looks like the indentation below isn't at the right level, but I wonder if there's also an issue with the `()` in the note title?

source/sdk/dotnet/fundamentals/data-binding.txt
dacharyc: The docs [style guide](https://www.mongodb.com/docs/meta/style-guide/style/titles-and-headings/) indicates we should use title case in section headings, but I think we're inconsistent about doing this.

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,97 @@
...
+.. note:: ToList()
+
+   Because Realm collections are live, you should not call ``ToList()`` when 
+	obtaining a collection of Realm objects. Doing so forces loading the 
+	collection from memory, so the resulting collection is no longer a live 
+	object.
+
+Binding to a Single Realm Object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+In the following example, we create a public property of type Employee. Elsewhere
```
dacharyc, line 50: This is a bit inconsistent in how we use monospace. I might expect `Employee` to be monospace, and I would definitely expect `EmployeeId` to be monospace since you do monospace `Name` in ln 56. In the next section, you bold `Employee`, and then further down the page you use monospace for ``Employee`` in ln 81.

I don't have a strong opinion about what styling you should use where, and the style guide doesn't provide a lot of guidance about when to use monospace, but I'd say we should be consistent across the page. If we fall back to the Microsoft Manual of Style, I see that class names should be bold and code samples, including keywords and variables within text and as separate paragraphs, should be monospace.

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,97 @@
...
+
+In the associated XAML file, we bind a ``ListView`` to the **Employees** 
+``IEnumerable`` to show a list of all Employees. In this case, we're listing 
+each employee's **EmployeeId** value:
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.snippet.bind-list-xaml.xaml
+   :language: xml
+
+We can also bind to a *collection property* of a Realm object. In this example, 
+we bind to the ``Items`` property of a specific Employee``. The ``Items``
```
dacharyc, line 80:
```suggestion
we bind to the ``Items`` property of a specific ``Employee``. The ``Items`` 
```

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,97 @@
...
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.snippet.bind-one-xaml.xaml
+   :language: xml
+
+Binding to a Realm Collection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+There are two types of collections we can bind to. First, we can bind to a 
+collection of Realm objects. In this code example, we create a public 
+``IEnumerable`` of the **Employee** Realm object and then populate that 
+collection with all of the documents in the Employees collection:
```
dacharyc, line 67: From context, I believe I am reading about a Realm collection here, but then you refer to documents in ln 67, which makes me wonder if we have switched context to a MongoDB collection.

I think, for clarity, we should stick with the term objects when discussing Realm collections and documents when discussing accessing MongoDB collections.

source/sdk/dotnet/fundamentals/data-binding.txt
```diff
@@ -0,0 +1,97 @@
...
+We can also bind to a *collection property* of a Realm object. In this example, 
+we bind to the ``Items`` property of a specific Employee``. The ``Items`` 
+property on the ``Employee`` class is of type ``IList<Item>``.
+
+.. literalinclude:: /examples/generated/dotnet/MainPage.xaml.snippet.bind-embedded-cs.cs
+   :language: csharp
+
+.. note:: Using AsRealmQueryable()
+
+	Note that we are calling the :dotnet-sdk:`AsRealmQueryable()
```
dacharyc, line 88: I'm just noting here that the `()` in the note title isn't an issue on the staged page, so I think it's the indent in the first line of the note above at ln 43 that's causing the weird bold styling in the staged page.

dacharyc (approved):
Found some grammar nits and a couple of style things. Mod those, the page LGTM.

PR #2120, opened 2022-08-24
Convert Nosync and sync realms

MongoCaleb:
### Jira
https://jira.mongodb.org/browse/DOCSP-20724

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/nosync-to-sync/sdk/dotnet/advanced-guides/convert-realm/


examples/dotnet/Examples/Convert.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            // When copying a Synced realm, you must ensure
+            // that there are no pending Sync operations. You do this
+            // by calling WaitForUploadAsync() and WaitForDownloadAsync()
+            // methods:
+            var session = syncedRealm.SyncSession;
+            await session.WaitForUploadAsync();
+            await session.WaitForDownloadAsync();
+
+            var nonsyncConfig = new RealmConfiguration("my.realm");
```
papafe, line 55: better call it nonSyncConfig

source/sdk/dotnet/advanced-guides/convert-realm.txt
```diff
@@ -0,0 +1,66 @@
...
+permanently stop Sync for a synced realm. However, the .NET SDK does provide a 
+method that enables you to copy a realm file for use with a different configuration. 
+With this method, you can easily duplicate a realm's data, which you can then 
+open with a sync or non-sync configuration. This lets you indirectly add Sync 
+to a non-synced realm, or permanently stop a realm from syncing.
+
+
+.. note:: Flexible Sync Only
+
+   Converting between non-synced and synced realms only works with Flexible Sync
```
papafe, line 25: Actually it works also with Partition Sync. The only non working case is when converting a Partition based sync realm to another partition based sync realm with a different partition key

papafe (changes requested):

source/sdk/dotnet/advanced-guides/convert-realm.txt
dacharyc: I was surprised to see this! In the other SDKs, this seems to only work with Partition-Based Sync. Way to go .NET team :)

source/sdk/dotnet/advanced-guides/convert-realm.txt
```diff
@@ -0,0 +1,66 @@
...
+.. literalinclude:: /examples/generated/dotnet/Convert.snippet.nonsync-to-sync.cs
+   :language: csharp
+   :copyable: true
+
+
+Convert a Synced Realm to a Non-Synced Realm
+--------------------------------------------
+In the following code, we first open a synced realm. To be sure that the most 
+recent data is copied to the non-synced realm, wait for the data to be synced 
+in both directions. We do this by calling
```
dacharyc, line 57: I had to stop and think about the implications of this, which makes me wonder if we should add some more explicit details around what this entails.

i.e. if we're awaiting an async upload/download, presumably we should block new writes to the realm? Otherwise, there could be new pending operations by the time these awaitable async calls complete and we try to copy the realm. 

I also wondered if it's worth explicitly calling out that this requires the user to be online. I know it's implied by the waiting for network activity, but it sort of breaks our "Realm is offline-first" marketing message, so I wonder if it's just worth making a note about it.

dacharyc (approved):
LGTM - just a couple of Qs about whether we should be more explicit about some potential gotchas.

source/sdk/dotnet/advanced-guides/convert-realm.txt
MongoCaleb: Ultimately it shouldn't matter -- we're explicitly saying we not longer want shared, fresh data. So...the code shows getting the "latest", which might not include something being written concurrently from another device. I think that's fine, and have added a note to the end. LMK what you think!

papafe (changes requested):

source/sdk/dotnet/advanced-guides/convert-realm.txt
```diff
@@ -0,0 +1,69 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Realm does not have a direct mechanism to add sync to a non-synced realm, or to 
+permanently stop Sync for a synced realm. However, the .NET SDK does provide a 
+method that enables you to copy a realm file for use with a different configuration. 
+With this method, you can easily duplicate a realm's data, which you can then
```
dacharyc, line 18: Nit: "easily" sounds dangerously close to "simple." Maybe remove the subjective difficulty modifier and just say "you can duplicate a realm's data"?

dacharyc (approved):
One tiny nit, but LGTM 👍

PR #2142, opened 2022-08-31
(DOCSP-24966): Kotlin tutorial improvements after testing template

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-24966

### Changes in this PR:
- Deleted the legacy Android SDK (i.e Java SDK) tutorial using Partition Based Sync.
- Changed the wording for several sentence to be more specific as to where users should paste code.
- Fixed a typo where i used a nonexistence variable (called 'mSpinner', the actual variable is called 'spinner')
- Fixed a bug where I updated the name of the initial subscription to "User's High Priority Tasks" (the name of the initial subscription should be the same, otherwise it keeps the old subscription and will cause issues in your app that may even crash the app)
- Specified that users can use Logcat to find error logs
- Fixed the error message: `client provided query with bad syntax: unsupported query for table "Item": key "priority"...` to be a Kotlin SDK specific error
- Add a sentence telling users to import 2 required packages
- Set "rerunOnOpen" and "updateExisting" to true so that the previous query is overwritten and closure is reran on the next opening of the realm, subscribing to the new query.

- [Android with Kotlin](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/kotlin-tutorial-improvements-after-testing-template/tutorial/kotlin/)

### Review Guidelines


dacharyc (approved):
These changes look good from a copy perspective, but I didn't actually run through and test the tutorial.

PR #2156, opened 2022-09-07
(DOCSP-24421): Replace Swift quick starts with new Flexible Sync/single quick start

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-24421

- [Quick Start - Swift SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24421/sdk/swift/quick-start/)

### Review Guidelines


source/sdk/swift/quick-start.txt
```diff
@@ -112,11 +110,72 @@ Later, when done observing:
...
+you can do with App Services, see: :ref:`App Services - Swift SDK <ios-application-services>`.
+
+Prerequisites
+~~~~~~~~~~~~~
+
+Before you can sync Realm data, you must:
+
+- :ref:`Create an App Services App <create-a-realm-app>`
+- :ref:`Enable anonymous authentication <anonymous-authentication>`
+- :ref:`Enable Flexible Sync <enable-flexible-sync>` with  :ref:`Development Mode <enable-development-mode>` toggled to ``On`` and an ``ownerId`` field in the :guilabel:`Queryable Fields` section.
```
MongoCaleb, line 127: An "ownerId" is not a requirement. It can be any field on which you want to query. OwnerId is specific only to our template app/ tutorial.

source/sdk/swift/quick-start.txt
```diff
@@ -112,11 +110,72 @@ Later, when done observing:
...
+
+.. _ios-quick-start-init-app:
+
+Initialize the App
+~~~~~~~~~~~~~~~~~~
+
+To use App Services features such as authentication and sync, you must
+access your App Services App using your App ID. You can :ref:`find your
```
MongoCaleb, line 135: "you must access" could be softened to "you first access".

examples/ios/Examples/QuickStartFlexSync.swift
```diff
@@ -0,0 +1,163 @@
...
+// :snippet-end:
+
+ class QuickStartFlexSync: XCTestCase {
+    func testRunExample() async {
+        await flexibleSyncQuickStart()
+        // Entrypoint. Call this to run the quick start.
+        func flexibleSyncQuickStart() async {
+            // Instantiate the app
+            // :snippet-start: connect-to-backend
+            let app = App(id: FS_APP_ID) // Replace FS_APP_ID with your Atlas App ID
```
MongoCaleb, line 37: In our code examples elsewhere, I think we use "my_app_id" or similar. It's a bit more understandable than what you have here, IMHO

source/sdk/swift/quick-start.txt
```diff
@@ -112,11 +110,72 @@ Later, when done observing:
...
+
+.. _ios-quick-start-authenticate:
 
-Run the complete example by calling ``runLocalOnlyExample()``.
+Authenticate a User
+~~~~~~~~~~~~~~~~~~~
 
-.. literalinclude:: /examples/generated/code/start/LocalOnlyCompleteQuickStart.snippet.complete-quick-start-local.swift
+When you have enabled :ref:`anonymous authentication <anonymous-authentication>` in the
+App Services UI, users can immediately log into your app without providing any identifying
```
MongoCaleb, line 147: These two sentences are kind of disconected. I'm not sure the first one is needed at all, really. Can we just say "You now provide code that authenticates a user and opens a realm for that user. In the following code, we are using :ref:`anonymous authentication <anonymous-authentication>`."

source/sdk/swift/quick-start.txt
```diff
@@ -112,11 +110,72 @@ Later, when done observing:
...
+.. literalinclude:: /examples/generated/code/start/QuickStartFlexSync.snippet.authenticate-user.swift
    :language: swift
 
+The Realm Swift SDK provides many additional ways to authenticate, register, and link users.
+For other authentication providers, see: :ref:`ios-authenticate-users`
+
+Open a Realm
+~~~~~~~~~~~~
+
+Once you have enabled Device Sync and authenticated a user, you can create
```
MongoCaleb, line 159: The paragraph switches between "you can" and "we can".

source/sdk/swift/quick-start.txt
```diff
@@ -112,11 +110,72 @@ Later, when done observing:
...
+a :swift-sdk:`Configuration <Structs/Realm/Configuration.html>` object and 
+open the realm. We can then add a the :ref:`Flexible Sync
+subscription <swift-manage-flexible-sync-subscriptions>` that determines 
+what data the realm can read and write.
+
+Once we have a realm with a subscription, this example passes the realm
```
MongoCaleb, line 165: And then we go from "you can" to "we can" to "this example {verb}".

source/sdk/swift/quick-start.txt
```diff
@@ -112,11 +110,72 @@ Later, when done observing:
...
+.. literalinclude:: /examples/generated/code/start/QuickStartFlexSync.snippet.open-synced-realm.swift
+   :language: swift
+
+The syntax to :ref:`read <ios-read-operations>`, :ref:`write
+<ios-write-operations>`, and :ref:`watch for changes <ios-react-to-changes>` on a
+synced realm is identical to the syntax for non-synced realms above. While 
+you work with local data, a background thread efficiently integrates, 
+uploads, and downloads changesets. 
+
+To use Sync with this Quick Start model, add the ``ownerId`` property
```
MongoCaleb, line 177: This last sentence brings me back to my first comment; it feels disjointed to mention this property specifically.

source/sdk/swift/quick-start.txt
```diff
@@ -112,11 +110,72 @@ Later, when done observing:
...
    :language: swift
 
-Complete Example
-----------------
+Add Device Sync (Optional)
+--------------------------
+
+If you want to sync Realm data across devices, you can set up an Atlas 
+App Services App and enable Device Sync. For more information on what 
+you can do with App Services, see: :ref:`App Services - Swift SDK <ios-application-services>`.
```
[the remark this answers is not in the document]
dacharyc: Yeah, this is a known platform bug. Nate has filed tickets about this and they have been deemed low-priority.

source/sdk/swift/quick-start.txt
dacharyc: It's specific to this quick start, too, as that is the queryable field we're syncing on. This is consistent with the [Kotlin](https://www.mongodb.com/docs/realm/sdk/kotlin/quick-start/#prerequisites) and [Flutter](https://www.mongodb.com/docs/realm/sdk/flutter/quick-start/#prerequisites) Flexible Sync quick starts.

examples/ios/Examples/QuickStartFlexSync.swift
dacharyc: The Swift unit tests share a namespace across the entire unit test suite, and the more normal "my_app_id" is already used for the partition-based sync app ID. This alternate name isn't in danger of a namespace violation, and also helps me remember that this is the Flexible Sync App ID instead of the default PBS one.

If you think this is truly difficult to understand, I can use Bluehawk replace to change the name of the App ID in the generated example, but I tend to minimize my use of Bluehawk replace based on comments that Chris has left on some of my early PRs where I used it more extensively.

source/sdk/swift/quick-start.txt
dacharyc: This is consistent with the new FS quick starts in the other SDKs and the old Sync Quick Start. I think it's useful to give context around what anonymous authentication actually means if we're expecting this to be a first contact situation for folks who are new to Realm.

source/sdk/swift/quick-start.txt
```diff
@@ -37,9 +34,10 @@ Define Your Object Model
...
 For a local-only Realm Database you can define your :ref:`object model
-<ios-realm-objects>` directly in code.
+<ios-realm-objects>` directly in code. In this quick start, you can remove
+``ownerId`` unless you want to add the optional Device Sync.
 
-.. literalinclude:: /examples/generated/code/start/LocalOnlyCompleteQuickStart.snippet.quick-start-local-define-object-model.swift
+.. literalinclude:: /examples/generated/code/start/QuickStartFlexSync.snippet.model.swift
    :language: swift
 
 Open a Realm
```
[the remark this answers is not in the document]
dacharyc: The config parameter is linked after the code example as it provides additional context for options when opening a realm. I think it's "close enough" that the first sentence doesn't hang without an explanation.

source/examples/generated/code/start/LocalOnlyCompleteQuickStart.snippet.complete-quick-start-local.swift
```diff
@@ -15,9 +15,9 @@ class LocalOnlyQsTask: Object {
...
 func runLocalOnlyExample() {
     // Open the local-only default realm
-    let localRealm = try! Realm()
+    let realm = try! Realm()
     // Get all tasks in the realm
-    let tasks = localRealm.objects(LocalOnlyQsTask.self)
+    let tasks = realm.objects(LocalOnlyQsTask.self)
 
     // Retain notificationToken as long as you want to observe
     let notificationToken = tasks.observe { (changes) in
```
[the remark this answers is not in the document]
dacharyc: This is explained in the "CRUD" operations above, so if the person is following the entire quick start, they have already encountered an explanation of "todos":

> You can retrieve a live :ref:`collection <ios-client-collections>` of all todos in the realm:
>
>.. literalinclude:: /examples/generated/code/start/QuickStartFlexSync.snippet.get-all-todos.swift
>   :language: swift

source/sdk/swift/quick-start.txt
```diff
@@ -37,9 +34,10 @@ Define Your Object Model
 ------------------------
 
 For a local-only Realm Database you can define your :ref:`object model
-<ios-realm-objects>` directly in code.
+<ios-realm-objects>` directly in code. In this quick start, you can remove
+``ownerId`` unless you want to add the optional Device Sync.
```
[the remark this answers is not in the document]
dacharyc: I wanted some sort of explanation that the model doesn't need an ownerId unless you're using Flexible Sync. I don't like it being shoehorned in with no explanation at all, because down below we set the ownerId to the user.id and I don't want folks who aren't doing Sync to think a local realm has any concept of users or owners. My original draft of this page had a separate model for Sync that showed the ownerId, but I removed it because we were trying to get away from too much verbiage/explanation. 

I am happy to revise if you've got a suggestion. If it's just adding more info here around using `ownerId` as a Flexible Sync queryable field, I am happy to do that, but was trying to follow the model of "less is more" and not give too much Sync info up front for folks who might not care about using it.

source/sdk/swift/quick-start.txt
dacharyc: I think "first" is unnecessarily complicated. I have to think about whether this means "first I have to do this" or "I must use this the first time I access it". Must is simpler.

source/sdk/swift/quick-start.txt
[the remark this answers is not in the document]
MongoCaleb: As a reader, it's jarring. I stumbled on "wait, what configuration parameter?" It stopped me from reading and looking at the code example, let alone looking below it for more information.

source/sdk/swift/quick-start.txt
```diff
@@ -65,29 +63,29 @@ Create, Read, Update, and Delete Objects
 Once you have opened a realm, you can modify it and its :ref:`objects <ios-realm-objects>` 
 in a :ref:`write transaction <ios-write-transactions>` block.
 
-To create a new Task, instantiate the Task class and add it to the realm in a write block:
+To create a new Todo, instantiate the Todo class and add it to the realm in a write block:
```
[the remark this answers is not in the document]
MongoCaleb: Why are we keeping a "todo" as an object?

source/examples/generated/code/start/LocalOnlyCompleteQuickStart.snippet.complete-quick-start-local.swift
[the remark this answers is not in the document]
MongoCaleb: True, but it's in a different section of the doc. Clarity over assumption?

source/sdk/swift/quick-start.txt
MongoCaleb: ...or remove "must" altogether to improve readability further?

examples/ios/Examples/QuickStartFlexSync.swift
MongoCaleb: I can let this one go, for sure. Red flag: "helps me remember" is not necessarily in the reader's best interest.

source/sdk/swift/quick-start.txt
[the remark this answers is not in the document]
MongoCaleb: Why keep "Todo" throughout?

source/sdk/swift/quick-start.txt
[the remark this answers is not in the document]
dacharyc: I see it as shoehorned because we set its value to user.id down below for the Sync element. There is no user.id if you're using a local realm. I don't want people to infer that they should have access to a user.id if their app _only_ uses a local realm. If we present it in the local model and then use it with a user.id in the Sync element, I think it's reasonable to infer by transitive property that the same should work for a local realm - and that isn't true in our case.

source/sdk/swift/quick-start.txt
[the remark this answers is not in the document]
dacharyc: I think `Item` is a poor name for this class because it's so generic. Anything could be an item. A shoe is an item. A bullet list could have an item. I think `Task` is the best name for this example, but that has become a keyword for Swift's concurrency model, so Todo seems the closest because it's still descriptive and illustrative of the example we present here.

source/sdk/swift/quick-start.txt
dacharyc: I don't think it's possible to present a quick start that gives someone a working Sync example without giving them an example of how to set a queryable field and use that to Sync. This is the field we use in the subscription below, and when creating the todo object, so we have to set it when we enable Flexible Sync for this quick start.

source/sdk/swift/quick-start.txt
dacharyc: Updated the paragraph - let me know if this works better for you?

source/sdk/swift/quick-start.txt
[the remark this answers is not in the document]
dacharyc: Reworded this sentence - let me know if this works better for you.

source/examples/generated/code/start/LocalOnlyCompleteQuickStart.snippet.complete-quick-start-local.swift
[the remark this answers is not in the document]
dacharyc: This feels like an arbitrary place to add more information, since we had a lengthy discussion about making this quick start brief on explanatory text and there is _already_ an explanation of this on the page. Can you help me understand why you think this part of the example needs more explanation, but you wanted less explanation for things like setting up the back end and queryable fields? I am trying to get a grasp on why we want more explanation in some places and less in others.

examples/ios/Examples/QuickStartFlexSync.swift
dacharyc: I used Bluehawk replace to rename this and regenerate the example since you seemed to feel strongly about it. I think I still have questions about what degree of Bluehawk replace is "ok" to use in examples since Bush had commented on some of my earlier PRs not to "stretch" it for more extensive renaming. Seems this is benign enough in this instance?

MongoCaleb (approved):
Approving because this all really looks good. It's not inline with _my_ vision of the quick starts, but this shouldn't be held up by that. We'll have a broader discussion to find team-wide consensus. :-)

PR #2169, opened 2022-09-14
Remove prerelease info, add info on all projects, update images

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-24913

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/24913/sdk/dotnet/install/


source/sdk/dotnet/install.txt
```diff
@@ -41,6 +41,11 @@ Installation
 
 Follow these steps to add the Realm .NET SDK to your project.
 
+.. important:: Install the SDK for all projects
+
+   Most Xamarin projects have 3 projects: one for Android, one for iOS, and one
```
papafe, line 46: I would keep this more general because now we also have MAUI, and there could be more than just iOS and Android. 
Something like 
"In case you have a multi platform solution (for example with Xamarin/MAUI), please remember to add Realm also to the platform specific projects, even if there is no Realm specific code there."

papafe (changes requested):
Looks fine apart from one small note

dacharyc (approved):
LGTM

PR #2170, opened 2022-09-15
(DOCSP-20093): Delete user - Kotlin

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-20093

- [Delete Users - Kotlin SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-20093-delete-user-kotlin/sdk/kotlin/app-services/delete-users/)

### Review Guidelines

### DIFF
--- a/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/AuthenticationTest.kt
+++ b/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/AuthenticationTest.kt
@@ -33,6 +33,26 @@ class AuthenticationTest: RealmTest() {
     }
 
     @Test
+    fun deleteUserTest(){
+        val email = getRandom()
+        val password = getRandom()
+        // :snippet-start: user-delete
+        val app: App = App.create(YOUR_APP_ID) // Replace this with your App ID
+        runBlocking { // use runBlocking sparingly -- it can delay UI interactions
+            // :remove-start:
+            app.emailPasswordAuth.registerUser(email, password)
+            val credentials = Credentials.emailPassword(email, password)
+            // :remove-end:
+            val user = app.login(credentials)
+            // use the user object ...
+
+            // later, delete the user object
+            user.delete() // regardless of which provider you used to login, you can logout using `delete()`
+        }
+        // :snippet-end:
+    }
+
+    @Test
     fun confirmEmailPasswordUserTest() {
         val token = getRandom()
         val tokenId = getRandom()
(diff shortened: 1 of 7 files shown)


examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/AuthenticationTest.kt
```diff
@@ -32,6 +32,26 @@ class AuthenticationTest: RealmTest() {
...
+        runBlocking { // use runBlocking sparingly -- it can delay UI interactions
+            // :remove-start:
+            app.emailPasswordAuth.registerUser(email, password)
+            val credentials = Credentials.emailPassword(email, password)
+            // :remove-end:
+            val user = app.login(credentials)
+            // use the user object ...
+
+            // later, delete the user object
+            user.delete() // regardless of which provider you used to login, you can logout using `delete()`
```
dacharyc, line 50: Could we use a hidden Assert or the Kotlin equivalent to verify this is doing what we think it does? i.e. in the Swift SDK I've got:

```swift
let syncUser = try await app.login(credentials: Credentials.anonymous)
// Now we have a user, and the total users in the app = 1
XCTAssertNotNil(syncUser)
XCTAssertEqual(app.allUsers.count, 1)
// Call the `delete` method to delete the user
try await syncUser.delete()
// When you delete the user, the SyncSession is destroyed and
// there is no current user.
XCTAssertNil(app.currentUser)
// Now that we've deleted the user, the app has no users.
XCTAssertEqual(app.allUsers.count, 0)
```

We could debate whether the nil/count checks are redundant or whether we should show them in the code example, but it's nice to have a code example backed by a test that will pass/fail based on the app actually doing what we say it will do.

source/sdk/kotlin/app-services/delete-users.txt
```diff
@@ -0,0 +1,24 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can call the ``delete`` method on a user object to delete the user object from your App.
```
dacharyc, line 15: Since this is the first mention on the page, we need to use the full name here per marketing guidelines.

```suggestion
You can call the ``delete`` method on a user object to delete the user object from your Atlas App Services App. 
```

dacharyc (approved):
Overall, LGTM - I just have a question about whether we should make the test more of a functionality test than just a syntax test, and a product naming change.

PR #2192, opened 2022-10-04
(DOCSP-24223): Flutter public realm models with `$`

mongodben:
## Pull Request Info

Public realm models with `$` prefix on class names. 

note that not heavily documenting this functionality b/c the SDK team advocates defining all models in a single file. 

### Jira

- https://jira.mongodb.org/browse/DOCSP-24223

- [Define Realm Object Schema (Flutter)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24223/sdk/flutter/realm-database/define-realm-object-schema)

### Review Guidelines

### DIFF
--- a/source/sdk/flutter/realm-database/define-realm-object-schema.txt
+++ b/source/sdk/flutter/realm-database/define-realm-object-schema.txt
@@ -62,11 +62,19 @@ Create Model
 
       Create the model for your Realm schema.
       You must include the annotation `RealmModel <https://pub.dev/documentation/realm_common/latest/realm_common/RealmModel-class.html>`__
-      at the top of the class definition. Prepend the class name with
-      an underscore (``_``) to make it private.
+      at the top of the class definition.
 
-      You'll generate the public ``RealmObject`` used throughout the application
-      in step 4.
+      You'll use the ``RealmModel`` to generate the public ``RealmObject``
+      used throughout the application in step 4.
+
+      You can make the model private or public. We recommend making
+      the all models private and defining them in a single file.
+      Prepend the class name with an underscore (``_``) to make it private.
+
+      If you need to define your schema across multiple files,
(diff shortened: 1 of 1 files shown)


source/sdk/flutter/realm-database/define-realm-object-schema.txt
```diff
@@ -73,6 +76,11 @@ Create Model
          :language: dart
          :caption: schemas.dart
 
+      You can also make the ``RealmModel`` public by prepending the name with a
+      dollar sign (``$``). You must do  this to generate the ``RealmObject``
```
dacharyc, line 80: I found this confusing for a couple of reasons. 

A few lines above, we tell people to prepend the class name with an underscore to make it private. Now here we're saying you can make it public. The way the above is written, I assumed that's "the" way to do it. But now here is another way.

The other thing that was confusing is we're saying here "You must do this to generate the ``RealmObject`` from the ``RealmModel``." So it makes me think we _have_ to do this, even though above we told you to make it private.

I wonder if we can structure this info a little differently to make it less confusing. Maybe something like:
> You can make the model private or public. We recommend making the model private and defining it in a single file. Prepend the class name with an underscore to make it private.
>
> If you need to define your schema across multiple files, you can make the ``RealmModel`` public. Prepend the name with a dollar sign to make the model public.

Something like that, maybe?

dacharyc (commented):
I found this a bit confusing as written. What do you think about presenting it slightly differently?

source/sdk/flutter/realm-database/define-realm-object-schema.txt
mongodben: makes sense to me. refactored around your suggestions

dacharyc (approved):
LGTM

PR #2209, opened 2022-10-11
(DOCSP-24581): Flutter - encrypt a realm docs

mongodben:
## Pull Request Info

Documentation on encrypting a realm for Flutter SDK.

### Jira

- https://jira.mongodb.org/browse/DOCSP-24581

- [Encrypt a Realm (Flutter)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24581/sdk/flutter/realm-database/encrypt)

### Review Guidelines


source/sdk/flutter/realm-database/open-and-close-a-realm.txt
```diff
@@ -138,6 +138,12 @@ to see where realm files are stored in your environment.
 .. literalinclude:: /examples/generated/flutter/open_realm_test.snippet.default-storage-path.dart
    :language: dart
 
+Encrypt a Realm
+~~~~~~~~~~~~~~~
+
+You can encrypt your local realm to ensure data security. For more information,
+refer to the :ref:`Encrypt a Realm documentation <flutter-encrypt>`.
```
nlarew, line 145:
```suggestion
see :ref:`Encrypt a Realm <flutter-encrypt>`.
```

nit: snappier, up to you

nlarew (approved):

examples/dart/test/encrypt_test.dart
```diff
@@ -0,0 +1,23 @@
...
+
+void main() {
+  group("Encryption - ", () {
+    test("Open Realm with encryption key", () {
+      // :snippet-start: encrypt-realm
+      // Generate encryption key
+      List<int> key = List<int>.generate(64, (i) => Random().nextInt(256));
+
+      Configuration encryptedConfig = Configuration.local([Car.schema],
+          encryptionKey: key, path: 'encrypted.realm');
```
desistefanova, line 15: I would suggest using the default path in this example. We don't support relative paths yet. A full path to a writable directory is required to be passed. IT is better just to skip it.
```
Configuration encryptedConfig = Configuration.local([Car.schema], encryptionKey: key);
```

source/examples/generated/flutter/encrypt_test.snippet.encrypt-realm.dart
```diff
@@ -0,0 +1,6 @@
+// Generate encryption key
+List<int> key = List<int>.generate(64, (i) => Random().nextInt(256));
+
+Configuration encryptedConfig = Configuration.local([Car.schema],
+    encryptionKey: key, path: 'encrypted.realm');
```
desistefanova, line 5: The same here. Better to use the default path.

desistefanova (commented):
Looks OK to me. I've just added a suggestion.

examples/dart/test/encrypt_test.dart
blagoev: this is not a relative path. This will rather use the current dir when creating that file. It is fully supported. 
But I agree if there is no technical reason to do this then perhaps we can just leave it out of the example. I suspect @mongodben needs this for the this test to run correctly.

examples/dart/test/encrypt_test.dart
```diff
@@ -0,0 +1,23 @@
...
+import './utils.dart';
+
+void main() {
+  group("Encryption - ", () {
+    test("Open Realm with encryption key", () {
+      // :snippet-start: encrypt-realm
+      // Generate encryption key
+      List<int> key = List<int>.generate(64, (i) => Random().nextInt(256));
+
+      Configuration encryptedConfig = Configuration.local([Car.schema],
```
blagoev, line 14: Since we have a comment on generating the key I think we better have a comment here as well explaining that we are passing it in the `Configuration`. It otherwise could be missed when reading the code.

examples/dart/test/encrypt_test.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'package:realm_dart/realm.dart';
+import 'dart:math';
+import './utils.dart';
+
+void main() {
+  group("Encryption - ", () {
+    test("Open Realm with encryption key", () {
+      // :snippet-start: encrypt-realm
+      // Generate encryption key
+      List<int> key = List<int>.generate(64, (i) => Random().nextInt(256));
```
blagoev, line 12:
```suggestion
      final key = List<int>.generate(64, (i) => Random().nextInt(256));
```

examples/dart/test/encrypt_test.dart
```diff
@@ -0,0 +1,23 @@
...
+
+      Configuration encryptedConfig = Configuration.local([Car.schema],
+          encryptionKey: key, path: 'encrypted.realm');
+      Realm encryptedRealm = Realm(encryptedConfig);
```
blagoev, line 16:
```suggestion
      final encryptedRealm = Realm(encryptedConfig);
```

examples/dart/test/encrypt_test.dart
blagoev: we can skip the type declaration as it is inferred from the assignment. I see we do this in many places I suggest we fix these since it makes the code shorter and this is how most Dart devs will write it ie it is common even suggested practice.

examples/dart/test/encrypt_test.dart
blagoev: actually path is required here. cause most probably the default realm is not encrypted.

examples/dart/test/encrypt_test.dart
```diff
@@ -0,0 +1,23 @@
...
+      // :snippet-end:
+      expect(encryptedRealm.isClosed, isFalse);
+      // Note: no way to test if Realm is encrypted per SDK PR - https://github.com/realm/realm-dart/blob/7b16c62fae4baf8b257531db6f1ed36093fa7309/test/configuration_test.dart#L550-L553
```
blagoev, line 19: There is a way and we have missed this test when merging the PR. 
I did a PR for that here is the test in question https://github.com/realm/realm-dart/pull/962/files#diff-bf48d02e4defff0f358c4c408815da36fa24d0d2a66ecb3cc0332d932e89dae9R793

source/sdk/flutter/realm-database/encrypt.txt
```diff
@@ -0,0 +1,62 @@
...
+
+You can encrypt a :ref:`synced realm <flutter-sync>`, too. Realm only
+encrypts the data on the device, and stores the data unencrypted in your
+Atlas data source. The transfer between client server is fully
+encrypted.
+
+If you need unique keys for each user of your application, you can use an OAuth
+provider, or use one of the :ref:`Realm Authentication providers <users-and-authentication>`
+and an :ref:`Authentication Trigger<authentication-triggers>`
+to create a 64-bit key and store that key in :ref:`a user object <user-objects>`.
```
blagoev, line 62: Should we give some sample here. Seems like very common use case. At least how to store that in the user object

examples/dart/test/encrypt_test.dart
blagoev: Ok, a correction from me.  Setting path to the filename will work in Dart. It will not work in Flutter apps. I don't think @mongodben is testing with a Flutter app, so it will work as is currently.

examples/dart/test/encrypt_test.dart
mongodben: i keep the types in the code examples as a form of in-line documentation, even though it's not normal dart best practice. 

i think it helps clarify what's going on with the code examples

source/sdk/flutter/realm-database/encrypt.txt
mongodben: that'd be a pretty complex example to include here. could make sense as a blog post or some other document that's linked to here. but i think it's outside the scope of this documentation beyond that.

examples/dart/test/encrypt_test.dart
mongodben: will update accordingly

examples/dart/test/encrypt_test.dart
mongodben: i deleted the `path: "encrypted.realm"` and it worked fine.

examples/dart/test/encrypt_test.dart
mongodben: updated to use a test that throws when opening same realm w/o encryption key

PR #2226, opened 2022-10-20
(DOCSP-16478) [RJS] Support Hermes JS Engine

nlarew:
### Jira

- (DOCSP-16478) [RJS] Support Hermes JS Engine

- [React Native > Install](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/hermes/sdk/react-native/install/)

### Review Guidelines

### DIFF
--- a/source/sdk/react-native/install.txt
+++ b/source/sdk/react-native/install.txt
@@ -77,10 +77,8 @@ and add the Realm React Native SDK to it.
             .. code-block:: bash
 
                cd MyRealmApp
-
-
-         .. step:: Install Realm with NPM
-
+         
+         .. step:: Install Realm with npm
 
             In your React Native project directory, add Realm to your
             project with the following command:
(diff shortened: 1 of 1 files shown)


source/sdk/react-native/install.txt
```diff
@@ -77,22 +77,33 @@ and add the Realm React Native SDK to it.
...
             In your React Native project directory, add Realm to your
             project with the following command:
 
             .. code-block:: bash
 
                npm install realm
-
-
+         
+         .. step:: Enable Hermes
```
mongomoe, line 90: If this is optional/only required for V11+ I would label this as optional. Especially since I imagine many of our users are on older versions of React Native and also not immediately going to adopt version 11 of realm-js

source/sdk/react-native/install.txt
```diff
@@ -77,22 +77,33 @@ and add the Realm React Native SDK to it.
...
+
+            Realm supports React Native's mobile-optimized JavaScript
+            engine, Hermes. By default, new apps created with the React
+            Native CLI already have Hermes enabled.
+            
+            We recommend that you use Hermes with Realm. However, Realm
+            also supports the JSC JavaScript engine if your app requires
```
mongomoe, line 97: Suggestion: Since this is the first use of JSC, let's spell out the word "JavaScriptCore" and link out to a definition/docs for it.

source/sdk/react-native/install.txt
cbullinger: And versioning aside, it seems like an optional step since it reads as though it only applies for use cases where you're also using JSC, which (I assume) isn't the default.

kneth: We should mention it requires RN 0.70.0 or later

kneth (approved):

PR #2236, opened 2022-10-24
(DOCSP-24576): Flutter open realm async

mongodben:
## Pull Request Info

Open a Realm asynchronously for Flutter SDK.

### Jira

- https://jira.mongodb.org/browse/DOCSP-24576

- [Open a Synced Realm (Flutter)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24576/sdk/flutter/sync/open-synced-realm/#sync-data-on-open)

### Review Guidelines

### DIFF
--- a/examples/dart/test/open_flexible_sync_realm_test.dart
+++ b/examples/dart/test/open_flexible_sync_realm_test.dart
@@ -89,9 +89,20 @@ void main() {
       // Cancel the open operation after 30 seconds.
       // Alternatively, you could display a loading dialog and bind the cancellation
       // to a button the user can click to stop the wait.
-      Future<void>.delayed(const Duration(seconds: 30), () => token.cancel());
+      Future<void>.delayed(
+          const Duration(seconds: 30),
+          () => token.cancel(CancelledException(
+              cancellationReason: "Realm took too long to open")));
 
-      final realm = await Realm.open(config, cancellationToken: token);
+      // If realm does not open after 30 seconds with asynchronous Realm.open(),
+      // open realm synchronously with Realm().
+      late Realm realm;
+      try {
+        realm = await Realm.open(config, cancellationToken: token);
+      } on CancelledException catch (err) {
+        print(err.cancellationReason); // prints "Realm took too long to open"
+        realm = Realm(config);
+      }
       // :snippet-end:
       expect(token.isCancelled, false);
       cleanUpRealm(realm, app);
--- a/examples/dart/test/open_flexible_sync_realm_test.dart
+++ b/examples/dart/test/open_flexible_sync_realm_test.dart
@@ -2,7 +2,6 @@ import 'dart:async';
 import 'package:test/test.dart';
 import 'package:realm_dart/realm.dart';
 import './utils.dart';
-
 part 'open_flexible_sync_realm_test.g.dart';
 
 @RealmModel()
@@ -17,16 +16,15 @@ class _Tricycle {
 void main() {
   group('Open Flexible Sync Realm', () {
     const APP_ID = "flex-config-tester-vwevn";
-    AppConfiguration appConfig = AppConfiguration(APP_ID);
-    App app = App(appConfig);
+    final appConfig = AppConfiguration(APP_ID);
+    final app = App(appConfig);
     test("Open Flexible Sync Realm", () async {
-      Credentials credentials = Credentials.anonymous();
+      final credentials = Credentials.anonymous();
       // :snippet-start: open-flexible-sync-realm
-      User currentUser = await app.logIn(credentials);
-      Configuration config = Configuration.flexibleSync(
-          currentUser, [Tricycle.schema],
+      final currentUser = await app.logIn(credentials);
+      final config = Configuration.flexibleSync(currentUser, [Tricycle.schema],
           path: 'flex.realm');
-      Realm realm = Realm(config);
+      final realm = Realm(config);
       // :snippet-end:
       expect(realm.isClosed, false);
       expect(app.currentUser?.id != null, true);
@@ -38,22 +36,32 @@ void main() {
       Credentials credentials = Credentials.anonymous();
       User currentUser = await app.logIn(credentials);
       // :snippet-start: async-open
-      Configuration config =
-          Configuration.flexibleSync(currentUser, [Tricycle.schema]);
-      Realm fullySyncedRealm = await Realm.open(config);
+      // Helper function to check if device is connected to the internet.
+      Future<bool> isDeviceOnline() async {
+        // ...logic to check if device is online
+        return true; // :remove:
+      }
+
+      final config = Configuration.flexibleSync(currentUser, [Tricycle.schema]);
+      // Only use asynchronous open if app is online.
+      late Realm realm;
+      if (await isDeviceOnline()) {
+        realm = await Realm.open(config);
+      } else {
+        realm = Realm(config);
+      }
       // :snippet-end:
-      expect(fullySyncedRealm.isClosed, false);
-      cleanUpRealm(fullySyncedRealm, app);
+      expect(realm.isClosed, false);
+      cleanUpRealm(realm, app);
     });
     test('Track download progress', () async {
       Credentials credentials = Credentials.anonymous();
       User currentUser = await app.logIn(credentials);
       late int transferred;
       late int transferable;
+      final config = Configuration.flexibleSync(currentUser, [Tricycle.schema]);
       // :snippet-start: async-open-track-progress
-      Configuration config =
-          Configuration.flexibleSync(currentUser, [Tricycle.schema]);
-      Realm fullySyncedRealm =
+      final realm =
           await Realm.open(config, onProgressCallback: (syncProgress) {
         if (syncProgress.transferableBytes == syncProgress.transferredBytes) {
           print('All bytes transferred!');
@@ -64,36 +72,35 @@ void main() {
         }
       });
       // :snippet-end:
-      expect(fullySyncedRealm.isClosed, false);
+      expect(realm.isClosed, false);
       expect(transferred, transferable);
       expect(transferred, greaterThanOrEqualTo(0));
-      cleanUpRealm(fullySyncedRealm, app);
+      cleanUpRealm(realm, app);
     });
     test('Cancel download in progress', () async {
-      Credentials credentials = Credentials.anonymous();
-      User currentUser = await app.logIn(credentials);
+      final credentials = Credentials.anonymous();
+      final currentUser = await app.logIn(credentials);
       late int transferred;
       late int transferable;
+      final config = Configuration.flexibleSync(currentUser, [Tricycle.schema]);
       // :snippet-start: async-open-cancel
-      Configuration config =
-          Configuration.flexibleSync(currentUser, [Tricycle.schema]);
+      final token = CancellationToken();
 
-      CancellationToken token = CancellationToken();
-      // Cancel operation after 5 seconds.
-      Future<void>.delayed(const Duration(seconds: 5), () => token.cancel());
+      // Cancel the open operation after 30 seconds.
+      // Alternatively, you could display a loading dialog and bind the cancellation
+      // to a button the user can click to stop the wait.
+      Future<void>.delayed(const Duration(seconds: 30), () => token.cancel());
 
-      Realm fullySyncedRealm =
-          await Realm.open(config, cancellationToken: token);
+      final realm = await Realm.open(config, cancellationToken: token);
       // :snippet-end:
       expect(token.isCancelled, false);
-      expect(fullySyncedRealm.isClosed, false);
-      cleanUpRealm(fullySyncedRealm, app);
+      cleanUpRealm(realm, app);
     });
 
     test("Handle Sync Error", () async {
       var handlerCalled = false;
-      Credentials credentials = Credentials.anonymous();
-      User currentUser = await app.logIn(credentials);
+      final credentials = Credentials.anonymous();
+      final currentUser = await app.logIn(credentials);
       // :snippet-start: sync-error-handler
       Configuration config = Configuration.flexibleSync(
           currentUser, [Tricycle.schema], syncErrorHandler: (SyncError error) {
--- a/examples/dart/test/open_flexible_sync_realm_test.dart
+++ b/examples/dart/test/open_flexible_sync_realm_test.dart
@@ -50,7 +50,7 @@ void main() {
         realm = await Realm.open(config);
       } else {
         // If the device is offline, open the realm immediately
-        // and sync changes in the background.
+        // and automatically sync changes in the background when the device is online.
         realm = Realm(config);
       }
       // :snippet-end:
@@ -98,7 +98,7 @@ void main() {
               cancellationReason: "Realm took too long to open")));
 
       // If realm does not open after 30 seconds with asynchronous Realm.open(),
-      // open realm synchronously with Realm().
+      // open realm immediately with Realm() and try to sync data in the background.
       late Realm realm;
       try {
         realm = await Realm.open(config, cancellationToken: token);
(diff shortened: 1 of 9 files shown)


examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -35,6 +34,61 @@ void main() {
...
+      Credentials credentials = Credentials.anonymous();
+      User currentUser = await app.logIn(credentials);
+      late int transferred;
+      late int transferable;
+      // :snippet-start: async-open-cancel
+      Configuration config =
+          Configuration.flexibleSync(currentUser, [Tricycle.schema]);
+
+      CancellationToken token = CancellationToken();
+      // Cancel operation after 5 seconds.
```
desistefanova, line 82: It is too late after 5 seconds. The realm is already downloaded, since it is empty. If you want to show how cancel works, you have to cancel before to obtain the realm. Then the realm should be closed.

source/sdk/flutter/sync/open-synced-realm.txt
```diff
@@ -33,14 +33,42 @@ Realm object schemas, and additional optional named arguments to the
...
+to the optional named argument ``onProgressCallback``.
+
+.. literalinclude:: /examples/generated/flutter/open_flexible_sync_realm_test.snippet.async-open-track-progress.dart
+   :language: dart
+
+To be able to cancel a synchronization in progress, pass a `CancellationToken
+<https://pub.dev/documentation/cancellation_token/latest/cancellation_token/CancellationToken-class.html>`__
+instance to the optional named argument ``cancellationToken``.
+Call `CancellationToken.cancel() <https://pub.dev/documentation/cancellation_token/latest/cancellation_token/CancellationToken/cancel.html>`__
+to cancel the synchronization.
```
desistefanova, line 67: In addition, it is also possible to cancel more than one opening realm async operations with the came cancellation token.

mongodben: i don't think that needs inclusion in the docs. this wouldn't be a very common use case, would it? 

as long as we're not saying this is forbidden, i think it's ok as is. thoughts?

examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -35,6 +34,61 @@ void main() {
       expect(realm.isClosed, true);
       expect(app.currentUser, null);
     });
+    test('Async Open Flexible Sync Realm', () async {
+      Credentials credentials = Credentials.anonymous();
+      User currentUser = await app.logIn(credentials);
+      // :snippet-start: async-open
+      Configuration config =
+          Configuration.flexibleSync(currentUser, [Tricycle.schema]);
+      Realm fullySyncedRealm = await Realm.open(config);
```
nirinchev, line 43: I don't know what the common practice is for the docs, but I personally prefer to omit the types as it makes the code easier to read - e.g. here we have "Configuration config equals Configuration dot ...", "Realm realm equals await realm dot..." - that's a lot of repetition. I find the following a little lighter:

```suggestion
      final config = Configuration.flexibleSync(currentUser, [Tricycle.schema]);
      final realm = await Realm.open(config);
```

I also suggest renaming `fullySyncedRealm` as those snippets should be as copy-pastable as possible - I know we want to be explicit about what's going on there, but I'd suggest using the text around the snippet or possibly adding a comment rather than naming variables descriptive names.

examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -35,6 +34,61 @@ void main() {
...
+      User currentUser = await app.logIn(credentials);
+      late int transferred;
+      late int transferable;
+      // :snippet-start: async-open-cancel
+      Configuration config =
+          Configuration.flexibleSync(currentUser, [Tricycle.schema]);
+
+      CancellationToken token = CancellationToken();
+      // Cancel operation after 5 seconds.
+      Future<void>.delayed(const Duration(seconds: 5), () => token.cancel());
```
nirinchev, line 83: [nit] While using `const` is probably a good practice here, it increases the verbosity of the example without improving users' understanding of the concept being explained (namely cancellation).

source/sdk/flutter/sync/open-synced-realm.txt
nirinchev: Agree - I think docs should avoid getting too much into technicalities and cover the most common use cases.

examples/dart/test/open_flexible_sync_realm_test.dart
nirinchev: Here I think having a comment in the snippet suggesting an alternative use case could be helpful. Something like:

```dart
// Cancel the open operation after x seconds. Alternatively, display a loading
// dialog and bind the cancelation to a button the user can click to stop the wait
Future<void>.delayed(...);
```

nirinchev (commented):
I have a few minor comments. One thing that seems to be missing and I think is valuable to be explicit about is that async open only works when online and will not complete - either with an error or successfully - if the device is offline. So it's recommended to either build a timeout after which the app falls back to sync open, or have a check for connectivity like:

```dart
late realm;
if (ConnectivityService.isDeviceOnline) {
  realm = await Realm.open(...);
} else {
  realm = Realm(...);
}
```

source/sdk/flutter/sync/open-synced-realm.txt
desistefanova: Ok. I've just mentioned it.

examples/dart/test/open_flexible_sync_realm_test.dart
desistefanova: The issue for me is that we demonstrate here that the Realm is opened even after we canceled. It could be confusing. We can at least delete the check that the Realm is opened.

nirinchev: The rest of the test doesn't make it into the docs, so I think we can do whatever we want there, but you're right - we can remove all the expects.

mongodben: > 

this i did not know! will add to the docs.

examples/dart/test/open_flexible_sync_realm_test.dart
mongodben: > I don't know what the common practice is for the docs, but I personally prefer to omit the types as it makes the code easier to read

lubo commented the same a few weeks ago. for the sake of consistency, i'm going to keep as is. but made ticket to update this everywhere across the flutter docs https://jira.mongodb.org/browse/DOCSP-26111

mongodben: > I also suggest renaming fullySyncedRealm as those snippets should be as copy-pastable as possible

and for this, sure can change

examples/dart/test/open_flexible_sync_realm_test.dart
mongodben: > The issue for me is that we demonstrate here that the Realm is opened even after we canceled. It could be confusing. We can at least delete the check that the Realm is opened.

deleting the expect statement `expect(fullySyncedRealm.isClosed, false);`

mongodben:
@nirinchev and @desistefanova responded to and implemented your feedback. could you take another look?

examples/dart/test/open_flexible_sync_realm_test.dart
desistefanova: Ok

source/sdk/flutter/sync/open-synced-realm.txt
```diff
@@ -33,14 +33,46 @@ Realm object schemas, and additional optional named arguments to the
...
 
 .. literalinclude:: /examples/generated/flutter/open_flexible_sync_realm_test.snippet.open-flexible-sync-realm.dart
    :language: dart
 
 Once you open a synced realm, :ref:`configure and manage the sync session
 <flutter-manage-sync-session>`.
 
+.. _flutter-open-realm-async:
+
+Sync Data on Open
```
dacharyc, line 47: I wonder if we might want to rename this section to something that makes it clearer that you're opening the realm _after_ downloading changes. I wouldn't necessarily know from this heading that this will block opening the realm. Maybe something like "Open a Realm After Downloading Changes" or similar?

source/sdk/flutter/sync/open-synced-realm.txt
```diff
@@ -33,14 +33,46 @@ Realm object schemas, and additional optional named arguments to the
...
+~~~~~~~~~~~~~~~~~
+
+To sync all data with App Services when you open a realm, use the asynchronous method
+:flutter-sdk:`Realm.open() <realm/Realm/open.html>`. The operation syncs all available
+data before returning the ``Realm``. ``Realm.open()`` only resolves if the app
```
dacharyc, line 52: It's very similar to what you already have here, but when the Swift SDK had a similar method, some of our docs readers found it clearer to say something like "If the user is offline, Realm.open() does not resolve" vs. something like "Realm.open() only resolves if the app is online." Make it explicit that this method does not work when offline. This also speaks more directly to our positioning Realm w/Sync as an "offline-first DB."

dacharyc (commented):
Overall, it looks good - I've just got a couple of big-picture questions based on many iterations of a similar page for the Swift SDK. A couple of other things we might want to consider adding:

- Information about first open/subsequent opens. Maybe something like the first open means downloading the entire realm, while subsequent opens only involve downloading the latest changes, so performance on first open may be slower while subsequent opens may be relatively fast, depending on initial realm size & data set churn.
- A use case about why you might want to use this. i.e. an example of when you might want to block opening the realm until after data has downloaded - I went with a game - vs. when you might want to make the app available immediately and sync data in the background.

examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -18,28 +16,102 @@ class _Tricycle {
...
+        return true; // :remove:
+      }
 
-    test("Handle Sync Error", () async {
-      var handlerCalled = false;
+      final config = Configuration.flexibleSync(currentUser, [Tricycle.schema]);
+      // Only use asynchronous open if app is online.
+      late Realm realm;
+      if (await isDeviceOnline()) {
+        realm = await Realm.open(config);
```
dacharyc, line 49: This is super subjective, but I wonder if we might want to put a comment in here and above ln 51 explicitly stating what we're doing? If people copy-paste this without really reading the docs & understanding what these APIs do, they may not realize exactly what this logic does. i.e.:

```suggestion
        // If the device is online, download changes and then open the realm
        realm = await Realm.open(config);
```

And then down below, something like:

```
} else {
   // If the device is offline, open a local realm
   realm = Realm(config);
```

source/sdk/flutter/sync/open-synced-realm.txt
```diff
@@ -33,14 +33,71 @@ Realm object schemas, and additional optional named arguments to the
...
+
+.. literalinclude:: /examples/generated/flutter/open_flexible_sync_realm_test.snippet.async-open-cancel.dart
+   :language: dart
+
+.. example:: Realm() vs. Realm.open()
+
+   This section compares examples of scenarios when you might want to use ``Realm()``
+   versus ``Realm.open()`` to open a realm in an application.
+
+   Say you have an app that allowed the user to record and save their
```
dacharyc, line 87: Nit: "Say" is idiomatic here, and may not be the best phrasing for people outside the US or people for whom English is a second language. We might instead want to say something like "Consider an app that allows" or "For example, your app might allow the user..."

Same for the paragraph down below.

dacharyc (approved):
Overall, LGTM. I did have one subjective suggestion that you may or may not want to take, and a nit that I hope you'll tweak.

Nice work!

source/sdk/flutter/sync/open-synced-realm.txt
mongodben: absolutely.

btw this phrasing was taken from the legacy swift page https://www.mongodb.com/docs/realm/sdk/swift/sync/sync-realm-open-legacy/#open-a-synced-realm-after-downloading-changes

maybe not worth making changes to legacy page, but wanted to put on your radar.

dacharyc: Hah, what was I thinking? 😉 

I am trying very hard not to touch that page ever again, so I'll leave it for now, but thanks for the heads-up!

examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -18,28 +16,105 @@ class _Tricycle {
...
+      // Cancel the open operation after 30 seconds.
+      // Alternatively, you could display a loading dialog and bind the cancellation
+      // to a button the user can click to stop the wait.
+      Future<void>.delayed(
+          const Duration(seconds: 30),
+          () => token.cancel(CancelledException(
+              cancellationReason: "Realm took too long to open")));
+
+      // If realm does not open after 30 seconds with asynchronous Realm.open(),
+      // open realm synchronously with Realm().
```
desistefanova, line 101:
```suggestion
      // open realm locally with Realm().
```

examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -18,28 +16,105 @@ class _Tricycle {
...
-    test("Handle Sync Error", () async {
-      var handlerCalled = false;
+      final config = Configuration.flexibleSync(currentUser, [Tricycle.schema]);
+      // Only use asynchronous open if app is online.
+      late Realm realm;
+      if (await isDeviceOnline()) {
+        // If the device is online, download changes and then open the realm.
+        realm = await Realm.open(config);
+      } else {
+        // If the device is offline, open the realm immediately
```
desistefanova, line 52:
```suggestion
        // If the device is offline, open the realm from your local device.
```

examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -18,28 +16,105 @@ class _Tricycle {
...
+        // and sync changes in the background.
```
desistefanova, line 53:
```suggestion
        // The date will be synced automatically later when the device goes online
```

source/sdk/flutter/sync/open-synced-realm.txt
```diff
@@ -33,14 +33,72 @@ Realm object schemas, and additional optional named arguments to the
...
+:flutter-sdk:`Realm.open() <realm/Realm/open.html>`. The operation syncs all available
+data before returning the realm.
+
+On first open, ``Realm.open()`` downloads all data that matches your sync subscriptions.
+Subsequent opens only download the latest changes.
+Depending on initial realm size and updates to the data set while the device is not syncing,
+performance may be slower on first open and faster on subsequent opens.
+
+If the client application is offline, ``Realm.open()`` does not resolve.
+You should check if the device is connected to the internet before using ``Realm.open()``.
```
desistefanova, line 60:
```suggestion
You could check if the device is connected to the internet before using ``Realm.open()`` or just to cancel the operation ``Realm.open()`` after some timeout.
```

source/sdk/flutter/sync/open-synced-realm.txt
```diff
@@ -33,14 +33,72 @@ Realm object schemas, and additional optional named arguments to the
...
+If the device is not connected to the internet, you can still use ``Realm()``
```
desistefanova, line 61:
```suggestion
If the device is not connected to the internet, you can always open your local realm using ``Realm()``
```

source/sdk/flutter/sync/open-synced-realm.txt
```diff
@@ -33,14 +33,72 @@ Realm object schemas, and additional optional named arguments to the
...
+to open a realm and sync data in the background when an internet connection is available.
```
desistefanova, line 62:
```suggestion
The data will be automatically synced in background once the  internet connection is available.
```

desistefanova (commented):
I added some suggestions trying to explain to the users what opening a realm synchronously means. It is really easy to misunderstand the difference between sync and synchronously. It is good to say that Realm() just opens the local file and they can continue to work without Atlas cloud. And once the connection is up, they really don't have to do anything. It will sync automatically. Maybe there is a better way to explain it. I think that the word "synchronously" is the issue here.

PR #2237, opened 2022-10-24
(DOCSP-24587): Add writeAsync example to Flutter docs

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-24587

- [Read & Write Data](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24587/sdk/flutter/realm-database/read-and-write-data/#background-writes): New "Background Writes" section w/writeAsync example

### Review Guidelines

### DIFF
--- a/examples/dart/test/read_write_data_test.dart
+++ b/examples/dart/test/read_write_data_test.dart
@@ -83,4 +83,19 @@ void main() {
     expect(prius!.miles, 500);
     cleanUpRealm(realm);
   });
+
+  test('Write async', () async {
+    final config = Configuration.local([Car.schema]);
+    final realm = Realm(config);
+    // :snippet-start: write-async
+    // Add Subaru Outback to the realm using `writeAsync`
+    Car newOutback = Car("Subaru", model: "Outback Touring XT", miles: 2);
+    await realm.writeAsync(() {
+      realm.add<Car>(newOutback);
+    });
+    // :snippet-end:
+    final outback = realm.find<Car>("Subaru");
+    expect(outback!.miles, 2);
+    cleanUpRealm(realm);
+  });
 }
--- a/examples/dart/test/read_write_data_test.dart
+++ b/examples/dart/test/read_write_data_test.dart
@@ -90,12 +90,17 @@ void main() {
     // :snippet-start: write-async
     // Add Subaru Outback to the realm using `writeAsync`
     Car newOutback = Car("Subaru", model: "Outback Touring XT", miles: 2);
-    await realm.writeAsync(() {
+   realm.writeAsync(() {
       realm.add<Car>(newOutback);
     });
     // :snippet-end:
     final outback = realm.find<Car>("Subaru");
-    expect(outback!.miles, 2);
+    expect(outback, isNull);
+    expect(realm.isInTransaction, true);
+    // let transaction resolve
+    await Future.delayed(Duration(milliseconds: 500));
+    expect(realm.isInTransaction, false);
+    expect(realm.find<Car>("Subaru")!.miles, 2);
     cleanUpRealm(realm);
   });
 }
--- a/examples/dart/test/read_write_data_test.dart
+++ b/examples/dart/test/read_write_data_test.dart
@@ -4,6 +4,7 @@ import 'package:test/test.dart';
 import 'package:realm_dart/realm.dart';
 import '../bin/models/car.dart';
 import 'utils.dart';
+import 'dart:io';
 
 part 'read_write_data_test.g.dart';
 
@@ -90,7 +91,7 @@ void main() {
     // :snippet-start: write-async
     // Add Subaru Outback to the realm using `writeAsync`
     Car newOutback = Car("Subaru", model: "Outback Touring XT", miles: 2);
-   realm.writeAsync(() {
+    realm.writeAsync(() {
       realm.add<Car>(newOutback);
     });
     // :snippet-end:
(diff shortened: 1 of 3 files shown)


examples/dart/test/read_write_data_test.dart
```diff
@@ -83,4 +83,19 @@ void main() {
...
     cleanUpRealm(realm);
   });
+
+  test('Write async', () async {
+    final config = Configuration.local([Car.schema]);
+    final realm = Realm(config);
+    // :snippet-start: write-async
+    // Add Subaru Outback to the realm using `writeAsync`
+    Car newOutback = Car("Subaru", model: "Outback Touring XT", miles: 2);
+    await realm.writeAsync(() {
```
mongodben, line 93:
```suggestion
   realm.writeAsync(() {
```

you actually don't want to use `await` here b/c that makes the operation thread blocking.

source/sdk/flutter/realm-database/read-and-write-data.txt
```diff
@@ -162,3 +162,15 @@ Delete multiple objects with the :flutter-sdk:`Realm.deleteMany()
...
    :language: dart
+
+.. _flutter-write-async:
+
+Background Writes
+-----------------
+
+You can add, modify, or delete objects in the background using 
+:flutter-sdk:`Realm.writeAsync() <realm/Realm/writeAsync.html>`. With 
+``writeAsync``, you don't need to pass frozen objects across threads.
```
mongodben, line 173: i think we need to rethink the sentence "With ``writeAsync``, you don't need to pass frozen objects across threads."

a couple of considerations:

1. first, we don't talk about threading in the Flutter docs. Dart actually doesn't technically have multiple threads. concurrency is achieved with [isolates](https://dart.dev/guides/language/concurrency). and, we don't talk about passing frozen objects btwn isolates in the docs. for now, we're avoiding talking abt concurrency in sdk docs b/c it's not a super prominent aspect of flutter development. 
2. rather, `Realm.writeAsync()` is intended for performing non-blocking operations. for example, you might want to use `Realm.writeAasync()` to perform a very large write operation but have the UI (which is running on the same thread) still reactive while you're doing this.

mongodben, line 173: i think we need to rethink the paragraph.

a couple of considerations:

1. in the first sentence it says "in the background". i'd change this to say "asynchronously", as that's a more accurate technical description of what's happening.
1.we don't talk about threading in the Flutter docs elsewhere. Dart actually doesn't technically have multiple threads. concurrency is achieved with [isolates](https://dart.dev/guides/language/concurrency). and, we don't talk about passing frozen objects btwn isolates in the docs. for now, we're avoiding talking abt concurrency in sdk docs b/c it's not a super prominent aspect of flutter development. 
1. rather, `Realm.writeAsync()` is intended for performing non-blocking operations. for example, you might want to use `Realm.writeAsync()` to perform a very large write operation but have the UI (which is running on the same thread) still reactive while you're doing this.

per these considerations, the paragraphh could be rephrased to something like:

```rst
You can add, modify, or delete objects asynchronously with 
:flutter-sdk:`Realm.writeAsync() <realm/Realm/writeAsync.html>`.
Use ``Realm.writeAsync()`` to perform write operations without blocking execution of the main process. For example, this can be useful when you perform a large write operations while you need to keep the UI responsive.
```

examples/dart/test/read_write_data_test.dart
```diff
@@ -83,4 +83,19 @@ void main() {
...
+      realm.add<Car>(newOutback);
+    });
+    // :snippet-end:
+    final outback = realm.find<Car>("Subaru");
+    expect(outback!.miles, 2);
```
mongodben, line 98: with the above suggestion accepted, you can test this by seeing that the car has not yet been added to the realm. 

so smthn like:

``dart
final outback = realm.find<Car>("Subaru");
expect(outback, isNull);
```

Comment (mongodben) [inline examples/dart/test/read_write_data_test.dart:98]: with the above suggestion accepted, you can test this by seeing that the car has not yet been added to the realm. then subsequently wait for a little while, and see that the car has been populated. 

so smthn like:

```suggestion
    final outback = realm.find<Car>("Subaru");
    expect(outback, isNull);
    expect(realm.isInTransaction, true);
    // let transaction resolve
    await Future.delayed(Duration(milliseconds: 500));
    expect(realm.isInTransaction, false);
    expect(realm.find<Car>("Subaru")!.miles, 2);
```

mongodben (commented):
back to you! this is a tricky one, and the docs needed is curiously different from swift.

would probably also be good to share this w #realm-dart-team. interested in their feedback on this one too since it's a bit of a *funky* feature.

mongodben:
oh one more thing, based on looking at the SDK PR, it's probably also worth documenting the second parameter of Realm.writeAsync(), the cancellation token which you can use to cancel a transaction.

https://github.com/realm/realm-dart/pull/930/files#diff-bf48d02e4defff0f358c4c408815da36fa24d0d2a66ecb3cc0332d932e89dae9R997

source/sdk/flutter/realm-database/read-and-write-data.txt
dacharyc: That makes sense. We've talked about some of this before, but I made assumptions when I saw the "Freeze" page based on the fact that freezing in other SDKs is largely related to alleviating threading issues. Updating.

mongodben (approved):
lgtm

PR #2242, opened 2022-10-26
(DOCSP-25673): Link credentials Kotlin

cbullinger:
### Jira

- https://jira.mongodb.org/browse/DOCSP-25673

- [Link User Identities](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-25673-link-credentials-kotlin/sdk/kotlin/app-services/link-credentials/)

### Review Guidelines

### DIFF
--- a/source/includes/link-user-identities.rst
+++ b/source/includes/link-user-identities.rst
@@ -1,0 +1,25 @@
+Realm provides :ref:`many authentication providers <authentication-providers>` 
+to log users into your app. Each provider creates a unique user identity. 
+Realm lets you merge multiple credentials into one user identity.
+
+However, you can only link credentials that are not already in use by another 
+user. Once credentials are used to login a user, they can no longer be merged 
+into another user identity. You cannot link multiple email/password credentials
+together.
+
+Example
+-------
+
+Consider an application that offers :ref:`anonymous login
+<anonymous-authentication>`, which allows users to explore the app without 
+registering. If a user wants to continue using the application, they can create 
+a permanent account by using another authentication provider. Realm 
+creates a new ``User`` object. The app can then link the new identity with the 
+current user.
+
+.. note:: 
+
+   Depending on how you have configured email/password authentication, there may 
+   be additional steps (confirming the email address, for example) before the 
+   new account is created and can be linked.
+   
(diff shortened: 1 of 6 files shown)


source/sdk/kotlin/app-services/link-credentials.txt
```diff
@@ -0,0 +1,40 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Realm provides :ref:`many authentication providers <authentication-providers>` 
+to log users into your app. Each provider creates a unique user identity. 
+Realm lets you merge multiple credentials into one user identity.
```
cmelchior, line 15: I think there needs to be some description of caveats here, i.e. you cannot link credentials if it was already used to login (and thus fully create the user). So this would not work:

```
val anonUser = app.login(Credentials.anonymous()) // logs in with an anonymous user
app.emailPasswordAuth.registerUser(email, password)
val credentials = Credentials.emailPassword(email, password)
val emailUser = app.login(credentials)
anonUser.linkCredentials(credentials); // Will fail
```

I believe it is the same for other auth providers as well. If you once logged in with them (and thus fully created a user). You can no longer merge them.

cbullinger: Good catch. I've updated to specify that you can't link credentials if they've already been used to login a user and that you can't link multiple email/pwd credentials.

source/includes/link-user-identities.rst
```diff
@@ -0,0 +1,25 @@
+Realm provides :ref:`many authentication providers <authentication-providers>` 
+to log users into your app. Each provider creates a unique user identity. 
+Realm lets you merge multiple credentials into one user identity.
+
+However, you can only link credentials that are not already in use by another
```
mongomoe, line 5: Suggestion/Not Required for LGTM: 

I don't like the use of "However" in docs, it feels very "essay-like", can we rephrase this to sound more natural, maybe:

"Credentials must be linked prior to logging a user in, if a user logs in then you cannot link that credential with another credential. Also, you cannot link multiple email/password credentials together."

cbullinger: Agreed. Updating to avoid 'however'.

mongomoe (commented):
1 needed fix + 1 suggestion, close to LGTM. Great work so far

mongomoe (approved):
LGTM + 1 more suggestion

PR #2254, opened 2022-10-31
(DOCSP-24568): Flutter - call Atlas Functions

mongodben:
## Pull Request Info

Call Atlas Functions from the Flutter SDK.

### Jira

- https://jira.mongodb.org/browse/DOCSP-24568

- [Call a Function](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24568/sdk/flutter/app-services/call-function)
- [App Services (landing page)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24568/sdk/flutter/app-services/)

### Review Guidelines


cbush (approved):

examples/dart/test/functions_test.dart
```diff
@@ -0,0 +1,40 @@
...
+main() {
+  late App app;
+  late EmailPasswordAuthProvider authProvider;
+  setUpAll(() async {
+    app = App(AppConfiguration(APP_ID));
+    authProvider = EmailPasswordAuthProvider(app);
+    final email = "lisa@example.com";
+    final password = "myStr0ngPassw0rd";
+    try {
+      await authProvider.registerUser(email, password);
```
dacharyc, line 15: I presume this is going to throw an error every time from now on since this user is already registered. Should we be deleting this user during teardown instead of just logging them out?

source/sdk/flutter/app-services/call-function.txt
```diff
@@ -0,0 +1,48 @@
...
+
+For more information on configuring and writing Atlas Functions, refer to
+:ref:`Atlas Functions in the App Services documentation <functions>`.
+
+Before You Begin
+----------------
+
+#. In an App Services App, :ref:`define an Atlas Function <define-a-function>`.
+#. :ref:`Initialize the App client <flutter-connect-to-backend>`.
+#. In your client project, :ref:`Authenticate a user <flutter-authenticate>`.
```
dacharyc, line 29: The "In your client project" jumped out at me here - it made me go "But wait, aren't we also initializing the App client in the client project? We make no mention of it there."

dacharyc (commented):
Just a couple of comments based on things that jumped out at me - nothing that need block merging, though.

examples/dart/test/functions_test.dart
mongodben: i don't think it really matters. i believe i added this flow before the ability to delete users was added to SDK. there's this pattern in the auth examples as well. 

keeping for now.

source/sdk/flutter/app-services/call-function.txt
mongodben: will clarify this text a bit. good spot.

source/examples/generated/flutter/functions_test.snippet.call-function.dart
```diff
@@ -0,0 +1,6 @@
+final response = await user.functions.call("addition", [1, 2]);
+
+// convert EJSON response to Dart number
+final responseAsNum = num.tryParse(response["\$numberLong"]);
+
+prints(responseAsNum); // prints 3
```
desistefanova, line 6: This is just a suggestion.
It would be much clearer for me if I could see where these arguments [1,2] are passed and where `numberLong` is returned from. Do you think it's a good idea here to have a very short sample of the Atlas function with an empty body but with the signature and the returned object?

PR #2256, opened 2022-11-01
Point to the correct Realm Kotlin dependency when used across Gradle Modules

cmelchior:
The callout to manually including the Realm Kotlin SDK dependency when used across modules where pointing to a non-existing dependency.


cbush (approved):
Thank you!

PR #2260, opened 2022-11-01
Add info about cancellationToken

MongoCaleb:
## Pull Request Info

Added a note indicating that the new methods take a cancellationToken. Since these methods use the token as all other C# async methods do, no need to add explanation or code example to our docs.

### Jira

https://jira.mongodb.org/browse/DOCSP-22721

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/cancel_token/sdk/dotnet/fundamentals/write-transactions/

### DIFF
--- a/source/sdk/dotnet/fundamentals/write-transactions.txt
+++ b/source/sdk/dotnet/fundamentals/write-transactions.txt
@@ -75,8 +75,8 @@ The following code shows how to use both approaches:
 
    As with most C# async methods, the 
    :dotnet-sdk:`WriteAsync <reference/Realms.Realm.html#Realms_Realm_WriteAsync_System_Action_System_Threading_CancellationToken_>`,
-   :dotnet-sdk:`BeginWriteAsync() <reference/Realms.Realm.html#Realms_Realm_BeginWriteAsync_System_Threading_CancellationToken_>`,
-   and :dotnet-sdk:`CommitAsync() <reference/Realms.Transaction.html#Realms_Transaction_CommitAsync_System_Threading_CancellationToken_>`
+   :dotnet-sdk:`BeginWriteAsync <reference/Realms.Realm.html#Realms_Realm_BeginWriteAsync_System_Threading_CancellationToken_>`,
+   and :dotnet-sdk:`CommitAsync <reference/Realms.Transaction.html#Realms_Transaction_CommitAsync_System_Threading_CancellationToken_>`
    methods can take a 
    `CancellationToken <https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-6.0>`_ 
    as a parameter, providing you with the option cancel the thread before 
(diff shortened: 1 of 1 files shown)


source/sdk/dotnet/fundamentals/write-transactions.txt
```diff
@@ -69,3 +71,13 @@ The following code shows how to use both approaches:
...
 
+.. note:: CancellationTokens
+
+   As with most C# async methods, the 
+   :dotnet-sdk:`WriteAsync <reference/Realms.Realm.html#Realms_Realm_WriteAsync_System_Action_System_Threading_CancellationToken_>`,
+   :dotnet-sdk:`BeginWriteAsync <reference/Realms.Realm.html#Realms_Realm_BeginWriteAsync_System_Threading_CancellationToken_>`,
+   and :dotnet-sdk:`CommitAsync <reference/Realms.Transaction.html#Realms_Transaction_CommitAsync_System_Threading_CancellationToken_>`
+   methods can take a 
+   `CancellationToken <https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-6.0>`_ 
+   as a parameter, providing you with the option cancel the thread before
```
nirinchev, line 82:
```suggestion
   as a parameter, providing you with the option cancel the operation before 
```

nirinchev (commented):
So this information is mostly correct, though there are some nuances. E.g. if you cancel `CommitAsync`, the commit doesn't actually get canceled - you just cancel waiting for the commit to complete (it'll continue writing to disk in the background). This is very niche though and I'm not sure we need to explicitly call it out. We had a similar discussion with @dacharyc for the dart docs and there we decided not to go into that much detail.

LaPeste (commented):
I agree with Nikola's comment, I think that keeping it simple as is covers the majority of users' needs.
I made just 2 very small suggestions.

source/sdk/dotnet/fundamentals/write-transactions.txt
```diff
@@ -49,15 +49,17 @@ Initiate a Transaction
 ~~~~~~~~~~~~~~~~~~~~~~
 
 The .NET SDK provides a simple API you can use for most writes.  The 
-:dotnet-sdk:`Write() <reference/Realms.Realm.html#Realms_Realm_Write__1_System_Func___0__>` 
-method wraps all commands into a single transaction and then commits the 
+:dotnet-sdk:`Write() <reference/Realms.Realm.html#Realms_Realm_Write__1_System_Func___0__>` and
+:dotnet-sdk:`WriteAsync() <reference/Realms.Realm.html#Realms_Realm_WriteAsync_System_Action_System_Threading_CancellationToken_>` and
+methods wrap all commands into a single transaction and then commits the
```
mongodben, line 54:
```suggestion
methods wrap all commands into a single transaction and then commit the 
```

plural

PR #2261, opened 2022-11-02
(DOCSP-26316): Flutter - new Compact a Realm page

dacharyc:
## Pull Request Info

This adds a page for compacting a Realm, showing the use of the `shouldCompactCallback()` API that is currently available.

### Jira

- https://jira.mongodb.org/browse/DOCSP-26316

- [Compact a Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24586/sdk/flutter/realm-database/compact/): New page
- [Open & Close a Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-24586/sdk/flutter/realm-database/open-and-close-a-realm/#compact-a-realm): New subsection mentioning compacting a realm and linking to the new page

### Review Guidelines

### DIFF
--- a/examples/dart/test/compact_realm_test.dart
+++ b/examples/dart/test/compact_realm_test.dart
@@ -15,7 +15,7 @@ void main() {
         // For convenience, this example defines a const
         // representing a byte to MB conversion for compaction
         // at an arbitrary 10MB file size.
-        const tenMB = 10 * 1024 * 1024;
+        const tenMB = 10 * 1048576;
         return totalSize > tenMB;
       }));
       var realm = Realm(config);
@@ -28,7 +28,7 @@ void main() {
       var config = Configuration.local([Car.schema],
           shouldCompactCallback: ((totalSize, usedSize) {
         // Compact if the file is over 10MB in size and less than 50% 'used'
-        const tenMB = 10 * 1024 * 1024;
+        const tenMB = 10 * 1048576;
         return (totalSize > tenMB) &&
             (usedSize.toDouble() / totalSize.toDouble()) < 0.5;
       }));
--- a/examples/dart/test/compact_realm_test.dart
+++ b/examples/dart/test/compact_realm_test.dart
@@ -1,0 +1,41 @@
+import 'package:test/test.dart';
+import '../bin/models/car.dart';
+import 'package:realm_dart/realm.dart';
+import 'package:path/path.dart' as path;
+import 'dart:io';
+import './utils.dart';
+
+void main() {
+  group('Compact a Realm', () {
+    test('Compact a Realm with callback', () async {
+      // :snippet-start: compact-with-callback
+      var config = Configuration.local([Car.schema],
+          shouldCompactCallback: ((totalSize, usedSize) {
+        // shouldCompactCallback sizes are in bytes.
+        // For convenience, this example defines a const
+        // representing a byte to MB conversion for compaction
+        // at an arbitrary 10MB file size.
+        const tenMB = 10 * 1024 * 1024;
+        return totalSize > tenMB;
+      }));
+      var realm = Realm(config);
+      // :snippet-end:
+      realm.close();
+      await cleanUpRealm(realm);
+    });
+    test('Compact a Realm with callback plus logic', () async {
+      // :snippet-start: compact-with-callback-and-logic
+      var config = Configuration.local([Car.schema],
+          shouldCompactCallback: ((totalSize, usedSize) {
+        // Compact if the file is over 10MB in size and less than 50% 'used'
+        const tenMB = 10 * 1024 * 1024;
+        return (totalSize > tenMB) &&
+            (usedSize.toDouble() / totalSize.toDouble()) < 0.5;
+      }));
+      var realm = Realm(config);
+      // :snippet-end:
+      realm.close();
+      await cleanUpRealm(realm);
+    });
+  });
+}
--- a/source/sdk/flutter/realm-database/compact.txt
+++ b/source/sdk/flutter/realm-database/compact.txt
@@ -10,25 +10,35 @@ Compact a Realm - Flutter SDK
    :depth: 2
    :class: singlecol
 
+If you want to reduce the size of a realm file to improve performance, you 
+can compact the realm. Compacting a realm reduces the amount of unused 
+empty space within the realm. Every production application should implement 
+compaction to periodically reduce the realm file size.
+
 The size of a Realm Database file is always larger than the total size of 
 the objects stored within it. This architecture enables some of realm's 
-great performance, concurrency, and safety benefits.
+performance, concurrency, and safety benefits.
 
 Realm writes new data within unused space tracked inside the file. In some 
-situations, unused space may comprise a significant portion of a realm file. 
+situations, a realm file may contain a significant amount of unused space. 
 If file size grows large enough to negatively impact performance, you can 
 compact the realm to reduce its file size.
 
-.. tip:: Implement Compaction in Your Production Application
+Compact When Opening a Realm
+----------------------------
 
-   Every production application should implement compaction to periodically 
-   reduce the realm file size.
+.. note::
+
+   You cannot compact a :ref:`Flexible Sync realm <flutter-open-synced-realm>`.
 
 You can define a :flutter-sdk:`shouldCompactCallback() 
 <realm/LocalConfiguration/shouldCompactCallback.html>` as a property of a 
-local realm configuration. This callback takes two ``int`` values representing 
+local realm's ``configuration``. This callback takes two ``int`` values representing 
 the total number of bytes and the used bytes of the realm file on disk. The 
-most basic usage is to define a file size at which compaction should occur.
+callback returns a ``bool``. Compaction only occurs if the ``bool`` returns 
+``true``, and another process is not currently accessing the realm file.
+
+The most basic usage is to define a file size at which compaction should occur.
 
 .. literalinclude:: /examples/generated/flutter/compact_realm_test.snippet.compact-with-callback.dart
    :language: dart
--- a/source/sdk/flutter/realm-database/compact.txt
+++ b/source/sdk/flutter/realm-database/compact.txt
@@ -19,16 +19,16 @@ situations, unused space may comprise a significant portion of a realm file.
 If file size grows large enough to negatively impact performance, you can 
 compact the realm to reduce its file size.
 
-.. tip:: Implement Compacting in Your Production Application
+.. tip:: Implement Compaction in Your Production Application
 
-   Every production application should implement compacting to periodically 
+   Every production application should implement compaction to periodically 
    reduce the realm file size.
 
 You can define a :flutter-sdk:`shouldCompactCallback() 
 <realm/LocalConfiguration/shouldCompactCallback.html>` as a property of a 
-local realm configuration. This callback takes two ints representing the
-total number of bytes and the used bytes of the realm file on disk. The most 
-basic usage is to define a file size at which compaction should occur.
+local realm configuration. This callback takes two ``int`` values representing 
+the total number of bytes and the used bytes of the realm file on disk. The 
+most basic usage is to define a file size at which compaction should occur.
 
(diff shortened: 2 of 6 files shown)


source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+The size of a Realm Database file is always larger than the total size of 
+the objects stored within it. This architecture enables some of realm's 
+great performance, concurrency, and safety benefits.
+
+Realm writes new data within unused space tracked inside the file. In some 
+situations, unused space may comprise a significant portion of a realm file. 
+If file size grows large enough to negatively impact performance, you can 
+compact the realm to reduce its file size.
+
+.. tip:: Implement Compacting in Your Production Application
```
krollins-mdb, line 22: Terminology check! This is probably out of scope for this PR, but I noticed that the [Swift](https://www.mongodb.com/docs/realm/sdk/swift/realm-files/compacting/) and [.NET](https://www.mongodb.com/docs/realm/sdk/dotnet/advanced-guides/compact-realm/) versions do this too. We seem to use "compacting" and "compaction" interchangeably. Do these refer to the same thing? If so, we should probably stick with one to avoid confusion.

If they're the same, I think "compaction" makes more sense. We should try to avoid verbing nouns, as it's not always clear which context they're being used in.

krollins-mdb (approved):
Nice! I have one small suggestion that shouldn't block you from merging. Also a note about terminology that's likely out of scope for this PR.

source/sdk/flutter/realm-database/compact.txt
dacharyc: Ahh, yeah, good point! I'll change it in this PR, but yeah, I know I personally have used these terms interchangeably - not sure if that is because of the underlying SDK docs doing that or if it's just me. Good call out!

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+The size of a Realm Database file is always larger than the total size of 
+the objects stored within it. This architecture enables some of realm's 
+great performance, concurrency, and safety benefits.
```
mongodben, line 15:
```suggestion
performance, concurrency, and safety benefits.
```

felt marketing-y

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+
+Realm writes new data within unused space tracked inside the file. In some 
+situations, unused space may comprise a significant portion of a realm file.
```
mongodben, line 18: can we find synonym for "comprise"? something at lower reading level.

perhaps (courtesy of thesaurus.com):

- take up
- be

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+Compact a Realm - Flutter SDK
+=============================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+The size of a Realm Database file is always larger than the total size of
```
mongodben, line 13: can you add a sentence at the top saying what compacting a realm is on a high level?

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+The size of a Realm Database file is always larger than the total size of 
+the objects stored within it. This architecture enables some of realm's 
+great performance, concurrency, and safety benefits.
+
+Realm writes new data within unused space tracked inside the file. In some 
+situations, unused space may comprise a significant portion of a realm file. 
+If file size grows large enough to negatively impact performance, you can 
+compact the realm to reduce its file size.
+
+.. tip:: Implement Compaction in Your Production Application
```
mongodben, line 22: i don't like this as a tip since it's important.

instead, can you make it the 2nd sentence, following the sentence that's proposed in this above comment https://github.com/mongodb/docs-realm/pull/2261/files#r1012903187

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+
+   Every production application should implement compaction to periodically 
+   reduce the realm file size.
+
+You can define a :flutter-sdk:`shouldCompactCallback() 
+<realm/LocalConfiguration/shouldCompactCallback.html>` as a property of a 
+local realm configuration. This callback takes two ``int`` values representing
```
mongodben, line 29: only local? doesn't work with sync?

mongodben, line 29:
```suggestion
local realm's ``Configuration``. This callback takes two ``int`` values representing 
```

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+Realm writes new data within unused space tracked inside the file. In some 
+situations, unused space may comprise a significant portion of a realm file. 
+If file size grows large enough to negatively impact performance, you can 
+compact the realm to reduce its file size.
+
+.. tip:: Implement Compaction in Your Production Application
+
+   Every production application should implement compaction to periodically 
+   reduce the realm file size.
+
```
mongodben, line 26: Can you add a header here, smthn like:

```rst
Compact When Opening Realm
--------------------------
```

examples/dart/test/compact_realm_test.dart
```diff
@@ -0,0 +1,41 @@
...
+      // :snippet-end:
+      realm.close();
+      await cleanUpRealm(realm);
+    });
+    test('Compact a Realm with callback plus logic', () async {
+      // :snippet-start: compact-with-callback-and-logic
+      var config = Configuration.local([Car.schema],
+          shouldCompactCallback: ((totalSize, usedSize) {
+        // Compact if the file is over 10MB in size and less than 50% 'used'
+        const tenMB = 10 * 1024 * 1024;
```
mongodben, line 31: nit: why is 1mb calculated as `1024 * 1024` (1kb * 1kb)?

imo it'd make more sense to calculate as something like `2^20` or alternatively just the number of bytes `1048576`.

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+You can define a :flutter-sdk:`shouldCompactCallback() 
+<realm/LocalConfiguration/shouldCompactCallback.html>` as a property of a 
+local realm configuration. This callback takes two ``int`` values representing 
+the total number of bytes and the used bytes of the realm file on disk. The
```
mongodben, line 30: there should be a sentence here specifying that the callback returns a boolean, and compaction only occurs if the bool returns true.

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,41 @@
...
+situations, unused space may comprise a significant portion of a realm file. 
+If file size grows large enough to negatively impact performance, you can 
+compact the realm to reduce its file size.
+
+.. tip:: Implement Compaction in Your Production Application
+
+   Every production application should implement compaction to periodically 
+   reduce the realm file size.
+
+You can define a :flutter-sdk:`shouldCompactCallback()
```
mongodben, line 27: somewhere in discussion of this method, could you provide a sentence saying why you **do not** want to compact on *every* open like `shouldCompactCallback: () => true`.

i assume it's for performance reasons b.c compaction is a costly thread blocking operation or smthn, but worth explicitly calling out.

mongodben (commented):
looking good! some small things throughout, noted in comments.

also can you add a very small section pointing to the 'Open & Close a Realm' page, similar to [Encryption](https://www.mongodb.com/docs/realm/sdk/flutter/realm-database/open-and-close-a-realm/#encrypt-a-realm) (also just noticed that ref issue, which will fix now)

source/sdk/flutter/realm-database/compact.txt
dacharyc: I don't think any of those synonyms are quite right - rewording it altogether.

source/sdk/flutter/realm-database/compact.txt
dacharyc: I can do that, but we added this as a tip a while ago because Ian wanted additional attention on this. I worry it may get lost as a plain text sentence. Could it be both? Or maybe it warrants an "Important" callout?

source/sdk/flutter/realm-database/compact.txt
dacharyc: It's only available on the local configuration. This is consistent with the Swift SDK, so I assume it's intentional and not an inadvertent omission or not-yet-implemented detail.

examples/dart/test/compact_realm_test.dart
dacharyc: 🤷‍♀️ 

This is the code example we use in the other SDKs - I'm not sure where the original came from. I don't have a strong opinion on this so I'm happy to change it. Making it bytes.

source/sdk/flutter/realm-database/compact.txt
mongodben: let's see how both looks. and i like the idea of changing to "important"

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,51 @@
...
+situations, a realm file may contain a significant amount of unused space. 
+If file size grows large enough to negatively impact performance, you can 
+compact the realm to reduce its file size.
+
+Compact When Opening a Realm
+----------------------------
+
+.. note::
+
+   You cannot compact a :ref:`Flexible Sync realm <flutter-open-synced-realm>`.
```
mongodben, line 32: nit: can you put the note above the heading?

source/sdk/flutter/realm-database/compact.txt
mongodben: and i think this paragraph should be above the `Compact When Opening a Realm` section b/c it'll also be applicable to the `Realm.compact()` static method when that's released.

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,51 @@
...
+
+You can define a :flutter-sdk:`shouldCompactCallback() 
+<realm/LocalConfiguration/shouldCompactCallback.html>` as a property of a 
+local realm's ``configuration``. This callback takes two ``int`` values representing
```
mongodben, line 36:
```suggestion
local realm's ``Configuration``. This callback takes two ``int`` values representing 
```

mongodben: actually this should be a `LocalConfiguration`, sorry for confusion with previous comment.

source/sdk/flutter/realm-database/compact.txt
```diff
@@ -0,0 +1,51 @@
...
+.. note::
+
+   You cannot compact a :ref:`Flexible Sync realm <flutter-open-synced-realm>`.
+
+You can define a :flutter-sdk:`shouldCompactCallback() 
+<realm/LocalConfiguration/shouldCompactCallback.html>` as a property of a 
+local realm's ``configuration``. This callback takes two ``int`` values representing 
+the total number of bytes and the used bytes of the realm file on disk. The 
+callback returns a ``bool``. Compaction only occurs if the ``bool`` returns 
+``true``, and another process is not currently accessing the realm file.
```
mongodben, line 39:
```suggestion
``true`` and another process is not currently accessing the realm file.
```

mongodben (commented):
looking way better! responded to the open convos, and a couple other small suggestions.

source/sdk/flutter/realm-database/compact.txt
dacharyc: With the callout about not compacting Flex Sync realms, I don't think there's really room on this page for a second callout about implementing this in prod apps. Leaving it as just text in the overview.

mongodben (approved):
lgtm! nice work here

PR #2272, opened 2022-11-07
Update the Carthage installation instructions to use xcframeworks

tgoyne:
This is required for simulators on m1 macs and is quite a bit simpler as a bonus.


dacharyc (approved):
Nice - thank you!

PR #2277, opened 2022-11-08
Removed note related to old versions

papafe:
I am just removing a note that is related to versions of Xamarin.iOS and Realm that are now more than 1 year old, so it can be safely removed.


papafe:
This is my first PR for the docs, so I wasn't sure if I was supposed to open a JIRA ticket necessarily or to do something else. Please let me know if I'm missing something :)

dacharyc (approved):
Awesome, this is great - thank you!

PR #2287, opened 2022-11-10
add cancellationtoken info

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-20211 

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-20211/sdk/dotnet/examples/sync-changes-between-devices/#cancel-a-sync-session

### Review Guidelines

### DIFF
--- a/examples/dotnet/Examples/Convert.cs
+++ b/examples/dotnet/Examples/Convert.cs
@@ -25,7 +25,9 @@ namespace Examples
             existingRealm.WriteCopy(syncConfig);
 
             // You can now delete the nonsynced realm:
-            Realm.DeleteRealm(existingConfig);
+            // :uncomment-start:
+            //Realm.DeleteRealm(existingConfig);
+            // :uncomment-end:
 
             // You can now use the synced realm:
             var syncedRealm = Realm.GetInstance(syncConfig);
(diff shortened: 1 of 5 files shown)


examples/dotnet/Examples/Convert.cs
```diff
@@ -25,7 +25,9 @@ public async Task TestNoSyncToSync()
             existingRealm.WriteCopy(syncConfig);
 
             // You can now delete the nonsynced realm:
-            Realm.DeleteRealm(existingConfig);
+            // :uncomment-start:
```
cbush, line 28: Why is DeleteRealm() commented out?

MongoCaleb: because it was borking something on the build server tests....but only intermittently.

MongoCaleb: Doesn't change the example code.

MongoCaleb:
+1 from Monsoor.

examples/dotnet/Examples/Convert.cs
cbush: Here's what I suggest instead: call the configuration "my.realm.nosyncToSyncExample" (or something unique to this test). Then it should not interfere with the other examples running concurrently.

The problem with uncommenting is that the commented code is not really tested and mistakes can seep in that way.

MongoCaleb: fixed

cbush (approved):

MongoCaleb:
Merging with the expectation that @cbush is fixing the ref issue.

PR #2307, opened 2022-11-17
(DOCSP-26590): Flutter model data section

mongodben:
## Pull Request Info

Move Flutter content to 'Model Data' section

also remove some `TODO:` statements that definitely should not have made the docs.

### Jira

- https://jira.mongodb.org/browse/DOCSP-26590

- [Model Data (Flutter)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26590/sdk/flutter/model-data/)

### Review Guidelines


source/sdk/flutter/model-data/define-realm-object-schema.txt
```diff
@@ -169,196 +169,19 @@ Supported Data Types
...
-   in the Atlas App Services documentation.
-
-.. _flutter-property-annotations:
-
 Property Annotations
 --------------------
 
 Use annotations to add functionality to properties in your Realm object models.
+To learn more about the available property annotations,
+refer to :ref:`Property Annotations <flutter-property-annotations>`.
```
dacharyc, line 177: I would at least give some example of what this entails so docs readers can decide whether this applies to them. I.e.

> You can use annotations for things like marking a property as nullable, setting a primary key, ignoring a property, and more. To learn more...

source/sdk/flutter/model-data/define-realm-object-schema.txt
```diff
@@ -169,196 +169,19 @@ Supported Data Types
...
-indicates that a property should be persisted under a different name.
-It's useful when opening a Realm across different bindings where code style conventions can differ.
+Define Relationship Properties
+------------------------------
 
-.. literalinclude:: /examples/generated/flutter/schemas.snippet.property-annotations.dart
-   :language: dart
-   :emphasize-lines: 15-16
+You can define relationships between Realm objects in your schema.
+To learn more about how to define relationships in your Realm object schema,
```
dacharyc, line 183: Same as above - I think I'd add a little more info here to give readers context on whether they want to click through. i.e.:

> The Realm Flutter SDK supports to-one relationships, to-many relationships, inverse relationships, and embedding objects within other objects. For more information about these relationships and how to define them, refer to:

source/sdk/flutter/model-data.txt
```diff
@@ -0,0 +1,65 @@
...
+-----------------------
+
+When you define your Realm object model, you specify a set of of properties
+to include in the schema. You can define properties with the following characteristics:
+
+- Its data type
+- If it is optional or required
+- If it is a primary key
+- If it is indexed
+- If the property defines a relationship to another Realm object type.
```
dacharyc, line 50: Our style guidance on punctuation in lists is a little oddly written, but I interpret this to mean either use no periods, or use periods after every list item - don't mix-and-match some with periods but others without: https://www.mongodb.com/docs/meta/style-guide/style/lists/#list-items

source/sdk/flutter/model-data.txt
```diff
@@ -0,0 +1,65 @@
...
+refer to the following documentation:
+
+- :ref:`Data Types <flutter-data-types>`
+- :ref:`Relationships <flutter-client-relationships>`
+- :ref:`Property Annotations <flutter-property-annotations>`
+
+Updating a Realm Object Schema
+------------------------------
+
+You can update your Realm Database schema overtime as your application changes.
```
dacharyc, line 62: "Overtime" = working more than some number of hours, "over time" = across a passage of time.
```suggestion
You can update your Realm Database schema over time as your application changes.
```

source/sdk/flutter/model-data.txt
```diff
@@ -0,0 +1,65 @@
...
+Depending on the type of change you make to the schema, you must take different actions.
```
dacharyc, line 63: "...you must take different actions."

To - do what? This sentence is a bit unclear. Maybe we could reword to something like:

> The steps to update the schema and your data vary depending on the type of schema change.

source/sdk/flutter/model-data/property-annotations.txt
```diff
@@ -0,0 +1,105 @@
...
+.. literalinclude:: /examples/generated/flutter/schemas.snippet.property-annotations.dart
+   :language: dart
+   :emphasize-lines: 2-3
+
+Ignored
+-------
+
+If you add the `Ignored <https://pub.dev/documentation/realm_common/latest/realm_common/Ignored-class.html>`__
+annotation to a property in your ``RealmModel``, the :ref:`realm object generator
+<flutter-generate-realm-object>` doesn't include the property in the ``RealmObject`` schema.
```
dacharyc, line 67: Should we go a step farther and add that Realm does not persist `Ignored` properties and their values to the database? I know that's the logical conclusion if you understand that the schema determines what gets stored in the DB, but wonder if it would be helpful to be explicit here for newcomers.

source/sdk/flutter/model-data/property-annotations.txt
```diff
@@ -0,0 +1,105 @@
...
+- ``String``
+- ``ObjectId``
+- ``Uuid``
+- ``DateTime``
+
+.. literalinclude:: /examples/generated/flutter/schemas.snippet.property-annotations.dart
+   :language: dart
+   :emphasize-lines: 12-13
+
+MapTo
```
dacharyc, line 96: I know the rest of the headings on this page match the annotation generally, but I wonder if we should change the heading for this one. I saw "Map" and immediately thought it was something related to dictionary types, and then wondered why it was on this page instead of on the types page. Might we call this something like "Map a Property to a Different Name" or "Use a Custom Property Name"? I also wonder if this should be up with Optional/Default value instead of at the bottom.

dacharyc (approved):
Nice! Looks good - one typo suggestion and I wonder about a couple of wording tweaks in places, but I think this is a good improvement.

source/sdk/flutter/model-data/property-annotations.txt
mongodben: def will shift up. will look at changing the other headers to not just be the annotation name as well.

source/sdk/flutter/model-data.txt
mongodben: much improved! thx

PR #2330, opened 2022-11-29
Fix Readability workflow

dacharyc:
In an attempt to fix the readability workflow check, I'm updating the workflow to a more recent version of the setup-python GitHub action, updating to a newer version of Python, and removing the x64 architecture since that's the default when unspecified. 🤞


mongodben (approved):
lgtm(?)!

give it a try 🤞

PR #2338, opened 2022-11-30
add new snippet & bluehawk it

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-26771

### Staged Changes

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26771/sdk/dotnet/manage-users/authenticate/#get-a-user-access-token

### DIFF
--- a/source/sdk/dotnet/manage-users/authenticate.txt
+++ b/source/sdk/dotnet/manage-users/authenticate.txt
@@ -168,10 +168,7 @@ work offline.
 Get a User Access Token
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Every User object contains a JSON Web Token (JWT) that you can use to access 
-Atlas App Services. You can also use the access token to authenticate HTTP 
-requests with Bearer Authentication, such as GraphQL. The following code 
-demonstrates getting and refreshing the user access token:
+.. include:: /includes/user-access-token.rst
 
 .. literalinclude:: /examples/generated/dotnet/AuthenticationExamples.snippet.get_user_token.cs
    :language: csharp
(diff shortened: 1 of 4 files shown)


source/sdk/dotnet/manage-users/authenticate.txt
```diff
@@ -165,3 +165,13 @@ where ``config`` is a :dotnet-sdk:`PartitionSyncConfiguration <reference/Realms.
 object. This approach results in a faster start-up and also enables the user to 
 work offline.
 
+Get a User Access Token
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Every User object contains a JSON Web Token (JWT) that you can use to access 
+Atlas App Services. You can also use the access token to authenticate HTTP 
+requests with Bearer Authentication, such as GraphQL. The following code 
+demonstrates getting and refreshing the user access token:
```
nlarew, line 174: - I think we can de-emphasize the JWT part here - we can talk about the implementation details more on the App Services side. Instead we can give context that an access token === logged in user.
- We could also be a little clearer that this isn't relevant for regular SDK use since it manages the token automatically.

```rst
When a user logs in, App Services creates an access token for the user that
grants them access to your App. The Realm SDK automatically manages access
tokens, refreshes them when they expire, and includes a valid access token for
the current user with each request. If you send requests outside of the SDK (for
example, through the GraphQL API) then you need to include the user's access
token with each request and manually refresh the token when it expires.

You can access and refresh a logged in user's access token in the SDK from their
``Realm.User`` object, as in the following example:
```

nlarew (approved):

PR #2346, opened 2022-12-01
(DOCSP-26727): [CPP] Configure and open a realm page

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26727

### Staged Changes

- [Configure & Open a Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26727/sdk/cpp/realm-files/configure-and-open-a-realm/): New page based on the Swift SDK "Configure & Open a Realm" page

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/examples/cpp/examples.cpp
+++ b/examples/cpp/examples.cpp
@@ -2,10 +2,13 @@
 #include <chrono>
 #include <string>
 #include <thread>
+#include <future>
 
 // :snippet-start: includes
 #include <cpprealm/sdk.hpp>
 // :snippet-end:
+
+static std::string APP_ID = "cpp-tester-foylc";
 
 // :snippet-start: define-models
 // Define your models like regular structs.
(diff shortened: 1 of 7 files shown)


cbush (approved):

examples/cpp/examples.cpp
```diff
@@ -51,6 +65,49 @@ TEST_CASE("first test case", "[test]") {
...
+    // :snippet-end:
+}
+
+TEST_CASE("open a realm at a path", "[realm]") {
+    // Construct an arbitrary path to use in the example
+    auto relative_realm_path = "custom_path_directory/";
+    std::filesystem::create_directories(relative_realm_path);
+    // :snippet-start: open-realm-at-path
+    // Construct a path, and then convert it to a string so you can append a name for the realm file
+    auto path = std::filesystem::current_path().append(relative_realm_path).string();
```
cbush, line 81: I would include the path again in the actual snippet - "custom_path_directory"

examples/cpp/examples.cpp
```diff
@@ -51,6 +65,49 @@ TEST_CASE("first test case", "[test]") {
...
+    /* Add a name for the file. When you provide a path, the `db_config` 
+     * constructor removes the last path element and makes that 
+     * the realm file name. */
+    path = path + "dog_objects";
+    auto config = realm::db_config{ path = path };
+    auto realm = realm::open<Dog>(config);
+    // :snippet-end:
+    // Write something to the realm to confirm this worked as expected.
```
cbush, line 89: What happens if it doesn't work as expected?

cbush (approved):

examples/cpp/examples.cpp
dacharyc: Well, adding it showed me the behavior about removing the last path element and making that the realm name. I suppose I could remove it since I've confirmed that this is correct. Or maybe I should add a `REQUIRE` to confirm that I can read from the realm and the object has been added? I dunno how much you'd advocate for "testing that the SDK works as expected." Given that it's in alpha and we know things will be changing, maybe it makes sense to be more aggressive about testing even basic stuff to make sure the code hasn't moved on from our code examples.

PR #2357, opened 2022-12-02
Minor C++ example readme updates

krollins-mdb:
## Pull Request Info

This PR doesn't have its own Jira, as I was working through the example repo to learn. I noticed some minor ways we can improve the readme, so this is kind of a follow-up.

### Jira

- https://jira.mongodb.org/browse/DOCSP-26490

### Review Guidelines

### DIFF
--- a/examples/cpp/README.md
+++ b/examples/cpp/README.md
@@ -79,7 +79,7 @@ FetchContent_Declare(
 )
 ```
 
-To change the version of the SDK your build, change the value
+To change the version of the SDK we use in the build, change the value
 of the `GIT_TAG`. While in early Alpha, this is a commit hash, but it could 
 be a version tag once the SDK starts doing tagged releases. For more 
 information, refer to the 
@@ -124,7 +124,9 @@ Open `CMakeLists.txt` and add the new file name to the `add_executable` function
 add_executable(examples examples.cpp my_new_test_file.cpp)
 ```
 
-To verify your new test file works, [build the project](https://github.com/mongodb/docs-realm/tree/master/examples/cpp#build-the-project) again and then [run the tests](https://github.com/mongodb/docs-realm/tree/master/examples/cpp#run-the-tests) again.
+To verify your new test file works, [build the project](https://github.com/mongodb/
+docs-realm/tree/master/examples/cpp#build-the-project) again and then [run the tests]
+(https://github.com/mongodb/docs-realm/tree/master/examples/cpp#run-the-tests) again.
 
 ## Add a New Example
 
@@ -171,10 +173,12 @@ Does not match the generator used previously: Unix Makefiles
 Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
 ```
 
-You can either run `cmake ../ -G Xcode` from a clean build directory, or remove the 
-files specified in the error. 
+To address this error, you can run `cmake ../ -G Xcode` from a clean build directory,
+create a new subdirectory for xcode (`/build/xcode`), or remove the files specified
+in the error. 
 
-On a successful build, you should see `examples.xcodeproj` in the `/build` directory.
+On a successful build, you should see `examples.xcodeproj` in the directory 
+Xcode builds to.
 
 Open the project in Xcode.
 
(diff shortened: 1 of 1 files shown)


examples/cpp/README.md
```diff
@@ -118,13 +118,13 @@ Include the relevant headers. They probably look something like:
...
 
-Open `CMakeLists.txt` and add the new file name to the `add_executable` at ln 20:
+Open `CMakeLists.txt` and add the new file name to the `add_executable` function:
 
 ```txt
 add_executable(examples examples.cpp my_new_test_file.cpp)
 ```
 
-Run `cmake` and `make` to build the project again, and then run the tests.
+To verify your new test file works, [build the project](https://github.com/mongodb/docs-realm/tree/master/examples/cpp#build-the-project) again and then [run the tests](https://github.com/mongodb/docs-realm/tree/master/examples/cpp#run-the-tests) again.
```
krollins-mdb, line 127: Note for reviewer: I know we tend to avoid "wiki" style links, but the process for this is exactly the same as what we've already defined. I think it makes sense to link back to those sections if readers need a reminder of what to do.

examples/cpp/README.md
```diff
@@ -148,30 +148,33 @@ TEST_CASE("add a dog object to a realm", "[test]") {
...
 
 From the `/build` directory, run CMake with a `-G Xcode` flag:
 
 ```shell
 cmake ../ -G Xcode
 ```
 
-Note: If you have previously run `CMake` for this project, you may see an
-error:
+If you have previously run `CMake` for this project, you may see this error:
```
krollins-mdb, line 166: I removed the "Note:" here because lines 166-175 are related content. Because GitHub doesn't have a good way of handling admonitions, it may not be clear that this content is related but 177 and further on isn't. (You can have a successful build and not see the error.)

examples/cpp/README.md
```diff
@@ -80,16 +79,16 @@ FetchContent_Declare(
 )
 ```
 
-To change the version of the SDK we use in the build, change the value
+To change the version of the SDK your build, change the value
```
dacharyc, line 82: Not sure if this was a typo of "To change the version of the SDK you build" or if it's "in your build" so I picked one.

```suggestion
To change the version of the SDK in your build, change the value
```

I do think it's worth noting, though, that if you commit these changes, it changes the Realm SDK version for anyone using this project. That's why I went with "we use in the build" here. I'm fine with not using first person here if you want to otherwise indicate that changing this and committing it means the project builds with the new version.

examples/cpp/README.md
dacharyc, line 127: Could we wrap this line, please? It's pretty long.

examples/cpp/README.md
```diff
@@ -148,30 +148,33 @@ TEST_CASE("add a dog object to a realm", "[test]") {
...
 
 ```shell
 CMake Error: Error: generator : Xcode
 Does not match the generator used previously: Unix Makefiles
 Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
 ```
 
-Run this from a clean build directory, or remove the specified files. On a 
-successful build, you should see `examples.xcodeproj` in the `/build` directory.
+You can either run `cmake ../ -G Xcode` from a clean build directory, or remove the
```
dacharyc, line 174: Bush pointed out another option here, which is to just make another directory in the `build` directory for the Xcode build. i.e. `/build/xcode`. Curious if you think it's worth adding. If you do, we might want to remove the specific `/build` in ln 177 below, and just say "build directory."

dacharyc (changes requested):
Nice improvements - thanks! Back to you with a couple of Qs and requests.

examples/cpp/README.md
krollins-mdb: Ooh, I wasn't aware that the Realm SDK version would change for people using this project. The original makes more sense now. Thanks for clarifying. I'll change this back to the original.

examples/cpp/README.md
krollins-mdb: I think that's worth adding. Seems like a better approach.

dacharyc (approved):
Nice, looks good! Good improvements 👍

PR #2377, opened 2022-12-07
(DOCSP-26728): C++ Model Data Section

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26728

### Staged Changes

- [Object Types and Schemas](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26728/sdk/cpp/model-data/object-models/)
- [Supported Data Types](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26728/sdk/cpp/model-data/supported-types/)
- [Relationships](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26728/sdk/cpp/model-data/relationships/)

### Test Run
Since we don't yet have CI to automatically run the tests, here is the output of running these tests locally:

```shell
Randomness seeded to: 3770930234
Business: {
	: 
	: MongoDB
	: {
	: email@example.com
	: 123-456-7890
}
}
Point of Interest: Grand Canyon Village
POI Latitude: 36.0554
Company named:Dunder Mifflin
dog: {
	: Rex
	: 1
}
===============================================================================
All tests passed (12 assertions in 8 test cases)

Program ended with exit code: 0
```

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/examples/cpp/README.md
+++ b/examples/cpp/README.md
@@ -68,6 +68,35 @@ To run the tests, execute the `examples` file from the `build` directory:
 ./examples
 ```
 
+### How to Skip Tests Using Device Sync
+
+Tests that use Device Sync require the example app to make network calls.
+When you run tests involving Device Sync, you'll see a modal requesting 
+you to enter your laptop password. Even when you select *Always Allow*, 
+you'll be prompted to re-enter this password whenever you build a new 
+version of the executable.
+
+This is because macOS enforces access control technology at the kernel 
+level. It does this using App Sandbox. When developing in Xcode, you can 
+[enable access to restricted 
+resources](https://developer.apple.com/documentation/xcode/configuring-the-macos-app-sandbox/). 
+Because this example project does not use Xcode, I'm not aware of a way 
+to set the required entitlements for this project. This is probably 
+something that would be useful to research when someone gets some downtime.
+
+Meanwhile, if you're doing a lot of building and running, and you don't
+want to enter your password every time you rebuild the executable, you 
+can exclude sync tests from your test run. To exclude the sync tests, 
+pass a flag that contains the tag or tags you want to exclude when you
+run the tests:
+
+```shell
+./examples ~[sync]
+```
+
+This requires us to tag every example using sync with the `sync` tag. For
+an example, see the "open a synced realm" test case in `examples.cpp`.
+
 ## Update the Realm SDK Version
 
 CMakeLists.txt has a FetchContent block that pulls in the `realm-cpp` repository
@@ -135,10 +164,17 @@ Find the relevant test case file for the section or category you wish to write
 an example for.
 
 Define a new `TEST_CASE`. This should have a unique string name - as in the 
-example below `"add a dog object to a realm"`, and a `"[test]"` tag.
+example below `"add a dog object to a realm"`, and a tag that designates it 
+as part of a relevant group of tests. In the example below, the test is tagged
+with `"[write]"`. 
+
+To enable us to run a group of tests, or exclude a group of tests, use 
+relevant tags when creating new test cases. For example, tests that use 
+Device Sync could be tagged with `"[sync]"` so we can easily skip them or
+run them specifically, depending on what we're testing.
 
 ```cpp
-TEST_CASE("add a dog object to a realm", "[test]") {
+TEST_CASE("add a dog object to a realm", "[write]") {
     auto dog = Dog { .name = "Floof", .age = 3 };
     
     std::cout << "dog: " << dog << "\n";
--- a/examples/cpp/README.md
+++ b/examples/cpp/README.md
@@ -80,7 +80,7 @@ This is because macOS enforces access control technology at the kernel
 level. It does this using App Sandbox. When developing in Xcode, you can 
 [enable access to restricted 
 resources](https://developer.apple.com/documentation/xcode/configuring-the-macos-app-sandbox/). 
-Because this example project does not use Xcode, I'm not aware of a way 
+Because this example project does not necessarily use Xcode, I'm not aware of a way 
 to set the required entitlements for this project. This is probably 
 something that would be useful to research when someone gets some downtime.
 
--- a/examples/cpp/define-object-model.cpp
+++ b/examples/cpp/define-object-model.cpp
@@ -142,8 +142,10 @@ TEST_CASE("create object with to-one relationship", "[model][write][relationship
     auto grandCanyonVillage = namedGrandCanyonVillage[0];
     std::cout << "Point of Interest: " << grandCanyonVillage->name << "\n";
     REQUIRE(grandCanyonVillage->name == "Grand Canyon Village");
-    auto const &theseGpsCoordinates = **grandCanyonVillage->gpsCoordinates;
-    auto const &latitude = gpsCoordinates.latitude;
+    auto const &theseGpsCoordinates = *(grandCanyonVillage->gpsCoordinates);
+    static_assert(std::is_same_v<decltype(theseGpsCoordinates), std::optional<GPSCoordinates> const &>, "Dereference fail!"); // :remove:
+    auto latitude = *(theseGpsCoordinates->latitude);
+    static_assert(std::is_same_v<decltype(latitude), double>, "Dereference fail!"); // :remove:
     std::cout << "POI Latitude: " << latitude << "\n";
     REQUIRE(latitude == 36.0554);
 };
--- a/examples/cpp/define-object-model.cpp
+++ b/examples/cpp/define-object-model.cpp
@@ -28,23 +28,23 @@ struct Business : realm::object {
 };
 // :snippet-end:
 
- // :snippet-start: model-with-ignored-field
- struct Employee : realm::object {
-     realm::persisted<realm::uuid> _id;
-     realm::persisted<std::string> firstName;
-     realm::persisted<std::string> lastName;
-     // Omitting the `realm::persisted` annotation means
-     // realm ignores this property
-     std::string ignoredField;
+// :snippet-start: model-with-ignored-field
+struct Employee : realm::object {
+    realm::persisted<realm::uuid> _id;
+    realm::persisted<std::string> firstName;
+    realm::persisted<std::string> lastName;
+    // Omitting the `realm::persisted` annotation means
+    // realm ignores this property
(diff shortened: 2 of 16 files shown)


examples/cpp/define-object-model.cpp
```diff
@@ -0,0 +1,159 @@
...
+    auto employeesNamedLeslie = employees.where("firstName == $0", {"Leslie"});
+    CHECK(employeesNamedLeslie.size() >= 1);
+    std::unique_ptr<Employee> leslieKnopePointer = employeesNamedLeslie[0];
+    REQUIRE(employee.ignoredField == "Organizer-In-Chief");
+    REQUIRE(leslieKnopePointer->ignoredField.empty());
+};
+
+#if 0
+// Can't currently get this to work
+TEST_CASE("create object with to-one relationship", "[model][write][relationship]") {
```
dacharyc, line 137: FYI, @cbush , I'm trying to work out this example but I can't quite get it to work. Feels like I may be missing something obvious. Do you spot something obvious I'm missing?

(Ideally I'd like the `GPSCoordinate` model properties `latitude` and `longitude` to be some more appropriate type like doubles, but I thought that might be part of my problem so I made them strings temporarily. Still no joy.)

examples/cpp/define-object-model.cpp
```diff
@@ -0,0 +1,180 @@
...
+    
+    std::cout << "Business: " << business << "\n";
+
+    realm.write([&realm, &business] {
+        realm.add(business);
+    });
+    auto businesses = realm.objects<Business>();
+    auto mongoDBPointer = businesses[0];
+    REQUIRE(businesses.size() >= 1);
+    REQUIRE((*mongoDBPointer->contactDetails).emailAddress == "email@example.com");
```
cbush, line 105: Is this test not made into a snippet?

examples/cpp/define-object-model.cpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+TEST_CASE("create object with ignored property", "[model][write]") {
+    auto realm = realm::open<Employee>();
+
+    auto employee = Employee { 
+        .firstName = "Leslie", 
+        .lastName = "Knope", 
+        .ignoredField = "Organizer-In-Chief" };
```
cbush, line 114: Er, the naming here threw me for a loop. I would call this `.someIgnoredField`, `.exampleIgnoredField`, or even `.jobTitle_notPersisted` or something. Otherwise I thought it was some magic realm object thing.

examples/cpp/define-object-model.cpp
```diff
@@ -0,0 +1,180 @@
...
+        realm.add(pointOfInterest);
+    });
+
+    auto pointsOfInterest = realm.objects<PointOfInterest>();
+    auto namedGrandCanyonVillage = pointsOfInterest.where("name == $0", {"Grand Canyon Village"});
+    CHECK(namedGrandCanyonVillage.size() >= 1);
+    auto grandCanyonVillage = namedGrandCanyonVillage[0];
+    std::cout << "Point of Interest: " << grandCanyonVillage->name << "\n";
+    REQUIRE(grandCanyonVillage->name == "Grand Canyon Village");
+    auto grandCanyonVillageLatitude = *(*grandCanyonVillage->gpsCoordinates).value().latitude;
```
cbush, line 145: Not much we can do about it, but this expression is nasty. We could break it into smaller statements to make it more understandable.
```cpp
auto const &gpsCoordinates = **grandCanyonVillage->gpsCoordinates;
auto const &latitude = gpsCoordinates.value().latitude;
```
Not sure of exact syntax that will work...

cbush (commented):
Part one of review: code

examples/cpp/README.md
```diff
@@ -68,6 +68,35 @@ To run the tests, execute the `examples` file from the `build` directory:
...
+something that would be useful to research when someone gets some downtime.
+
+Meanwhile, if you're doing a lot of building and running, and you don't
+want to enter your password every time you rebuild the executable, you 
+can exclude sync tests from your test run. To exclude the sync tests, 
+pass a flag that contains the tag or tags you want to exclude when you
+run the tests:
+
+```shell
+./examples ~[sync]
```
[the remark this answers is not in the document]
dacharyc: It does not, weirdly. I'm not sure why.

examples/cpp/define-object-model.cpp
dacharyc: This was mainly for me to verify that my understanding of how to define the model was correct. I will probably use it as a snippet in the CRUD section that I'll work on next. (Or may tweak it slightly - not sure yet.) i.e. using the model we defined, here's how to create an embedded object, read it, etc.

cbush (commented):
Important comments

examples/cpp/define-object-model.cpp
cbush: There's quite a bit of stuff going on, so this is actually not quite right either. First of all, very subtly, you forgot to change the subsequent line to use `theseGpsCoordinates`, hence "not needing" the .value(). Playing around with it I came up with this:
```cpp
    auto const &theseGpsCoordinates = *(grandCanyonVillage->gpsCoordinates);
    static_assert(std::is_same_v<decltype(theseGpsCoordinates), std::optional<GPSCoordinates> const &>, "Dereference fail!"); // :remove:
    auto latitude = *(theseGpsCoordinates->latitude);
    static_assert(std::is_same_v<decltype(latitude), double>, "Dereference fail!"); // :remove:
```

The first line uses the std::persisted dereference operator overload (`operator*()`) to give us the optional GPSCoordinates. Actually, we can say "unwraps". The first line unwraps the `persisted<optional<GPSCoordinates>>` to give us `optional<GPSCoordinates>`. Note the `->` is because grandCanyonVillage is a unique_ptr (also weird, but let's ignore that for now).

The second line checks that the above line did in fact give us an `optional<GPSCoordinates>`.

The third line *unwraps* the `optional<GPSCoordinates>` to get a `GPSCoordinates` with `->`. It then *unwraps* the `persisted<double>` member "latitude" with the persisted's `operator*()`.

The fourth line checks that it is indeed a double.

I got rid of the `const &` on the double because small values like doubles are better off being deliberately copied anyway.

cbush: The std types unique_ptr and optional offer "struct dereferencing" via `->`. `realm::persisted` only has the dereference operator `*`. Makes interoperation a bit messier. Would prefer if I could do `grandCanyonVillage-> gpsCoordinates->latitude`.

The reason for the lack of `->` is probably that it MUST return a pointer (`operator*()` can return whatever). From a cursory glance at how persisted is currently implemented, it doesn't really have something it can return as a pointer -- what it does return from `operator*()` is a temporary value.

Still, this makes the API pretty clunky. Will raise with team.

source/sdk/cpp/model-data/supported-types.txt
```diff
@@ -0,0 +1,144 @@
...
+          realm::persisted<std::optional<realm::BinaryData>>> opt_data_name;
+   * - Date
+     - .. code-block:: cpp
+          :copyable: false
+
+          realm::persisted<realm::Timestamp>> date_name;
+     - .. code-block:: cpp
+          :copyable: false
+
+          realm::persisted<std::optional<realm::Timestamp>> opt_date_name;
```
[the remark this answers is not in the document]
cbush: I have an update that basically steady_clock should never be used. They'll add a fix. For now we should say you have to use `std::chrono::time_point<std::chrono::system_clock>`.

source/sdk/cpp/model-data/object-models.txt
```diff
@@ -0,0 +1,177 @@
...
+is embedded. This differs from a :ref:`to-one <cpp-to-one-relationship>` 
+or :ref:`to-many <cpp-to-many-relationship>` relationship, in which the 
+related objects have independent lifecycles.
+
+To declare an object as an embedded object, inherit from 
+:cpp-sdk:`realm::embedded_object <structrealm_1_1embedded__object.html>`.
+
+.. literalinclude:: /examples/generated/cpp/define-object-model.snippet.model-with-embedded-object.cpp
+   :language: cpp
+   :emphasize-lines: 2, 17, 22
```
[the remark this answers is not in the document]
dacharyc: At the moment, there are not a lot of examples. While building things out, I thought I might want to reuse parts of examples in different permutations. For that purpose, it seemed better to set the emphasize in the doc where I use it instead of in the code example, since I might want to emphasize something different when I use it in a different context.

It's possible that this one is specific enough that I wouldn't re-use it, though? 🤷‍♀️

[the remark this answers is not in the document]
cbush: fwiw bluehawk generates the original file as well as an rST wrapper when you use the emphasize tag... so you can always re-use the original file elsewhere and treat the emphasized one as the "main" one.

source/sdk/cpp/model-data/supported-types.txt
```diff
@@ -0,0 +1,144 @@
...
+
+          realm::persisted<std::optional<MyEmbeddedClass>> opt_embedded_object_name;
+
+Supported Type Implementation Details
+`````````````````````````````````````
+
+Some of the supported types above are aliases for:
+
+- ``int``: An alias for an int-type object that encompasses: 
+  ``int8_t``, ``int16_t``, ``int32_t``, ``int64_t``
```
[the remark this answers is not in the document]
dacharyc: Ah, you're right, I misunderstood what I was seeing in one of the tests and this isn't right. Removing it.

source/sdk/cpp/model-data/supported-types.txt
```diff
@@ -0,0 +1,147 @@
...
+          :copyable: false
+
+          realm::persisted<std::optional<MyEmbeddedClass>> opt_embedded_object_name;
+
+Supported Type Implementation Details
+`````````````````````````````````````
+
+Some of the supported types above are aliases for:
+
+- ``mixed``: An alias for an object that could contain any of the other
```
cbush, line 139:
```suggestion
- ``mixed``: A union-like object that can represent a value any of the 
```

cbush (approved):
Great!

PR #2394, opened 2022-12-13
Corrected QuickStart page for .NET

papafe:
I have corrected some small things and changed the name of the example object, given that `Task` has a specific meaning in .NET and it could be confusing. I'm definitely open to other names, this was just an example. 

Also I'm not sure if in a quick start page we should show `MapTo` attributes that are probably not very important when just starting, and that developer will probably start to use only when looking at sync. For the same reason if we want a very slim quick start we could also remove the `partition` property and change `ObjectId` to `Guid`, given that the first is a new type for a newcomer and just adds to the confusion. Those are my two cents 😁 Obviously if we want to show already how to work with sync then this doesn't matter, but definitely we need to change the name and remove that enum that wasn't being used

### DIFF
--- a/source/examples/generated/dotnet/Program.snippet.read-filter.cs
+++ b/source/examples/generated/dotnet/Program.snippet.read-filter.cs
@@ -1,5 +1,5 @@
 var lessExpensiveGuitars = realm.All<Guitar>().Where(g => g.Price < 400);
 
-var guitarsSortedByMake = realm.All<Guitar>().OrderBy(g => g.Make);
+var guitarsSortedByMaker = realm.All<Guitar>().OrderBy(g => g.Maker);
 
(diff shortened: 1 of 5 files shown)


papafe:
NOTE: I just noticed that we are using `Task` also in other pages, and I think we should modify it in all of them if we agree for a name. If we want to keep it consistent with the template app we could also call it `Item`.

MongoCaleb (commented):
Rejecting because
1. "Make" is correct when referring to a brand.
2. I will rename "Task" to "Item" to match the template apps.
3. We don't change the /generated/ files. I'll update the source for those. 
4. I'll update the QuickStart.txt file with your suggestions.

MongoCaleb:
Fixed in https://github.com/mongodb/docs-realm/pull/2430

PR #2404, opened 2022-12-14
Small changes to compatibility for .NET

papafe:
I did some small changes:
- Made explicit we support .NET 5.0 and older versions. Strictly speaking .NET Core 2.0 and later includes .NET 5.0 but because the naming on .NET is very confusing, I think it makes sense to make it clear
- Ordered the top part, so that iOS and MacOS sections are close, same for iOS and Android+
- Removed the word "framework". The thing is that there is already .NET Framework, so the word framework here could be confusing... I could be convinced it makes sense though 😁 
- Modified a little the windows section so it seems slightly more consistent


MongoCaleb (approved):

MongoCaleb:
@papafe can you please rebase from upstream/master? That should fix the build errors.

MongoCaleb:
@papafe can you please rebase from upstream/master? That should fix the build errors.

papafe:
@MongoCaleb Sure, I've just done it

PR #2418, opened 2022-12-16
Add backend source

cbush:


examples/backend/data_sources/mongodb-atlas/todo/Item/schema.json
```diff
@@ -0,0 +1,24 @@
+{
```
dacharyc, line 1: We probably don't need this handful of schemas, as development mode is enabled and other apps will likely have different models anyway.

examples/backend/sync/config.json
```diff
@@ -0,0 +1,26 @@
...
+                },
+                "write": {
+                    "owner_id": "%%user.id"
+                }
+            }
+        ]
+    },
+    "queryable_fields_names": [
+        "_id",
+        "owner_id"
```
dacharyc, line 24: If we remove the schemas, I don't think we need an `owner_id` queryable field. Folks can add them as needed for the objects they're using in the examples.

examples/backend/auth/providers.json
```diff
@@ -0,0 +1,23 @@
+{
```
dacharyc, line 1: Big-picture, I'm not sure if this is the right selection of auth providers to have enabled by default. I'm wondering if we should also have custom-function and custom-jwt enabled. But some (most?) of the apps also have Google & FB enabled, although I know exposing those here would expose the client ID & secret name. Can we use those client IDs and secrets across clients? Curious if they do/should all share the same IDs.

I'm also now wondering if we should have our tester apps backing up their configs to private git repositories for backup purposes. When I was learning how to make changes to these, I worried occasionally about mucking up the backends. It might be nice to be able to restore them with all of their custom stuff - i.e. FB/Google auth, specific models, custom funcs, etc. if one of us accidentally messes with the wrong one or otherwise causes a problem with it.

dacharyc (approved):
This is probably fine and useful to add as-is. But I think we could make a couple of tweaks if we wanted to, and now I'm wondering if we should be backing up all of our tester configs.

examples/backend/sync/config.json
cbush: I think that defeats the purpose of exporting the backend. It should be that whatever in this can be imported and used right away.

cbush:
@dacharyc I'm thinking we can merge this as-is (just to have something) or cancel this and set up a private repo with all the exports like you suggested. Thoughts?

I don't want to tweak this on the fly because I feel it would defeat the purpose of exporting the backend exactly as is.

dacharyc:
> @dacharyc I'm thinking we can merge this as-is (just to have something) or cancel this and set up a private repo with all the exports like you suggested. Thoughts?

I think we could do both. I think there's value in having this here in case folks need to spin up a new backend. But separately, we should probably maintain a private repo with the configs for the backend apps we use in our tests.

PR #2420, opened 2022-12-16
(DOCSP-26729): C++ CRUD section

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26729

### Staged Changes

- [Create](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26729/sdk/cpp/crud/create/)
- [Read](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26729/sdk/cpp/crud/read/)
- [Update](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26729/sdk/cpp/crud/update/)
- [Delete](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26729/sdk/cpp/crud/delete/)

### Test Run

Since we don't yet have CI to automatically run the tests, here is the output of running these tests locally:

```shell
Randomness seeded to: 1000146614
Business: {
	: 
	: MongoDB
	: {
	: email@example.com
	: 123-456-7890
}
}
Point of Interest: Grand Canyon Village
POI Latitude: 36.0554
Company named: Dunder Mifflin
New email address: info@example.com
dog: {
	: Rex
	: 1
}
dog: {
	: Rex
	: 1
}
Dog Maui is 1 years old
Dog Maui is 2 years old
===============================================================================
All tests passed (20 assertions in 11 test cases)

Program ended with exit code: 0
```

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/examples/cpp/define-object-model.cpp
+++ b/examples/cpp/define-object-model.cpp
@@ -86,6 +86,7 @@ struct Company : realm::object {
 // :snippet-end:
                                                  
 TEST_CASE("create an embedded object", "[model][write]") {
+    // :snippet-start: create-embedded-object
     auto realm = realm::open<Business, ContactDetails>();
 
     auto business = Business { .name = "MongoDB" };
@@ -99,6 +100,8 @@ TEST_CASE("create an embedded object", "[model][write]") {
     realm.write([&realm, &business] {
         realm.add(business);
     });
+    // :snippet-end:
+
     auto businesses = realm.objects<Business>();
     auto mongoDBPointer = businesses[0];
     REQUIRE(businesses.size() >= 1);
@@ -126,6 +129,7 @@ TEST_CASE("create object with ignored property", "[model][write]") {
 };
 
 TEST_CASE("create object with to-one relationship", "[model][write][relationship]") {
+    // :snippet-start: create-object-with-to-one-relationship
     auto realm = realm::open<PointOfInterest, GPSCoordinates>();
 
     auto gpsCoordinates = GPSCoordinates { .latitude = 36.0554, .longitude = 112.1401 };
@@ -135,6 +139,7 @@ TEST_CASE("create object with to-one relationship", "[model][write][relationship
     realm.write([&realm, &pointOfInterest] {
         realm.add(pointOfInterest);
     });
+    // :snippet-end:
 
     auto pointsOfInterest = realm.objects<PointOfInterest>();
     auto namedGrandCanyonVillage = pointsOfInterest.where("name == $0", {"Grand Canyon Village"});
@@ -151,6 +156,7 @@ TEST_CASE("create object with to-one relationship", "[model][write][relationship
 };
 
 TEST_CASE("create object with to-many relationship", "[model][write][relationship]") {
+    // :snippet-start: create-object-with-to-many-relationship
     auto realm = realm::open<Company, Employee>();
 
     auto employee1 = Employee {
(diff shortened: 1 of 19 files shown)


examples/cpp/define-object-model.cpp
```diff
@@ -99,10 +100,16 @@ TEST_CASE("create an embedded object", "[model][write]") {
     realm.write([&realm, &business] {
         realm.add(business);
     });
+    // :snippet-end:
+
     auto businesses = realm.objects<Business>();
     auto mongoDBPointer = businesses[0];
     REQUIRE(businesses.size() >= 1);
     REQUIRE((*mongoDBPointer->contactDetails).emailAddress == "email@example.com");
+    // Clean up after the test
```
cbush, line 109: If the test fails, this won't run. I think we need Sections: https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#test-cases-and-sections

dacharyc: Good call - I've pushed an update that should isolate the failable bits in sections but leave the cleanup code to execute even when the test sections fail.

examples/cpp/define-object-model.cpp
```diff
@@ -99,10 +100,17 @@ TEST_CASE("create an embedded object", "[model][write]") {
...
+    // :snippet-end:
+    SECTION("Test code example functions as intended") {
+        auto businesses = realm.objects<Business>();
+        auto mongoDBPointer = businesses[0];
+        REQUIRE(businesses.size() >= 1);
+        REQUIRE((*mongoDBPointer->contactDetails).emailAddress == "email@example.com");
+    }
+    // Clean up after test
+    realm.write([&realm, &business] {
+        realm.remove(business);
```
cbush, line 112: Does C++ realm have a removeAll()? I would just use that - then you don't need the reference to the business object. You can open the realm both in the SECTION (for snippet purposes) and down here for cleanup.

examples/cpp/examples.cpp
```diff
@@ -65,6 +67,73 @@ TEST_CASE("first test case", "[test]") {
...
+    auto realm = realm::open<Dog>();
+
+    // Persist your data in a write transaction
+    realm.write([&realm, &dog] {
+        realm.add(dog);
+    });
+    // :snippet-end:
+    auto dogsCount = realm.objects<Dog>().size();
+    SECTION("Test code example functions as intended") {
+        REQUIRE(dogsCount >= 1);
```
cbush, line 93: It seems more of the test should actually be in the section, so that the cleanup can clean up at any point of failure.

cbush (approved):
Generally looks good. Would prefer to see the tests tightened up, but it's not a blocker.

source/sdk/cpp/crud/create.txt
```diff
@@ -0,0 +1,171 @@
...
+Create a New Object
+-------------------
+
+.. _cpp-create-a-new-object:
+
+Create an Object
+~~~~~~~~~~~~~~~~
+
+To add an object to a realm, instantiate it as you would any other
+object. Then pass it to the :cpp-sdk:`Realm.add() template
```
cbush, line 61: Minor: I would call these `template functions` or just `functions`.

examples/cpp/define-object-model.cpp
dacharyc: Alas, not yet. The sadness of a new SDK that doesn't yet have all the niceties I've gotten spoiled by in other SDKs!

examples/cpp/examples.cpp
dacharyc: Could you elaborate on what you would suggest including in the section here? I assumed that if the write failed, there would be nothing for the cleanup to clean up, so it was fine that neither of those be in the section. And I assume that the .size() method isn't something we need to expect a failure in. Am I assuming invalid things?

PR #2421, opened 2022-12-16
Fix copy-paste error causing ambiguous target

dacharyc:
Oops - I missed changing this and introduced an ambiguous ref target. This PR should fix any Snooty autobuilder failures resulting from this.


mongomoe (approved):
LGTM 👍

PR #2426, opened 2022-12-19
(DOCSP-27038): Flutter SDK throws error if you try to write one realm object to multiple realms

mongodben:
## Pull Request Info

Document that SDK throws error if you try to write one realm object to multiple realms.

### Jira

- https://jira.mongodb.org/browse/DOCSP-27038

### Staged Changes

- [Read & Write Data - Flutter SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-27038/sdk/flutter/read-and-write-data)

### Reminder Checklist

### Review Guidelines


source/sdk/flutter/read-and-write-data.txt
```diff
@@ -95,6 +95,10 @@ You can also return values from the write transaction callback function.
 .. literalinclude:: /examples/generated/flutter/read_write_data_test.snippet.return-from-write.dart
    :language: dart
 
+You can only write ``RealmObjects`` to a single realm. If a ``RealmObject`` is already
+written to one realm, the SDK throws a ``RealmException`` if you try to add it
```
mongomoe, line 99: Can we rephrase this without passive voice?

mongomoe: Also, would it be useful to show how to write similar objects to multiple realms? Even if it's just showing that the ObjectID is a different id in each object that you're adding.

mongomoe (commented):
1 bigger question, and a smaller wording suggestion. We could technically merge this as is, but depending on your answer to the question, it maybe better to convey this information differently/with an actual code example.

source/sdk/flutter/read-and-write-data.txt
mongomoe: Also, should this be a warning admonition since it's about an error?

mongodben: > Smaller suggestion: Can we rephrase this without passive voice? I.e, "is already written"

sure, did this

> Question: Would it be useful to show how to write similar objects to multiple realms? Even if it's just showing that the ObjectID is a different id in each object that you're adding.

i drafted a version of the text with examples, and it got quite long and i think was distracting from the core point of the page. i made it it's own section, but it really felt like overkill. you can see the draft here 54a1c91

> Also, should this be a warning admonition since it's about a possible error?

added as a warning. i think this strikes a good balance btwn calling out but not making it too central to the page. 

i'd initially thought about doing this text in an admonition, but went against it b/c i've been trying to do fewer admonitions on the Flutter docs. but this is the only admonition on the page, and it could cause an error as you note, so seems like fine usage to me.

mongomoe: Fair point about the drafted version distracting from the core point. This works for me.

mongomoe (approved):
LGTM 👍

PR #2427, opened 2022-12-19
(DOCSP-26117): Flutter RealmValue

mongodben:
## Pull Request Info

Flutter RealmValue documentation.

Note that the reference docs link won't work until the next release which'll include reference docs for this feature.

Note also that the description makes reference to the fact that RealmValue can't be a 'collection', w no further explanation of collections. There is future work for documenting collections, and i'll hyperlink to that here ([ticket](https://jira.mongodb.org/browse/DOCSP-27153)).

### Jira

- https://jira.mongodb.org/browse/DOCSP-26117

### Staged Changes

- [Data Types (Flutter) > 'RealmValue' section](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26117/sdk/flutter/model-data/data-types#realmvalue)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/flutter/model-data/data-types.txt
+++ b/source/sdk/flutter/model-data/data-types.txt
@@ -114,7 +114,7 @@ RealmValue
 The `RealmValue <https://pub.dev/documentation/realm_common/latest/realm_common/RealmValue-class.html>`__
 data type is a mixed data type that can represent any other valid Realm data type except a collection.
 You can create collections  of type ``RealmValue``, but a ``RealmValue`` itself
-cannot be a collection. ``RealmValue`` is also indexable.
+cannot be a collection. ``RealmValue`` is indexable, but cannot be a primary key.
 
 .. note::
 
--- a/source/sdk/flutter/model-data/data-types.txt
+++ b/source/sdk/flutter/model-data/data-types.txt
@@ -112,11 +112,9 @@ RealmValue
 ~~~~~~~~~~
 
 The `RealmValue <https://pub.dev/documentation/realm_common/latest/realm_common/RealmValue-class.html>`__
-data type is a mixed data type. ``RealmValue`` is indexable.
-
-``RealmValue`` can represent any other valid Realm data type except a collection.
+data type is a mixed data that can represent any other valid Realm data type except a collection.
 You can create collections  of type ``RealmValue``, but a ``RealmValue`` itself
-cannot be a collection.
+cannot be a collection. ``RealmValue`` is also indexable.
 
 .. note::
 
--- a/source/sdk/flutter/model-data/data-types.txt
+++ b/source/sdk/flutter/model-data/data-types.txt
@@ -112,7 +112,7 @@ RealmValue
 ~~~~~~~~~~
 
 The `RealmValue <https://pub.dev/documentation/realm_common/latest/realm_common/RealmValue-class.html>`__
-data type is a mixed data that can represent any other valid Realm data type except a collection.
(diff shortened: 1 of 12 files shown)


source/sdk/flutter/model-data/data-types.txt
```diff
@@ -102,6 +106,37 @@ such as ``Uuid.v4()``.
...
+
+.. literalinclude:: /examples/generated/flutter/data_types_test.snippet.realm-value-model.dart
+   :language: dart
+
+To add a ``RealmValue`` to a Realm object, call ``RealmValue.from()`` on the data.
+
+.. literalinclude:: /examples/generated/flutter/data_types_test.snippet.realm-value-from.dart
+   :language: dart
+
+Access the type of data with ``RealmValue.type`` and the value with ``RealmValue.value``.
```
dacharyc, line 135: The way this is written, it wasn't obvious to me this instruction was "Do _this_ then _that_." It kind of read to me like here are the two options for interacting with a `RealmValue` - you could check its type or you could access the value. I wonder if we want to be more explicit here - i.e. "you must check the type of the value using ``RealmValue.type`` and then access the value itself with the appropriate handling using ``RealmValue.value``?"

source/sdk/flutter/model-data/data-types.txt
```diff
@@ -102,6 +106,37 @@ such as ``Uuid.v4()``.
...
+   you can use the special ``RealmValue.nullValue()`` property.
+
+To define a property as ``RealmValue``, set its type in your Realm object model.
+
+.. literalinclude:: /examples/generated/flutter/data_types_test.snippet.realm-value-model.dart
+   :language: dart
+
+To add a ``RealmValue`` to a Realm object, call ``RealmValue.from()`` on the data.
+
+.. literalinclude:: /examples/generated/flutter/data_types_test.snippet.realm-value-from.dart
```
dacharyc, line 132: Maybe this is morning brain that hasn't soaked in enough coffee yet, but I wonder if we may want to tweak the naming a bit to make it more obvious what this code example is doing. Like, for the `anyValue` property, maybe we want to call that `singleAnyValue` and for the `mixedAnyValue` maybe it could be `listOfMixedAnyValues`. I know it's obvious if you look at the model, but when I was looking at the code example, my brain skipped right past `anyValue` and then I was looking at the map on `mixedAnyValue` and thinking that looked convoluted, and then looked back and forth at the model a few times and really parsed what was happening. I think we could make it clearer with some naming tweaks, even though that won't really be idiomatic naming.

Or better yet, maybe make it a real example. Like, if we used the dog model, age could be an int but could also be a spelled-out string, but it would be obvious that age is a single-value property. Maybe the list property could be something like favoriteToys, where some of them are just string names but others may be DogToy objects with a name and a brand? Just a thought.

source/sdk/flutter/model-data/data-types.txt
mongodben: it doesn't need to be this *then* that. that's just what's done in this example.

source/sdk/flutter/model-data/data-types.txt
mongodben: > want to call that singleAnyValue and for the mixedAnyValue maybe it could be listOfMixedAnyValues.

i like this approach. aligns more with other examples on the page.

desistefanova (commented):
Isn't it good to mention that RealmValue can not be annotated as a Primary Key?

PR #2419, opened 2022-12-16
(DOCSP-26733): Add email/password page for C++

krollins-mdb:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26733

### Staged Changes

- [Manage Users](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26733/sdk/cpp/users/manage-email-password-users/)

### Reminder Checklist

If your PR modifies the docs, you might need to also update some corresponding
pages. Check if completed or N/A.

- [ n/a ] Create Jira ticket for corresponding docs-app-services update(s), if any
- [ n/a ] Checked/updated Admin API
- [ n/a ] Checked/updated CLI reference

### Review Guidelines

[REVIEWING.md](https://github.com/mongodb/docs-realm/blob/master/REVIEWING.md)

### Animal in a hat

Beluga whales wearing santa hats

PR #2432, opened 2022-12-20
[C++] Make non-constants not capitalized

krollins-mdb:
## Pull Request Info

Super small fix for C++ that I didn't include in my [recent PR](https://github.com/mongodb/docs-realm/pull/2419).

### Review Guidelines

### DIFF
--- a/examples/cpp/authentication.cpp
+++ b/examples/cpp/authentication.cpp
@@ -10,10 +10,10 @@ TEST_CASE("create and log in a user", "[realm]") {
 
     auto app = realm::App(APP_ID);
 
-    std::string USER_EMAIL = "testUser@mongodb.com";
-    std::string USER_PASSWORD = "password1234";
+    std::string user_email = "testUser@mongodb.com";
+    std::string user_password = "password1234";
 
-    app.register_user(USER_EMAIL, USER_PASSWORD).get_future().get();
+    app.register_user(user_email, user_password).get_future().get();
     // :snippet-end:
 
     // :snippet-start: log-user-in
--- a/examples/cpp/authentication.cpp
+++ b/examples/cpp/authentication.cpp
@@ -17,7 +17,7 @@ TEST_CASE("create and log in a user", "[realm]") {
     // :snippet-end:
 
     // :snippet-start: log-user-in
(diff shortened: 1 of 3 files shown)


examples/cpp/authentication.cpp
```diff
@@ -10,14 +10,14 @@ TEST_CASE("create and log in a user", "[realm]") {
 
     auto app = realm::App(APP_ID);
 
-    std::string USER_EMAIL = "testUser@mongodb.com";
-    std::string USER_PASSWORD = "password1234";
+    std::string user_email = "testUser@mongodb.com";
```
dacharyc, line 13: Just a question - is there a reason we're using `std::string` here instead of `auto`? I recall Bush saying he's team AAA or "Almost Always Auto" and I wondered if there was a reason to use the specific `std::string` type here.

examples/cpp/authentication.cpp
```diff
@@ -10,14 +10,14 @@ TEST_CASE("create and log in a user", "[realm]") {
...
+    std::string user_password = "password1234";
```
dacharyc, line 14: Re: the naming here - you may have already talked to Bush about this, but he pointed out the inconsistency in variable naming in some of my examples. I was using snake case in some places because that's what the SDK used, but camel case in other places. Bush said camel case is perhaps more idiomatic or preferred, so I've been trying to use camel case in my examples. Should we pick a convention and unify on them in our examples?

dacharyc (approved):
This seems fine, so I think it's ok to merge. But I noted a couple of things we may want to consider and incorporate here, and/or in our other C++ SDK examples.

examples/cpp/authentication.cpp
krollins-mdb: Ah, just my lack of idiomatic knowledge. I opted for `std::string` in the example because it was artificially contrived, but `auto` makes sense. I'll update in my next commit.

examples/cpp/authentication.cpp
krollins-mdb: Ooh, thanks for pointing this out. I also used snake case because that's what's in the SDK. I'll update to camel case to be consistent with the rest of the example code.

PR #2449, opened 2022-12-30
Bluehawkify fast follow updates

krollins-mdb:
## Pull Request Info

This is a fast follow update to clean up work done in a [previous PR](https://github.com/mongodb/docs-realm/pull/2437).

### Staged Changes

- [Define a Realm Object Model](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/bluehawkify-fast-follow/sdk/node/examples/define-a-realm-object-model/)

### Review Guidelines

### DIFF
--- a/source/sdk/node/examples/define-a-realm-object-model.txt
+++ b/source/sdk/node/examples/define-a-realm-object-model.txt
@@ -10,31 +10,15 @@ Define a Realm Object Schema - Node.js SDK
    :depth: 2
    :class: singlecol
 
-.. _node-define-a-new-object-type:
-
 To define a Realm object type, create a schema object that specifies the type's
 ``name`` and ``properties``. The type name must be unique among object types in
 a realm. For details on how to define specific properties, see :ref:`Define
 Object Properties <node-define-a-property>`.
 
-.. code-block:: javascript
-   
-   const Car = {
-     name: "Car",
-     properties: {
-       make: "string",
-       model: "string",
-       miles: "int",
-     }
-   }
+It's best to define your schemas with JavaScript classes, but you can also define
+them as JavaScript objects.
 
-Open a Realm with a Realm Object
---------------------------------
-
-After defining a Realm object, you can :ref:`open a realm <node-open-a-local-realm>`
-with it.
-
-.. literalinclude:: /examples/generated/node/open-and-close-a-local-realm.snippet.open-local-realm-with-car-schema.js
+.. literalinclude:: /examples/generated/node/define-a-realm-object-schema.snippet.define-schema-as-object.js
    :language: javascript
 
 .. _node-define-objects-with-js-classes:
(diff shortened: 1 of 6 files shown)


source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -10,31 +10,15 @@ Define a Realm Object Schema - Node.js SDK
...
-       miles: "int",
-     }
-   }
-
-Open a Realm with a Realm Object
---------------------------------
-
-After defining a Realm object, you can :ref:`open a realm <node-open-a-local-realm>`
-with it.
+It's best to define your schemas with JavaScript classes, but you can also define
```
mongomoe, line 18: Suggestion: alter this sentence to either explain why it's best to us classes or don't include it at all. Right now the reader is left wondering "why is it best?", "what if I don't use JS Classes?", "are there any pitfalls to using schema objects"?

source/sdk/node/examples/define-a-realm-object-model.txt
```diff
@@ -46,14 +30,23 @@ You can define Realm object types with JavaScript classes. To use a class as an
...
    
 Pass the class itself to the schema property of the :js-sdk:`Realm.Configuration
 <Realm.html#~Configuration>` object when opening a realm. You can then :doc:`read and write data
 </sdk/node/examples/read-and-write-data>` normally.
 
 .. literalinclude:: /examples/generated/node/define-a-realm-object-schema.snippet.define-a-realm-object-schema-js-classes-open-and-access-properties.js
-     :language: javascript
+   :language: javascript
+
+Open a Realm with a Realm Object
```
mongomoe, line 42: Suggestion: I think we could get rid of this "Open a Realm with a Realm Object" subsection for the following reasons:

1. The next page in this TOC section is "Open & Close a Realm" which shows how to open a realm with a given schema. 
2. We already show how to open a realm in the subsection of the page above this "Open a Realm with a Realm Object" page. 
3. I don't think it really needs to be its own subsection given that this page is talking about defining a R.O.M, not opening a realm with a R.O.M.

mongomoe (approved):
LGTM with a couple suggestions. Also if these typos (where it describes a 'dog' schema, but the example has actually been changed to a 'manufacturer' or 'car' schema) are also in React Native, would you mind updating those typos as well or creating a ticket to update them that someone else can do?

krollins-mdb:
Thanks, @mongomoe! I appreciate the suggestions. 😄  

The React Native pages still have in-line examples - they're not Bluehawked. We'll be refactoring those as part of the realm reactify project, right?

source/sdk/node/examples/define-a-realm-object-model.txt
krollins-mdb: Ah, great point! I've reworded to avoid using "best". I updated to this:
> You can define your schemas with JavaScript classes (like most of the examples on 
this page), but you can also define them as JavaScript objects.

source/sdk/node/examples/define-a-realm-object-model.txt
krollins-mdb: Ah, yeah. I think removing this section makes a lot of sense. Thanks, Moe!

PR #2463, opened 2023-01-04
Add info and code about querying dictionaries

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-19341

### Staged Changes

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-19341/sdk/dotnet/model-data/data-types/dictionaries/#usage-example

### IMPORTANT
We never manually edit files in the /generated/ directory. Code changes happen in the /examples/ directory only.

### DIFF
--- a/examples/dotnet/Examples/DataTypesSectionExamples.cs
+++ b/examples/dotnet/Examples/DataTypesSectionExamples.cs
@@ -91,8 +91,8 @@ namespace Examples
 
             realm.Write(() =>
             {
-                realm.Add<Inventory>(storeInventory);
-                realm.Add<Inventory>(storeInventory2);
+                realm.Add(storeInventory);
+                realm.Add(storeInventory2);
             });
 
             // Find all Inventory items that have "Petunia"
@@ -117,8 +117,8 @@ namespace Examples
             var myStoreInventory = realm
                 .All<Inventory>().FirstOrDefault();
 
-            var petunia = myStoreInventory.Plants
-                .FirstOrDefault(p => p.Key == "Petunia");
+            var petunia = myStoreInventory.Plants.AsRealmQueryable()
+                .Where(p => p.Name == "Petunia");
             //:snippet-end:
 
             Assert.IsNotNull(petunias);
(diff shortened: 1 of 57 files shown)


source/sdk/dotnet/model-data/data-types/dictionaries.txt
```diff
@@ -46,12 +46,8 @@ The following code shows examples of dictionary types:
 
 Usage Example
 -------------
-The following code shows how to create, write to, and read from Dictionaries.
-
-.. note::
-
-   You cannot use Linq to query against a Realm Dictionary. Instead, 
-   use a string query.
+The following code shows how to create, write to, and read from Dictionaries,
```
mongomoe, line 49:
```suggestion
The following code shows how to create, write to, and read from Dictionaries
```

cbush (approved):

examples/dotnet/Examples/DataTypesSectionExamples.cs
```diff
@@ -96,24 +96,33 @@ public async Task WorkWithDictionaries()
             });
```
papafe, line 96: It's unnecessary to specify the type when using `realm.Add`.

examples/dotnet/Examples/DataTypesSectionExamples.cs
```diff
@@ -78,16 +78,16 @@ public async Task WorkWithDictionaries()
                 Id = ObjectId.GenerateNewId().ToString()
```
papafe, line 78: The ID is set, but that's not in the schema that is shown in the document

papafe:
The changes that have been added are not showing the correct thing. What we want to show is that dictionaries can be transformed into a queryable live collection. For example like in https://github.com/realm/realm-dotnet/blob/ef50e4ce7c2359e4dcdef5110e6aecafcdd2f7ce/Tests/Realm.Tests/Database/RealmDictionaryTests.cs#L1076-L1111

Practically one can have a live collection of the Values of the dictionary using `AsRealmQueryable()`. Something like

```csharp
myStoreInventory.Plants.AsRealmQueryable().Where( p => p.Name == "Geranium") //This is a live collection 

```

papafe (changes requested):

MongoCaleb:
Latest version hasn't been released due to Core issues, so the updated code won't compile until then.

MongoCaleb:
This PR will fail unit tests until the latest version of the SDK is released.

PR #2473, opened 2023-01-06
(DOCSP-26992): @realm/react-ify "Relationships - React Native SDK"

mongomoe:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26992

### Staged Changes

- [Relationships - React Native SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26992-relationships/sdk/react-native/realm-database/schemas/relationships/)

### Reminder Checklist

- [N/A] Create Jira ticket for corresponding docs-app-services update(s), if any
- [N/A] Checked/updated Admin API
- [N/A] Checked/updated CLI reference

### Review Guidelines


source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -64,130 +64,197 @@ There are three primary types of relationships between objects:
 .. _react-native-define-a-to-one-relationship-property:
```
nlarew, line 64: nit: Can we update `/includes/see-also-define-relationship-in-app-services-ui.rst` to not be specific to the UI? You can do it with CLI/API too, so we can just say "Alternatively, you can define them in your App Services app"

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -64,130 +64,197 @@ There are three primary types of relationships between objects:
...
-   creates several users and task objects, and the application pushes the newly created tasks into a 
-   user's task field. 
+--------------------
+
+An inverse relationship links an object back to any other objects that refer to
+it in a defined to-one or to-many relationship. Relationship definitions are
+unidirectional by default. 
+
+Inverse relationships can either be :ref:`defined explicitly <react-native-define-an-inverse-relationship-property>` in your schema or
+:ref:`obtained dynamically <react-native-dynamically-obtain-an-inverse-relationship-object>` using a query.
```
nlarew, line 157: Let's add a little guidance on when you'd wanna do each of these. Also (nit) this uses passive voice.

Maybe we flip the script here and start with the implicit backlinks:

> You can reference the inverse of any relationship from any RQL query using the `@links` keyword (link to RQL reference).

then talk about mapping them from an object property via `LinkingObjects`

> You can also explicitly assign an inverse relationship to an object as a `LinkingObjects` property that you access like a regular property.

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -64,130 +64,197 @@ There are three primary types of relationships between objects:
...
+
+In the example of the "User" and "Post" classes from the :ref:`to-many
+relationship example <react-native-to-many-relationship>`,  the "User has many
+Posts" does not automatically create the inverse relationship "Post belongs to
+User". 
+
+To create an inverse relationship, you can define it in your object model by
+specifying the ``linkingObjects`` property. The ``linkingObjects`` property
+specifies the object type and property name of the relationship that it links
+back to.
```
nlarew, line 173: "does not automatically create the inverse relationship" isn't quite right - there's always a backlink for every relationship.

Here I think it's more like "`Post` doesn't automatically get a `Post.owner` field"

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -64,130 +64,197 @@ There are three primary types of relationships between objects:
...
+.. _react-native-define-an-inverse-relationship-property:
+
+Explicitly Define an Inverse Relationship
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+You can explicitly define an inverse relationship in your schema. 
+
+In the example of the "User" and "Post" classes from the :ref:`to-many
+relationship example <react-native-to-many-relationship>`,  the "User has many
+Posts" does not automatically create the inverse relationship "Post belongs to
+User".
```
nlarew, line 168: I think we could get rid of the opening line + this first paragraph, which is a bit redundant with the example

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -64,130 +64,197 @@ There are three primary types of relationships between objects:
...
+An inverse relationship links an object back to any other objects that refer to
+it in a defined to-one or to-many relationship. Relationship definitions are
+unidirectional by default. 
+
+Inverse relationships can either be :ref:`defined explicitly <react-native-define-an-inverse-relationship-property>` in your schema or
+:ref:`obtained dynamically <react-native-dynamically-obtain-an-inverse-relationship-object>` using a query.
+
+.. _react-native-define-an-inverse-relationship-property:
+
+Explicitly Define an Inverse Relationship
```
nlarew, line 161: - Rather than "explicit" vs "dynamic" can we frame these sections as "Implicit Backlink" vs "LinkingObjects Property"
- Can we flip these two subsections to match the order in my comment above (i.e. implicit backlinks before linkingobjects properties)

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -64,130 +64,197 @@ There are three primary types of relationships between objects:
...
+         :linenos:
+
+.. _react-native-dynamically-obtain-an-inverse-relationship-object:
+
+Dynamically Obtain an Inversely Linked Object Using a Query
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To dynamically obtain an inversely linked object **without** altering the
+object's schema, call the :js-sdk:`Realm.Object.linkingObjects()
+<Realm.Object.html#linkingObjects>` method.
```
nlarew, line 223: You can also use `@links` in the query filter string - should we mention that here too?

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -64,130 +64,197 @@ There are three primary types of relationships between objects:
...
 
 To-One Relationship
-~~~~~~~~~~~~~~~~~~~
+-------------------
 
-A **to-one** relationship means that an object is related to no more than 
+A **to-one** relationship means an object is related to no more than 
 one other object in an :ref:`object schema <react-native-object-schemas>`. 
 To define a to-one relationship, specify the property type as the related
 Realm object type.
```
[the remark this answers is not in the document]
mongomoe: To clarify, are you saying that we should specify that the relationship can be optional? Like in the Petowner and Pet example, a Petowner object doesn't have to have a Pet object?

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -64,130 +64,197 @@ There are three primary types of relationships between objects:
...
-   
-   In the following continuation from the inverse relationship example, we have removed the
-   ``assignee`` field with type 'linkingObjects' from the Task schema. An application developer
-   creates several users and task objects, and the application pushes the newly created tasks into a 
-   user's task field. 
+--------------------
+
+An inverse relationship links an object back to any other objects that refer to
+it in a defined to-one or to-many relationship. Relationship definitions are
+unidirectional by default.
```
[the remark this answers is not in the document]
mongomoe: I like that example sentence you gave, however I altered it slightly to correlate with the User and Post example that's actually used in our inverse example:

```
For example, if you define a relationship that maps ``User.posts`` to a list of
``Post`` objects, then you could use an inverse relationship to look up the
``User`` object related to a given ``Post`` object in this way.
```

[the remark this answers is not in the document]
mongomoe: Good idea about referencing the word backlinks, just added that to the intro sentence:
```
An inverse relationship :ref:`backlinks <rql-backlinks>` to any other objects
that refer to it in a defined to-one or to-many relationship. Relationship
definitions are unidirectional by default. 
```

takameyer (approved):
Just a few small things.  Other than that, LGTM 💯

source/sdk/react-native/realm-database/schemas/relationships.txt
mongomoe: Feels odd to include a mention to @links without an example, and the other SDKs don't mention the @links syntax on their respective relationships page, so I'd prefer not to. We could maybe link out to https://www.mongodb.com/docs/realm/realm-query-language/#backlink-queries but I find those examples kind of confusing.

source/sdk/react-native/realm-database/schemas/relationships.txt
mongomoe: Altered this bit from its original variation, but I did use the order you suggested (however, I don't talk about the `@links` syntax in this doc at all (and kind of want to keep it that way, instead the way to obtain a reference to an inversely linked object is to use the 'linkingObjects()' method).

```
Inverse relationships can be:

#. :ref:`Obtained dynamically <react-native-dynamically-obtain-an-inverse-relationship-object>` using the :js-sdk:`Realm.Object.linkingObjects() <Realm.Object.html#linkingObjects>` method.
#. :ref:`Defined explicitly <react-native-define-an-inverse-relationship-property>` in your schema
```

source/sdk/react-native/realm-database/schemas/relationships.txt
mongomoe: Got rid of this paragraph based on your [suggestion](https://github.com/mongodb/docs-realm/pull/2473#discussion_r1066322473).

source/sdk/react-native/realm-database/schemas/relationships.txt
mongomoe: I swapped the two subsections and framed them as "Obtain an Implicitely Backlinked Object" vs "Define an Inverse Relationship Explicitely in a Schema"

source/sdk/react-native/realm-database/schemas/relationships.txt
mongomoe: Decided against this, we link out to the backlinks docs (see: https://docs/mongodb.com/realm/realm-query-language/#backlink-queries) in the beginning of the "Inverse Relationship" section, but I don't find the @links syntax particularly user friendly, so if this isn't a requirement, I'd prefer to leave it out.

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -63,131 +63,198 @@ There are three primary types of relationships between objects:
...
-To-One Relationship
-~~~~~~~~~~~~~~~~~~~
+One-to-One Relationship
+-----------------------
 
-A **to-one** relationship means that an object is related to no more than 
-one other object in an :ref:`object schema <react-native-object-schemas>`. 
+A **one-to-one** relationship means an object is related to at most 
+one other object of a particular type. 
 To define a to-one relationship, specify the property type as the related
```
takameyer, line 71:
```suggestion
To define a one-to-one relationship, specify the property type as the related
```

source/sdk/react-native/realm-database/schemas/relationships.txt
```diff
@@ -63,131 +63,198 @@ There are three primary types of relationships between objects:
...
+definitions are unidirectional by default. 
+
+For example, if you define a relationship that maps ``User.posts`` to a list of
+``Post`` objects, then you could use an inverse relationship to look up the
+``User`` object related to a given ``Post`` object in this way.
+
+To use inverse relationships, you can: 
+
+#. :ref:`Obtain an implicitly backlinked object <react-native-obtain-an-implicitly-backlinked-object>` using the :js-sdk:`Realm.Object.linkingObjects() <Realm.Object.html#linkingObjects>` method.
+#. :ref:`Define an inverse relationship explicitly in a schema <react-native-define-an-inverse-relationship-property>` using the ``LinkingObjects`` property.
```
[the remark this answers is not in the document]
nlarew: I see you decided against this - see my follow up https://github.com/mongodb/docs-realm/pull/2473#discussion_r1082809020

examples/react-native/__tests__/js/realm-database/schemas/relationships-test.jsx
```diff
@@ -0,0 +1,168 @@
...
+  beforeEach(async () => {
+    // we will use this Realm for assertions to access Realm Objects outside of a Functional Component (like required by @realm/react)
+    assertionRealm = await Realm.open(realmConfig);
+
+    // delete every object in the realmConfig in the Realm to make test idempotent
+    assertionRealm.write(() => {
+      assertionRealm.delete(assertionRealm.objects(User));
+      assertionRealm.delete(assertionRealm.objects(Post));
+
+      const user1 = new User(assertionRealm, {
```
takameyer, line 29: We have recently decided to back track against using the constructor to create objects.  This has caused some problems with embedded objects. https://github.com/realm/realm-js/issues/5157

Therefore we should convert these examples to `const user1 = assertionRealm.create(User, {_id: new Realm.BSON.ObjectId(), etc...`

nlarew: Looks like we're not actually using the Babel plugin yet but the class still had `Realm.Object<User>` - updating it fixes this.

nlarew: @takameyer Ok got this working but I needed to update the TS models by defining `RequiredProperties`, the second type argument to `Realm.Object`.

Is it supposed to be a requirement that you define a union of required field names like `"_id" | "name"`?

```
class User extends Realm.Object<User, "_id" | "name"> {
  _id!: Realm.BSON.ObjectId;
  name!: string;
  birthdate?: Date;
  posts!: Realm.List<Post>;

  static schema = {
    name: 'User',
    properties: {
      _id: 'objectId',
      name: 'string',
      birthdate: 'date?',
      posts: 'Post[]',
    },
    primaryKey: '_id',
  };
}
```

takameyer: @nlarew Hmm...that's a bit non-ergonomic.  @kraenhansen Do you have an insights into this?  I'm trying to find out what the best recommendation is until we get things cleared up with class based models.

kraenhansen: Yeah. I believe this was the compromise Tom reached when finalizing the feature. The alternative was that users had to pass in a list of all optional property names. I agree it's not as elegant as we would have liked it to be 🙄

PR #2405, opened 2022-12-14
Kotlin Sync Feedback Document changes

edualonso:
Incorporated changes from https://docs.google.com/document/d/1M2QoA-auyOaWZ88PeL1_6QylcEkqt3Aqv3ppOvuuGtI/edit#heading=h.watv1kfii35

PR #2479, opened 2023-01-10
(DOCSP-27332): Regenerating Kotlin Open a Realm code

cbullinger:
## Pull Request Info
Regenerated the `open-a-realm` snippet and added small blurb about using `RealmConfiguration.create()` (per [Kotlin Sync Feedback Document changes](https://github.com/mongodb/docs-realm/pull/2405#discussion_r1049208862) PR)

### Jira

- https://jira.mongodb.org/browse/DOCSP-27332

### Staged Changes

- [Kotlin Quick Start: Open a Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-27332-open-realm-fix/sdk/kotlin/quick-start/#open-a-realm)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/kotlin/quick-start.txt
+++ b/source/sdk/kotlin/quick-start.txt
@@ -61,6 +61,12 @@ to generate an instance of that realm:
    :language: kotlin
    :copyable: false
 
+You can open a realm using the
+`RealmConfiguration.create() <{+kotlin-local-prefix+}io.realm.kotlin/-realm-configuration/-companion/create.html>`__
+function if no configuration parameters are passed. Only default values are used to open the realm. 
+Otherwise, you must build the configuration
+using `RealmConfiguration.builder() <{+kotlin-local-prefix+}io.realm.kotlin/-realm-configuration/-builder/index.html>`__.  
+
 Create, Read, Update, and Delete Objects
 ----------------------------------------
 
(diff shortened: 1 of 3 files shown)


source/sdk/kotlin/quick-start.txt
```diff
@@ -61,6 +61,10 @@ to generate an instance of that realm:
    :language: kotlin
    :copyable: false
 
+To open a realm using only default properties, you can use the
```
dacharyc, line 64: Might it be worth saying "instead" somewhere in here - i.e. is this an alternative to using `Realm.Configuration.Builder()` that we show in the code example?

Alternately, do we want to show `Realm.Configuration.create()` _instead_ of `.Builder()` here? i.e. are our reasons for using `.Builder()` here related to something internal to the test, but we would recommend that new users following a quick start just use `.create()`? If so, I wonder if we'd want to use Bluehawk replace to show this as `.create()` in the generated code snippet while using `.Builder()` in the test?

dacharyc (approved):
I think this is technically fine as-is, but I wonder if we want to make some tweaks to capture some details/make some things clearer to the reader?

source/sdk/kotlin/quick-start.txt
cbullinger: 👍 on the "instead". I'll reword. 

> are our reasons for using .Builder() here related to something internal to the test,  but we would recommend that new users following a quick start just use .create()?

Yes. I played around with the test code to get a better code snippet and went with the text blurb instead. Buuuut...since you pretty much repeated my initial thoughts, I'll take another pass and generate a more appropriate code snippet for the quick start.

PR #2484, opened 2023-01-11
(DOCSP-27219): (Swift SDK) Add a Write to a Synced Realm page

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-27219

### Staged Changes

- [Write to a Synced Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-27219/sdk/swift/sync/write-to-synced-realm/): New page

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/swift/sync/write-to-synced-realm.txt
+++ b/source/sdk/swift/sync/write-to-synced-realm.txt
@@ -7,7 +7,7 @@ Write Data to a Synced Realm - Swift SDK
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 2
+   :depth: 3
    :class: singlecol
 
 Overview
@@ -39,9 +39,11 @@ and other Device Sync error types, refer to :ref:`sync-errors` in the App Servic
 Determining What Data Syncs
 ---------------------------
 
-The examples on this page use an Atlas App Services App with the following 
-Device Sync configuration and a client app with the following Realm SDK 
-data model and subscriptions.
+The data that you can write to a synced realm is the intersection of your 
+Device Sync configuration - your queryable fields and permissions - and the
+Flexible Sync subscription query that you use when you open the realm.
+
+The examples on this page use the following configurations and models:
 
 Device Sync Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -104,44 +106,15 @@ value is greater than ``4``, cannot sync to this realm.
 Write to a Synced Realm
 -----------------------
 
-When you write to a synced realm, these are the most common cases:
+Writes to Flexible Sync realms may broadly fall into one of two categories: 
 
 - **Successful writes**: The written object matches both the query subscription
   and the user's permissions. The object writes successfully to the realm,
   and syncs successfully to the App Services backend and other devices.
-- **Writes that don't match the query subscription**: The written object matches 
-  the user's permissions, but does not match the query subscription. The
-  object writes to the realm, but when the realm syncs to the backend, it
-  reverts the write in a non-fatal error operation we call a compensating 
-  write.
-- **Writes that don't match permissions**: The written object matches the query
-  subscription, but does not match the user's permissions. The
-  object writes to the realm, but when the realm syncs to the backend, it
-  reverts the write in a non-fatal error operation we call a compensating 
-  write.
 
-If you perform a write that does not match the subscription query, or where
-the user does not have sufficient permissions to perform the write, Realm 
-performs a compensating write. This is a non-fatal operation that 
-essentially reverts an illegal write.
-
-In more detail, when you write data that is outside the bounds of a query
-subscription or does not match the user's permissions, the following occurs:
-
-#. Because the client realm has no concept of "illegal" writes,
-   the write initially succeeds until realm resolves the changeset
-   with the App Services backend.
-#. Upon sync, the server applies the rules and permissions.
-   The server determines that the user does not have authorization to perform the write.
-#. The server sends a revert operation, called a "compensating write", back to the client.
-#. The client's realm reverts the illegal write operation.
-
-Any client-side writes to a given object between an illegal write to that object
-and the corresponding compensating write will be lost.
-
-In practice, this may "look like" an object being written to the realm, and
-then disappearing after the server sends the compensating write back to 
-the client.
+- **Compensating writes**: When the written object does not match
+  the subscription query, or where the user does not have sufficient 
+  permissions to perform the write, Realm reverts the illegal write.
 
 .. _swift-successful-writes:
 
@@ -159,8 +132,42 @@ with the App Services backend when the device has a network connection.
 
 .. _swift-compensating-writes:
 
+Compensating Writes
+~~~~~~~~~~~~~~~~~~~
+
+In some cases, a write that initially appears to succeed is actually an 
+illegal write. In these cases, the object writes to the database, but when 
+the database syncs to the backend, Realm reverts the write in a non-fatal 
+error operation called a compensating write. Compensating writes can occur
+when:
+
+- **Writes that don't match the query subscription**: The written object matches 
+  the user's permissions, but does not match the query subscription. 
+- **Writes that don't match permissions**: The written object matches the query
+  subscription, but does not match the user's permissions.
+
+In more detail, when you write data that is outside the bounds of a query
+subscription or does not match the user's permissions, the following occurs:
+
+#. Because the client realm has no concept of "illegal" writes,
+   the write initially succeeds until realm resolves the changeset
+   with the App Services backend.
+#. Upon sync, the server applies the rules and permissions.
+   The server determines that the user does not have authorization to perform the write.
+#. The server sends a revert operation, called a "compensating write", back to the client.
+#. The client's realm reverts the illegal write operation.
+
+Any client-side writes to a given object between an illegal write to that object
+and the corresponding compensating write will be lost.
+
+In practice, this may look like an object being written to the realm, and
+then disappearing after the server sends the compensating write back to 
+the client.
+
+.. _swift-writes-outside-subscription:
+
 Writes that Don't Match the Query Subscription
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``````````````````````````````````````````````
 
 You can only write objects to a Flexible Sync realm if they match the 
 subscription query. If you perform a write that does not match the 
@@ -168,11 +175,19 @@ subscription query, Realm initially writes the object, but then performs
 a compensating write. This is a non-fatal operation that 
 reverts an illegal write that does not match the subscription query.
 
-In practice, this may look like the write succeeding, but then the object
-"disappears" when Realm performs the compensating write.
+In practice, this may look like the write succeeding, but then the 
+object "disappears" when Realm syncs with the App Services backend and 
+performs the compensating write.
+
+If you want to write an object that does not match the query subscription,
+you must open a different realm where the object matches the query 
+subscription. Alternately, you could write the object to a :ref:`non-synced 
+realm <swift-synced-realms-vs-non-synced-realms>` that does not enforce 
+permissions or subscription queries. There is no way to write an object 
+to a Flexible Sync realm that does not match the query subscription.
 
 Code Example
-````````````
+++++++++++++
 
 Given the configuration for the Flexible Sync realm above, attempting to 
 write this object does not match the query subscription:
@@ -181,7 +196,7 @@ write this object does not match the query subscription:
    :language: swift
 
 Client Error
-````````````
+++++++++++++
 
 The error message in the client-side logs in this scenario is: 
 
@@ -192,7 +207,7 @@ The error message in the client-side logs in this scenario is:
    reverted" (error_code=231, try_again=true, error_action=Warning)
 
 App Services Error
-``````````````````
+++++++++++++++++++
 
 The error message in the App Services logs in this scenario is:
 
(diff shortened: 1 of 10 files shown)


source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,256 @@
...
+The examples on this page use an Atlas App Services App with the following 
+Device Sync configuration and a client app with the following Realm SDK 
+data model and subscriptions.
+
+Device Sync Configuration
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Device Sync is configured with the following queryable fields:
+
+- ``_id`` (this is always a queryable field in every Device Sync configuration)
```
mongodben, line 51:
```suggestion
- ``_id`` (this is a queryable field in every Device Sync configuration)
```

soft nit: felt a bit redundant to have 'always' and 'every'. could also **bold** 'every' for more emphasis

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,256 @@
...
+   with the App Services backend.
+#. Upon sync, the server applies the rules and permissions.
+   The server determines that the user does not have authorization to perform the write.
+#. The server sends a revert operation, called a "compensating write", back to the client.
+#. The client's realm reverts the illegal write operation.
+
+Any client-side writes to a given object between an illegal write to that object
+and the corresponding compensating write will be lost.
+
+In practice, this may "look like" an object being written to the realm, and
```
mongodben, line 142: why 'look like' in quotes? isn't this literally what's happening?

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,256 @@
...
+         },
+         "write": {
+           "ownerId": "%%user.id"
+         }
+       }
+     ]
+   }
+
+Client Data Model and Configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
mongodben, line 77:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,256 @@
...
+      write filter in the role.
+    - To learn more about configuring Device Sync permissions for your app,
+      refer to :ref:`sync-rules` and the :ref:`flexible-sync-permissions-guide`
+      in the App Services documentation.
+
+To learn more about permission denied errors, compensating write errors
+and other Device Sync error types, refer to :ref:`sync-errors` in the App Services documentation.
+
+Determining What Data Syncs
+---------------------------
```
mongodben, line 40: I think it'd be clearer to call this smthn like 'Examples on this Page' to make clear that this is just an example config and explanation of it.

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,256 @@
...
+
+- The ``ownerId`` matches the ``user.id`` of the logged-in user (from the permissions)
+- The ``complexity`` property's value is less than or equal to ``4`` (from the subscription query)
+
+Any object in the Atlas collection where the ``ownerId`` does not match 
+the ``user.id`` of the logged-in user, or the ``complexity`` property's 
+value is greater than ``4``, cannot sync to this realm.
+
+Write to a Synced Realm
+-----------------------
```
mongodben, line 105: i think the subsections here could use a bit of refactoring w/o necessary making copy changes. 

the core of the confusion for me while reading this came from two things:
- the long description of the writes failing process before discussion of successful writes or the ways that the writes fail below.
- lack of conceptual description of 'reverted write', which can be caused by the permission denied or writing outside the query subscription.

for me, more logical structure here would be (pardon the Markdown but i think it makes the point more clearly about the headers):

````md
## Write to a Synced Realm

When you write to a synced realm, these are the most common cases:

- **Successful writes**: The written object matches both the query subscription
  and the user's permissions. The object writes successfully to the realm,
  and syncs successfully to the App Services backend and other devices.
- **Reverted writes**: If you perform a write that does not match
  the subscription query, or where the  user does not have sufficient permissions to 
  perform the write, Realm reverts the illegal write.

### Successful Writes

*keep content in this section the same*

### Reverted Writes

If you perform a write that does not match the subscription query, or where
the user does not have sufficient permissions to perform the write, Realm 
performs a compensating write. This is a non-fatal operation that 
essentially reverts an illegal write. The two most common reasons for reverted writes are the following: 

- **Writes that don't match the query subscription**: The written object matches 
  the user's permissions, but does not match the query subscription. The
  object writes to the realm, but when the realm syncs to the backend, it
  reverts the write in a non-fatal error operation we call a compensating 
  write.
- **Writes that don't match permissions**: The written object matches the query
  subscription, but does not match the user's permissions. The
  object writes to the realm, but when the realm syncs to the backend, it
  reverts the write in a non-fatal error operation we call a compensating 
  write.

In more detail, when you write data that is outside the bounds of a query
subscription or does not match the user's permissions, the following occurs:

#. Because the client realm has no concept of "illegal" writes,
   the write initially succeeds until realm resolves the changeset
   with the App Services backend.
#. Upon sync, the server applies the rules and permissions.
   The server determines that the user does not have authorization to perform the write.
#. The server sends a revert operation, called a "compensating write", back to the client.
#. The client's realm reverts the illegal write operation.

Any client-side writes to a given object between an illegal write to that object
and the corresponding compensating write will be lost.

In practice, this may "look like" an object being written to the realm, and
then disappearing after the server sends the compensating write back to 
the client.

#### Writes that Don't Match the Query Subscription

*keep content in this section the same*

#### Writes That Don't Match Permissions

*keep content in this section the same*


````

File: source/sdk/swift/sync/write-to-synced-realm.txt [hunk 6]
```diff
@@ -0,0 +1,256 @@
...
+.. _swift-compensating-writes:
+
+Writes that Don't Match the Query Subscription
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can only write objects to a Flexible Sync realm if they match the 
+subscription query. If you perform a write that does not match the 
+subscription query, Realm initially writes the object, but then performs 
+a compensating write. This is a non-fatal operation that essentially 
+reverts an illegal write that does not match the subscription query.
```
mongodben, line 169:
```suggestion
a compensating write. This is a non-fatal operation that 
reverts an illegal write that does not match the subscription query.
```

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,256 @@
...
+
+In practice, this may look like the write succeeding, but then the object
+"disappearing" when Realm performs the compensating write.
```
mongodben, line 172:
```suggestion
In practice, this may look like the write succeeding, but then the object
"disappears" when Realm performs the compensating write.
```

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,256 @@
...
+
+Writes That Don't Match Permissions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Attempting to write to the client can also trigger a compensating write 
+error when the object does not match the user's write permissions. 
+
+On the client, this type of write behaves the same as a write that doesn't 
+match the query subscription. In practice, this may look like the write 
+succeeding, but then the object "disappearing" when Realm syncs with the
```
mongodben, line 217:
```suggestion
succeeding, but then the object "disappears" when Realm syncs with the 
```

source/sdk/swift/sync/write-to-synced-realm.txt
mongodben: also i liked your phrasing on [lines 217-218](https://github.com/mongodb/docs-realm/pull/2484/files#diff-e2d8c506c5da4619322a5fb7fec7dc60573f5642ba09573913f868f317dabb1fR217-R218) about App Services initiating the the compensating write. i think it clearly conveyed the idea that the App Services sync server triggers the compensating write. could you add that here as well?

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,256 @@
...
+   write that is outside of permissions or query filters; it has been 
+   reverted" (error_code=231, try_again=true, error_action=Warning)
+
+App Services Error
+``````````````````
+
+The error message in the App Services logs provides some additional 
+information to help you determine that it is a permissions issue - in this 
+case the object does not match the user's role - and not a query subscription 
+issue:
```
mongodben, line 248:
```suggestion
The error message in the App Services logs provides some additional 
information to help you determine that it is a permissions issue,
and not a query subscription issue. In this example, the error message shows
that the the object does not match the user's role:
```

nit: break up sentence for readability.

source/sdk/swift/sync/write-to-synced-realm.txt
dacharyc: I went with quotes for a couple of reasons which in retrospect don't really seem to matter so much. Removing the quotes.

source/sdk/swift/sync/write-to-synced-realm.txt
dacharyc: I renamed this away from "Examples on this Page" because I wanted it to be more about you as a developer understanding what data syncs, and why. And to emphasize that it's because of these two components - the Device Sync Configuration w/permissions and the client model/subscription query. It's really the intersection of these things that's important.

So I think the fact that these are examples that support the code on the page is secondary; I think the important aspect is that the intersection of these two things determines what data syncs.

That said, I kept the intro paragraph from when this was more focused on the "Examples on this Page" title so I could tweak that a bit. I'll rework this a bit and then I'd love to hear whether that makes sense or if you still think it's unclear or otherwise bad naming.

source/sdk/swift/sync/write-to-synced-realm.txt
dacharyc: I see what you mean here. I've taken your suggestion and restructured the page a bit. I didn't go with exactly what you suggested here, but it's close - just some minor differences. Curious what you think now.

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,271 @@
...
+In more detail, when you write data that is outside the bounds of a query
+subscription or does not match the user's permissions, the following occurs:
+
+#. Because the client realm has no concept of "illegal" writes,
+   the write initially succeeds until realm resolves the changeset
+   with the App Services backend.
+#. Upon sync, the server applies the rules and permissions.
+   The server determines that the user does not have authorization to perform the write.
+#. The server sends a revert operation, called a "compensating write", back to the client.
+#. The client's realm reverts the illegal write operation.
```
mongodben, line 158: i think it could be helpful if in this sentence it makes clear that "reverting illegal write" basically means "delete the illegal object from the realm". so perhaps rephrase sentence to smthn like: 

> The client's realm reverts the illegal write operation by deleting the illegal object from the realm.

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,271 @@
...
+       the current query view"
+   }
+
+.. _swift-writes-outside-permissions:
+
+Writes That Don't Match Permissions
+```````````````````````````````````
+
+Attempting to write to the client can also trigger a compensating write 
+error when the object does not match the user's write permissions.
```
mongodben, line 228:
```suggestion
error when the object does not match the user's server-side write permissions. 
```

nit: clarity where the permissions are

source/sdk/swift/sync/write-to-synced-realm.txt
mongodben: great changes! 

i think i previously had a slightly off mental model, where i was incorrectly thinking that only writes that don't match the query subscription are compensating writes. and then permission denied reverts are a separate but similar thing. 

this clarifies things for me (and hopefully our users too 😄). will soon update the flutter page w this in mind.

source/sdk/swift/sync/write-to-synced-realm.txt
mongodben: great changes! 

i think i previously had a slightly off mental model, where i was incorrectly thinking that only writes that don't match the query subscription are compensating writes. and then permission denied reverts are a separate but similar thing. 

this clarifies things for me (and hopefully our users too 😄). will soon update the flutter page w this in mind.

mongodben (approved):
lgtm! couple small things here and there to address noted in comments.

i think this is a really important and clear page that makes working w flexible sync much easier. i think all sdks w flex sync should have this.

source/sdk/swift/sync/write-to-synced-realm.txt
```diff
@@ -0,0 +1,271 @@
...
+      the write reverts with a non-fatal permission denied error.
+      In the client, this shows as an (ErrorCompensatingWrite). On the server,
+      you can see more details about how the write was denied was by a 
+      write filter in the role.
+    - To learn more about configuring Device Sync permissions for your app,
+      refer to :ref:`sync-rules` and the :ref:`flexible-sync-permissions-guide`
+      in the App Services documentation.
+
+To learn more about permission denied errors, compensating write errors
+and other Device Sync error types, refer to :ref:`sync-errors` in the App Services documentation.
```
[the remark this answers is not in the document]
dacharyc: Hmm, good point that it doesn't fit well here anymore - I actually think it fits better in the "Compensating Writes" section (ln 164 ish) so moving it directly into there.

PR #2488, opened 2023-01-12
Add find-unused action

cbush:


source/index.txt
```diff
@@ -1,3 +1,5 @@
+.. Don't commit this
+
```
mongodben, line 2: assuming this shouldn't be here

.github/workflows/find-unused.yml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+    paths:
+      - "source/**"
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: cbush/docdoctor/actions/find-unused@find-unused-action
```
mongodben, line 13: where is the source code for this GH action?

mongodben: follow up question: will the above PR be merged prior to this one?

source/index.txt
cbush: Correct, but it's required to trigger the workflow. I'll remove it before merge.

mongodben (commented):
lgtm

---

out of curiosity, how'd you test this out? i always struggle w testing for GH actions.

cbush:
The trick is to push it to an upstream branch.

mongodben (approved):

PR #2518, opened 2023-01-25
Update to Kotlin 1.6.0

cmelchior:
Update the documentation version to latest Kotlin release, 1.6.0

For reviewers: I don't know if there is any other places I need to update. So I might have forgotten something


cbullinger (approved):
lgtm (And as far as I know, this should be the only file we need to update)

PR #2556, opened 2023-02-03
(DOCSP-26755): Kotlin Embedded Objects page

cbullinger:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26755

### Staged Changes

- [Embedded Objects - Kotlin SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-26755-linking-obj/sdk/kotlin/realm-database/schemas/embedded-objects/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/DataTypesTest.kt
+++ b/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/DataTypesTest.kt
@@ -18,7 +18,7 @@ import kotlin.test.Test
 import kotlin.test.assertEquals
 
 // :snippet-start: embedded-object-model
-// Define an embedded object
+// Define an embedded object (cannot have primary key)
 class Address() : EmbeddedRealmObject {
     var street: String? = null
     var city: String? = null
@@ -81,14 +81,22 @@ class DataTypesTest : RealmTest() {
                 val address: Address = realm.query<Address>().find().first()
                 val contact: Contact = realm.query<Contact>().find().first()
                 // :snippet-start: update-embedded-object
-                // Modify the property of the embedded object in a write transaction
+                // Modify embedded object properties in a write transaction
                 realm.write {
                     // Fetch the objects
                     val addressToUpdate = findLatest(address)?: error("Cannot find latest version of embedded object")
                     val contactToUpdate = findLatest(contact)?: error("Cannot find latest version of parent object")
 
-                    // Update the embedded object property directly
+                    // Update a single embedded object property directly
                     addressToUpdate.street = "100 10th St N"
+
+                    // Update multiple properties
+                    addressToUpdate.apply {
+                        street = "202 Coconut Court"
+                        city = "Los Angeles"
+                        state = "CA"
+                        postalCode = "90210"
+                    }
 
                     // Update property through the parent object
                     contactToUpdate.address?.state = "NY"
@@ -98,12 +106,12 @@ class DataTypesTest : RealmTest() {
             // :snippet-start: overwrite-embedded-object
             // Overwrite the embedded object in a write transaction
             realm.write {
-                // Fetch the object
-                val addressToOverwrite: Address =
-                    realm.query<Address>("street == '123 Fake St'").find().first()
+                // Fetch the parent object
+                val parentObject: Contact =
+                    realm.query<Contact>("name == 'Nick Riviera'").find().first()
 
-                // Overwrite the embedded object (updates the existing object)
-                addressToOverwrite.apply {
+                // Overwrite the embedded object (deletes the existing object)
+                parentObject.address = Address().apply {
                     street = "202 Coconut Court"
                     city = "Los Angeles"
                     state = "CA"
@@ -209,7 +217,7 @@ class DataTypesTest : RealmTest() {
                 val propertyToClear: Contact =
                     this.query<Contact>("name == 'Nick Riviera'").find().first()
 
-                // Clear the parent property (deletes the embedded class)
+                // Clear the parent property (deletes the embedded object instance)
                 propertyToClear.address = null
             }
 
--- a/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/DataTypesTest.kt
+++ b/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/DataTypesTest.kt
@@ -11,8 +11,6 @@ import io.realm.kotlin.types.EmbeddedRealmObject
 import io.realm.kotlin.types.RealmList
 import io.realm.kotlin.types.RealmObject
 import io.realm.kotlin.types.annotations.PrimaryKey
-import kotlinx.coroutines.Deferred
-import kotlinx.coroutines.async
 import org.mongodb.kbson.ObjectId
 import kotlin.test.Test
 import kotlin.test.assertEquals
@@ -51,7 +49,7 @@ class DataTypesTest : RealmTest() {
     fun createEmbeddedObject() {
         runBlocking {
             // :snippet-start: open-realm-embedded-object
-            // Include parent and embedded objects in schema
+            // Include parent and embedded object classes in schema
             val config = RealmConfiguration.Builder(
                 setOf(Contact::class, Address::class)
             )
@@ -81,18 +79,23 @@ class DataTypesTest : RealmTest() {
             // :snippet-end:
 
             // :snippet-start: update-embedded-object
-            // ... Fetch the parent object
+            // ... Fetch the object
 
+            // Modify the property of the embedded object in a write transaction
             realm.write {
-                // Update the embedded object directly through the contact
-                nickRiviera.address?.street = "100 10th St N"
+
+                // Update the embedded object property directly
+                nicksAddress.street = "100 10th St N"
+
+                // Update property through the parent object
+                nickRiviera.address?.state = "NY"
             }
             // :snippet-end:
 
             // :snippet-start: overwrite-embedded-object
-            // ... Fetch the parent object
+            // ... Fetch the object
 
-            // Overwrite the embedded object
+            // Overwrite the embedded object (deletes the original object)
             realm.write {
                 nicksAddress.apply {
                     street = "202 Coconut Court"
@@ -101,14 +104,6 @@ class DataTypesTest : RealmTest() {
                     postalCode = "90210"
                 }
             }
-            // :snippet-end:
-
-            // :snippet-start: query-embedded-objects
-            // Access the embedded object through the parent object
-            val contactsInNY: RealmResults<Contact> =
-                realm.query<Contact>("address.state == 'NY'")
-                    .sort("name")
-                    .find()
             // :snippet-end:
 
             realm.close()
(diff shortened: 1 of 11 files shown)


source/examples/generated/kotlin/DataTypesTest.snippet.create-embedded-object.kt
```diff
@@ -0,0 +1,18 @@
+val nickRiviera = Contact()
+val nicksAddress = Address()
+
+// Create a parent object with one embedded address
+realm.write {
+    this.copyToRealm(
```
rorbech, line 6: Maybe a bit confusing with the external defined objects 🤔 

Maybe better with 
```
val contact = copyToRealm(Contact())

contact.apply {
    name ...
    
    address = Address().apply { ...  }
}
```
I would think that it makes it more obvious what is actually going on, but maybe I am biased from doing the SDK 😅

source/examples/generated/kotlin/DataTypesTest.snippet.update-embedded-object.kt
```diff
@@ -0,0 +1,11 @@
+// ... Fetch the object
```
rorbech, line 1: You need to fetch the object inside the `write` block with `findLatest(object)` otherwise you would not be able to update it.

source/examples/generated/kotlin/DataTypesTest.snippet.overwrite-embedded-object.kt
```diff
@@ -0,0 +1,11 @@
+// ... Fetch the object
+
+// Overwrite the embedded object (deletes the original object)
+realm.write {
```
rorbech, line 4: This is not overwriting the `address` with a new object but updating the existing one. And again you need to fetch the most recent version to be able to update it. 

You should do something like

```
realm.write {
     val contact = findLatest(contactReference)?: error("Could not retrieve the latest version of parent object")
     contact.address = Address().apply { ... }
}
```

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
```diff
@@ -0,0 +1,122 @@
...
+
+.. literalinclude:: /examples/generated/kotlin/DataTypesTest.snippet.overwrite-embedded-object.kt 
+    :language: kotlin
+
+
+Query an Embedded Object
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+To query a collection of embedded objects, use dot notation to filter or sort a 
+collection based on an embedded object property value.
```
rorbech, line 97: This sounds like a weird paragraph to start out with. It is a general possibility for all queries that you can filter on member, so can't see why this should be mentioned specifically here.

source/examples/generated/kotlin/DataTypesTest.snippet.query-embedded-objects.kt
```diff
@@ -0,0 +1,13 @@
+// Query through the parent object
```
rorbech, line 1: I think it would make most sense to show that you can query the embedded objects directly first and then follow up with the filter on linked property afterwards. Thus, swapping this example with the below.

source/examples/generated/kotlin/DataTypesTest.snippet.delete-embedded-object.kt
```diff
@@ -0,0 +1,13 @@
+//  Delete an embedded object directly from parent
```
rorbech, line 1: Maybe a bit off with `directly from parent` as the embedded object are deleted directly without touching the parent. This a valid use case though, so just keep with an updated text. But maybe also include the use case of deleting it directly from the parent by clearing the parents property, which would delete the embedded class
```
realm.write {
    val parent = findLatest(parentRef)?: error("Cannot find parent reference")
    parent.address = null
}
```
This will delete the object so that it isn't returned from a query directly on the specific embedded class (`realm.query<Address>()`). This is fundamentally different from properties referencing normal (non-embedded) objects where clearing a reference would not delete the actual object, but only the reference to it.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/DataTypesTest.kt
```diff
@@ -0,0 +1,179 @@
...
+            val nicksAddress = Address()
+
+            // Create a parent object with one embedded address
+            realm.write {
+                this.copyToRealm(
+                    nickRiviera.apply {
+                        name = "Nick Riviera"
+
+                        // Embed the address in the contact object
+                        address = nicksAddress.apply {
```
rorbech, line 71: This `apply` will update the unmanaged `nicksAddress` object and import it by creating a copy inside the realm and assign the properties of the clone according to the values of `nicksAddress`. Is that what you expect? and intend to show?

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/DataTypesTest.kt
```diff
@@ -0,0 +1,179 @@
...
+            // :snippet-end:
+
+            // :snippet-start: update-embedded-object
+            // ... Fetch the object
+
+            // Modify the property of the embedded object in a write transaction
+            realm.write {
+
+                // Update the embedded object property directly
+                nicksAddress.street = "100 10th St N"
```
rorbech, line 88: This doesn't update anything inside the realm. You just update the unmanaged instance. If you want to update the address of `Nick Riviera` inside the Realm then you need to fetch the reference inside the realm with a query or by looking up a previous managed instance with `findLatest(...)`

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/DataTypesTest.kt
```diff
@@ -0,0 +1,179 @@
...
+                nickRiviera.address?.state = "NY"
+            }
+            // :snippet-end:
+
+            // :snippet-start: overwrite-embedded-object
+            // ... Fetch the object
+
+            // Overwrite the embedded object (deletes the original object)
+            realm.write {
+                nicksAddress.apply {
```
rorbech, line 100: Again, this it not updating anything inside the realm, but just the unmanaed `nicksAddress` instance.

rorbech (changes requested):
There are some errors in the usage of embedded objects. I have tried to highlight the errors.

source/examples/generated/kotlin/DataTypesTest.snippet.delete-embedded-object.kt
cbullinger: Added/clarified use case

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
cbullinger: I copied from the other sdks' versions of the page, but agree that it's not needed here. Removed.

source/examples/generated/kotlin/DataTypesTest.snippet.create-embedded-object.kt
cbullinger: Updated (hopefully) for better clarity

source/examples/generated/kotlin/DataTypesTest.snippet.overwrite-embedded-object.kt
cbullinger: Updated in this and all following instances

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
```diff
@@ -0,0 +1,125 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+An embedded object is a special type of :ref:`Realm object
+<kotlin-object-type>` that models complex data about a specific object.
+Embedded objects are similar to :ref:`relationships
+<kotlin-relationships>`, but they provide additional constraints.
```
cbush, line 19: Not sure if this was inherited text from elsewhere, but anyway: I'm not sure "provide ... constraints" is the right verb? I might simply say "have additional constraints" here. Depends what we're trying to say about the quality of the constraints I guess.

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
```diff
@@ -0,0 +1,125 @@
...
+
+Realm treats each embedded object as nested data inside of a single specific
+parent object.
+An embedded object inherits the lifecycle of its parent object.
+It cannot exist as an independent Realm object. 
+
+Embedded objects have the following properties: 
+
+- Embedded objects are deleted when their parent object is deleted 
+  or their parent no longer references them.
```
cbush, line 29: It seems this list and the paragraph above have some overlapping information. Could it be tightened up?

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
```diff
@@ -0,0 +1,125 @@
...
+   You can use the same embedded object type in multiple parent object types, and
+   you can embed objects inside other embedded objects. You can even
+   recursively reference an embedded object type as an optional property in its
+   own definition.
+
+Define an Embedded Object
+-------------------------
+
+To define an embedded object, derive a class from 
+`EmbeddedRealmObject <{+kotlin-local-prefix+}/io.realm.kotlin.types/-embedded-realm-object/index.html>`__:
```
cbush, line 45: Is the link correct? I'm getting 403

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
```diff
@@ -0,0 +1,125 @@
...
+- Embedded objects cannot have a :ref:`primary key <kotlin-primary-keys>`. 
+- You cannot reassign an embedded object to a different parent object.
+- You cannot link to an embedded object from multiple parent objects. 
+
+.. tip:: Embedded Object Types are Reusable and Composable
+
+   You can use the same embedded object type in multiple parent object types, and
+   you can embed objects inside other embedded objects. You can even
+   recursively reference an embedded object type as an optional property in its
+   own definition.
```
cbush, line 39: Neat!

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
```diff
@@ -0,0 +1,125 @@
...
+
+.. literalinclude:: /examples/generated/kotlin/DataTypesTest.snippet.update-embedded-object.kt 
+    :language: kotlin
+
+
+Overwrite an Embedded Object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To overwrite an embedded object, reassign the embedded object properties to a new 
+instance in a write transaction:
```
cbush, line 87: This is interesting. Does this imply that simply assigning a new embedded object instance to the property (as in create an embedded object) would not work?

I would almost argue this overwrite example is an antipattern -- as the schema evolves, this code might "forget" to update some new property and therefore leave old data in the supposedly new instance.

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
```diff
@@ -0,0 +1,125 @@
...
+   If you want the referenced objects to persist after the deletion of the 
+   parent object, use a regular Realm object with a :ref:`to-one relationship 
+   <kotlin-to-one-relationship>` instead.
+
+To delete an embedded object: 
+
+- Delete only the specific embedded object (without touching the parent object) 
+- Delete through the parent: 
+
+  - Deleting the parent object also deletes all of its embedded objects
```
cbush, line 121: I think the list could use a bit more context. Are these two ways to approach the problem?

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
```diff
@@ -0,0 +1,125 @@
...
+  - Removing the parent's reference to the embedded object also deletes the embedded object 
+
+.. literalinclude:: /examples/generated/kotlin/DataTypesTest.snippet.delete-embedded-object.kt
```
cbush, line 124: What consequence does `delete(addressToDelete)` have on the parent object that holds it? Is the parent object's embedded property automatically nullified?

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/DataTypesTest.kt
```diff
@@ -0,0 +1,231 @@
...
+                    this.query<Address>("street == '123 Fake St'").find().first()
+                delete(addressToDelete)
+            }
+
+            // Delete an embedded object through the parent
+            realm.write {
+                val propertyToClear: Contact =
+                    this.query<Contact>("name == 'Nick Riviera'").find().first()
+
+                // Clear the parent property (deletes the embedded class)
```
cbush, line 212:
```suggestion
                // Clear the parent property (deletes the embedded object instance)
```

cbush (approved):
Looks great! I have a few minor comments I'd like to see addressed before merge

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
cbullinger: Fixed. Thanks for catching

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
cbullinger: Yeah, the wording was taken from the text shared by the other sdks' pages, but I agree it's awkward. Updated to "have additional constraints"

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
cbullinger: Yes, after delete, parent's address property is null

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
cbullinger: Yes, insofar as they're the approaches available in kotlin sdk (other than Flutter, we only instruct on deleting the parent but IDK if that's also because you can't query the embedded objects directly in the other sdks). 
Suggestion from tech review was to include the 3rd bullet (clear through the parent property) since that's also a valid use case.

cbullinger: I've tweaked slightly

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
cbullinger: Updated for clarification and a better code example for overwrite (deleting & creating a new object)

source/sdk/kotlin/realm-database/schemas/embedded-objects.txt
cbullinger: I've tweaked slightly -- lmk if it there's still concern of overlapping/redundant info

cbush (approved):

PR #2561, opened 2023-02-06
(DOCSP-26120): Realm.refresh() and Realm.refreshAsync() for Flutter

mongodben:
## Pull Request Info

Document `Realm.refresh()` and `Realm.refreshAsync()` for Flutter SDK. 

Note that the docs here are pretty limited b/c this method is only really needed for multi-processing, and we've decided to only talk about multi-processing in the context of syncing on multiple processes *for now*. 

In future docs changes where we talk more about working with Realm on multiple Dart/Flutter processes, there'll be further documentation of these methods.

### Jira

- https://jira.mongodb.org/browse/DOCSP-26120

### Staged Changes

- [Sync Multiple Processes (Flutter)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26120/sdk/flutter/sync/sync-multiple-processes/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/examples/dart/test/sync_multiple_processes_test.dart
+++ b/examples/dart/test/sync_multiple_processes_test.dart
@@ -7,10 +7,18 @@ import 'package:path/path.dart' as path;
 import './utils.dart';
 import './open_flexible_sync_realm_test.dart';
 
+part 'sync_multiple_processes_test.g.dart';
+
 late App app;
 late Realm realm;
 late User currentUser;
 const APP_ID = "flutter-flexible-luccm";
+
+@RealmModel()
+class _Person {
+  @PrimaryKey()
+  late String name;
+}
 
 void main() {
   group('Sync multiple processes', () {
@@ -59,5 +67,34 @@ void main() {
       realmWithDisconnectedSync.close();
       // since both realm connections are only for 1 realm, this deletes both
     });
+  });
+  // Note: This test doesn't actually test what is being documented, b/c the
+  // documentation is about working across multiple processes and it's not
+  // intuitive to do that within a unit test (which runs in only 1 process)
+  test("Realm.refresh()", () async {
+    final realm = Realm(Configuration.local([Person.schema]));
+    realm.write(() => realm.add(Person('John')));
+    // :snippet-start: refresh-main-process
+    // Delete object in one process
+    realm.write(() {
+      realm.add(Person('John'));
+    });
+    // :snippet-end:
+    // :snippet-start: refresh-secondary-process
+    // You must call realm.refresh() in the secondary process
+    // before the data written in the main process registers
+    // in the secondary process.
+    realm.refresh();
+    realm.find<Person>('John');
+    // :snippet-end:
+    // :snippet-start: refresh-async
+    // Asynchronously refresh the realm. Once the refresh operation occurs
+    // in the background.
+    realm.refreshAsync();
+    // :snippet-end:
+    // wait for async operation to finish before clean up
+    await realm.refreshAsync();
+
+    cleanUpRealm(realm);
   });
 }
--- a/examples/dart/test/sync_multiple_processes_test.dart
+++ b/examples/dart/test/sync_multiple_processes_test.dart
@@ -88,8 +88,7 @@ void main() {
     realm.find<Person>('John');
     // :snippet-end:
     // :snippet-start: refresh-async
-    // Asynchronously refresh the realm. Once the refresh operation occurs
-    // in the background.
+    // Asynchronously refresh the realm in the background.
     realm.refreshAsync();
     // :snippet-end:
     // wait for async operation to finish before clean up
--- a/examples/dart/test/sync_multiple_processes_test.dart
+++ b/examples/dart/test/sync_multiple_processes_test.dart
@@ -10,7 +10,6 @@ import './open_flexible_sync_realm_test.dart';
 part 'sync_multiple_processes_test.g.dart';
 
 late App app;
-late Realm realm;
 late User currentUser;
 const APP_ID = "flutter-flexible-luccm";
 
@@ -28,10 +27,6 @@ void main() {
       final credentials =
           Credentials.emailPassword("bart@example.com", "abc123");
       currentUser = await app.logIn(credentials);
-    });
-    tearDown(() async {
-      await currentUser.logOut();
-      Realm.shutdown();
     });
     // Note: This test doesn't actually test what is being documented, b/c the
     // documentation is about working across multiple processes and it's not
@@ -66,6 +61,8 @@ void main() {
       expect(myTri, isNotNull);
       realmWithDisconnectedSync.close();
       // since both realm connections are only for 1 realm, this deletes both
+      await cleanUpRealm(realmWithSync, app);
+      await cleanUpRealm(realmWithDisconnectedSync, app);
     });
   });
   // Note: This test doesn't actually test what is being documented, b/c the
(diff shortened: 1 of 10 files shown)


source/sdk/flutter/sync/sync-multiple-processes.txt
```diff
@@ -13,8 +13,10 @@ in multiple processes include:
 - A multi-window desktop application where each window writes to the same synced realm.
 - A server application that writes to a synced realm from multiple processes.
 
-On a high level, syncing data to a realm from multiple processes consists
-of the following:
+Open a Synced Realm in Multiple Processes
+-----------------------------------------
+
+To open a single synced realm from multiple processes, perform the following:
```
mongodben, line 19: refactored this header/intro to make addition of later section in the PR make more sense.

examples/dart/test/sync_multiple_processes_test.dart
```diff
@@ -60,4 +68,32 @@ void main() {
       // since both realm connections are only for 1 realm, this deletes both
     });
   });
+  // Note: This test doesn't actually test what is being documented, b/c the
```
blagoev, line 71: Do we need a test for this in the docs? Its a complex scenario to test which could be done in the SDK only.

blagoev: we have a test for that in the SDK

source/examples/generated/flutter/sync_multiple_processes_test.snippet.refresh-secondary-process.dart
```diff
@@ -0,0 +1,5 @@
+// You must call realm.refresh() in the secondary process
```
blagoev, line 1: Not sure to whom this comment is but it is not a must to do it. This methos is intendent only if people find problems with the automatic refresh Realm has built-in

source/sdk/flutter/sync/sync-multiple-processes.txt
```diff
@@ -41,3 +43,28 @@ constructor. Include the schema and any additional optional named arguments.
    :language: dart
    :caption: secondary_process.dart
    :emphasize-lines: 5,6
+
+.. _flutter-realm-refresh:
+
+Refresh Data across Processes
+-----------------------------
+
+When you are working with a realm across multiple processes, the realm does not
```
blagoev, line 52: This is incorrect. Realm has auto-refresh built-in. It works very well and is on by default. Developers don't need to do anything to sync data between processes. Still in some cases they might need to call realm.refresh() if they notice that data is not available in their  Realm instance. This should happen very rarely so the API is marked as `Typically should not be used.`

blagoev: So here we better write something like

> When sharing the same realm file between multiple processes, Realm has auto-refresh built-in and data written from one process is visible immediately from the other processes. Occasionally this may not happen immediately and thus Realm exposes a refresh() and refreshAsync() methods to trigger the data refresh mechanism manually.

blagoev (changes requested):
added a suggestion how to fix the dos for the `refresh` methods

examples/dart/test/sync_multiple_processes_test.dart
mongodben: yeah i saw your tests. i don't think it matters very much here since the actual logic in the code example is so simple.

source/sdk/flutter/sync/sync-multiple-processes.txt
mongodben: oh ok, thanks for that clarification. what are some circumstances when the refresh might not occur automatically? 

would be good to specify in the PR.

examples/dart/test/sync_multiple_processes_test.dart
blagoev: would suggest to remove this test here

mongodben: going to keep b/c it's important for us to make sure that our code examples in the docs run even if they're not fully testing the behavior.

blagoev: then the comments needs to be fixed I suppose.

examples/dart/test/sync_multiple_processes_test.dart
```diff
@@ -60,4 +68,32 @@ void main() {
...
+    realm.write(() => realm.add(Person('John')));
+    // :snippet-start: refresh-main-process
+    // Delete object in one process
+    realm.write(() {
+      realm.add(Person('John'));
+    });
+    // :snippet-end:
+    // :snippet-start: refresh-secondary-process
+    // You must call realm.refresh() in the secondary process
+    // before the data written in the main process registers
```
blagoev, line 85:
```suggestion
```

examples/dart/test/sync_multiple_processes_test.dart
```diff
@@ -60,4 +68,32 @@ void main() {
...
+    // in the secondary process.
+    realm.refresh();
+    realm.find<Person>('John');
+    // :snippet-end:
+    // :snippet-start: refresh-async
+    // Asynchronously refresh the realm in the background.
+    realm.refreshAsync();
+    // :snippet-end:
+    // wait for async operation to finish before clean up
+    await realm.refreshAsync();
```
blagoev, line 95: the second call is not needed

examples/dart/test/sync_multiple_processes_test.dart
```diff
@@ -60,4 +68,32 @@ void main() {
...
+    // :snippet-start: refresh-secondary-process
+    // You must call realm.refresh() in the secondary process
+    // before the data written in the main process registers
+    // in the secondary process.
+    realm.refresh();
+    realm.find<Person>('John');
+    // :snippet-end:
+    // :snippet-start: refresh-async
+    // Asynchronously refresh the realm in the background.
+    realm.refreshAsync();
```
blagoev, line 92:
```suggestion
    realm.refreshAsync();
```
This is a bad practice. One should really use await with async calls.

source/sdk/flutter/sync/sync-multiple-processes.txt
mongodben: per slack convo, updating the text but not going into detail about what might cause the refresh to not happen.

source/examples/generated/flutter/sync_multiple_processes_test.snippet.refresh-secondary-process.dart
blagoev: its definitely not a `must` to call this method. We should reword this comment

mongodben: yep, reworded

examples/dart/test/sync_multiple_processes_test.dart
blagoev: when you call a method with await you release the main isolate (thread/event loop) to handle any other async events and microtasks. Then when the result is available the code continues with the next statement.

mongodben: thx, taking conversation to slack for additional clarification.

blagoev (approved):

dacharyc (approved):
One grammar nit we should fix, and a question about whether we can add more info, but otherwise LGTM.

source/sdk/flutter/sync/sync-multiple-processes.txt
```diff
@@ -41,3 +43,35 @@ constructor. Include the schema and any additional optional named arguments.
...
+
+Refresh Data across Processes
+-----------------------------
+
+When sharing the same realm file between multiple processes,
+Realm has auto-refresh built-in. Data written from one process is visible
+to the other processes. Generally, you do not need to write additional logic
+to refresh data across processes.
+
+However, occasionally the refresh may not happen immediately. In this case,
```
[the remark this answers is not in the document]
mongodben: yeah i had the same question. lubo said there's no clear answer, which is why the above sentence is 

> Generally, you do not need to write additional logic
to refresh data across processes.

so this is what folks generally should do

PR #2574, opened 2023-02-09
(DOCSP-27007): Link user identities @realm/react

mongodben:
## Pull Request Info

Link user identities @realm/react page

### Jira

- https://jira.mongodb.org/browse/DOCSP-27007

### Staged Changes

- [Link User Identities (React Native)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-27007/sdk/react-native/manage-users/link-user-identities/)

### Reminder Checklist

### Review Guidelines


source/sdk/react-native/manage-users/link-user-identities.txt
```diff
@@ -21,21 +18,22 @@ accounts. They sign up with SSO or email/password authentication. By default,
...
+
 
    .. tab::
       :tabid: typescript
 
-      .. literalinclude:: /examples/generated/code/start/link-identities.snippet.link-identities.ts
+      .. literalinclude:: /examples/generated/react-native/ts/link-identities.test.snippet.link-identities.tsx
          :language: typescript
-         :emphasize-lines: 10
+         :emphasize-lines: 13
```
mongodben, line 39:
```suggestion
         :emphasize-lines: 14
```

source/examples/generated/react-native/js/link-identities.test.snippet.link-identities.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {useUser} from '@realm/react';
+import Realm from 'realm';
+
+function LinkUserIdentities({username, password}) {
```
cbush, line 5: I think it's a bit lacking as an example, because without the return statement it's not clear what this function is supposed to be, where it sits in a component lifecycle, etc.

In order to use useUser(), it must either be a component or a hook.

It shows us defining a linkIdentities(), but it doesn't show where to put it.

One simple idea could be to turn this into an explicit custom hook example (by naming it `use...`):

```js
function useIdentityLinker() { // ... 
```

then return your callback
```js
const linkIdentities = useCallback((newCredentials) => /* ... */)
return {linkIdentities};
```

But probably a more clear example would be defining a component that registers a email/password user while linking your anonymous user with the new email/password user.

mongodben: thanks for the feedback. thoughts: 

>  One simple idea could be to turn this into an explicit custom hook example (by naming it use...):

i tried implementing a custom hook, but had trouble following what you meant here. 

> But probably a more clear example would be defining a component that registers a email/password user while linking your anonymous user with the new email/password user.

anyways decided to go w this approach to have some other components, describing how they relate to one another. this approach was actually something that i'd considered when first drafting this PR, but decided not to go w it out of fear it'd bloat the example and also keeping scope of the @realm/react-ification limited.

though now that you suggest the change as well, and i've actually implemented it, it seems more manageable to me 😄 

i've implemented this in latest pushed commit 6e1b3bec4

cbush: I should have said "minimal" instead of "simple"

cbush:
I like the new example! I expected something like:
>The following example illustrates a common scenario. In this app, users are automatically logged in anonymously. After exploring the app, a user might decide to register an account. The content from their anonymous session(s) should be persisted under the new account. So we link user accounts in the registration component.

takameyer (approved):
LGTM

cbush (approved):

PR #2581, opened 2023-02-13
Update for Rules v2: remove "Device Sync Permissions"

cbush:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26779

### Staged Changes

- https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/permissions-v2/

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/flutter/sync/write-to-synced-realm.txt
+++ b/source/sdk/flutter/sync/write-to-synced-realm.txt
@@ -23,31 +23,20 @@ The examples on this page use an Atlas App Services App with the following
 Device Sync configuration and a client app with the following Realm SDK 
 data model and subscriptions.
 
-Device Sync is configured with the following queryable fields:
-
-- ``_id`` (always included)
-- ``miles``
-- ``ownerId``
-
-Device Sync has permissions configured to let users read and write only their own
-data:
+The App Services App has permissions configured to let users read and write only
+their own data:
 
 .. code-block:: json
 
    {
-     "rules": {},
-     "defaultRoles": [
-       {
-         "name": "owner-read-write",
-         "applyWhen": {},
-         "read": {
-           "ownerId": "%%user.id"
-         },
-         "write": {
-           "ownerId": "%%user.id"
-         }
-       }
-     ]
+     "name": "owner-read-write", 
+     "apply_when": {},
+     "document_filters": {
+       "read": { "ownerId": "%%user.id" },
+       "write": { "ownerId": "%%user.id" }
+     },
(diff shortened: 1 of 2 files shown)


source/sdk/flutter/sync/write-to-synced-realm.txt
```diff
@@ -23,31 +23,20 @@ The examples on this page use an Atlas App Services App with the following
 Device Sync configuration and a client app with the following Realm SDK 
 data model and subscriptions.
 
-Device Sync is configured with the following queryable fields:
```
cbush, line 26: Removed because queryable fields are now automatically determined.

dacharyc: I thought this was only when Development Mode is enabled?

If this is true when Development Mode is enabled, we should call that out instead of just removing this - i.e. "If Development Mode is enabled, Sync can automatically determine queryable fields from your Flexible Sync query. If Development Mode is not enabled, add these queryable fields."

Related, if this is true when Development Mode is enabled, then we can probably remove this info from all of the Quick Starts? I _think_ they all have a bullet in the [Device Sync prerequisites](https://www.mongodb.com/docs/realm/sdk/swift/quick-start/#prerequisites) about configuring queryable field, but they also say to enable Development Mode, so this would be unnecessary there.

dacharyc: Related, there is also mention of configuring queryable fields on the "Manage Flexible Sync Subscription" pages: https://www.mongodb.com/docs/realm/sdk/swift/sync/flexible-sync/#subscribe-to-queryable-fields

We might want to add something there about not needing to do this if Development Mode is enabled?

dacharyc: And in the "Add Device Sync to an App" page that exists in Swift: https://www.mongodb.com/docs/realm/sdk/swift/sync/add-sync-to-app/

Flutter has the same page but the verbiage there is slightly different so I don't think it needs any updates relative to queryable fields.

dacharyc (approved):
The changes here look fine, except I have a Q about the queryable fields. If I've misunderstood, feel free to disregard - but if my understanding is correct, I'd suggest we tweak this update and also update some related places throughout the SDK docs.

cbush:
@dacharyc you're right, the change to queryable fields only applies to Development Mode. I undid that change, and we'll address it in the proper ticket.

PR #2601, opened 2023-02-16
(DOCSP-26737): C++: Add a Threading page

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26737

### Staged Changes

- [Threading](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-26737/sdk/cpp/crud/threading/)

### Reminder Checklist

### Review Guidelines


examples/cpp/threading.cpp
```diff
@@ -0,0 +1,88 @@
+// :replace-start: {
+//   "terms": {
+//     "ThreadingItem_": ""
```
dacharyc, line 3: Ah, I thought it was odd that this naming was in the generated code example - looks like we need to tweak the `replace` here and then regenerate the Bluehawk snippet.

```suggestion
//     "ThreadingExample_": ""
```

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,254 @@
...
+  :ref:`notifications <cpp-react-to-changes>`.
+
+- To see changes from other threads in the realm on the current
+  thread, :ref:`refresh <cpp-refreshing-realms>` your realm instance.
+
+- To share an instance of a realm or specific object with another thread, 
+  share a :ref:`thread_safe_reference <cpp-thread-safe-reference>` to the 
+  realm or object.
+
+.. TODO: Add a bullet about the scheduler https://www.mongodb.com/docs/realm-sdks/cpp/latest/structrealm_1_1platform__default__scheduler.html#a96336dad0a9ee38af907cb7c9bed4870
```
dacharyc, line 83: Is there something we should say here about the scheduler - i.e. "To schedule realm transactions on a specific run loop, something something scheduler."

dacharyc (commented):
Couple of minor Qs. Seems I can't approve this PR as it's still just a Draft.

examples/cpp/threading.cpp
cbush: Blar I was messing with this and forgot to fix it

source/sdk/cpp/crud/threading.txt
cbush: Oops, I meant to remove this comment. I don't think so. I think the scheduler has a bit more to do with how threads/tasks/events are dispatched/handled, and we would use the other mechanisms more directly.

examples/cpp/threading.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+TEST_CASE("scheduler", "[write]") {
+    auto shouldQuitProgram = true;
+    
+    // :snippet-start: scheduler
+    // :uncomment-start:
+    // int main() {
+    // :uncomment-end:
+        // Set up a custom scheduler
+        auto scheduler = realm::NoPlatformScheduler::make();
```
leemaguire, line 63: The `NoPlatformScheduler`isn't tested. For the time being users should implement their own `UV` or `CFRunLoop` scheduler derived from `realm::scheduler`, for the scenario where they need to install a run loop inside a thread. See https://github.com/realm/realm-cpp/blob/main/tests/run_loop_tests.cpp as a guide

source/examples/generated/cpp/threading.snippet.refresh.cpp
```diff
@@ -0,0 +1 @@
+realm.refresh();
```
leemaguire, line 1: this won't be required when a run loop is setup

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+Communication Across Threads
+----------------------------
+
+To access the same realm file from different threads, you must instantiate a
+realm instance for every thread of your app. As long as you specify the same
+configuration, all realm instances will map to the same file on disk.
+
+One of the key rules when working with Realm Database in a multithreaded
+environment is that objects are thread-confined: **you cannot access the
+instances of a realm, collection, or object that originated on other threads.**
```
krollins-mdb, line 61: > you cannot access the instances of a realm, collection, or object that originated on other threads.

This is a strong statement that seems to be untrue? Below, we have a section about `thread_safe_reference` and present it as a way to do precisely what we here say we cannot do.

I'm not sure how to reconcile this and have no idea how `thread_safe_reference` works with MVCC. Maybe it makes sense to introduce `thread_safe_reference` here as an exception?

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+  share a :ref:`thread_safe_reference <cpp-thread-safe-reference>` to the 
+  realm or object.
+
+.. _cpp-thread-safe-reference:
+
+Pass Instances Across Threads
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Instances of ``realm::realm``, ``realm::results``, and ``realm::object`` (when
+managed) are *thread-confined*. That means you may only use them on the thread
```
krollins-mdb, line 89: > (when managed)

Just checking that our expected audience will know what this means. We don't mention int anywhere else on the page.

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+   short-lived.
+
+.. literalinclude:: /examples/generated/cpp/threading.snippet.thread-safe-reference.cpp
+    :language: cpp
+
+Another way to work with an object on another thread is to query for it
+again on that thread. But if the object does not have a primary
+key, it is not trivial to query for it. You can use ``thread_safe_reference`` 
+on any object, regardless of whether it has a primary key. You can also 
+use it with lists and results.
```
krollins-mdb, line 116: > You can also use it with lists and results.

I think we should make this clearer earlier in the section.

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+
+.. _ios-use-realm-across-threads:
+
+Use the Same Realm Across Threads
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You cannot share realm instances across threads.
```
krollins-mdb, line 123: Except with `thread_safe_reference`, right?

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+instances will map to the same file on disk.
+
+.. _cpp-scheduler-thread-management:
+
+Schedulers (Run Loops)
+----------------------
+
+Some platforms or frameworks automatically set up a **scheduler** (or **run
+loop**), which continuously processes events during the lifetime of your
+program. The Realm C++ SDK detects and uses schedulers on the following
```
krollins-mdb, line 137: We use "app" in the rest of the page.

```suggestion
app. The Realm C++ SDK detects and uses schedulers on the following
```

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+
+Similarly, Realm Database has atomically-committed writes in the form
+of transactions. Realm Database also has many different versions of the 
+history at any given time, like branches.
+
+Unlike Git, which actively supports distribution and
+divergence through forking, a realm only has one true latest
+version at any given time and always writes to the head of
+that latest version. Realm Database cannot write to a previous
+version. This makes sense: your data should converge on one
```
krollins-mdb, line 211: We shouldn't presume to know what makes sense to readers.

```suggestion
version. This means your data converges on one
```

krollins-mdb (commented):
This is really well done! Threading is a complicated topic. 🧠 🍳 

I have one larger concern about how we talk about not accessing realm instances in other threads, but also have the ability to do so. Right now, we seem to be contradicting ourselves.

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+--------
+
+To create performant apps, developers must write thread-safe and maintainable
+multithreaded code that avoids issues like deadlocking and race conditions.
+Realm provides tools specifically designed for this.
+
+.. _cpp-threading-three-rules:
+
+Three Rules to Keep in Mind
+---------------------------
```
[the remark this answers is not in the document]
cbush: Took the first because in some cases violating the rules would go beyond unsafe multithreading. Some violations are simply disallowed by Realm altogether.

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+
+Don't lock to read:
+  Realm Database's :ref:`Multiversion Concurrency Control (MVCC) <cpp-mvcc>`
+  architecture eliminates the need to lock for read operations. The values you
+  read will never be corrupted or in a partially-modified state. You can freely
+  read from realms on any thread without the need for locks or mutexes.
+  Unnecessarily locking would be a performance bottleneck since each thread
+  might need to wait its turn before reading.
+
+Avoid synchronous writes on the UI thread if you write on a background thread:
```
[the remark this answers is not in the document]
cbush: I see your point, but I think it's only really a concern if you are in fact using Device Sync or doing something with multithreading. Simple apps that only use Realm as a local data store can get away with writing directly on the UI thread. Then again, at this point, most SDKs offer pretty straightforward async write APIs. I'm not against generalizing this as you suggested.

source/sdk/cpp/crud/threading.txt
```diff
@@ -0,0 +1,248 @@
...
+When you need to communicate across threads, you have several options depending
+on your use case:
+
+- To modify the data on two threads, :ref:`query <cpp-client-query-engine>`
+  for the object on both threads.
+
+- To react to changes made on any thread, use Realm Database's
+  :ref:`notifications <cpp-react-to-changes>`.
+
+- To see changes from other threads in the realm on the current
```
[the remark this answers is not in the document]
cbush: What I'm trying to say here is:
>To see [changes that happened on another thread] in [the realm on the current thread]

And I don't think your suggestion quite captures it...

source/sdk/cpp/crud/threading.txt
cbush: It is true, but the description of what thread safe reference does is a bit leading. It's more of a way to copy an instance to another thread as if it were created on that other thread.

source/sdk/cpp/crud/threading.txt
cbush: I'll just remove the line. The point being made here was that lists and results don't have a handle (like a primary key) so fetching them again on the other thread could be complex. We already enumerate the things you can use this with earlier.

source/sdk/cpp/crud/threading.txt
cbush: No. Need to clarify that thread-safe references just copy instances to the other thread. You aren't using the same instance across threads.

cbush:
Will apply text changes to sibling pages in a subsequent PR.

PR #2612, opened 2023-02-21
(DOCSP-28151): Add Sync Session documentation for C++

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-28151

### Staged Changes

- [Manage Sync Sessions](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28151/sdk/cpp/sync/manage-sync-session/): New page
- [Sync Data](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28151/sdk/cpp/sync/): New page containing existing content to serve as a container page for the Sync session, now that we have more than a couple of sync-related pages

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/cpp/sync.txt
+++ b/source/sdk/cpp/sync.txt
@@ -54,8 +54,8 @@ Subscription sets are based on a specific type of :ref:`Realm object
 <cpp-object-model>`. You might have multiple subscriptions if you 
 have many types of Realm objects.
 
-To use Flexible Sync in your client application, open a synced realm 
-with a flexible sync configuration. Then, manage subscriptions
+To use Device Sync in your client application, open a synced realm 
+with a Flexible Sync configuration. Then, manage subscriptions
 to determine which documents to sync. For more information, refer to
 :ref:`cpp-manage-flexible-sync-subscriptions`.
 
--- a/source/sdk/cpp/sync.txt
+++ b/source/sdk/cpp/sync.txt
@@ -1,0 +1,76 @@
+.. _cpp-sync:
+.. _cpp-realm-sync:
+
+===========================================
+Sync Data Between Devices - C++ SDK (Alpha)
+===========================================
+
+.. toctree::
+   :titlesonly:
+
+   Manage Sync Subscriptions </sdk/cpp/sync/sync-subscriptions>
+   Manage Sync Sessions </sdk/cpp/sync/manage-sync-session>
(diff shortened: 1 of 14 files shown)


source/sdk/cpp/sync.txt
```diff
@@ -0,0 +1,76 @@
...
+You can form queries using :ref:`Realm Query Language <realm-query-language>`.
+
+.. include:: /includes/note-unsupported-flex-sync-rql-operators.rst
+
+Subscription sets are based on a specific type of :ref:`Realm object 
+<cpp-object-model>`. You might have multiple subscriptions if you 
+have many types of Realm objects.
+
+To use Flexible Sync in your client application, open a synced realm 
+with a flexible sync configuration. Then, manage subscriptions
```
cbush, line 58: Should this capitalization be consistent?

source/sdk/cpp/sync.txt
```diff
@@ -0,0 +1,76 @@
...
+
+You can form queries using :ref:`Realm Query Language <realm-query-language>`.
+
+.. include:: /includes/note-unsupported-flex-sync-rql-operators.rst
+
+Subscription sets are based on a specific type of :ref:`Realm object 
+<cpp-object-model>`. You might have multiple subscriptions if you 
+have many types of Realm objects.
+
+To use Flexible Sync in your client application, open a synced realm
```
cbush, line 57: I think we should start referring to Device Sync rather than "Flexible Sync".

cbush (approved):
One question I have is if the sync session object is a "lightweight handle". We are passing it around by value in an optional. This is all we can do given that the SDK returns it that way. I think we might want to be explicit that it is a lightweight handle that can be passed around by value.

source/sdk/cpp/sync.txt
dacharyc: Ahh, yeah, I suspect there was some conflating with flexibleSyncConfiguration or otherwise. Will fix 👍

source/sdk/cpp/sync.txt
dacharyc: Yeah, I think this came from page copy where there is both Flexible Sync and Partition-based Sync, and the instructions are different depending on which Sync mode. But since there's only Flexible Sync in C++, it makes sense to just refer to it as Device Sync. 👍

PR #2653, opened 2023-03-07
(DOCSP-28423): Add information to docs around reflection

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-28423

### Staged Changes

- [Install](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28423/sdk/swift/install/#prerequisites): Added a prerequisite around this
- [Define an Object Model](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28423/sdk/swift/model-data/define-model/object-models/#key-concept--properties): Added a paragraph to the "Properties" concept about this
- [Test and Debug](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28423/sdk/swift/test-and-debug/#no-properties-are-defined-for-model): New section "No Properties are Defined for Model" about this

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/swift/install.txt
+++ b/source/sdk/swift/install.txt
@@ -31,7 +31,9 @@ meets the following prerequisites:
 - When using Xcode 14, a target of iOS 11.0 or higher, macOS 10.13 or higher, tvOS 11.0 or higher, or watchOS 4.0 or higher.
 - Reflection is enabled in your project. The Realm Swift SDK uses reflection to 
   determine your model's properties. Your project must not set 
-  ``SWIFT_REFLECTION_METADATA_LEVEL = none`` or Realm cannot see properties in your model.
+  ``SWIFT_REFLECTION_METADATA_LEVEL = none`` or Realm cannot see properties 
+  in your model. Reflection is enabled by default if your project does 
+  not specifically set a level for this setting. 
 
 .. note::
 
--- a/source/sdk/swift/test-and-debug.txt
+++ b/source/sdk/swift/test-and-debug.txt
@@ -262,4 +262,6 @@ the following, confirm that your project has not disabled reflection metadata:
    Terminating app due to uncaught exception 'RLMException', reason: 'No properties are defined for 'ObjectName'.
 
 If you set ``SWIFT_REFLECTION_METADATA_LEVEL = none``, Realm cannot 
-discover children of types, such as properties and enums.
+discover children of types, such as properties and enums. Reflection is 
+enabled by default if your project does not specifically set a level for 
+this setting.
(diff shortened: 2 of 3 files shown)


source/sdk/swift/install.txt
```diff
@@ -29,6 +29,9 @@ meets the following prerequisites:
 - When using Xcode 13, a target of iOS 9.0 or higher, macOS 10.9 or higher, tvOS 9.0 or higher, or watchOS 2.0 or higher.
 - If using Swift Package Manager with Xcode 13, target of iOS 11+ or macOS 10.10+ is required.
 - When using Xcode 14, a target of iOS 11.0 or higher, macOS 10.13 or higher, tvOS 11.0 or higher, or watchOS 4.0 or higher.
+- Reflection is enabled in your project. The Realm Swift SDK uses reflection to 
+  determine your model's properties. Your project must not set 
+  ``SWIFT_REFLECTION_METADATA_LEVEL = none`` or Realm cannot see properties in your model.
```
cbullinger, line 34: since this is in the prereqs, should this be inverted to what *must be* set instead of what you must *not* (i.e. reflection set to All or Without Names, I'm guessing)?

source/sdk/swift/test-and-debug.txt
```diff
@@ -243,3 +246,20 @@ use a property, and there aren't any objects in memory that could have
...
+
+The Realm Swift SDK uses the Swift language reflection feature to determine
+the properties in your model at runtime. If you get a crash similar to:
+
+.. code-block:: shell
+   :copyable: false
+
+   Terminating app due to uncaught exception 'RLMException', reason: 'No properties are defined for 'ObjectName'.
+
+Confirm that your project has not disabled reflection metadata. If you set
```
cbullinger, line 263: can we tweak this so the sentence isn't split by the error block (or that the second half isn't capital 'C' Confirm ... maybe that's what's tripping me up?)

cbullinger (approved):
overall lgtm

source/sdk/swift/install.txt
dacharyc: By default, you don't actually have to set it at all. If you don't add this setting to your project, the default behavior is `all`. It's only the case where a user explicitly sets it to `none` - typically for performance reasons - that they have issues with their Realm models. So I don't think it makes sense to tell users to set `all` when that's the default behavior and they shouldn't have to do anything with it.

tgoyne: It might be worth explicitly calling out that it's enabled by default just to avoid any confusion?

tgoyne (approved):

PR #2675, opened 2023-03-21
(DOCSP-27600): SwiftUI Bluehawk & Code Example Cleanup

dacharyc:
## Pull Request Info

Removing unused SwiftUI code examples & tests where possible, removing Bluehawk markup where the code itself is still required.

### Jira

- https://jira.mongodb.org/browse/DOCSP-27600

### DIFF
--- a/examples/ios/SwiftUICatalog/Views/QuickWrite.swift
+++ b/examples/ios/SwiftUICatalog/Views/QuickWrite.swift
@@ -29,7 +29,7 @@ struct DogDetailsView: View {
     }
 }
 
-// :snippet-start: edit-property
+// :snippet-start: quick-write-property
 struct EditDogDetails: View {
     @ObservedRealmObject var dog: SwiftUI_Dog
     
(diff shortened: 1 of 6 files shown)


examples/ios/SwiftUICatalog/Views/QuickWrite.swift
```diff
@@ -29,7 +29,7 @@ struct DogDetailsView: View {
     }
 }
 
-// :snippet-start: edit-property
+// :snippet-start: quick-write-property
```
MongoCaleb, line 32: I don't see the txt file that references this snippet. Did you also update it to point to the new name?

MongoCaleb (approved):
LGTM, if you also updated the text file that references the renamed "quick-write-property" snippet.

examples/ios/SwiftUICatalog/Views/QuickWrite.swift
dacharyc: Yeah, this snippet already exists as one named `QuickWrite.snippet.quick-write-property.swift` and is used on the Write Data page: https://github.com/mongodb/docs-realm/blob/a838a3c4ae975ff6fdc789091252508052f4f8f9/source/sdk/swift/swiftui/write.txt#L38

I must have renamed it at some point, so I'm just changing the Bluehawk markup name to match the existing snippet & where it's used.

PR #2676, opened 2023-03-22
(DOCSP-26500): Add Set the Client Log Level - Kotlin SDK page

cbullinger:
### Jira

- https://jira.mongodb.org/browse/DOCSP-26500

### Staged Changes

- [Set the Client Log Level - Kotlin SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-26500-log-level/sdk/kotlin/sync/log-level/) - new page

### Reminder Checklist

### Review Guidelines


examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/SyncTest.kt
```diff
@@ -336,5 +336,63 @@ class SyncTest: RealmTest() {
...
+            // Access your app
+            val app = App.create(yourFlexAppId)
+            val user = app.login(credentials)
+
+            // Access the configuration builder for the app
+            val config = SyncConfiguration.Builder(user, setOf(Toad::class))
+
+                // Set the custom logger and log level
+                // Must be set BEFORE you open a synced realm
+                .log(LogLevel.ALL, customLoggers = listOf(customLogger))
```
dacharyc, line 383: There's no example of what this `customLogger` might look like, and I didn't see a link to anything describing what this implementation should be. It looks like `customLoggers` is a list of RealmLogger objects, so I wonder if we want to link to [RealmLogger](https://www.mongodb.com/docs/realm-sdks/kotlin/latest/library-base/-realm%20-kotlin%20-s-d-k/io.realm.kotlin.log/-realm-logger/index.html) copy that goes along with this example. Additionally, I wonder if we want to provide just a basic placeholder `customLogger` to show what that might look like, similar to [the Swift example](https://github.com/mongodb/docs-realm/blob/272fe781ac6d65e21eefc7b54879b1d321faec87/examples/ios/Examples/Sync.swift#L364). (Just showing that this exposes a LogLevel and a String message, plus whatever additional stuff you want to throw in.)

dacharyc (changes requested):
We should tweak the "enum" wording because it's not quite right, and I also wonder if we want to elaborate on the custom logger example, or at least add a link to the API reference.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/SyncTest.kt
cbullinger: Updated to include link to RealmLogger docs and example to better match Swift (sdk engg didn't think it necessary to include more than that -- e.g. full custom logger implementation/logic)

source/sdk/kotlin/sync/log-level.txt
```diff
@@ -0,0 +1,49 @@
+.. _kotlin-client-log-level:
+
+=====================================
+Set the Client Log Level - Kotlin SDK
```
dacharyc, line 4: I notice in the ToC we call this "Set the Sync Log Level" - should we update the title here to match?

dacharyc (approved):
Overall LGTM, just a Q about the page title and the ToC entry.

source/sdk/kotlin/sync/log-level.txt
cbullinger: Was on purpose, but I changed TOC entry to match page title

PR #2680, opened 2023-03-22
(DOCSP-28060): C++: Organize and clean up docs test suite

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-28060

### Staged Changes

### Reminder Checklist

### Review Guidelines


examples/cpp/crud.cpp
```diff
@@ -0,0 +1,102 @@
...
+                { realm.add(dog); });
+    // :snippet-end:
+    auto dogsCount = realm.objects<Dog>().size();
+    SECTION("Test code example functions as intended")
+    {
+        REQUIRE(dogsCount >= 1);
+    }
+    // :snippet-start: delete-an-object
+    realm.write([&realm, &dog]
+                { realm.remove(dog); });
```
cbush, line 58: Hmm, can we configure the formatter to not align the next line with the open paren? This takes up a lot of horizontal real estate in the limited code example block on the site.

dacharyc: I couldn't spot an obvious option for this in the [Prettier configuration options](https://prettier.io/docs/en/options.html). The closest thing I found was to [ignore entire files](https://prettier.io/docs/en/ignore.html) using a .prettierignore at the root of the project which seems to function similar to a .gitignore. There's also a [Range option](https://prettier.io/docs/en/options.html#range) to format only a segment of a file, but you have to define the offsets. That seems pretty brittle and annoying to maintain as the files change. If you have a specific preference or recommendation here, I'm happy to adopt it.

cbush: I don't think you'd be using Prettier for a C++ file?

PR #2683, opened 2023-03-23
Add tvOS support

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-27760

### Staged Changes
https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/tvos/sdk/dotnet/compatibility/

### DIFF
--- a/source/sdk/dotnet/compatibility.txt
+++ b/source/sdk/dotnet/compatibility.txt
@@ -27,10 +27,10 @@ version(s) you can use with Realm on your target platforms:
    * - | Debian 8 or later
        | RHEL 7.1 or later
        | Ubuntu 16.04 or later
-     - .NET Core 2.0 or later (.NET 5.0 or later)
+     - .NET Core 2.0 or later (.NET Framework 5.0 or later)
 
    * - Windows 8.1 or later
-     - | .NET Core 2.0 or later (.NET 5.0 or later)
+     - | .NET Core 2.0 or later (.NET 5.0 Framework or later)
        | .NET Framework 4.6.1 or later
        | MAUI requires Windows 11 and Windows 10 version 1809 or higher, using 
          Windows UI Library (WinUI) 3.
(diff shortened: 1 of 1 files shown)


krollins-mdb (approved):

source/sdk/dotnet/compatibility.txt
```diff
@@ -50,6 +50,11 @@ version(s) you can use with Realm on your target platforms:
    * - Android
      - | Xamarin.Android for Android 4.1 (API level 16) or later. 
        | MAUI requires Android 5.0 (API level 21) or later.
+  
+   * - tvOS
+     - | Xamarin and Unity for tvOS 9.0 or later. 
+       | MAUI does not support tvOS.
```
nirinchev, line 56: We don't need to explicitly call that out - e.g. we don't do that for Linux or UWP.

source/sdk/dotnet/compatibility.txt
```diff
@@ -50,6 +50,11 @@ version(s) you can use with Realm on your target platforms:
...
+       | For more information on developing for tvOS, see :ref:`<ios-tvos>`.
```
nirinchev, line 57: This links to the swift page - not sure if we need to recreate it for .NET, but there's a swift code snippet at the bottom, which can be rewritten in C#.

nirinchev (commented):
A few minor comments. Also, completely unrelated to this page, but I saw the supported versions for desktop platforms say ".NET Core 2.0 or later (.NET 5.0 or later)", which is kind of weird - i.e. .NET 5.0 would be implicitly included in the "or later" part of .NET Core 2.0 - any reason we're calling it out explicitly?

source/sdk/dotnet/compatibility.txt
MongoCaleb: I think for now we won't recreate it, unless tvOS support becomes broad enough that we have people asking for code examples.

source/sdk/dotnet/compatibility.txt
MongoCaleb: @nirinchev  do we no longer need to differentiate between the .NET framework and .NET Core, which have different versioning schemes?

MongoCaleb:
@nirinchev do we no longer need to differentiate between the .NET framework and .NET Core, which have different versioning schemes?

source/sdk/dotnet/compatibility.txt
nirinchev: Sounds good!

nirinchev:
I think we should differentiate between .NET Framework and .NET Core, but not between .NET Core and .NET (i.e. the implication is that anything that supports .NET Core 2.0 supports .NET 5.0 and later).

PR #2692, opened 2023-03-27
(DOCSP-28190) Implement Andrew's RN SDK docs feedback - Round 1

krollins-mdb:
## Pull Request Info

Addresses Andrew's comments on the [Realm React Guidance feature branch](https://github.com/mongodb/docs-realm/pull/2442).

The Change an Object Model page's example tests needed to be completely re-written because the previous tests... didn't actually test anything.

### Jira

- https://jira.mongodb.org/browse/DOCSP-28190

### Staged Changes

- [Change an Object Model](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28190/sdk/react-native/realm-database/change-an-object-model/)

### Reminder Checklist

### Review Guidelines


examples/react-native/__tests__/ts/CRUD/read-test.tsx
```diff
@@ -120,6 +122,8 @@ describe('Read Data Tests', () => {
         '1 <= progressMinutes && progressMinutes < 10',
       );
 
+      console.log('lowProgressTasks', lowProgressTasks.length);
+
```
mongodben, line 126: this seems like it shouldn't be there

examples/react-native/__tests__/ts/CRUD/update-test.tsx
```diff
@@ -129,11 +132,10 @@ describe('Update Data Tests', () => {
     const CreateTaskItem = () => {
       const realm = useRealm();
 
-      let myTask: Realm.Object;
-      realm.write(() => {
+      const myTask: Realm.Object = realm.write(() => {
```
mongodben, line 135: i didn't know that the JS realm.write callback returns a value to realm.write()!

source/sdk/react-native/realm-database/change-an-object-model.txt
```diff
@@ -198,14 +198,14 @@ property, and then delete the old property.
  
          .. literalinclude:: /examples/generated/react-native/ts/change-an-object-model-test.snippet.rename-a-property-of-a-schema.tsx
            :language: typescript
-           :emphasize-lines: 2, 9, 18, 28
+           :emphasize-lines: 3, 10, 20, 30
```
mongodben, line 201:
```suggestion
           :emphasize-lines: 3, 11, 20, 30
```

i think you wanna move this one down a line 

Screen Shot 2023-03-28 at 3 57 52 PM

examples/react-native/__tests__/ts/realm-database/change-an-object-model-test.tsx
```diff
@@ -105,49 +202,65 @@ describe('Change an Object Model Tests', () => {
...
     // :snippet-start: modify-a-property-type
-    // :replace-start: {
-    //  "terms": {
-    //   "MyTask": "Task"
-    //   }
-    // }
-    class MyTask extends Realm.Object<MyTask> {
+
+    class Person extends Realm.Object<Person> {
       _id: Realm.BSON.ObjectId = new Realm.BSON.ObjectId();
```
mongodben, line 239: i think you shouldn't set a default value here as it currently does. the JS example doesn't have this.

```suggestion
      _id: Realm.BSON.ObjectId;
```

mongodben (approved):
lgtm w a few small things to address before merge 👍

examples/react-native/__tests__/ts/realm-database/change-an-object-model-test.tsx
```diff
@@ -1,23 +1,80 @@
...
+//       "schemaVersion: 3": "schemaVersion: 2",
+//       "schemaVersion: 4": "schemaVersion: 2",
+//       "schemaVersion: 5": "schemaVersion: 2",
+//       "oldRealm.schemaVersion < 4": "oldRealm.schemaVersion < 2",
+//       "oldRealm.schemaVersion < 5": "oldRealm.schemaVersion < 2",
+//        "RealmProvider, useRealm": "RealmProvider"
+//    }
+// }
+
+class Person extends Realm.Object<Person> {
```
takameyer, line 21:
```suggestion
class Person extends Realm.Object {
```

I'm going to clear this up with the team now, but I believe since we are not documenting creating objects with the constructor, we might be able to remove this generic declaration

takameyer: I looked into the types found some small typing issues with not including `<Person>`, though it seems possible that we could fix our types to make it work without.
One thing happens is that when you do:
```
person.addListener((obj, changes) => {
 //...
});
```
That `obj` would be typed as `unknown` instead of `Person`.  `linkingObject` method is also effected by this.

Another thing to keep in mind is that we have an `RequiredTypes` generic as well (which answers one of your questions elsewhere).
```
class Person extends Realm.Object<Person, 'name' | 'age'> {
//...
```
This is used since currently when creating an object, all the fields are optional.  The `RequiredTypes` say which fields are non-optional and will show a type error if they are not included.  For the example above, you would get a type error if name or age were not defined when trying to create an object.  Unfortunately, it's only doing this on the "constructor" creation (using `new Person(...)`) and no on the `create` method (using `realm.create(Person...)`).  

That being said, I suggest, we keep the generic in.  I'll make open an issue to clean this up.

krollins-mdb: Thanks for sharing this info! We'll keep the generic then. I think I'll also experiment with a dedicated TypeScript page that explains things like this.

PR #2697, opened 2023-03-28
(DOCSP-27850): Flutter Upgrade Write to a Synced Realm page

mongodben:
## Pull Request Info

Flutter Upgrade Write to a Synced Realm page

### Jira

- https://jira.mongodb.org/browse/DOCSP-27850

### Staged Changes

- [Write to a Synced Realm (Flutter)](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-27850/sdk/flutter/sync/write-to-synced-realm/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/flutter/sync/write-to-synced-realm.txt
+++ b/source/sdk/flutter/sync/write-to-synced-realm.txt
@@ -134,7 +134,7 @@ subscription or does not match the user's permissions, the following occurs:
 #. Upon sync, the server applies the rules and permissions.
    The server determines that the user does not have authorization to perform the write.
 #. The server sends a revert operation, called a "compensating write", back to the client.
-#. The client's realm reverts the illegal write operation by deleting the illegal object from the realm.
+#. The client's realm reverts the illegal write operation.
 
 Any client-side writes to a given object between an illegal write to that object
 and the corresponding compensating write will be lost.
@@ -164,8 +164,7 @@ If you want to write an object that does not match the query subscription,
 open a different realm where the object matches the query
 subscription. Alternately, you could write the object to a :ref:`local realm
 <flutter-open-local-only-realm>` that does not enforce
-permissions or subscription queries. There is no way to write an object
-to a Flexible Sync realm that does not match the query subscription.
+permissions or subscription queries.
 
 Code Example
 ````````````
--- a/source/sdk/flutter/sync/write-to-synced-realm.txt
+++ b/source/sdk/flutter/sync/write-to-synced-realm.txt
@@ -16,8 +16,28 @@ in behavior to keep in mind as you develop your application.
 To learn more about reading and writing data to a realm,
 refer to :ref:`Read & Write Data <flutter-read-write-data>`.
 
-Examples on This Page
----------------------
+When you write to a synced realm, your write operations must match *both*
+of the following:
+
+- **The sync subscription query.**
+    - If your write operation doesn't match the query in the subscription,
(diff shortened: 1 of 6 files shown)


source/sdk/flutter/sync/write-to-synced-realm.txt
```diff
@@ -87,21 +114,167 @@ and other Device Sync error types, refer to :ref:`sync-errors` in the App Servic
...
+subscription or does not match the user's permissions, the following occurs:
 
 #. Because the client realm has no concept of "illegal" writes,
    the write initially succeeds until realm resolves the changeset
    with the App Services backend.
 #. Upon sync, the server applies the rules and permissions.
    The server determines that the user does not have authorization to perform the write.
 #. The server sends a revert operation, called a "compensating write", back to the client.
-#. The client's realm reverts the illegal write operation.
+#. The client's realm reverts the illegal write operation by deleting the illegal object from the realm.
```
dacharyc, line 137: I suppose technically, this wouldn't _have_ to be deleting the illegal object from the realm. It could also be reverting an illegal modification, maybe? I'm not sure I've tested this scenario. It's probably more accurate to say that it reverts the illegal write operation by... rejecting the changeset and reverting the object state to the last permitted state? That's awkward to say, though. Open to ideas but I think "deleting the illegal object" is not the only possible revert operation.

source/sdk/flutter/sync/write-to-synced-realm.txt
```diff
@@ -87,21 +114,167 @@ and other Device Sync error types, refer to :ref:`sync-errors` in the App Servic
...
+subscription query. If you perform a write that does not match the
+subscription query, Realm initially writes the object, but then performs
+a compensating write. This is a non-fatal operation that
+reverts an illegal write that does not match the subscription query.
+
+If you want to write an object that does not match the query subscription,
+open a different realm where the object matches the query
+subscription. Alternately, you could write the object to a :ref:`local realm
+<flutter-open-local-only-realm>` that does not enforce
+permissions or subscription queries. There is no way to write an object
```
dacharyc, line 167: Instead of saying "There is no way to write an object... that does not match the query subscription" here, I wonder if we would want to say instead something like "Any attempt to write an object to a Flexible Sync realm that does not match the query subscription will be reverted." Technically, you can write objects that don't match the subscription all day long and it doesn't actually prevent you writing it, but just reverts the writes when it gets the message back from the server. (And if this is the wording in Swift, I think we should probably tweak it on that page, too, because it's not _entirely_ correct.)

dacharyc (approved):
Overall, I think this LGTM. There are a couple of small wording Qs where I don't think we're being entirely accurate (and if that wording is in the Swift page we should probably correct it there, too).

source/sdk/flutter/sync/write-to-synced-realm.txt
mongodben: i think it's fine to just say 'The client's realm reverts the illegal write operation' w/o additional text

source/sdk/flutter/sync/write-to-synced-realm.txt
mongodben: actually sentence can be deleted b/c it just repeats content of previous paragraph

dacharyc (approved):
Awesome - looks good. Thank you for applying these changes to the Swift page, too!

PR #2724, opened 2023-04-10
(DOCSP-29048): Swift: Update docs for encrypted realm multiprocess support

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-29048

### Staged Changes

- [Encrypt a Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29048/sdk/swift/realm-files/encrypt-a-realm/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/swift/realm-files/encrypt-a-realm.txt
+++ b/source/sdk/swift/realm-files/encrypt-a-realm.txt
@@ -61,9 +61,18 @@ to create a 64-bit key and store that key in a :ref:`user object <user-objects>`
 Accessing an Encrypted Realm from Multiple Processes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. include:: /includes/encrypt-mult-processes.rst
+.. versionchanged:: 10.38.0
 
-As an alternative, you can store data that you want to encrypt as 
+
+Prior to Realm Swift SDK version 10.38.0, Realm did not support opening
+the same encrypted realm from multiple processes. Attempting to do so 
+throws this error:
+``Encrypted interprocess sharing is currently unsupported.``
+
+If your app uses Realm Swift SDK version 10.37.2 or older and needs to
+access a realm from multiple processes, use an unencrypted realm.
+
+Alternately, you can store data that you want to encrypt as 
 ``NSData`` properties on realm objects. Then, you can encrypt and 
 decrypt individual fields. 
 
--- a/source/sdk/swift/realm-files/encrypt-a-realm.txt
+++ b/source/sdk/swift/realm-files/encrypt-a-realm.txt
@@ -63,18 +63,19 @@ Accessing an Encrypted Realm from Multiple Processes
 
 .. versionchanged:: 10.38.0
 
+Starting with Realm Swift SDK version 10.38.0, Realm supports opening
+the same encrypted realm in multiple processes. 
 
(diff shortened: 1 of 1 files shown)


source/sdk/swift/realm-files/encrypt-a-realm.txt
```diff
@@ -61,9 +61,18 @@ to create a 64-bit key and store that key in a :ref:`user object <user-objects>`
...
+.. versionchanged:: 10.38.0
 
-As an alternative, you can store data that you want to encrypt as 
+
+Prior to Realm Swift SDK version 10.38.0, Realm did not support opening
+the same encrypted realm from multiple processes. Attempting to do so 
+throws this error:
+``Encrypted interprocess sharing is currently unsupported.``
+
+If your app uses Realm Swift SDK version 10.37.2 or older and needs to
```
krollins-mdb, line 72: I believe we should use earlier/later when referring to versions instead of older/newer. I couldn't find a direct reference in the style guide, but it does recommend [avoiding before/after](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#b) and prefers earlier/later (look for the "before" entry).

older/newer has an inverse relationship with the actual version numbers. Smaller numbers == older, which goes against how most folks think about time and numbers. "earlier/later" doesn't have that issue.

```suggestion
If your app uses Realm Swift SDK version 10.37.2 or earlier and needs to
```

krollins-mdb: Not sure if the team has discussed this before, but we could bring it up in lobby if it seems appropriate.

source/sdk/swift/realm-files/encrypt-a-realm.txt
```diff
@@ -61,9 +61,18 @@ to create a 64-bit key and store that key in a :ref:`user object <user-objects>`
 Accessing an Encrypted Realm from Multiple Processes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. include:: /includes/encrypt-mult-processes.rst
+.. versionchanged:: 10.38.0
 
-As an alternative, you can store data that you want to encrypt as 
+
+Prior to Realm Swift SDK version 10.38.0, Realm did not support opening
```
krollins-mdb, line 67: I know this was a quick update, so maybe this suggestion falls out of scope.

The way this currently reads is still predominantly that this feature isn't supported. The opening sentence implies that v10.38.0 supports it, but doesn't exactly call it out. Then the rest of the section is all about pre-v10.38.0 and the feature not being supported.

Consider starting this section with a confirmation that the feature works in v10.38.0 (and maybe an example?). Then have a subsection that clearly specifies what to do if readers aren't on v10.38.0 or later.

krollins-mdb (approved):
LGTM, with a couple questions.

source/sdk/swift/realm-files/encrypt-a-realm.txt
dacharyc: I don't have strong feelings about this so I'm happy to change it here. But I agree it would be good to get consensus, because I'm sure there are a ton of places I've used "older" (it's just the way my brain thinks about it) and I may forget and do it again if we don't have an official policy on it. 

Even in just making the change you suggested here, I tweaked this mention but went on to use "older" immediately in a new sentence. 🤦‍♀️

PR #2726, opened 2023-04-10
(DOCSP-28190) Andrew's RN SDK docs feedback - Round 4: Smaller item roundup

krollins-mdb:
### Jira

### Staged Changes

### Review Guidelines


source/sdk/react-native/realm-database/schemas/uuid.txt
```diff
@@ -15,7 +15,10 @@ Overview
 --------
 
 ``UUID`` (Universal Unique Identifier) is a 16-byte :wikipedia:`unique value
-<Universally_unique_identifier>`. You can use ``UUID`` as an identifier for
+<Universally_unique_identifier>`. ``UUID`` is bundled with the Realm package as
+part of BSON (``Realm.BSON.UUID``).
+
+You can use ``UUID`` as an identifier for
```
takameyer, line 21:
```suggestion
You can use ``UUID`` as an unique identifier for
```

source/sdk/react-native/realm-database/schemas/mixed.txt
```diff
@@ -120,4 +121,12 @@ The ``CatInfoCard`` component does the following:
...
 
+Mixed Properties and Types
+--------------------------
+
+Because mixed properties can be more than one type, you can't rely on the
+property's value being a specific type.
 
+To get a mixed property's type, you can use :js-sdk:`Object.getPropertyType()
+<Realm.Object.html#getPropertyType>`. This method gives you the underlying type
+of a property value.
```
takameyer, line 132: Maybe an example would be valuable

takameyer (approved):
Just a couple suggestions!

examples/react-native/__tests__/ts/realm-database/schemas/mixed-test.tsx
```diff
@@ -29,13 +29,18 @@ describe('Mixed Tests', () => {
         name: 'Clover',
         birthDate: new Date('January 21, 2016'),
       });
+      assertionRealm.create('Cat', {
+        name: 'Yoshi',
+        birthDate: 'November 22nd, 2020',
```
krollins-mdb, line 34: @takameyer, do we have a way to define a realm type for JS Numbers? Like realm.defineAsInt(<number>)?

When I tried the below example, Realm cast `birthDate` to a Double, when I would have expected an Int.
Because Realm translates JS Number into Int, Float, and Double, do we have a way to tell Realm which one it should be?

``` js
realm.create("Cat", {
  name: "Yoshi",
  birthDate: 11222020,
});
```

source/sdk/react-native/realm-database/schemas/mixed.txt
krollins-mdb: @takameyer, I added an example. Would you mind taking a look?

examples/react-native/__tests__/ts/realm-database/schemas/mixed-test.tsx
takameyer: From what I remember, `mixed` defaults to double and there currently isn't a way to make it do otherwise.  I think this was done, as it would be the most flexible digit type to use, when we are not sure what it should be.  
Perhaps we could make a feature request to open the discussion.

source/sdk/react-native/realm-database/schemas/mixed.txt
takameyer: The example looks great!
However, the highlighting seems to be a bit off in the staged document.

examples/react-native/__tests__/ts/realm-database/schemas/mixed-test.tsx
krollins-mdb: Ah, that's good to know. Probably good to mention in the docs if `mixed` does indeed default to Double on the realm side.

source/sdk/react-native/realm-database/schemas/mixed.txt
krollins-mdb: Thanks for pointing that out. Fixed it.

takameyer (approved):
LGTM!

source/sdk/react-native/realm-database/schemas/mixed.txt
```diff
@@ -11,22 +11,22 @@ Mixed - React Native SDK
...
 dictionaries) of type ``mixed``, but a ``mixed`` type itself cannot be a
 collection. 
 
-The mixed data type is indexable, but you can't use it as a primary key.
+The Mixed type is indexable, but you can't use it as a primary key.
 
-Properties using the mixed data type can also hold null values. You can't
-declare a Mixed property as optional.
+Properties using the Mixed type can hold null values and cannot be defined
+as optional. All JavaScript ``Number``s in a Mixed type are
```
dacharyc, line 22: This needs a formatting fix - unfortunately, rST doesn't like it when you put a letter next to double backticks, so this  is causing the formatting on the staged page to have an issue. I'm not sure if `types` is the best way of addressing this here, but wanted to add a suggestion for convenience.

```suggestion
as optional. All JavaScript ``Number`` types in a Mixed type are
```

source/sdk/react-native/realm-database/schemas/mixed.txt
```diff
@@ -11,22 +11,22 @@ Mixed - React Native SDK
...
+mapped to a Realm ``double`` type.
 
 Realm Object Models
 -------------------
+
 To :ref:`set a property of your object model
-<react-native-define-a-realm-object-schema>` as ``Mixed``, set the property's type to
+<react-native-define-a-realm-object-schema>` as Mixed, set the property's type to
 "``mixed``".
```
dacharyc, line 30: I know you're not touching this line in this PR, but since you _are_ removing the monospace from "Mixed" in ln 29 here, it's probably worth considering this case. It struck me as odd to have both monospace _and_ quotes. I can't find any examples showing this in our style guide - I think just monospace is sufficient?

```suggestion
``mixed``.
```

source/sdk/react-native/realm-database/schemas/mixed.txt
```diff
@@ -120,4 +122,16 @@ The ``CatInfoCard`` component does the following:
...
+Because Mixed properties can be more than one type, you can't rely on the
+property's value being a specific type.
+
+With :js-sdk:`Object.getPropertyType() <Realm.Object.html#getPropertyType>`, you
+can get a Mixed property's underlying type. This allows you build your own type
+checking.
 
+.. literalinclude:: /examples/generated/react-native/ts/mixed-test.snippet.type-check.tsx
+   :language: javascript
+   :emphasize-lines: 2-9, 17-19
```
dacharyc, line 137: I think the emphasize here is a little off. Probably a good case for using Bluehawk emphasize.

```suggestion
   :emphasize-lines: 3-9, 16-18
```

source/sdk/react-native/sync-data/stream-data-to-atlas.txt
```diff
@@ -44,11 +49,3 @@ collection with its ``asymmetric`` property set to ``true``.
          :language: javascript
          :emphasize-lines: 5
          :linenos:
-
-
-.. note:: Attempting to Read Asymmetric Objects
```
dacharyc, line 49: I'm curious about removing this note. It seems worth stating that you can't add a flexible sync subscription query for an asymmetric object type, since we say in an `Important` callout on the [Configure a Synced Realm page](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28190-round4/sdk/react-native/sync-data/configure-a-synced-realm/#configure--realmprovider-) that Flexible Sync requires a subscription. Asymmetric objects are a weird exception to that in that you _cannot_ subscribe to an asymmetric object type (at least in Swift - not sure what the behavior is in JS). If this is similar in JS, it's probably worth noting this on the page.

dacharyc (approved):
Overall LGTM, but I had a couple of Qs and I spotted a couple of formatting things that need a fix before we merge.

source/sdk/react-native/realm-database/schemas/mixed.txt
krollins-mdb: Thanks for pointing this out! There is only one ``Number`` type in JavaScript. I'll rewrite to avoid this formatting issue.

source/sdk/react-native/sync-data/stream-data-to-atlas.txt
krollins-mdb: I added a bit more context, using "query" and "subscription", as that's how we typically talk about sync things.

dacharyc: I think the updated verbiage is still potentially ambiguous or misleading. You _can_ query for this object type using MongoDB Remote Access, GraphQL or Data API - you just can't query it in a sync subscription. So I think "subscribe or query" is probably not the right way to characterize this behavior. 

Maybe instead just something like "you cannot add a Flexible Sync subscription for an asymmetric object type."

krollins-mdb: Excellent point. Thank you!

PR #2730, opened 2023-04-13
Update Kotlin test suite to 1.7.0

cbullinger:
## Pull Request Info

Update to Kotlin SDK v1.7.0
- convert val to var for any persisted fields

### Staged Changes

- [staged build](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/no-jira-update-kotlin-test-suite/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/RQLTest.kt
+++ b/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/RQLTest.kt
@@ -1,5 +1,6 @@
 package com.mongodb.realm.realmkmmapp
 
+import io.realm.kotlin.ext.realmListOf
 import io.realm.kotlin.types.RealmList
 import io.realm.kotlin.types.RealmObject
 import io.realm.kotlin.types.annotations.PrimaryKey
@@ -21,7 +22,7 @@ class RQLTest: RealmTest() {
         @PrimaryKey
         var _id: ObjectId = ObjectId()
         var name: String = ""
-        var items: RealmList<Item>? = null
+        var items: RealmList<Item> = realmListOf<Item>()
         var quota: Int? = null
     }
     // :snippet-end:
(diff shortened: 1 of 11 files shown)


examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/RQLTest.kt
```diff
@@ -20,8 +20,8 @@ class RQLTest: RealmTest() {
     class Project(): RealmObject {
         @PrimaryKey
         var _id: ObjectId = ObjectId()
-        lateinit var name: String
-        lateinit var items: RealmList<Item>
+        var name: String = ""
+        var items: RealmList<Item>? = null
```
dacharyc, line 24: As a point of curiosity since you're changing this here, is there a reason we're making this an optional list and setting it to null, instead of just initializing it as an empty list?

In the [to-many relationship example](https://www.mongodb.com/docs/realm/sdk/kotlin/realm-database/schemas/relationships/#to-many-relationship), we're initializing an empty list property:

`val fishes: RealmList<Fish> = realmListOf<Fish>()`

In the [inverse relationships example](https://www.mongodb.com/docs/realm/sdk/kotlin/realm-database/schemas/relationships/#inverse-relationships), we're initializing an optional list property and setting it to null, similar to what you're doing here:

`var posts: RealmList<Post>? = null`

I'm wondering if there are performance or optional unwrapping considerations for either of these forms of initialization, and if initializing it this way is the right choice in this case?

dacharyc (approved):
Overall LGTM. I do have a Q about a change we're making to a list property here, but I see both types of initialization in the docs examples so I'm assuming the test works after this change and this is more of a non-blocking Q.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/RQLTest.kt
cbullinger: That's an excellent question. I can confirm re. performance or unwrapping considerations.

PR #2737, opened 2023-04-17
(DOCSP-29286): Node.js: Update content type IA to topic-based IA

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-29286

### Staged Changes

- [Node.js SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29286/sdk/node/)

### Reminder Checklist

### Review Guidelines


source/sdk/node/realm-database.txt
```diff
@@ -1,16 +1,48 @@
-.. _node-realm-database-reads:
-.. _node-live-queries:
+.. _node-realm-database:
 
-==========================
-Live Queries - Node.js SDK
-==========================
+============================
+Realm Database - Node.js SDK
+============================
```
krollins-mdb, line 5: This page doesn't seem to be in the SDK's TOC. We could reframe it as "Why Realm Database?" and put it at the top, like the Swift SDK.

source/sdk/node/model-data.txt
```diff
@@ -0,0 +1,110 @@
...
+Model Data - Node.js SDK
+========================
+
+.. toctree::
+   :titlesonly:
+
+   Define an Object Model </sdk/node/model-data/define-a-realm-object-model>
+   Relationships & Embedded Objects </sdk/node/model-data/relationships-and-embedded-objects>
+   Change an Object Model </sdk/node/model-data/modify-an-object-schema>
+   Data Types </sdk/node/model-data/data-types>
```
krollins-mdb, line 11: I think we're missing the "Modify an Object Schema" page.

source/sdk/node/model-data/define-a-realm-object-model.txt
```diff
@@ -10,6 +10,11 @@ Define a Realm Object Schema - Node.js SDK
    :depth: 2
    :class: singlecol
 
+.. _node-define-realm-object-type:
+
```
krollins-mdb, line 14: We should change the page title to match the TOC: "Define a Realm Object Model".

source/sdk/node/sync/flexible-sync.txt
krollins-mdb, line 1: We should update this page's title to match the TOC: "Manage Flexible Sync Subscriptions".

source/sdk/node/sync/reset-a-client-realm.txt
krollins-mdb, line 1: We should update this page's title to match the TOC: "Client Resets".

source/sdk/node.txt
```diff
@@ -14,16 +14,18 @@ Realm Node.js SDK
...
+   Manage Users </sdk/node/manage-users>
+   Sync Data </sdk/node/sync>
    Integration Guides </sdk/node/integrations>
-   Advanced Guides </sdk/node/advanced>
    SDK Telemetry </sdk/node/telemetry>
    API Reference <https://www.mongodb.com/docs/realm-sdks/js/latest/>
-   Upgrade from Stitch to Realm </sdk/node/migrate>
    Release Notes <https://github.com/realm/realm-js/releases>
-
+   Upgrade from Stitch to Realm </sdk/node/migrate>
```
krollins-mdb, line 28: I suggest we orphan this page. It's been long enough that I don't think this content needs to be particularly discoverable.

krollins-mdb (commented):
Thanks, @dacharyc! I appreciate you working on this update. Just a few things to address.

source/sdk/node/realm-database.txt
dacharyc: D'oh, good catch, that is what I meant to do!

source/sdk/node/model-data.txt
dacharyc: Actually, it's "Change an Object Model" here. Do you have a preference on the name? I went with "Change an Object Model" in Swift because it seemed to be plainer/simpler language than "Modify an Object Schema" - I renamed it in the ToC here but left the page title and page name for the old version of the page. Which do you like better?

source/sdk/node/model-data/define-a-realm-object-model.txt
dacharyc: Hmm. I left this title as a hedge. All of the examples on the page are defining a schema, so I wondered if that was the preferred language to use in the JS SDKs vs. the more generic "Object Model." Do you use these terms interchangeably in RN? Arguably I should use Schema in the ToC, but I was just copying across from one of the other SDKs and didn't think too much about the name.

krollins-mdb: I've definitely been moving towards using "object model" in the RN docs. We have too many meanings of "schema", IMO.

But, after thinking about this some more, we could also potentially go with "object type". The first section on this page says "Define a Realm Object Type". lol.

For now, I think matching the main TOC and page title is all that matters. That way readers don't click a link and see a different title, thinking they somehow didn't go where they intended. I don't particularly care which form we go with right now.

Long term, we should re-visit "object schema" vs "object model" vs "object type" and their nuances. But for all the SDK docs.

dacharyc: That's fair. Alright, I'll go with "Object Model" since it's more consistent with the "Change an Object Model" page naming.

PR #2742, opened 2023-04-19
fix TS test + add missing JS asserts

nlarew:


examples/node/Examples/mongodb.js
```diff
@@ -22,8 +22,15 @@ const app = new Realm.App({ id: "example-testers-kvjdy" });
 
 function getPlantsCollection() {
   // :snippet-start: plants-collection-handle
+  // :replace-start: {
+  //   "terms": {
+  //     "mongodb-atlas": "mongodb-atlas",
```
krollins-mdb, line 27: I don't think we need this replace? Same thing in the .ts file.

krollins-mdb (approved):
LGTM! One minor thing to look at.

PR #2746, opened 2023-04-19
(DOCSP-28989) React Native IA update

krollins-mdb:
## Pull Request Info

In addition to the work in the Jira ticket, I've updated old React Native SDK redirects so that we're not chaining.

### Jira

- https://jira.mongodb.org/browse/DOCSP-28989

### Staged Changes

- [Main landing page](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28989/sdk/react-native/)

### Review Guidelines


source/sdk/react-native.txt
```diff
@@ -12,19 +12,21 @@ Realm React Native SDK
...
    Bootstrap with Expo </sdk/react-native/bootstrap-with-expo>
-   Realm Database </sdk/react-native/realm-database>
+   Realm Files </sdk/react-native/realm-files>
+   Model Data </sdk/react-native/model-data>
+   CRUD </sdk/react-native/crud>
+   React to Changes </sdk/react-native/react-to-changes>
    Atlas App Services </sdk/react-native/app-services>
    Manage Users </sdk/react-native/manage-users>
    Sync Data </sdk/react-native/sync-data>
-   Test & Debug </sdk/react-native/test-and-debug/>
```
dacharyc, line 22: It looks like we've removed Test & Debug from the ToC here, but I don't see any changes to those files - we don't seem to be getting rid of the pages. Was it moving but got lost along the way? Or did I miss where it moved to?

source/sdk/react-native.txt
```diff
@@ -12,19 +12,21 @@ Realm React Native SDK
...
    Integration Guides </sdk/react-native/integrations>
-   SDK Telemetry </sdk/react-native/telemetry>
    @realm/react Reference </sdk/react-native/use-realm-react>
-   Realm Node.js API Reference <https://www.mongodb.com/docs/realm-sdks/js/latest/>
-   Upgrade from Stitch to Realm </sdk/react-native/migrate>
```
dacharyc, line 27: [boulder] If we're intentionally removing the "Upgrade from Stitch to Realm" page but not deleting it outright, we should intentionally orphan this page. There are a couple of steps to do that properly:

- Rename the `migrate` page to `index.txt` and move it into a directory called `migrate` (the former page name). This prevents the "Find Unused" checker from seeing it as an unused file - it's a special pattern Bush built into the checker to let us intentionally orphan files.
- Add a redirect for the `migrate` page to its new `migrate/index/` location (and I realize now I think I forgot to do that for Node and Java, so maybe you could add redirects for those two, also?)
- Add markup to the top of the page itself so it won't be indexed and won't show up in the previous/next breadcrumbs:
```
:noprevnext:
:orphan:

.. meta::
   :robots: noindex, nosnippet
```

source/sdk/react-native.txt
```diff
@@ -12,19 +12,21 @@ Realm React Native SDK
...
+   SDK Telemetry </sdk/react-native/telemetry>
```
dacharyc, line 28: [sand] Should we keep the two API reference links together? I thought it was odd to see the realm/react Reference, then some unrelated page, then the JS API reference. (We should probably decide how we want to present multiple API reference links in the ToC and unify. Swift also has two API reference links - Swift and Objective-C - and I've put them together under an `API Reference` link in the ToC, which does add a level of nesting but the ToC was getting pretty long.)

source/sdk/react-native/app-services/connect-to-app-services-app.txt
```diff
@@ -10,9 +10,6 @@ Connect to an Atlas App Services App - React Native SDK
    :depth: 2
    :class: singlecol
 
-Overview
```
dacharyc, line 13: Ooo nice de-Overview-ing! :)

source/sdk/react-native/crud/read.txt
```diff
@@ -10,8 +10,17 @@ CRUD - Read - React Native SDK
    :depth: 2
    :class: singlecol
 
-About the Examples on this Page
```
dacharyc, line 13: [pebble] It looks like we're removing the `About the Examples on this Page` here, but the rest of the pages in this section have kept this header. (Although the "Query Data" page is just `Examples on this Page`.) Should we be consistent in our application of this header?

source/sdk/react-native/manage-users.txt
```diff
@@ -11,4 +11,100 @@ Manage Users - React Native SDK
...
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+When you use Atlas App Services to back your client app, you get access to a
+:ref:`user object <user-objects>`. Use Swift SDK methods with this user 
+object to conveniently:
```
dacharyc, line 24: Oops, copypasta error. (I wonder if you got this from the Node.js page, in which case, sorry! I noticed this error when I did the Java updates but fixed it in Node when I made the Java PR. But I don't know if there's a good analog for "Swift SDK methods" in RN - react hooks? Or just genericize like so?)

```suggestion
:ref:`user object <user-objects>`. With this user object,
you can:
```

source/sdk/react-native/model-data.txt
```diff
@@ -0,0 +1,110 @@
+.. _react-native-model-data:
+
+=============================
+Model Data - React Native SDK
+=============================
+
+.. toctree::
+   :titlesonly:
+
+   Define an Object Model </sdk/react-native/model-data/define-a-realm-object-model>
```
dacharyc, line 10: I really like this organization compared with what's currently live - this is a nice improvement! And I think you've chosen well to re-integrate some of those smaller pages in "Schemas" - the way this is chunked feels more tractable, I think. Nice work!

source/sdk/react-native/realm-database/schemas/overview.txt
```diff
@@ -1,49 +0,0 @@
-.. _react-native-schemas-overview:
```
dacharyc, line 1: It's possible I missed it, but I didn't see this ref target moved to one of the pages where I might expect to see it (model data or define a realm object model). If it has gotten lost entirely, we should add it back somewhere just in case there's a ref somewhere looking for this target.

source/sdk/react-native/model-data/change-an-object-model.txt
```diff
@@ -43,8 +40,8 @@ in a ``migration`` function. This might include changes such as:
 
 .. _react-native-schema-version:
 
-Key Concept: Schema Version
```
dacharyc, line 46: [sand] I don't mind removing it, but there are places elsewhere in the docs where I used "Key Concept" to indicate purely concept-related content that moved from the old Fundamentals stuff. I'll add a note to our sync doc to discuss whether we want to remove this from the docs entirely, and if so, we can cut a ticket or tickets to do this in the future.

source/sdk/react-native/model-data/data-types.txt
```diff
@@ -0,0 +1,30 @@
+.. _react-native-data-types:
```
dacharyc, line 1: [dust] I think we probably don't want to put ref targets on container pages. This is more like a "this page has the `doc` links _in case_ someone finds their way here," not that "we want to intentionally send people to a container page." If we want a handle for RN data types, I'd probably put this on the higher-level model-data page, or maybe the property types page.

source/sdk/react-native/model-data/define-a-realm-object-model.txt
```diff
@@ -0,0 +1,262 @@
...
+         :linenos:
+
+   .. tab::
+      :tabid: javascript
+
+      .. literalinclude:: /examples/generated/react-native/js/Task.snippet.js-task-schema.js
+         :language: javascript
+         :emphasize-lines: 5, 11
+         :linenos:
+.. _react-native-index-a-property:
```
dacharyc, line 139: [dust] This doesn't appear to have caused an issue, but I'd probably put a newline before this ref target.

dacharyc (changes requested):
This looks great, Kyle! I've scattered a handful of comments throughout, but only a few are blocking changes. I've tried using the Netlify Feedback Ladder to indicate the priority of the feedback, except in a few places where it's just a question. Let me know if this was helpful!

source/sdk/react-native.txt
krollins-mdb: Thanks for catching this mistake. It should indeed be here. Was testing different IAs and forgot to move this back in.

source/sdk/react-native.txt
krollins-mdb: Oh, this is definitely good to know. Thanks, Dachary!

krollins-mdb: I added redirects for the node and java /migrate/index files.

source/sdk/react-native.txt
krollins-mdb: Good point. We should definitely create a new section. I wasn't really considering the @realm/react link an API reference, but it does server that purpose. And they're working on building out proper API docs.

I'll add an `API Reference` section. Seems like a good, expandable pattern.

source/sdk/react-native/crud/read.txt
krollins-mdb: Ah, yeah. I removed it for this page because the in-page TOC just seemed bizarre. I tried to avoid making these page-level changes in this PR, but couldn't help myself on this one.

We should definitely be consistent. I'll look at the other pages.

source/sdk/react-native/manage-users.txt
krollins-mdb: Shoot. I thought I fixed this. Maybe I did in a different spot. Thanks for catching!

source/sdk/react-native/model-data/change-an-object-model.txt
krollins-mdb: Yeah, let's discuss more. I'm not totally opposed to it, but it felt unnecessary.

source/sdk/react-native/api-reference.txt
```diff
@@ -0,0 +1,9 @@
+================================
+API Reference - React Native SDK
+================================
+
+.. toctree::
+   :titlesonly:
+
+   API Reference <https://www.mongodb.com/docs/realm-sdks/js/latest/>
```
dacharyc, line 8: [dust] Seeing "API Reference" inside an "API Reference" ToC entry (API Reference -> API Reference) confused me for a minute. Maybe we should rename to JS API Reference or SDK API Reference or... something - to more easily discern its purpose without having to click it?

dacharyc (approved):
One more tiny but non-blocking comment, but otherwise LGTM! Awesome work! 🎉

PR #2756, opened 2023-04-25
Add nullability info to .NET docs

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-27348

### Staged Changes

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/nullability/sdk/dotnet/model-data/define-object-model/#required-and-optional-properties

This PR also includes moving the Data Binding MVVM to a more appropriate location (https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/nullability/sdk/dotnet/model-data/data-binding/#data-binding-and-mvvm). There are no content changes.


source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,51 @@ Schema, you can update your configuration to include the specific classes you wa
...
+.. _dotnet-optional-field:
+.. _dotnet-required-optional-property:
+
+Required and Optional Properties
+--------------------------------
 
-However, when you bind UI elements to a :dotnet-sdk:`RealmObject <reference/Realms.RealmObject.html>`, 
-this notification happens automatically, thus greatly simplifying your ViewModel code. 
+In C#, value types are implicitly non-nullable, but can be made optional by appending 
+`?  <https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types>`__.
```
dacharyc, line 255: Nit: having just the `?` here as the source of a link looks odd to me on the page. I wonder if, in this case, we want to just make this monospace and then add a line explicitly calling out what the link entails? Ala:

```suggestion
In C#, value types are implicitly non-nullable, but can be made optional by appending ``?``.
For more information, refer to the `Nullable Value Types C# Reference documentation. <https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types>`__
```

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,51 @@ Schema, you can update your configuration to include the specific classes you wa
...
-interface.
+Beginning with C# 8.0, new projects are run in a 
+`nullable aware context <https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types>`__.
+In this case, reference types, such as ``string`` and ``byte[]``, must also be declared 
+as nullable with ``?``.
 
-For example, if you bind a ListView to a query that returns all Product 
-objects, the list of available products updates automatically 
-when a product is added, changed, or removed from the realm:
+The Realm .NET SDK fully supports the nullable aware context and uses nullability
```
dacharyc, line 262:
```suggestion
The Realm .NET SDK fully supports the nullable-aware context and uses nullability 
```

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,51 @@ Schema, you can update your configuration to include the specific classes you wa
...
+`?  <https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types>`__.
 
-You can also bind a UI element to a realm :ref:`Live Query <dotnet-live-queries>`. 
-For example, if you bind a 
-`ListView <https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.listview?view=xamarin-forms>`_ 
-to a live query, then the list will update automatically when the results of the 
-query change; you do not need to implement the ``INotifyPropertyChange`` 
-interface.
+Beginning with C# 8.0, new projects are run in a 
+`nullable aware context <https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types>`__.
```
dacharyc, line 258: Nit: I see that the C# docs are missing it, but I think "nullable-aware" should be hyphenated in this case because it's a compound adjective that modifies context.

```suggestion
`nullable-aware context <https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types>`__.
```

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,51 @@ Schema, you can update your configuration to include the specific classes you wa
...
-when a product is added, changed, or removed from the realm:
+The Realm .NET SDK fully supports the nullable aware context and uses nullability 
+to determine whether a property is required or optional. The SDK has the 
+following rules:
 
-.. code-block:: csharp
-   :copyable: false
+- Realm assumes that both value and reference type properties are required if you 
+  do not designate them as nullable. If you designate them as nullable 
+  (by using ``?``), Realm considers them optional.
```
dacharyc, line 268: Nit: I don't think these are strictly required, and generally writing style guides call for leaving them out when they're not needed.

```suggestion
  by using ``?``, Realm considers them optional.
```

dacharyc (approved):
Overall, no big blockers. There are a handful of grammar nits, though, and a few questions about things we might want to consider refining.

source/sdk/dotnet/model-data/data-binding.txt
```diff
@@ -94,3 +94,41 @@ The XAML implementation is similar to the example above:
 
 .. literalinclude:: /examples/generated/dotnet/MainPage.snippet.bind-embedded-xaml.xaml
    :language: xml
+
+.. _mvvm_data_binding:
+
+Data Binding and MVVM
```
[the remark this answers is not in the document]
MongoCaleb: Good point. It definitely has a fit in both places. I'll add a link.

source/sdk/dotnet/model-data/define-object-model.txt
MongoCaleb: Good point, You suggest feels a bit busy and highlights the external link more than I like. I hope this change works better!

examples/dotnet/Examples/NullabilityTest.cs
```diff
@@ -0,0 +1,69 @@
...
+        public string RequiredName { get; set; }
+        public string? NullableName { get; set; }
+        public byte[] RequiredArray { get; set; }
+        public byte[]? NullableArray { get; set; }
+
+        /* Realm Objects */
+
+        public Nullable_Bar? ANullableNullable_Bar { get; set; }
+        // :uncomment-start:
+        //public Dog ANonNullableDog { get; set; } // Compile-time error
```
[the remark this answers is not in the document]
MongoCaleb: great points

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,52 @@ Schema, you can update your configuration to include the specific classes you wa
...
+.. _dotnet-required-field:
+.. _dotnet-optional-field:
+.. _dotnet-required-optional-property:
+
+Required and Optional Properties
+--------------------------------
 
-However, when you bind UI elements to a :dotnet-sdk:`RealmObject <reference/Realms.RealmObject.html>`, 
-this notification happens automatically, thus greatly simplifying your ViewModel code. 
+In C#, value types are implicitly non-nullable, but can be made optional by using
```
papafe, line 254:
```suggestion
In C#, value types (such as ``int`` or ``bool``) are implicitly non-nullable, but can be made optional by using 
```

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,52 @@ Schema, you can update your configuration to include the specific classes you wa
...
-   AllProducts = realm.All<Products>();
+- You cannot declare collections (list, sets, and dictionaries) as nullable. 
+  Therefore, Realm considers them required properties. 
 
-.. code-block:: xaml
-   :copyable: false
-   
-   <!-- Somewhere in your View -->
-   <ListView ItemsSource="{Binding AllProducts}">
+  - If the collection contains a list of primatives, the primatives *can* be
```
papafe, line 274: I think it makes sense to move this to his own bullet point

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,52 @@ Schema, you can update your configuration to include the specific classes you wa
...
-   :copyable: false
-   
-   <!-- Somewhere in your View -->
-   <ListView ItemsSource="{Binding AllProducts}">
+  - If the collection contains a list of primatives, the primatives *can* be 
+    nullable.
+  - If the collection contains a list of Realm objects, the objects *cannot* be 
+    nullable.
+
+- You cannot declare dictionaries as nullable, but their values are inherently
```
papafe, line 279: I think we need to restructure this bullet points regarding collections/objects, because it's a little confusing the part regarding collections and objects.

An idea:

- Collections (list, sets, dictionaries, and backlinks) will never be null, and thus they cannot be declared nullable, but their parameters may be:
	- Value types (int, bool, ...) and string, byte[] can be declared as nullable or not, depending if it's a required property or not. 
	- Lists, sets, and backlinks of Realm objects cannot contain null values, and thus the type parameter must be non-nullable.
	- Dictionaries of Realm object values can contain null, and thus the value type parameter must be nullable

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,52 @@ Schema, you can update your configuration to include the specific classes you wa
...
+  you must define the property as nullable.
+
+The following code snippet demonstrates these rules:
+
+.. literalinclude:: /examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
+   :language: csharp
+
+.. note::
 
+   If your classes derive from the ``RealmObject`` base class, the rules above do
```
papafe, line 292: I would say something like:

"If you are using the old schema types definition (link) with classes deriving from "RealmObject", or you do not have nullability annotations enabled, then you need to use the `[Required]` attribute with `string` and `byte[]` properties to specify if they are required

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,52 @@ Schema, you can update your configuration to include the specific classes you wa
...
+  nullable and you must declare the values as such.
+
+- Properties that are Realm object types are inherently nullable, and so 
+  you must define the property as nullable.
+
+The following code snippet demonstrates these rules:
+
+.. literalinclude:: /examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
+   :language: csharp
+
```
papafe, line 289: We need to add a note also about this (taken from the blog post):

We realize that some developers would prefer to have more freedom in the nullability annotation of Realm object properties, and it is possible to do so by setting realm.ignore_objects_nullability = true in a global configuration file (more information about this can be found in the 
[.NET documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files)
). If this option is enabled, all the Realm object properties (including collections) will be considered valid, and the nullability annotations will be ignored.

source/examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
```diff
@@ -0,0 +1,30 @@
...
+    public string? NullableName { get; set; }
+    public byte[] RequiredArray { get; set; }
+    public byte[]? NullableArray { get; set; }
+
+    /* Realm Objects */
+
+    public Dog? ANullableDog { get; set; }
+    // public Dog ANonNullableDog { get; set; } // Compile-time error
+
+    /* Collections of Primatives */
```
papafe, line 16:
```suggestion
    /* Collections of Primitives */
```

source/examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
```diff
@@ -0,0 +1,30 @@
+#nullable enable
+public partial class Person : IRealmObject
+{
+    /* Reference Types */
```
papafe, line 4: I would also add a small section with value types:

public int? NullableInt {get; set;}
public int NonNullableInt {get; set;}

source/examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
```diff
@@ -0,0 +1,30 @@
...
+    public byte[]? NullableArray { get; set; }
+
+    /* Realm Objects */
+
+    public Dog? ANullableDog { get; set; }
+    // public Dog ANonNullableDog { get; set; } // Compile-time error
+
+    /* Collections of Primatives */
+
+    public IList<int> RequiredIntList { get; }
```
papafe, line 18: I would call this IntListWithNonNullableValues, to be consistent

Also probably we should either called them all NonNullable or Required (I'd go for the first)

source/examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
```diff
@@ -0,0 +1,30 @@
...
+    public IList<int?> IntListWithNullableValues { get; }
+    // public IList<int>? NullableListOfInts { get; } // Compile-time error
+
+    /* Collections of Realm Objects */
+
+    public IList<Dog> AListOfNonNullableDogs { get; }
+    // public IList<Dog?> AListOfNNullableDogs { get; } // Compile-time error
+    public IDictionary<string, Dog?> MyDictionaryOfNullableObjects { get; }
+    public IDictionary<string, Dog> MyDictionaryOfNonNullableObjects { get; }
```
papafe, line 27: This line is not supported

source/examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    public IList<int> RequiredIntList { get; }
+    public IList<int?> IntListWithNullableValues { get; }
+    // public IList<int>? NullableListOfInts { get; } // Compile-time error
+
+    /* Collections of Realm Objects */
+
+    public IList<Dog> AListOfNonNullableDogs { get; }
+    // public IList<Dog?> AListOfNNullableDogs { get; } // Compile-time error
+    public IDictionary<string, Dog?> MyDictionaryOfNullableObjects { get; }
```
papafe, line 26: Let's put a space between the different collection types, and add examples for sets and backlinks

papafe (changes requested):

papafe:
Apart from the comments I've left, I think we need a couple things more:

- Check all the code snippets that we have and use the correct nullability annotations
- Add another note somewhere about the old schema type (this could be done in another PR too)

Regarding the second point, we need to say something like:

Prior to Realm .NET version 10.18.0, all realm objects had to be declared as derived from "RealmObject", "EmbeddedObject" and "AsymmetricObject". The classic Realm model definition is still supported, but it will not receive some of the new updates, such as the support for nullability annotations, and will be phased out in the future. For this reason we really recommend migrating to the new interface based class definitions (classes need to be declared implementing IRealmObject, IEmbeddedObject or IAsymmetricObject). 
(Here we could just copy the section about how to migrate from the blog post: https://www.mongodb.com/developer/products/realm/source-generated-classes-nullability-realm/)

source/sdk/dotnet/model-data/define-object-model.txt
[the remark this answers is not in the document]
MongoCaleb: doh! Good catch!

source/examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
MongoCaleb: @papafe These all compile and run correctly. I raised this in the thread in the ask-sdks channel; I think it's an oversight/bug in the SDK.

MongoCaleb: Interestingly, it doesn't compile with Set<> or Backlink, so it must just be an issue with the Dictionary implementation.

MongoCaleb:
@papafe I think I've addressed all of your comments. Note my comment about an apparent bug in the Dictionary implementation.

papafe (changes requested):

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -263,41 +243,68 @@ Schema, you can update your configuration to include the specific classes you wa
...
+and ``byte[]``, as nullable with ``?``. 
 
-.. code-block:: csharp
-   :copyable: false
+.. note:: 
 
-   // Somewhere in your ViewModel
-   AllProducts = realm.All<Products>();
+   Beginning with .NET 6.0, the nullable context is enabled by default for new 
+   projects. For older projects, need to manually enable it. For more information,
```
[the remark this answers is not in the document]
papafe: Or "it needs to be manually enabled"

source/examples/generated/dotnet/NullabilityTest.snippet.nullability.cs
```diff
@@ -0,0 +1,46 @@
...
+    public IList<int?> IntListWithNullableValues { get; }
+    // public IList<int>? NullableListOfInts { get; } // Compile-time error
+
+    /* Collections of Realm Objects */
+
+    public IList<Dog> ListOfNonNullableObjects { get; }
+    // public IList<Dog>? NullableListOfObjects { get; }  // Compile-time error
+    // public IList<Dog?> ListOfNullableObjects { get; } // Compile-time error
+
+    public IDictionary<string, Dog> DictionaryOfNonNullableObjects { get; }
```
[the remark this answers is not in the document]
papafe: And yes, I will create a ticket for the bug you found.

source/includes/dotnet-implement-interface.rst
```diff
@@ -6,10 +6,13 @@
    :dotnet-sdk:`IAsymmetricObject <reference/Realms.IAsymmetricObject.html>`
    interface and must be declared ``partial`` classes.
 
-   You can also derive from the 
+   In older versions of the .NET SDK, objects derive from
```
papafe, line 9: I would add a note that this is for versions lower than 10.18.0, so it's clear where is the cutoff.

source/includes/dotnet-implement-interface.rst
```diff
@@ -6,10 +6,13 @@
...
-   base classes, but note that in the future we may deprecate the 
-   base classes. You should use the interfaces for any new classes
-   that you write.
\ No newline at end of file
+   base classes. This approach to Realm model definition is still supported, but 
+   does not include new features such as the :ref:`nullability annotations 
+   <dotnet-required-optional-property>`. In a future SDK release, the 
+   base classes will become deprecated. You should use the interfaces for any 
+   new classes that you write and should consider migrating your existing 
+   classes.
```
papafe, line 18: What do you think about putting a link to the blog post or just copying it here how to do the conversion? I think it would be quite useful

source/sdk/dotnet/model-data/define-object-model.txt
```diff
@@ -289,6 +294,17 @@ The following code snippet demonstrates these rules:
 
 .. note::
 
-   If your classes derive from the ``RealmObject`` base class, the rules above do 
-   not apply to reference types. Instead, use the 
-   :dotnet-sdk:`[Required] <reference/Realms.RequiredAttribute.html>` attribute.
+   If you are using the older schema type definition (your classes derive from
```
papafe, line 297: Is it possible to add a link that goes to the new section about the old schema type you just added?

papafe (changes requested):

papafe:
After those small changes it looks good for me! 

Another thing that I think we need to do is to change the nullability on all the code snippets in the whole .NET docs, so that we are consistent. I understand that would be a bigger work, so I think we could have it done in a separate PR

source/sdk/dotnet/model-data/define-object-model.txt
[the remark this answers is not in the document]
MongoCaleb: We don't use passive voice, but changed to "can" instead of "need"

source/sdk/dotnet/model-data/define-object-model.txt
papafe: I was referring to the section that starts with "In older versions of the .NET SDK, objects derive from...."

PR #2763, opened 2023-04-27
(DOCSP-29401): Kotlin: Update Bluehawk markup, update and remove code example snippets

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-29401

### Staged Changes

- [Staged Kotlin docs](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29401/sdk/kotlin/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/examples/generated/kotlin/NotificationsTest.snippet.realm-list-change-listener.kt
+++ b/source/examples/generated/kotlin/NotificationsTest.snippet.realm-list-change-listener.kt
@@ -18,8 +18,10 @@ val job = CoroutineScope(Dispatchers.Default).launch {
             is DeletedList -> {
                 // if the list was deleted
             }
-            else -> {
-                throw Exception("Some Error Occurred")
+            is InitialList -> {
+                // Initial event observed on a RealmList flow. It contains a reference
+                // to the starting list state.
+                changes.list
             }
         }
     }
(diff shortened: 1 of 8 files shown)


source/examples/generated/kotlin/NotificationsTest.snippet.realm-list-change-listener.kt
```diff
@@ -18,6 +18,9 @@ val job = CoroutineScope(Dispatchers.Default).launch {
             is DeletedList -> {
                 // if the list was deleted
             }
+            else -> {
+                throw Exception("Some Error Occurred")
```
cbush, line 22: What error? Do we have any info about that?

dacharyc: Moe added this in a PR last December but then never re-generated the code example: https://github.com/mongodb/docs-realm/pull/2413

If I had to guess, I'd assume this was something the compiler suggested at the time but I'm not sure why it was added. Should I check with the Kotlin SDK team to see if this is actually needed?

cbush: You're probably right that the compiler complained about a case that has not been covered... Is there a specific one missing that we can add to avoid the catch-all `if` altogether? Otherwise, can we remove it? Otherwise, can we be more clear about what the error is in this message?

dacharyc: Ahh, got it. Yeah, I've figured out what we're missing and will update this. Thank you for catching this!

cbush (approved):

PR #2766, opened 2023-04-28
(DOCSP-28928) Document asymmetric sync/stream for Node.js

krollins-mdb:
## Pull Request Info

This PR adds a new page to the Node.js SDK docs. It expands on how to implement Data Ingest with asymmetric objects.

Should open another PR for the React Native SDK to update its Stream Data to Atlas page.

Note about Node.js CI test: the `authenticate.js` file is failing a test, but it's not relevant to this PR.

### Jira

- https://jira.mongodb.org/browse/DOCSP-28928

### Staged Changes

- [Stream Data to Atlas](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28928/sdk/node/sync/stream-data-to-atlas/)
- [Define a Realm Object Model](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28928/sdk/node/model-data/define-a-realm-object-model/#define-an-asymmetric-object)

### Review Guidelines


source/includes/data-ingest-overview.rst
```diff
@@ -0,0 +1,12 @@
+You can use :ref:`Data Ingest <optimize-data-ingest>` to stream
```
dacharyc, line 1: [Blocking] I think it makes sense to refactor this to an include as it is used in multiple places. But if we're changing it on the Swift page, we should probably also apply this change to the C++ and .NET pages.

source/sdk/node/model-data/define-a-realm-object-model.txt
```diff
@@ -307,9 +307,21 @@ that invoice data does not need to be read after employees have recorded it. To
 satisfy this requirement, the application developer defines an invoice
```
dacharyc, line 307: [Blocking] You've updated the examples here to show a weather sensor, but the copy in this paragraph still refers to an invoice. We should update this copy.

examples/node/Examples/asymmetric-sync.js
```diff
@@ -1,56 +1,86 @@
...
-const app = new Realm.App({ id: "asymmetric-sync-emuon" });
+const app = new Realm.App({ id: "js-flexible-oseso" });
 
-describe("Asymmetric Sync", () => {
-  afterEach(async () => {});
+// :snippet-start: asymmetric-sync-object
+class WeatherSensor extends Realm.Object {
+  static schema = {
+    name: "WeatherSensor",
+    // sync WeatherSensor objects one way from your device
```
dacharyc, line 9:
```suggestion
    // Sync WeatherSensor objects one way from your device
```

source/sdk/node/sync/stream-data-to-atlas.txt
```diff
@@ -0,0 +1,84 @@
...
+               :emphasize-lines: 12
+         
+         .. tab::
+            :tabid: javascript
+
+            .. literalinclude::  /examples/generated/node/asymmetric-sync.snippet.asymmetric-sync-object.js
+               :language: javascript
+               :emphasize-lines: 6
+
+      For more information on how to define an asymmetric object, see:
```
dacharyc, line 42: I suspect this was just a copy-from-Swift situation, but have we decided to avoid "see" verbiage in our docs? I think Ben and I have been refactoring to "refer to" when we come across these instances.

```suggestion
      For more information on how to define an asymmetric object, refer to 
```

dacharyc (changes requested):
I've left some comments throughout, and have tried using [Blocking] to indicate which ones are blocking vs. which ones should be considered but don't strictly _require_ changes. Let me know if this works for you or if there's another distinction that would be more helpful!

examples/node/Examples/asymmetric-sync.js
```diff
@@ -1,56 +1,86 @@
...
-    // Clean up
+    const weatherSensorCollection = await getWeatherSensors();
+    const weatherSensor = await weatherSensorCollection.findOne({
+      deviceId: "WX1278UIT",
+    });
+
+    expect(weatherSensor?.deviceId).toBe("WX1278UIT");
+
+    // Delete weather sensor documents.
+    await weatherSensorCollection.deleteMany({
```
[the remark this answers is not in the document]
krollins-mdb: Good point! I should test to make sure that there aren't any weatherSensor documents remaining.

Changes to the returned collection do indeed sync back to Atlas. I confirmed it when writing the test by looking in the logs and UI.

examples/node/Examples/asymmetric-sync.js
```diff
@@ -1,56 +1,86 @@
...
+        temperatureInFahrenheit: 66.7,
+        barometricPressureInHg: 29.65,
+        windSpeedInMph: 2,
       });
     });
 
-    // Clean up
+    const weatherSensorCollection = await getWeatherSensors();
+    const weatherSensor = await weatherSensorCollection.findOne({
+      deviceId: "WX1278UIT",
```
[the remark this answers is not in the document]
krollins-mdb: Great point! I've refactored to address this.

examples/node/Examples/asymmetric-sync.js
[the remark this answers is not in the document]
krollins-mdb: In the JS SDK, you only get a count of documents deleted, no info about them. So, we can't match _id properties. But we can count how many documents are in the collection. After deleting, it should be 0.

examples/node/Examples/asymmetric-sync.ts
```diff
@@ -0,0 +1,98 @@
...
+        temperatureInFahrenheit: 66.7,
+        barometricPressureInHg: 29.65,
+        windSpeedInMph: 2,
+      });
+    });
+    // :snippet-end:
+
+    const weatherSensorCollection = await getWeatherSensors();
+    const weatherSensor = await weatherSensorCollection.findOne({
+      deviceId: "WX1278UIT",
```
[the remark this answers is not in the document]
krollins-mdb: One of the most annoying things about writing tests for the Realm.js SDKs. We're in a limbo state where we need to duplicate a lot of things for JS and TS.

I'll make sure both are updated.

source/sdk/node/model-data/define-a-realm-object-model.txt
krollins-mdb: Thanks for giving me an excuse to remove that paragraph. I don't think we need elaborate scenarios in this context.

examples/node/Examples/asymmetric-sync.js
```diff
@@ -1,56 +1,92 @@
 import Realm, { BSON } from "realm";
 
-const app = new Realm.App({ id: "asymmetric-sync-emuon" });
+const app = new Realm.App({ id: "js-flexible-oseso" });
+const weatherSensorPrimaryKey = new BSON.ObjectID();
```
dacharyc, line 4: [Non-blocking] This is purely a style question for my edification - is there a reason to define the `weatherSensorPrimaryKey` as a `const` for the entire file instead of just within the test? I have generally been in the habit of defining constants/variables in the smallest scope where they are needed so I'm wondering if that's not the typical practice in JS dev.

examples/node/Examples/asymmetric-sync.js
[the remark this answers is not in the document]
dacharyc: Reasonable approach 👍

dacharyc (approved):
Looks good - nice work! Just one Q for my curiosity but doesn't affect merging the PR.

examples/node/Examples/asymmetric-sync.js
krollins-mdb: That's probably better practice. I got in the habit of putting things like this at the highest level when working with multiple tests, but it doesn't need to be all the way at the top.

I'll move it lower in scope, as that does improve readability and puts it closer to its proper context.

krollins-mdb: We haven't really standardized the format for JS tests. Cleaning up all of the tests is on my bucket list. 😄

PR #2771, opened 2023-05-01
DOCSP-29577 Link to C++ Filter Page from RQL Page

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-29577

### Staged Changes

- [Realm Query Language](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29577/realm-query-language/)
- [Filter Data - C++ SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29577/sdk/cpp/crud/filter-data/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/realm-query-language.txt
+++ b/source/realm-query-language.txt
@@ -69,8 +69,8 @@ for your SDK:
   .. tab:: C++ SDK
     :tabid: cpp
 
-    The full range of RQL expression are not currently supported by the 
-    C++ SDK. For examples querying Realm in the C++ SDK, refer to:
+    The C++ SDK Alpha implements only a subset of RQL. For examples 
+    querying Realm in the C++ SDK, refer to:
 
     - :ref:`Filter Data - C++ SDK <cpp-filter-data>`
 
--- a/source/realm-query-language.txt
+++ b/source/realm-query-language.txt
@@ -33,38 +33,38 @@ for your SDK:
 
 .. tabs::
 
-   .. tab:: Flutter SDK
-      :tabid: flutter
+  .. tab:: Flutter SDK
+    :tabid: flutter
 
-      - :ref:`Filter & Sort Results - Flutter SDK <flutter-filter-results>`
+    - :ref:`Filter & Sort Results - Flutter SDK <flutter-filter-results>`
 
-   .. tab:: Java SDK
-      :tabid: java
+  .. tab:: Java SDK
+    :tabid: java
 
-      - :ref:`Query Engine - Java SDK <java-client-query-engine>`
-      - :ref:`Query with Realm Query Language - Java SDK <java-filter-with-realm-query-language>`
+    - :ref:`Query Engine - Java SDK <java-client-query-engine>`
+    - :ref:`Query with Realm Query Language - Java SDK <java-filter-with-realm-query-language>`
 
-   .. tab:: Kotlin SDK
-      :tabid: kotlin
+  .. tab:: Kotlin SDK
+    :tabid: kotlin
 
-      - :ref:`Filter Data - Kotlin SDK <kotlin-filter-data>`
+    - :ref:`Filter Data - Kotlin SDK <kotlin-filter-data>`
 
-   .. tab:: .NET SDK
-      :tabid: dotnet
+  .. tab:: .NET SDK
+    :tabid: dotnet
 
-      - :ref:`Query Engine - .NET SDK <dotnet-client-query-engine>`
+    - :ref:`Query Engine - .NET SDK <dotnet-client-query-engine>`
 
-   .. tab:: Node.JS SDK
-      :tabid: node
+  .. tab:: Node.JS SDK
+    :tabid: node
 
-      - :ref:`Query Engine - Node.Js SDK <node-client-query-engine>`
-      - :ref:`Filter Queries - Node.Js SDK <node-filter-queries>`
+    - :ref:`Query Engine - Node.Js SDK <node-client-query-engine>`
+    - :ref:`Filter Queries - Node.Js SDK <node-filter-queries>`
 
-   .. tab:: React Native SDK
-      :tabid: react-native
+  .. tab:: React Native SDK
+    :tabid: react-native
 
-      - :ref:`Query Engine - React Native SDK <react-native-client-query-engine>`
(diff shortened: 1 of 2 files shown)


source/realm-query-language.txt
```diff
@@ -66,11 +66,13 @@ for your SDK:
       - :ref:`Query Engine - React Native SDK <react-native-client-query-engine>`
       - :ref:`Filter Queries - React Native SDK <react-native-filter-queries>`
 
-.. note:: Swift SDK does not support Realm Query Language
+.. note:: Swift SDK and C++ SDK do not support Realm Query Language
 
-   The Swift SDK does not support querying with Realm Query Language.
-   You can instead use NSPredicate to query Realm. Refer to 
-   :ref:`Filter Data - Swift SDK <ios-client-query-engine>`.
+   The Swift SDK and C++ SDK do not support querying with Realm Query Language.
```
dacharyc, line 71: This isn't strictly true. Swift uses NSPredicate. C++ does not use NSPredicate, but does not yet support the full range of RQL expressions as it is currently still in development. We should probably instead just title this note something generic like:

> "Swift and C++" 

And in the note body say something like: 
> "For examples of querying Realm in the Swift SDK or C++ SDK, refer to:"

And then keep the links below.

source/sdk/cpp/crud/filter-data.txt
```diff
@@ -47,6 +47,14 @@ You can set up the realm for these examples with the following code:
 .. literalinclude:: /examples/generated/cpp/filter-data.snippet.set-up-filter-data-tests.cpp
    :language: cpp
 
+.. note:: Sort Functionality for the C++ SDK
+
+  The C++ SDK has not yet implemented sort functionality. Instead, use the
```
dacharyc, line 52: I don't think we want to _recommend_ using `std::sort` necessarily. I think we just want to state that if a developer uses this to sort results, it pulls the results into memory. This is worth calling out here because it's not what they might expect when Realm implements `sort` in C++, and it negates the lazy loading stuff that Realm does. Some developers use C++ _because_ the environment is resource-constrained or performance-constrained, so they'll need to know that using `std::sort` in lieu of Realm implementing sort functionality negates some of the memory benefits of using Realm.

dacharyc (changes requested):
I think we need to make some updates here, but there's a bit of nuance and background involved. Let me know if you need more info!

source/realm-query-language.txt
lindseymoore: Decided to split into two notes, but open to combining!

dacharyc: After seeing the two notes together on the page, it seems like not the right solution. We try to avoid stacking callouts on top of each other, although there are places where it's unavoidable. But I think we can avoid it here.

Maybe the best option is to just change these tabs starting on ln 34 into a nested bullet list? So remove the tabs entirely, and have bullets like:

```
- C++ SDK

  - :ref:`Filter Data - C++ SDK <cpp-filter-data>`

- Flutter

  - :ref:`Filter & Sort Results - Flutter SDK <flutter-filter-results>`
  
- Java SDK

  - :ref:`Query Engine - Java SDK <java-client-query-engine>`
  - :ref:`Query with Realm Query Language - Java SDK <java-filter-with-realm-query-language>` 
```

The spacing may be a little odd - I know you have to leave a blank line between the first-level indent and second-level indent or it messes up the formatting when it renders on the page. But I don't know if you need a blank line between the second level of indent and the first-level indent of the next SDK. You might have to play with this. When we provide a bullet list of SDKs, we like to keep them in alphabetical order by SDK.

I suggest bullets instead of adding a C++ SDK tab because I think we're at the max tab width here without cutting off the tab names. We've had to go to bullet lists in the App Services docs because the tab names were getting cut off. But I don't think we necessarily need to do anything more in this case than just link out to the C++ page - that page already has a note at the top about not supporting the full range of RQL expressions.

source/sdk/cpp/crud/filter-data.txt
```diff
@@ -47,6 +47,14 @@ You can set up the realm for these examples with the following code:
 .. literalinclude:: /examples/generated/cpp/filter-data.snippet.set-up-filter-data-tests.cpp
    :language: cpp
 
+.. note:: Sort Functionality for the C++ SDK
+
+  The C++ SDK has not yet implemented sort functionality. Use
+  `std::sort <https://en.cppreference.com/w/cpp/algorithm/sort>`__ with
+  caution, as the function pulls results into memory. If the data 
+  set is large, ``std::sort`` might not work in a resource-constrained or 
+  performance-constrained environment.
```
dacharyc, line 56: I think the updated version of this note is better, but now we're getting into nitpick territory. I don't think we should start the sentence with "Use `std::sort`" even if we follow that with "with caution", because it sounds like we're advising people to use std::sort and we probably wouldn't recommend that. What about something like:
```suggestion
  The C++ SDK has not yet implemented sort functionality. If you use
  `std::sort <https://en.cppreference.com/w/cpp/algorithm/sort>`__ , 
  this pulls the entire results set into memory. If your data set is 
  large, this could cause problems in a resource-constrained or 
  performance-constrained environment.
```

source/realm-query-language.txt
```diff
@@ -33,44 +33,54 @@ for your SDK:
...
+    :tabid: react-native
+
+    - :ref:`Query Engine - React Native SDK <react-native-client-query-engine>`
+    - :ref:`Filter Queries - React Native SDK <react-native-filter-queries>`
+
+  .. tab:: C++ SDK
+    :tabid: cpp
+
+    The full range of RQL expression are not currently supported by the 
+    C++ SDK. For examples querying Realm in the C++ SDK, refer to:
```
dacharyc, line 73: As written, this is awkward grammatically. It has some is/are agreement issues (and a typo). We can sidestep this by rewriting slightly:

```suggestion
    The C++ SDK Alpha implements only a subset of RQL. 
    For examples querying Realm in the C++ SDK, refer to:
```

dacharyc (approved):
I think it works as a tab, but there's a little bit of grammar weirdness there so we could rewrite it to just bypass that issue. Otherwise LGTM! Thank you for reworking this :)

PR #2784, opened 2023-05-10
(DOCSP-29690) Node.js test update: authentication

krollins-mdb:
## Pull Request Info

I fixed the server API key test and adjusted all the other auth tests so that they would properly fail if something goes wrong. Lots of the tests had `expect()` in a try/catch block, after the login attempt. This means if the login failed, the `expect()` code would never run because it would immediately move to the `catch()` block. And for some reason, `throw new Error()` in the catch block wasn't stopping the test.

I'm considering the skipped tests out of scope for now. I'll come back to them if I have time.

### Jira

- https://jira.mongodb.org/browse/DOCSP-29690

### Staged Changes

- [Authenticate Users](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29690-auth-test/sdk/node/users/authenticate-users/)

### Review Guidelines

### DIFF
--- a/examples/node/Examples/authenticate.ts
+++ b/examples/node/Examples/authenticate.ts
@@ -134,27 +134,6 @@ describe("USER AUTHENTICATION", () => {
     expect(app.currentUser?.id).not.toBe(undefined);
     expect(user?.id).toBe(app.currentUser?.id);
   });
-
-  test("logout", async () => {
-    // Ensure all users are logged out.
-    await Promise.all(Object.values(app.allUsers).map((user) => user.logOut()));
-
-    const credentials = Realm.Credentials.anonymous();
-    const user = await app.logIn(credentials);
-
-    // If login succeeds, user.id and currentUser.id should exist and match.
-    expect(user?.id).not.toBe(undefined);
-    expect(app.currentUser?.id).not.toBe(undefined);
-    expect(user.id).toBe(app.currentUser?.id);
-
-    // :snippet-start: logout-current-user
-    // Log out the current user
-    await app.currentUser?.logOut();
-    // :snippet-end:
-
-    // There shouldn't be any current user.
-    expect(app.currentUser).toBe(null);
-  });
 });
 
 describe("USER SESSIONS", () => {
--- a/source/examples/generated/node/authenticate.snippet.logout-current-user.ts
+++ b/source/examples/generated/node/authenticate.snippet.logout-current-user.ts
@@ -1,2 +1,0 @@
-// Log out the current user
-await app.currentUser?.logOut();


examples/node/Examples/authenticate.js
```diff
@@ -174,56 +180,65 @@ describe("user authentication", () => {
...
     // :snippet-end:
+
+    expect(user?.id).not.toBe(undefined);
+    expect(app.currentUser?.id).not.toBe(undefined);
+    expect(user?.id).toBe(app.currentUser?.id);
   });
 
   test("custom jwt login", async () => {
-    const authenticateWithExternalSystem = () => {
+    const authenticateWithExternalSystem = async () => {
```
krollins-mdb, line 241: In the future, I'd like to set up a proper flow for this. Kind of hard to do in an agnostic way though.

examples/node/Examples/authenticate.ts
```diff
@@ -14,122 +14,148 @@ const randomInt = Math.floor(Math.random() * Math.floor(200000));
...
+
+    expect(user?.id).not.toBe(undefined);
+    expect(app.currentUser?.id).not.toBe(undefined);
+    expect(user?.id).toBe(app.currentUser?.id);
   });
 
   test("custom jwt login", async () => {
-    const authenticateWithExternalSystem = () => {
+    const authenticateWithExternalSystem = async () => {
       // Simulates returning the following JWT information from an external auth service
```
krollins-mdb, line 159: In the future, I'd like to set up a proper flow for this. Kind of hard to do in an agnostic way though.

examples/node/Examples/authenticate.ts
```diff
@@ -14,122 +14,148 @@ const randomInt = Math.floor(Math.random() * Math.floor(200000));
...
 const testPassword = "Pa55w0rd!";
 
-describe("user authentication", () => {
+describe("USER AUTHENTICATION", () => {
   afterEach(async () => {
     await app.currentUser?.logOut();
     jest.runAllTimers();
   });
 
+  // TODO: Figure out why this afterAll causes the following error on tests:
```
krollins-mdb, line 23: I think there may be some timing issue. If I run this to clean up users created during the tests, it seems that `afterAll()` isn't waiting until the tests have finished. It's deleting users before tests have completed. That's my theory right now, anyway.

Is it worth getting this working, or just letting it be, like it has been up until now.

examples/node/Examples/authenticate.js
```diff
@@ -14,22 +14,22 @@ describe("user authentication", () => {
...
-      expect(user.id).toBe(app.currentUser.id);
-      // :remove-end:
-      console.log("Successfully logged in!", user.id);
-      return user;
-    } catch (err) {
-      console.error("Failed to log in", err.message);
-      // :remove-start:
-      throw new Error("Anonymous login failed:" + err.message);
-      // :remove-end:
+      user = await app.logIn(credentials);
```
cbush, line 23: I would add a line after this like
```
// ... use user ...
```

examples/node/Examples/authenticate.js
```diff
@@ -174,56 +180,65 @@ describe("user authentication", () => {
...
+
+    await testUser.logOut();
+
+    process.env.appServicesApiKey = serverApiKey;
+
     // :snippet-start: server-api-key-login
     // Get the API key from the local environment
-    const apiKey = process.env.appServicesApiKey;
+    const apiKey = process.env?.appServicesApiKey;
+    let user;
```
cbush, line 196: In all of these cases I would probably declare `user` in the actual `try` block (and put the user expectation checks in the `try` block as well). There's no reason to scope it outside.
```js
const user = await app.logIn(...)
```

cbush (commented):
I would actually remove all of the try/catch blocks. We are just logging an error and rethrowing - it's distracting from the actual example and introduces a lot of scoping choices. In a real app, we should encourage people to catch and handle errors, but I don't think it's needed for the example.

examples/node/Examples/authenticate.js
```diff
@@ -242,54 +257,48 @@ describe("user authentication", () => {
...
-      // Log out a specific user by ID
-      await app.allUsers[app.currentUser.id].logOut();
-      // :snippet-end:
-    } catch (err) {
-      console.error(err.message);
-      // :remove-start:
-      throw new Error("Logout failed:" + err.message);
-      // :remove-end:
+    // Ensure all users are logged out.
+    if (app.allUsers.length) {
```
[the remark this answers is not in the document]
krollins-mdb: Ah, yeah, thanks for pointing that out. Definitely not needed.

examples/node/Examples/authenticate.ts
```diff
@@ -14,122 +14,148 @@ const randomInt = Math.floor(Math.random() * Math.floor(200000));
...
-      return user;
-    } catch (err) {
-      if (err instanceof Error) {
-        console.error("Failed to log in", err.message);
-        // :remove-start:
-        throw new Error("Email/password login failed:" + err.message);
-        // :remove-end:
+      user = await app.logIn(credentials);
+    } catch (error) {
+      if (error instanceof Error) {
```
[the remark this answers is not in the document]
krollins-mdb: Yeah, that's the code that was already there. It seemed odd to me, but I wasn't entirely sure if it was idiomatic or not.

I'll remove all the try catch blocks, so it's a moot point. Thanks for letting me know that it was an odd pattern.

examples/node/Examples/authenticate.js
krollins-mdb: I moved `user` outside the try block so that I could better test whether the login succeeded or not. Removing the try/catch blocks altogether will fix this.

cbush (approved):

PR #2804, opened 2023-05-18
(DOCSP-28803): Kotlin: Update upsert code example

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-28803

### Staged Changes

- [Update Realm Objects](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28803/sdk/kotlin/realm-database/crud/update/#upsert-objects)

### Reminder Checklist

### Review Guidelines


examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/UpdateTest.kt
```diff
@@ -83,6 +85,61 @@ class UpdateTest: RealmTest() {
...
+                    species = "Green"
+                    owner = "Jim"
+                })
+            }
+            val frogsNamedWirt: RealmResults<CRUDTest.Frog> =
+                realm.query<CRUDTest.Frog>("name == 'Wirt'").find()
+            assertEquals(1, frogsNamedWirt.count())
+
+            // :snippet-start: upsert-an-object
+            realm.write {
```
cmelchior, line 119: This example is a bit artificial because in this case where you search for the object, it would be easier to just update the object you found rather than upserting a newly created object.

I assume this is to make it reproducible, but I would consider maybe doing something like this instead:

```
 // :snippet-start: upsert-an-object
realm.write {
                // Id of existing frog that can either already exists or be new.
                val frogId = ObjectId()

                // If a frog with this id already exists, update its properties, otherwise create it
                this.copyToRealm(CRUDTest.Frog().apply {
                    _id = frogId
                    name = "Wirt"
                    age = 4
                    species = "Greyfrog"
                    owner = "L'oric"}
```

But I'm not 100% sure if that will work with the testing setup you have around this.

cmelchior (commented):
Looking fine. I have a small suggestion to perhaps make it more clear.

I noticed that the Issue also mentions a lot of other things the user feels are underdocumented. Should those be pulled out into its own issue?

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/UpdateTest.kt
```diff
@@ -83,6 +86,60 @@ class UpdateTest: RealmTest() {
...
+                    age = 4
+                    species = "Greyfrog"
+                    owner = "L'oric"
+                }, updatePolicy = UpdatePolicy.ALL)
+            }
+            // :snippet-end:
+            val updatedFrogsNamedWirt: RealmResults<CRUDTest.Frog> =
+                realm.query<CRUDTest.Frog>("name == 'Wirt'").find()
+            assertEquals(1, updatedFrogsNamedWirt.count())
+            assertEquals(4, updatedFrogsNamedWirt.first().age)
```
krollins-mdb, line 137: [non-blocking] Does the Kotlin SDK have a way to find objects based on primary key? I've seen this kind of logic in other SDK tests, where we assume the first object in the results is the one we want because there will only be one object. I think it would be better to look for the object by specific primary key if possible. That way, the test isn't operating on an assumption.

source/sdk/kotlin/realm-database/crud/update.txt
```diff
@@ -65,20 +65,22 @@ To upsert into a realm:
...
-   Filter your criteria with unique identifying criteria for the
-   object you want to upsert.
-
-#. Insert a new object with the same primary key as the object returned
-   by the previous query (if it exists) with `copyToRealm()
+#. Insert an object with a primary key using `copyToRealm()
    <{+kotlin-local-prefix+}io.realm.kotlin/-mutable-realm/copy-to-realm.html>`__.
-   If an object matching your filter already exists, realm updates the
-   existing object. If no object exists that matches your filter, realm
+   If an object matching the primary key already exists, realm updates the
```
krollins-mdb, line 70: [non-blocking] I think this would be a case of capital "R" Realm. But, maybe something like "the Kotlin SDK updates" is better? Saying "Realm" does things feels a little odd. Maybe out of scope for this PR though, as I know that is existing phrasing.

krollins-mdb (approved):
LGTM! Just a couple of non-blocking comments.

examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/UpdateTest.kt
dacharyc: Fair point - updated to find the object matching the ObjectId so there should only be one result, and we can safely work with `.first()` knowing it's correct.

PR #2806, opened 2023-05-18
(DOCSP-29536): C++: update test suite to v0.1.1-alpha

dacharyc:
## Pull Request Info

Updating the C++ test suite to use versioned releases!

A few minor code changes to fix build errors, but the only visible changes are listed in the Staged Changes section.

### Jira

- https://jira.mongodb.org/browse/DOCSP-29536

### Staged Changes

- [Handle Sync Errors](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29536/sdk/cpp/sync/handle-sync-errors/): Remove `error_code()` as it is no longer available 
- [Threading](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29536/sdk/cpp/crud/threading/#schedulers--run-loops-): Overriding `invoke` in a custom scheduler now requires `realm::Function` instead of `std::function`. I also added some minimal code to get past build errors, but that code is not reflected in the published code snippet.

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/examples/cpp/quick-start.cpp
+++ b/examples/cpp/quick-start.cpp
@@ -101,6 +101,7 @@ TEST_CASE("local quick start", "[realm][write]") {
         realm.remove(todo);
     });
     // :snippet-end:
+    token.unregister();
 }
 
 TEST_CASE("sync quick start", "[realm][write][sync]") {
--- a/examples/cpp/threading.cpp
+++ b/examples/cpp/threading.cpp
@@ -65,12 +65,6 @@ TEST_CASE("scheduler", "[write]") {
         
         void invoke(realm::Function<void()> &&task) override {
             // ... Add the task to the (lock-free) processor queue ...
-            // :remove-start:
-            // Had to implement some code in here in order to override invoke
-            // This is arbitrary code that isn't shown in the example just to
-            // get the tests to compile and run.
-            task.release(); 
-            // :remove-end:
         }
 
         [[nodiscard]] bool is_on_thread() const noexcept override {
--- a/examples/cpp/threading.cpp
+++ b/examples/cpp/threading.cpp
@@ -63,20 +63,29 @@ TEST_CASE("scheduler", "[write]") {
             // ... Call in the processor thread(s) and block until return ...
         }
         
-        void invoke(std::function<void()> &&task) override {
+        void invoke(realm::Function<void()> &&task) override {
             // ... Add the task to the (lock-free) processor queue ...
+            // :remove-start:
+            // Had to implement some code in here in order to override invoke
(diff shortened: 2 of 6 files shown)


examples/cpp/quick-start.cpp
```diff
@@ -101,6 +101,7 @@ TEST_CASE("local quick start", "[realm][write]") {
         realm.remove(todo);
     });
     // :snippet-end:
+    token.unregister();
```
cbush, line 104: Why is this needed?

examples/cpp/threading.cpp
```diff
@@ -63,20 +63,29 @@ TEST_CASE("scheduler", "[write]") {
...
         }
         
-        void invoke(std::function<void()> &&task) override {
+        void invoke(realm::Function<void()> &&task) override {
             // ... Add the task to the (lock-free) processor queue ...
+            // :remove-start:
+            // Had to implement some code in here in order to override invoke
+            // This is arbitrary code that isn't shown in the example just to
+            // get the tests to compile and run.
+            task.release();
```
cbush, line 72: Why is this needed?

examples/cpp/quick-start.cpp
dacharyc: After upgrading the test suite to the new version, subsequent tests were failing with an error that there was already a realm open with a different schema. Unregistering the observation token appears to let the realm properly go out of scope and resolves test failures in those subsequent tests.

examples/cpp/threading.cpp
dacharyc: With no executable code in this overridden function, `make` was failing to build the test suite with an error that this was not a valid override. Adding executable code allowed the test to build, and this seemed benign.

examples/cpp/quick-start.cpp
cbush: The fact that the token doesn't unregister itself when going out of scope appears to be a violation of [RAII](https://en.cppreference.com/w/cpp/language/raii). This would be an SDK issue. Worth raising with the team?

examples/cpp/threading.cpp
cbush: Can you post the full error? This doesn't seem right, and I don't want to get in the habit of adding meaningless code.

dacharyc: No, you're right! I guess the invalid override message was cleared by changing `std::function` to `realm::Function` and I did not realize that was the change that resolved the error. I was able to remove this and the test still passes. Thank you for pushing!

cbush (approved):

PR #2822, opened 2023-06-01
(DOCSP-30309) Remove [Required] attributes from .NET docs

nirinchev:
### Jira

- https://jira.mongodb.org/browse/DOCSP-30309

### Staged Changes

- Can't create this since I don't have access.

### Review Guidelines


MongoCaleb:
@nirinchev there are still numerous errors and warnings. See the build logs above. Thanks.

nirinchev:
Those seem to be test failures, not build errors. Not sure how you prefer to have those fixed - one option is to change those properties to be optional, but I feel this is going to complicate docs for no apparent benefit. For example, definitions will need to be updated to look like:

```diff
 public partial class UserProject : IRealmObject
{
    [PrimaryKey]
    [MapTo("_id")]
    public ObjectId ID { get; set; } = ObjectId.GenerateNewId();
+    public string? Name { get; set; }
-    public string Name { get; set; }
}
```

The `?` for `Name` doesn't really bring in meaningful information to readers of the docs. However, changing the schema is a breaking change which requires that the server side schema be updated - it appears you're using a hardcoded app for those tests, which means any such changes will be kind of disruptive.

Then there are some build-time warnings that we're missing constructors for some types. While those are valid, it seems a bit of a hassle to add them since those are just example classes and not really something used in a real application. I'd recommend adding a ignore rule for those warnings.

PR #2834, opened 2023-06-06
Remove Serializer not needed

cmelchior:
This PR updates the Kotlin Serializer section with updated information. The ObjectId serializer was needed in an earlier iteration of the API, but it turned out to not be needed for the final release.


dacharyc (approved):
Thanks, @cmelchior !

PR #2836, opened 2023-06-06
DOCSP-29875 Parameterized Queries on RQL page

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-29875

### Staged Changes

- [RQL](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-29875/realm-query-language/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/realm-query-language.txt
+++ b/source/realm-query-language.txt
@@ -215,6 +215,47 @@ Include multiple parameters with ascending integers starting at ``$0``.
 .. literalinclude:: /examples/generated/realm-query-language/realm-query-language.snippet.multiple-predicate.js
    :language: js
 
+Serialization
+~~~~~~~~~~~~~
+
+The following is a table of how a query will look when serialized for the
+available data types:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 33 34 33
+
+   * - Type
+     - Serialization Description
+     - Example
+
+   * - Boolean
+     - boolean
+     - "setting == $0", false
+
+   * - Int
+     - integer 
+     - "age > $0", 5
+
+   * - Double
+     - double 
+     - "height > $0", 5.5
+
+   * - Date
+     - YYYY-MM-DD@HH:mm:ss:nnnnnnnnnn (year-month-day@hours:minutes:seconds:nanoseconds)
+       Ts:n (s, seconds; n, nanoseconds)
+       Date data type
+     - "date < $0", 2022-08-14@12:02:44:00
+
+   * - String
+     - string or char
+     - "name == $0", "George"
+
+   * - ObjectID
+     - ObjectId(value as hex)
+     - "_id == $0", ObjectId("507f1f77bcf86cd799439011")
+    
+
 .. _rql-dot-notation:
 
 Dot Notation
@@ -610,14 +651,6 @@ You can also represent dates the following ways:
   since :wikipedia:`unix epoch <Unix_time>`.
 
 Date supports :ref:`comparison operators <rql-comparison-operators>`.
-
-.. example::
-
-   The following example shows how to compare dates with their different
-   representations.
-
-   .. literalinclude:: /examples/generated/realm-query-language/realm-query-language.snippet.date-alt-representation.js
-      :language: js
 
 .. _rql-aggregate-operators:
 
--- a/source/realm-query-language.txt
+++ b/source/realm-query-language.txt
@@ -247,7 +247,7 @@ parameterized for the following data types:
 
   * - :ref:`Date <rql-date-operators>`
     - "date < $0", dateObject
-    - "date < 2021-02-20@17:30:15:0 "
+    - "date < 2021-02-20@17:30:15:0"
     - For parameterized date queries, you must pass in a date object. 
       For serialized date queries, you can represented the date in the following formats:
   
@@ -283,7 +283,7 @@ parameterized for the following data types:
       for each member of the list.
   
   * - RealmObject
-    - "link == $0", obj('class_Foo', 'primary key of object')
+    - "link == $0", obj("class_Foo", "primary key of object")
     - ""link == obj('class_Foo', '*primary_key*')"
     - To pass in a RealmObject, you need the class and primary key of the object.
     
--- a/source/realm-query-language.txt
+++ b/source/realm-query-language.txt
@@ -218,7 +218,7 @@ Include multiple parameters with ascending integers starting at ``$0``.
 Serialization
 ~~~~~~~~~~~~~
 
-The following table shows how a query looks when serialized for the
+The following table shows how a query should be formatted for the
 available data types:
 
 .. list-table::
@@ -239,7 +239,7 @@ available data types:
 
    * - Number
      - "age > $0", 5.50
-     - applies for ``int``, ``short``, ``long``, ``double``, ``Decimal128``, and ``float`` data types
+     - Applies for ``int``, ``short``, ``long``, ``double``, ``Decimal128``, and ``float`` data types
 
    * - Date
      - YYYY-MM-DD@HH:mm:ss:nn (year-month-day@hours:minutes:seconds:nanoseconds)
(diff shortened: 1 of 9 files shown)


source/realm-query-language.txt
```diff
@@ -215,6 +215,66 @@ Include multiple parameters with ascending integers starting at ``$0``.
...
+        (T, designates the start of the time; ``s``, seconds; ``n``, nanoseconds)
+      - Parameterized ``Date`` object
+
+  * - :ref:`ObjectID <rql-objectid-uuid-operators>`
+    - "_id == $0", ObjectId("507f1f77bcf86cd799439011")
+    - Formatted as ObjectId(unique value as hex).
+
+  * - :ref:`UUID <rql-objectid-uuid-operators>`
+    - "id == $0", uuid(d1b186e1-e9e0-4768-a1a7-c492519d47ee)
+    - Formatted as uuid(unique 16-byte value).
```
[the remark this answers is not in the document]
MongoCaleb: Also, as with my comment about dates, does this value need to be wrapped in quotes as a string? Again, the unit tests will give you the answer.

source/realm-query-language.txt
```diff
@@ -215,6 +215,66 @@ Include multiple parameters with ascending integers starting at ``$0``.
...
+    
+  * - Binary
+    - "value == $0", "binary"
+
+      "value == $0", ImJpbmFyeSI=
+    - For ASCII characters, RQL serializes the binary value like a string,
```
MongoCaleb, line 266: This section confuses me. We show examples both in quotes and not in quotes, and the description leaves me scratching my head.

source/realm-query-language.txt
```diff
@@ -602,19 +664,19 @@ from the SDK language you are using to a query.
 .. literalinclude:: /examples/generated/realm-query-language/realm-query-language.snippet.date-parameterized-query.js
    :language: js
 
-You can also represent dates the following ways:
+You can also represent dates the following ways in your query:
```
MongoCaleb, line 667: Also? We don't precede this with an example, do we?

How about:
You can specify the date in two ways:

source/realm-query-language.txt
```diff
@@ -602,19 +664,19 @@ from the SDK language you are using to a query.
...
+- As a time in seconds since the :wikipedia:`unix epoch <Unix_time>`: ``Ts:n``, where ``T`` designates the start of the time, 
+  ``s`` is the number of seconds, and ``n`` the number of nanoseconds.
 
 Date supports :ref:`comparison operators <rql-comparison-operators>`.
 
 .. example::
 
-   The following example shows how to compare dates with their different
-   representations.
+   The following examples shows how to use a parameterized query with
```
MongoCaleb, line 678: example shows or examples show (I only see one example, but actually think you should show both styles)

MongoCaleb (commented):
One over-arching question is whether we refer to dates as datetimes (or similar).

source/realm-query-language.txt
```diff
@@ -215,6 +215,66 @@ Include multiple parameters with ascending integers starting at ``$0``.
 .. literalinclude:: /examples/generated/realm-query-language/realm-query-language.snippet.multiple-predicate.js
```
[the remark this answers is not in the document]
lindseymoore: Are you referring to the intro to the parameterized queries section?

source/realm-query-language.txt
```diff
@@ -215,6 +215,66 @@ Include multiple parameters with ascending integers starting at ``$0``.
...
+    - "name == $0", "George"
+    - Can be used for ``string`` or ``char`` data type.
+
+  * - :ref:`Number <rql-arithmetic-operators>`
+    - "age > $0", 5.50
+    - Applies for ``int``, ``short``, ``long``, ``double``, ``Decimal128``, and ``float`` data types.
+
+  * - :ref:`Date <rql-date-operators>`
+    - "date < $0", 2022-08-14@12:02:44:00
+    - The date can be represented in the following formats:
```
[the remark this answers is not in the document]
lindseymoore: Updated!

source/realm-query-language.txt
```diff
@@ -215,6 +215,66 @@ Include multiple parameters with ascending integers starting at ``$0``.
...
+  * - Binary
+    - "value == $0", "binary"
+
+      "value == $0", ImJpbmFyeSI=
+    - For ASCII characters, RQL serializes the binary value like a string,
+      with quotes. For non-printable characters, 
+      RQL serializes the binary to a base 64 value.
+  
+  * - RealmObject
+    - "link == $0", obj('class_Foo', 'primary key of object')
```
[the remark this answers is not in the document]
lindseymoore: It does use single quotes. I will also get a tech review to confirm the table

source/realm-query-language.txt
lindseymoore: There is a code example above this sentence. The previous mentioned way is to provide the date in the query is as a date object.

source/realm-query-language.txt
lindseymoore: Deleted the non-printable characters example since it will probably be very uncommon.

source/realm-query-language.txt
```diff
@@ -215,6 +215,78 @@ Include multiple parameters with ascending integers starting at ``$0``.
...
+
+  * - :ref:`Date <rql-date-operators>`
+    - "date < $0", dateObject
+    - "date < 2021-02-20@17:30:15:0"
+    - For parameterized date queries, you must pass in a date object. 
+      For serialized date queries, you can represented the date in the following formats:
+  
+      - As an explicit date and time: YYYY-MM-DD@HH:mm:ss:nn 
+        (year-month-day@hours:minutes:seconds:nanoseconds)
+      - As a datetime relative to the :wikipedia:`Unix epoch <Unix_time>`:: Ts:n
```
MongoCaleb, line 256: double :: here. Also see the note above about replacing the colon with a hyphen.

source/realm-query-language.txt
```diff
@@ -602,19 +676,19 @@ from the SDK language you are using to a query.
...
 
-You can also represent dates the following ways:
+You can also specify dates in the following two ways:
 
-- ``YYYY-MM-DD@HH:mm:ss:nnnnnnnnnn`` (year-month-day@hours:minutes:seconds:nanoseconds), UTC.
+- As a specific date (in UTC): ``YYYY-MM-DD@HH:mm:ss:nnnnnnnnnn`` (year-month-day@hours:minutes:seconds:nanoseconds), UTC.
   You can also use ``T`` instead of ``@`` to separate the date from the time.
-- ``Ts:n``, where ``s`` is the number of seconds and n the number of nanoseconds
-  since :wikipedia:`unix epoch <Unix_time>`.
+- As a time in seconds since the :wikipedia:`Unix epoch <Unix_time>`: ``Ts:n``, where ``T`` designates the start of the time,
```
MongoCaleb, line 683: Same suggest here as above - consider replacing the colon with hyphen.

MongoCaleb (approved):
LGTM with wee nits

source/realm-query-language.txt
```diff
@@ -215,6 +215,78 @@ Include multiple parameters with ascending integers starting at ``$0``.
...
+    - Applies to ``string`` and ``char`` data type.
+
+  * - :ref:`Number <rql-arithmetic-operators>`
+    - "age > $0", 5.50
+    - "age > 5.50"
+    - Applies to ``int``, ``short``, ``long``, ``double``, ``Decimal128``, and ``float`` data types.
+
+  * - :ref:`Date <rql-date-operators>`
+    - "date < $0", dateObject
+    - "date < 2021-02-20@17:30:15:0"
```
ironage, line 250: I noticed that this date is being presented as a mailto link in my browser (chrome) due to the "@" (same thing for the date in the notes column). Is there a way to prevent it from becoming a hyperlink?

source/realm-query-language.txt
```diff
@@ -215,6 +215,78 @@ Include multiple parameters with ascending integers starting at ``$0``.
...
+
+  * - List
+    - "ANY items.name == {$0, $1}", "milk", "bread"
+    - "ANY items.name == {'milk', 'bread'}"
+    - Applies for list, collections, and sets. A parameterized value should be used
+      for each member of the list.
+  
+  * - RealmObject
+    - "link == $0", obj("class_Foo", "primary key of object")
+    - ""link == obj('class_Foo', '*primary_key*')"
```
ironage, line 287: - there's an extra `"` at the beginning of this line
- It is correct to use the prefix "class_" eg: "class_Foo" or just simply the verbatim class name "Foo". Perhaps it would be less confusing to users to document the simpler version.
- In keeping with your examples using the model above, you could consider changing this to: `ANY items == obj('Item', oid(6489f036f7bd0546377303ab))`

ironage (approved):
A couple small suggestions, but overall LGTM 👍

source/realm-query-language.txt
lindseymoore: Hi @ironage, for your last bullet point, are you suggesting users to use the oid of the object rather than the primary key?

ironage: The value there should match whatever type the primary key is. I was following the example model from above for class "Item" which has a primary key property called "id" which is of type "objectId". If the primary key property were a string, then the syntax would use quotes `obj('Item', 'primary key value')` or if it were an integer it would use a number: `obj('Item', 42)`

PR #2840, opened 2023-06-08
(DOCSP-30369): Add docs release notes for w/e June 8

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-30369

### DIFF
--- a/docs-release-notes.md
+++ b/docs-release-notes.md
@@ -1,4 +1,4 @@
-# June 9, 2023
+# June 8, 2023
 
 ## React Native SDK
 
@@ -12,7 +12,6 @@
 ## C++ SDK
 
 -  Update Alpha naming to Beta throughout the SDK
-
 
 ## Flutter SDK
 
--- a/docs-release-notes.md
+++ b/docs-release-notes.md
@@ -1,3 +1,32 @@
+# June 9, 2023
+
+## React Native SDK
+
+- Remove 'Overview' headings from documentation pages for the SDK
+
+## Kotlin SDK
+
+- Data Types: Add `RealmDictionary` to list of supported data types
+- Serialization: Remove `ObjectId` serializer from the API
+
+## C++ SDK
+
+-  Update Alpha naming to Beta throughout the SDK
+
+
+## Flutter SDK
+
+- React to Changes: Add information about `isCleared` property for the query change listener
+- Data Types: Add documentation for `Decimal128` type
+
+## .NET SDK
+
+- Update Async Guidance for `WriteAsync`
+  - Write Transactions: Update overview description of transactions
+  - Create: Add methods of creating a new Realm object in a `WriteAsync` transaction
+  - Threading: Added details of the benefits of `WriteAsync` for threading
+
+
 # June 2, 2023
 
 ## C++ SDK
--- a/docs-release-notes.md
+++ b/docs-release-notes.md
@@ -1,30 +1,28 @@
 # June 8, 2023
 
-## React Native SDK
+## Flutter SDK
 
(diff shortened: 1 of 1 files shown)


docs-release-notes.md
```diff
@@ -1,3 +1,31 @@
...
+- Remove 'Overview' headings from documentation pages for the SDK
+
+## Kotlin SDK
+
+- Data Types: Add `RealmDictionary` to list of supported data types
+- Serialization: Remove `ObjectId` serializer from the API
+
+## C++ SDK
+
+-  Update Alpha naming to Beta throughout the SDK
```
dacharyc, line 14: We should remove this section because this work was merged into a feature branch and isn't yet live in the docs site. The intent of these docs release notes are to capture public-facing changes, so things we merge to a branch other than `master` shouldn't be reflected in the release notes until they're merged to master.

docs-release-notes.md
```diff
@@ -1,3 +1,31 @@
...
+
+## Flutter SDK
+
+- React to Changes: Add information about `isCleared` property for the query change listener
+- Data Types: Add documentation for `Decimal128` type
+
+## .NET SDK
+
+- Update Async Guidance for `WriteAsync`
```
dacharyc, line 23: To make it easier for folks to consume the docs release notes, I try to directly list the pages that are updated in a structure that matches the docs site. I don't nest them under feature-related headings. This makes it easier for stakeholders outside of the team to scan for docs for just a specific section, and easily understand which docs in the docs set have had updates.

So we might remove this ln 23 here, and instead make this a heading for `CRUD` to indicate the docs section where these changes were made. And then re-order them to match the table of contents. So this might look like:

- CRUD
  - Create: Add methods...
  - Threading: Add details...
  - Transactions: Update...

(Because that is the order in which they appear in the ToC.)

Also, you've used "Update," "Add" and "Added" here - it looks like all the rest of the release notes in this batch are present tense so we should probably change "Added" to "Add".

docs-release-notes.md
```diff
@@ -1,3 +1,31 @@
+# June 8, 2023
+
+## React Native SDK
```
dacharyc, line 3: I try to list the SDKs in alphabetical order to make it match the table of contents, make it consistent across release notes, and make it easier to scan for the section you want. So this should be re-ordered to:

- Flutter SDK
- Kotlin SDK
- .NET SDK
- React Native SDK

docs-release-notes.md
```diff
@@ -1,3 +1,31 @@
...
+
+- Remove 'Overview' headings from documentation pages for the SDK
+
+## Kotlin SDK
+
+- Data Types: Add `RealmDictionary` to list of supported data types
```
dacharyc, line 9: We're missing the section headings here. Only listing the page title requires interested parties to be very familiar with how our docs are structured and where exactly these pages are, and we shouldn't assume everyone will have that level of familiarity with our docs set. So these should be something like:

- Realm Database
  - Model Data/Data Types: Add...
  - Serialization: Remove...

This way, if someone wanted to browse to a page that changed in order to see the changes, they could just follow the docs structure to easily get to the page - i.e. Kotlin SDK -> Realm Database -> Model Data -> Data Types

dacharyc (changes requested):
Awesome work sussing out the changes for the week! I know going through all the PRs and figuring out what has changed isn't always easy, and I think you did a great job at pulling out the relevant details here. Just a few things to tweak for ordering and to make it easier for folks to find changed pages.

dacharyc (approved):
LGTM! Nice work!

PR #2858, opened 2023-06-21
Geospatial Data Types

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-30429

### Staged Changes

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/geospatial/sdk/dotnet/model-data/data-types/geospatials/


examples/dotnet/DataBinding/DataBinding.csproj
```diff
@@ -16,9 +16,9 @@
 
   <ItemGroup>
     <PackageReference Include="Xamarin.Forms" Version="5.0.0.2578" />
-    <PackageReference Include="Xamarin.Essentials" Version="1.7.5" />
-    <PackageReference Include="Realm" Version="10.21.1" />
-    <PackageReference Include="Realm.Fody" Version="10.21.1">
+    <PackageReference Include="Xamarin.Essentials" Version="1.7.6" />
+    <PackageReference Include="Realm" Version="11.1.1" />
+    <PackageReference Include="Realm.Fody" Version="11.0.0">
```
nirinchev, line 21: This should be removed

source/sdk/dotnet/model-data/data-types/geospatials.txt
```diff
@@ -0,0 +1,160 @@
...
+creates 2 boxes:
+
+.. literalinclude:: /examples/generated/dotnet/Geospatial.snippet.geobox.cs
+   :language: csharp
+
+.. figure:: /images/geoboxes.png
+   :alt: 2 GeoBoxes
+.. _persist-geopoint:
+
+Persisting GeoPoint Data
```
nirinchev, line 98: I think we probably want to move this above the geospatial data types as persisting is generally viewed as the first step and querying as the second.

examples/dotnet/Examples/Geospatial.cs
```diff
@@ -0,0 +1,153 @@
...
+    //:snippet-start:usingcustomgeopoint
+    public partial class Company : IRealmObject
+    {
+        [PrimaryKey]
+        [MapTo("_id")]
+        public Guid Id { get; private set; } = Guid.NewGuid();
+
+        public CustomGeoPoint? Location { get; set; }
+
+        public Company(double lat, double lon)
```
nirinchev, line 147: This is a somewhat weird constructor - I know we don't have other fields, it's just weird further down in the examples when you see `new Company(12.34, 56.78)` - it almost feels like the company is a geo point. Up to you whether you want to change it, but I personally find an example like:

```csharp
new Company
{
    Location = new(12.34, 56.78)
}
```

more understandable.

source/sdk/dotnet/model-data/data-types/geospatials.txt
```diff
@@ -0,0 +1,160 @@
...
+   :class: singlecol
+
+Geospatial data, or "geodata", specifies points and geometric objects on the Earth's 
+surface. With the geodata types, you can create queries that check whether a given  
+point is contained within a shape. For example, you can find all coffee shops within 
+15 km of a specified point.
+
+Geospatial Data Types
+---------------------
+
```
nirinchev, line 20: I think we need a comment here explaining that these types are only used for querying and not persistence.

examples/dotnet/Examples/Geospatial.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            // System.ArgumentException: not enough points defined
+            // var notEnoughPointsPolygon = new GeoPolygon((20, 20),
+            //    (10, 20), (0, 20));
+
+            // System.ArgumentException the first and the last points
+            // do not match
+            // var closingPointDoesNotMatchPolygon =
+            //    new GeoPolygon((20, 20), (10, 20), (0, 20), (10, 10));
+            // :snippet-end:
```
nirinchev, line 77: Do we want these int he docs? We already mention the restrictions in the text and if people don't abide by them, they'll get these exceptions, but it seems unusual to document them in this way.

source/sdk/dotnet/model-data/data-types/geospatials.txt
```diff
@@ -0,0 +1,160 @@
...
+.. figure:: /images/geopoints.png
+   :alt: 2 GeoPoints
+
+Querying Geospatial Data
+------------------------
+To query against geospatial data, you can use the 
+:dotnet-sdk:`GeoWithin <reference/Realms.QueryMethods.html#Realms_QueryMethods_GeoWithin_Realms_IEmbeddedObject_Realms_GeoShapeBase_>` 
+method, or you can use the ``geoWithin`` operator with :ref:`RQL <dotnet-realm-query-language>`. 
+The ``GeoWithin`` method takes a :ref:`GeoPoint <geopoint>` and any one of
```
nirinchev, line 137: This is not precise - it doesn't take a `GeoPoint`, but rather an embedded object that has a GeoJSON point shape.

nirinchev (commented):
Really nice, especially with the visualizations! I have mostly minor suggestions.

source/sdk/dotnet/model-data/data-types/geospatials.txt
```diff
@@ -0,0 +1,160 @@
...
+Persisting GeoPoint Data
+------------------------
+If you want to persist ``GeoPoint`` data, you must create your own custom realm object 
+that inherits from :dotnet-sdk:`IEmbeddedObject <reference/Realms.IEmbeddedObject.html>`.
+An embedded GeoPoint class must have a minimum of two fields:
+
+- A field of type ``IList<double>`` that maps to a "coordinates" (case sensitive) 
+  property in the realm schema. 
+
+- A string field that maps to a "type" property.
```
papafe, line 107: We need to specify that this needs to be equal to "Point"

examples/dotnet/Examples/Geospatial.cs
MongoCaleb: Decided to make it even more explicit:
```
new Company
                {
                    Location = new CustomGeoPoint(47.68, -122.35)
                });
```

source/sdk/dotnet/model-data/data-types/geospatials.txt
MongoCaleb: I had that and then removed it because we have a section specifically about persisting DataPoints. Other types, it seems, can be persisted with IList<double>, for example. No?

examples/dotnet/DataBinding/DataBinding.csproj
MongoCaleb: We no longer need Fody at all?

source/sdk/dotnet/model-data/data-types/geospatials.txt
nirinchev: What do you mean with "can be persisted with IList"?

source/sdk/dotnet/model-data/data-types/geospatials.txt
MongoCaleb: I disagree in this case because geospatial data types are (generally) not for persistence, right? The persistence info feels secondary, and unless I expand that section to how to persist the other data types, I think it should stay lower in the page. And, of course, we have the right-side menu for discovering it.

source/sdk/dotnet/model-data/data-types/geospatials.txt
MongoCaleb: Yep. Hard to explain that one, since the whole thing hinges on a custom class. But LMK what you think of this change!

source/sdk/dotnet/model-data/data-types/geospatials.txt
MongoCaleb: I mean something like a polygon could be persisted as a collection of doubles that are then converted to a polygon on the client. For example, if my company has a sales region, I can store that.

source/images/geocircles-query.png
lindseymoore, line 1: Is it possible to include the images in a way that when you click it, it expands in a pop up window? The images are currently a bit hard to read. I think I've seen this as possible somewhere in the docs!

source/sdk/dotnet/model-data/data-types/geospatials.txt
```diff
@@ -0,0 +1,190 @@
...
+1. Create an embedded realm object (a class that inherits from 
+   :dotnet-sdk:`IEmbeddedObject <reference/Realms.IEmbeddedObject.html>`).
+
+#. Add a minimum of two fields:
+
+   - A field of type ``IList<double>`` that maps to a "coordinates" (case sensitive) 
+     property in the realm schema. 
+
+   - A field of type ``string`` that maps to a "type" property. The value of this 
+     field must be "Point".
```
lindseymoore, line 118: Might be nice to add a few words on why it must be a "Point".

source/sdk/dotnet/model-data/data-types/geospatials.txt
```diff
@@ -0,0 +1,190 @@
...
+   the geodata region.
+
+The following example shows the difference between querying with the ``GeoWithin`` 
+method and RQL:
+
+.. literalinclude:: /examples/generated/dotnet/Geospatial.snippet.rql-geowithin.cs
+   :language: csharp
+
+The following examples show querying against various shapes to return a list of 
+companies within the shape.
```
lindseymoore, line 166:
```suggestion
companies within the shape:
```
I would end with a colon here rather than a period to show the following examples are related. I though something was missing because there was new subject heading and head to reread.

source/sdk/dotnet/model-data/data-types/geospatials.txt
```diff
@@ -0,0 +1,190 @@
...
+
+   The fourth point in a polygon must be the same as the first point.
+
+You can also exclude areas within a polygon by defining one or more "holes". A 
+hole is another polygon whose bounds fit completely within the outer polygon. 
+The following example creates 3 polygons: one is a basic polygon with 5 points, 
+one is the same polygon with a single hole, and the third is the same polygon 
+with two holes:
+
+.. literalinclude:: /examples/generated/dotnet/Geospatial.snippet.geopolygon.cs
```
lindseymoore, line 91: I would include the descriptions of the polygons as comments in the code snippets rather than in the paragraph above, so readers don't have to look back and forth comparing text with code.

lindseymoore (changes requested):
A few small comments and suggestions. If the images can't be made larger with a popup, I would recommend putting the captions as a comment in the code instead, so they are easier to read.

examples/dotnet/DataBinding/DataBinding.csproj
nirinchev: We moved the Fody weaver out of a dedicated package into the main Realm package in 11.1.0. That's why there hasn't been a 11.1.x release of Realm.Fody. Prior to 11.1.0, most projects would take a transitive dependency on Realm.Fody through the Realm package, so this change was seamless, but in cases like this one where both packages are explicitly referenced, you'll need to manually remove the Realm.Fody reference.

source/sdk/dotnet/model-data/data-types/geospatials.txt
nirinchev: Not sure what you mean - persistence is the #1 task of a database, querying is #2. Even if the shape types we expose are not used for persistence, they cannot be used for querying either until you have some data in the correct shape. If you think about the code as a stream of tasks, you have:
1. Define a model
2. Open a Realm
3. Add data to the Realm
4. Query that data

source/sdk/dotnet/model-data/data-types/geospatials.txt
nirinchev: No, polygons are not supported with the current data modeling capabilities of Realm. You can neither persist nor query them. The reason is that polygons are represented as an array of arrays which is not a data structure Realm supports today.

source/sdk/dotnet/model-data/data-types/geospatials.txt
MongoCaleb: It's an interesting discussion: how much we rely on comments in the code vs in the text. I had the comments in the code and I thought it just made the example really "messy".

source/sdk/dotnet/model-data/data-types/geospatials.txt
MongoCaleb: This is how the page is arranged: we define the types, we show how to persist the one type that can be persisted, and then we query against it.

PR #2872, opened 2023-06-28
(DOCSP-30998): Swift SDK migrate PBS to FS updates

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-30998

### Staged Changes

- [Sync Data](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-30998/sdk/swift/sync/#migrate-from-partition-based-sync-to-flexible-sync): Remove PBS info, link out to new PBS page
- [Partition-Based Sync](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-30998/sdk/swift/sync/partition-based-sync/): New page with Configure/Open content, Progress Notifications (they only work on PBS), migration info.
- [Configure & Open a Synced Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-30998/sdk/swift/sync/configure-and-open-a-synced-realm/#open-a-synced-realm-for-partition-based-sync): Remove PBS-only examples. Some of these examples still need to be updated to show FS instead, but that seems out of scope for this ticket.
- [Write to a Synced Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-30998/sdk/swift/sync/write-to-synced-realm/#group-writes-for-improved-performance): Moved "Group Updates" to this page where it makes more sense
- [SwiftUI/Open a Synced Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-30998/sdk/swift/swiftui/configure-and-open-realm/#open-a-synced-realm)
- [SwiftUI/Pass Realm Data Between Views](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-30998/sdk/swift/swiftui/pass-realm-data-between-views/#inject-the-partition-value)

I removed the "Legacy Realm Open" page. Caleb asked me about this page last week, and it has only gotten ~5 views over the last 6 weeks, deals with versions of Realm Swift that are two years old, so I think we can safely remove it.

### Reminder Checklist

### Review Guidelines


source/sdk/swift/sync.txt
```diff
@@ -94,7 +95,64 @@ The partition value determines which data the client application can access.
...
+
+If you need to make updates to your client code after migration, consider 
+updating your client codebase to remove hidden migration functionality.
+You might want update your client codebase when:
+
+- You add a new model or change a model in your client codebase
+- You add or change functionality that involves reading or writing Realm objects
+- You want to implement more fine-grained control over what data you sync
+
+Make these changes to convert your Partition-Based Sync client code to use
```
cbullinger, line 139: something about the wording of this and its placement after the previous paragraphs is off to me. I can't put my finger on it but I think it's the "convert your code to use Flex Sync" wording -- isn't my code technically already using Flex Sync after the migration? or it still using PBS but now there's a weird automagic layer on top to convert to Flex Sync?

source/sdk/swift/sync/flexible-sync.txt
```diff
@@ -354,3 +354,20 @@ Flexible Sync RQL Limitations
...
+Migrate from Partition-Based Sync
+---------------------------------
+
+In Realm Swift SDK version 10.40.0 or newer, you can automatically migrate 
+your App Services Device Sync Mode from Partition-Based Sync to Flexible
+Sync. For more information, refer to :ref:`ios-migrate-pbs-to-fs`.
+
+When you migrate from Partition-Based Sync to Flexible Sync, Realm
+automatically creates hidden Flexible Sync subscriptions for your app. The 
+next time you add or change subscriptions, we recommend removing the
```
cbullinger, line 367: maybe make these explicit steps?

> When you migrate from Partition-Based Sync to Flexible Sync, Realm
> automatically creates hidden Flexible Sync subscriptions for your app. The next time you add or change subscriptions, we recommend that you: 
> 
> 1. Remove the automatically generated subscriptions. 
> 2. Manually add the relevant subscriptions in your client codebase to see all of your subscription logic together and enable easy iteration and debugging.

source/sdk/swift/sync/flexible-sync.txt
```diff
@@ -354,3 +354,20 @@ Flexible Sync RQL Limitations
 -----------------------------
 
 .. include:: /includes/flex-sync-limitations.rst
+
+Migrate from Partition-Based Sync
```
cbullinger, line 358: I almost feel like there should be something at the top of this page acknowledging users who are here to remove their automatically generated subs and manually add them back.

source/sdk/swift/sync.txt
```diff
@@ -94,7 +95,64 @@ The partition value determines which data the client application can access.
...
+authentication configuration. Flexible Sync provides more versatile permissions
+configuration options and more granular data synchronization.
+
+For more information about how to migrate your App Services App from 
+Partition-Based Sync to Flexible Sync, refer to :ref:`Migrate Device Sync Modes 
+<realm-sync-migrate-modes>`.
+
+.. _ios-update-client-code-after-pbs-to-fs-migration:
+
+Updating Client Code After Migration
```
cbullinger, line 118:
```suggestion
Update Client Code After Migration
```

cbullinger (approved):
Overall updates look good. Hard for me to tell if there are any other places that should be updated too, but it looks like you and Kyle already had a good idea of which topics needed to be touched.

source/sdk/swift/sync/flexible-sync.txt
dacharyc: I'm not sure how common this use case will be, and it is reflected in the OTP, so I'm a little reluctant to add something at the top that most users won't actually need. Do you have thoughts on what this might look like?

source/sdk/swift/sync.txt
dacharyc: I intentionally chose this form because... "Update" seems like an imperative - i.e. "You must Update Client Code After Migration". But this is not actually required. One of the points Product wanted to emphasize about the migration is that you don't actually have to do anything with your client code.

You really only need to do this if you are actively changing your codebase. I.e. if you are updating your code base to add new models, optimizing for performance, etc. - you should also do these things to avoid hidden issues.

I am open to a different heading if you can think of a better way to communicate this?

source/sdk/swift/sync.txt
dacharyc: I don't disagree with you. I rewrote this section like three times to try to figure out what to communicate here.

Technically, both things are true - your app is using Flexible Sync after migration, but your codebase still has Partition-Based Sync stuff in it - i.e. you may be opening a realm with a Partition-Based Sync config, using a Partition key, you don't have to add Flexible Sync subscriptions, etc. The migration adds a weird automagic layer to convert to Flex Sync.

I think this is a big gotcha if you ever need to change your app in the future, because those automagic things won't apply to - for example, a new model you might want to add. 

I am definitely open to reworking this to improve clarity. Maybe we can chat about this?

source/sdk/swift/sync/flexible-sync.txt
cbullinger: No, I agree if it's an edge case we should leave it down here only. I think it stood out to me because it's where I landed after clicking the link from the previous Sync content, so it seemed weird to drop you off there with no acknowledgment.

source/sdk/swift/sync.txt
cbullinger: No, my suggestion was assuming this is more of an explicit task, which it sounds like it isn't. I'm good with how you've worded it ("Updating..."), though, as we were discussing, I still don't love this overall presentation. Open to help brainstorm another approach later.

source/sdk/swift/sync.txt
```diff
@@ -34,16 +33,6 @@ an :ref:`App Services backend <realm-cloud>`. When a client
 device is online, Sync asynchronously synchronizes data in a 
 background thread between the device and your backend App. 
 
-When you use Sync in your client application, your implementation must match
```
cbullinger, line 37: Can't comment on it, but if we're limiting this page to just FS, let's update the headings (e.g. remove Sync Data, make Flexible Sync an H1)

source/sdk/swift/sync.txt
```diff
@@ -80,21 +69,10 @@ To use Flexible Sync in your client application, open a synced realm
...
-
-When you select :ref:`Partition-Based Sync <partition-based-sync>` for your 
-backend App configuration, your client implementation must include a 
-partition value. This is the value of the :ref:`partition key 
-<partition-key>` field you select when you configure Partition-Based Sync. 
-
-The partition value determines which data the client application can access.
-
-You pass in the partition value when you open a synced realm.
+.. seealso::
```
cbullinger, line 72:
```suggestion
.. tip::
```

source/sdk/swift/sync/partition-based-sync.txt
```diff
@@ -0,0 +1,249 @@
...
+================================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Partition-Based Sync is an older mode for using Atlas Device Sync with the 
+Realm Swift SDK. We recommend using Flexible Sync for new apps. The information
```
cbullinger, line 14:
```suggestion
Realm Swift SDK. We recommend using :ref:`Flexible Sync <enable-realm-sync>` for new apps. The information 
```
link to the FS section on Sync Data Between Devices landing page?

source/sdk/swift/sync/partition-based-sync.txt
```diff
@@ -0,0 +1,249 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Partition-Based Sync is an older mode for using Atlas Device Sync with the 
+Realm Swift SDK. We recommend using Flexible Sync for new apps. The information 
+on this page is for users who are still using Partition-Based Sync.
+
```
cbullinger, line 16: Maybe add a blurb about migration, then link to Migrate section?

> The information on this page is for users who are still using Partition-Based Sync. 

> .. tip::

>   Realm Swift SDK v10.40.0 provides a way to easily migrate from PBS to FS. See the Migrate from PBS to FS section for more info.

source/sdk/swift/sync/partition-based-sync.txt
```diff
@@ -0,0 +1,249 @@
...
+You might want update your client codebase when:
+
+- You add a new model or change a model in your client codebase
+- You add or change functionality that involves reading or writing Realm objects
+- You want to implement more fine-grained control over what data you sync
+
+Make these changes to convert your Partition-Based Sync client code to use 
+Flexible Sync:
+
+- Switch to a ``flexibleSyncConfiguration()`` where you :ref:`open a synced realm
```
cbullinger, line 224: should this be an ordered list?

source/sdk/swift/sync/partition-based-sync.txt
```diff
@@ -0,0 +1,249 @@
...
+.. literalinclude:: /examples/generated/code/start/ConvertSyncLocalRealms.snippet.convert-sync-to-local.swift
+   :language: swift
+
+.. _swift-check-upload-download-progress:
+.. _ios-check-sync-progress:
+
+Check Upload & Download Progress for a Sync Session
+---------------------------------------------------
+
+.. include:: /includes/flex-sync-unsupported-progress-notifications.rst
```
cbullinger, line 142: kind of threw me off that this note refers to FS first (i.e. seems FS-focused) now that this content is PB-only

cbullinger (commented):
I like this approach. I think it makes sense if we're truly pushing FS, plus there's the added bonus that the content is more direct (vs. a single long, potentially confusing page lots of clarifying notes, etc.). I don't have a strong opinion on whether the PBS page should be in the toc, though I'm curious how many users are still using PBS and would need to refer to those docs.

source/sdk/swift/sync/partition-based-sync.txt
dacharyc: I don't think so. It doesn't necessarily matter which thing you do in which order. These are all just things you _ought_ to do.

PR #2889, opened 2023-07-14
Fix Node.js bluehawking

krollins-mdb:
## Pull Request Info

No ticket for this one. I didn't get the Bluehawk script quite right in my initial PR for the new Node.js SDK test suite.

For now, I'm leaving the existing generated samples in their original location so that I don't need to update the paths for every sample where it's included. I'll move all of these into the `legacy` directory and update their references in a later PR.

### Review Guidelines


dacharyc (approved):
Presuming you have run this locally and it works, LGTM!

PR #2894, opened 2023-07-18
(DOCSP-31037, DOCSP-30208) Address backlinks and unmanaged objects

krollins-mdb:
## Pull Request Info

Updated the `graphql` dependency for the Flutter SDK test suite. Also clarified how backlinks work (or, rather, don't) with unmanaged objects.

### Jira

- https://jira.mongodb.org/browse/DOCSP-31037
- https://jira.mongodb.org/browse/DOCSP-30208

### Staged Changes

- [Relationships](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-31037/sdk/flutter/realm-database/model-data/relationships/#inverse-relationship)

### Review Guidelines

### DIFF
--- a/examples/dart/pubspec.lock
+++ b/examples/dart/pubspec.lock
@@ -249,6 +249,70 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "2.1.2"
+  gql:
+    dependency: transitive
+    description:
+      name: gql
+      sha256: "07ae289bc5ccfbfd143883bfe9a380f5bf52c94559ded740cf3fb152fcbdbe2e"
+      url: "https://pub.dev"
+    source: hosted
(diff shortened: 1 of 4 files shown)


dacharyc (approved):
This looks fine. But outside the scope of this ticket, I'm not seeing where we're actually testing creating an object with a backlink, querying on it, etc. to confirm this is correct.

In looking at the Read & Write Data page, I see we don't have those examples in Flutter. 

I wonder if we should scope a project for next Q to build out the Flutter examples more?

krollins-mdb:
@dacharyc, yeah, that's a good point. The only backlinks example we have is the one that shows how to define a backlink in a model on the Relationships page. Would be good to expand these, for sure.

PR #2933, opened 2023-08-02
(DOCSP-29950): JS Geospatial Queries for Points

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-29950

### Staged Changes

### Reminder Checklist

### Review Guidelines


source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,305 @@
...
+GeoPoint
+~~~~~~~~
+
+A :js-sdk:`GeoPoint <INSERT-API-REFERENCE-LINK>` defines a specific 
+location on the Earth's surface. All of the geospatial data types use ``GeoPoints`` 
+to define their location. 
+
+.. note::
+   
+   In methods that take a ``GeoPoint``, you can alternatively provide a tuple of
```
krollins-mdb, line 34: > a tuple of doubles

sounds like a Bond move title or something. lol 😄 
(Not suggesting a change - it reads well. I just enjoyed these particular words together.)

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,305 @@
...
+.. _node-geopoint:
+
+GeoPoint
+~~~~~~~~
+
+A :js-sdk:`GeoPoint <INSERT-API-REFERENCE-LINK>` defines a specific 
+location on the Earth's surface. All of the geospatial data types use ``GeoPoints`` 
+to define their location. 
+
+.. note::
```
krollins-mdb, line 32: Does this need to be a note? It feels like a natural extension of the previous para and I'm not sure it's important enough to set it apart. It feels more a like a "oh, and you can do this" whereas setting it apart in a note implies it's important or different enough that readers should pay particular attention.

But that's my personal take. According to the style guide, you're using this perfectly! Regarding note admonitions:
> Provides information that emphasizes or supplements information in the text. A note can provide information that applies only in certain cases.

Still, I we should consider removing either this note or the one in the following section. Visually, they're very close together on the page and it interrupts reading a bit.

examples/node/v12/__tests__/geospatial.test.js
```diff
@@ -0,0 +1,208 @@
...
+      center: [-121.9, 47.3],
+      distance: 0.25,
+    };
+
+    const largeCircleCenter = {
+      longitude: -122.6,
+      latitude: 47.8,
+    };
+
+    const radius = kmToRadians(44.4);
```
krollins-mdb, line 68: It may be helpful to note that `kmToRadians()` is imported from Realm.

krollins-mdb (approved):
LGTM, though I have minor comments throughout.

Thank you for putting this together, @dacharyc!

examples/node/v12/__tests__/geospatial.test.js
```diff
@@ -0,0 +1,208 @@
...
+          _id: "int",
+          location: "MyGeoPoint",
+        },
+        primaryKey: "_id",
+      };
+    }
+    // :snippet-end:
+
+    const realm = await Realm.open({
+      schema: [Company, MyGeoPoint.schema],
```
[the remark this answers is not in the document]
dacharyc: So it seems like I _have_ to use just the `.schema` here - if I try removing it, I get:

```console
Type 'typeof MyGeoPoint' is not assignable to type 'ObjectSchema | RealmObjectConstructor<AnyRealmObject>'. 
Types of construct signatures are incompatible. 
Type 'new (long: number, lat: number) => MyGeoPoint' is not assignable to type 'new (...args: any) => AnyRealmObject'. 
Type 'MyGeoPoint' is missing the following properties from type 'RealmObject<any, never>': keys, entries, toJSON, isValid, and 9 more.
```

I have different issues if I try removing the constructor. I'm sure there's a way around this, but this was how it was defined in the JS PR test file, so I presumed it was the "correct" way to do this. If you have time to help me figure out how I might do this in a way that is more canonical for our examples, I'm happy to pair on it.

examples/node/v12/__tests__/geospatial.test.js
```diff
@@ -0,0 +1,208 @@
...
+      outerRing: outerRing,
+      holes: [hole, hole2],
+    };
+    // :snippet-end:
+
+    // :snippet-start: geocircle-query
+    const companiesInLargeCircle = realm
+      .objects(Company)
+      .filtered("location geoWithin $0", largeCircle);
+    console.log("Companies in large circle: ${companiesInLargeCircle.length}");
```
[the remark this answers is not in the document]
dacharyc: Ah! No, they don't show up - I was mostly doing it for the example syntax. Updating to console.debug.

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,305 @@
...
+
+Geospatial Data Types
+---------------------
+
+.. _node-geopoint:
+
+GeoPoint
+~~~~~~~~
+
+A :js-sdk:`GeoPoint <INSERT-API-REFERENCE-LINK>` defines a specific
```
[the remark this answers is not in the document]
dacharyc: I was not aware! I'll do that - it's probably easier to swap `next` for `latest` when the time comes than to have to look up all the links at that point. Better to front-load the link lookups now. Thanks!

source/sdk/node/model-data/data-types/geospatial.txt
dacharyc: Wish I could take credit but that is Caleb's wordsmithing from the .NET version of the page!

source/sdk/node/model-data/data-types/geospatial.txt
dacharyc: I don't feel strongly about this - it's just the wording from Caleb's original version of the page. I do agree that multiple notes close together on a page look heavy to me stylistically, so I'll un-note this instance.

examples/node/v12/__tests__/geospatial.test.js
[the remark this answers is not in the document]
krollins-mdb: Huh. That's odd and not idiomatic for the JS SDK. Though it's entirely possible that's just how it's implemented.

I'll ask the JS team.

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,305 @@
...
+
+.. figure:: /images/geopolygons.png
+   :alt: 3 GeoPolygons
+   :width: 250
+   :lightbox:
+
+.. _node-persist-geopoint:
+
+Persist GeoPoint Data
+---------------------
```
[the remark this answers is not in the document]
dacharyc: Let's chat about this at SDK docs sync.

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,303 @@
...
+
+GeoPoint
+~~~~~~~~
+
+A :js-sdk:`GeoPoint <Realm.GeoPoint.html>` defines a specific 
+location on the Earth's surface. All of the geospatial data types use ``GeoPoints`` 
+to define their location. 
+
+In methods that take a ``GeoPoint``, you can alternatively provide a tuple of 
+doubles, where the first double is the longitude and the second double is the
```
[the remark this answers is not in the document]
papafe: This was correct in .NET though

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,303 @@
...
+~~~~~~~~~
+
+A :js-sdk:`GeoCircle <Realm.GeoCircle.html>` defines a circle on 
+the Earth's surface. You define a ``GeoCircle`` by providing a ``GeoPoint`` 
+for the center of the circle and a number for the distance to specify the 
+radius of the circle. 
+
+.. note::
+   
+   You can define the radius in kilometers, miles, degrees, or radians.
```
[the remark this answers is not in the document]
papafe: The text was correct regarding .NET but not in JS

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,303 @@
...
+   :alt: 3 GeoPolygons
+   :width: 250
+   :lightbox:
+
+.. _node-persist-geopoint:
+
+Persist GeoPoint Data
+---------------------
+
+If you want to persist ``GeoPoint`` data, it must conform to the
```
papafe, line 145: As said before, we need to make clear that GeoPoint is only for queries, and this section is instead for saving "a geo point" to realm

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,303 @@
...
+
+#. At a minimum, add the two fields required by the GeoJSON spec:
+
+   - A field of type ``double[]`` that maps to a "coordinates" (case sensitive) 
+     property in the realm schema. 
+
+   - A field of type ``string`` that maps to a "type" property. The value of this 
+     field must be "Point".
+
+The following example shows an embedded class named "MyGeoPoint" that is used
```
papafe, line 164: We need to say that to make things simple, developers can just "implement CanonicalGeoPoint", like in the example, that has already the correct shape

examples/node/v12/__tests__/geospatial.test.js
```diff
@@ -0,0 +1,209 @@
...
+    const companiesInPolygonWithTwoHoles = realm
+      .objects(Company)
+      .filtered("location geoWithin $0", polygonWithTwoHoles);
+    console.debug(
+      `Companies in small circle: ${companiesInPolygonWithTwoHoles.length}`
+    );
+    // :snippet-end:
+
+    expect(companiesInLargeCircle.length).toBe(1);
+    expect(companiesInSmallCircle.length).toBe(2);
```
papafe, line 202: In the first two assertions you can see that actually the small circle is bigger than the large one.

papafe (changes requested):

examples/node/v12/__tests__/geospatial.test.js
dacharyc: Gotcha. I reversed the circles in my example to show the one with radians first, while the other SDKs show kilometers first, and I made a naming error. 

I was confused by the results, though - I have gotten different results from my GeoCircle queries than the other SDKs:

- Kotlin gets 1 and 0: https://github.com/mongodb/docs-realm/blob/78c376c92404c38d6cd661ac7b21a0c61bae6b83/examples/kotlin/shared/src/commonTest/kotlin/com/mongodb/realm/realmkmmapp/Geospatial.kt#L180
- .NET gets 1 and 0: https://github.com/mongodb/docs-realm/blob/97678776eda3a288ce887b478ebe711d9a38ab4e/examples/dotnet/Examples/Geospatial.cs#L121

I'm getting 2 and 1. As far as I can tell, I'm using the same coordinates as the other SDKs so I think I'm missing something to explain why my results are different.

papafe: Now I see the problem here. If you look at how the circles are defined in .NET, that 0.25 is actually degrees, while you're using radians (that's much bigger). In JS at the moment there's no conversion method between degrees and radians, so if you want to keep the same example, you need to do the conversion yourself. 
This is how it's converted in .NET https://github.com/realm/realm-dotnet/blob/daaccf298e10694fc95be197709c8db9d099288a/Realm/Realm/DatabaseTypes/Geospatial/Distance.cs#L85C28-L85C28

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,303 @@
...
+      .. literalinclude::  /examples/generated/node/v12/geospatial.test.snippet.geocircle.ts
+          :language: typescript
+    
+    .. tab::
+      :tabid: javascript
+
+      .. literalinclude::  /examples/generated/node/v12/geospatial.test.snippet.geocircle.js
+          :language: javascript
+
+.. figure:: /images/geocircles.png
```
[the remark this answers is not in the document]
papafe: @dacharyc I just understood that I was misreading! In the .NET example 0.25 is in degrees, not in radians. 
So the image is actually correct, is the text in JS that's wrong.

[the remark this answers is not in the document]
dacharyc: Ah, thank you! I've changed the names back and have updated the test so it's using radians instead of degrees (although the image still shows degrees). I'll talk to the team about the image.

examples/node/v12/__tests__/geospatial.test.js
dacharyc: Thank you! I knew I was missing something and that was it. After converting to radians, I'm getting the expected result. I've updated the tests and the query, and have added a comment clarifying that the radius is in radians just to make it super clear in the code.

source/sdk/node/model-data/data-types/geospatial.txt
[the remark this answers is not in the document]
dacharyc: Thank you! I've updated this, and have also added comments to the code example noting it just to be super clear.

source/sdk/node/model-data/data-types/geospatial.txt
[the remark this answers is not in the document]
dacharyc: I've updated this language - do you think the update is clearer and more accurate?

examples/node/v12/__tests__/geospatial.test.ts
```diff
@@ -12,8 +12,10 @@ import Realm, {
 describe("Geospatial", () => {
   test("should define an object model, open a realm, and perform geospatial queries", async () => {
     // :snippet-start: define-geopoint-class
+    // Implement `CanonicalGeoPoint` instead of extending
+    // `Realm.Object` for convenience when persisting geodata.
```
papafe, line 16: Implementing `CanonicalGeoPoint` is not an alternative to extending `Realm.Object`, they can actually do both. 
I would just leave the first part "Implement CanonicalGeoPoint for convenience when persisting geodata points"

examples/node/v12/__tests__/geospatial.test.js
```diff
@@ -3,6 +3,8 @@ import Realm, { kmToRadians } from "realm";
 describe("Geospatial", () => {
   test("should define an object model, open a realm, and perform geospatial queries", async () => {
     // :snippet-start: define-geopoint-class
+    // Implement `CanonicalGeoPoint` instead of extending
```
papafe, line 6: Given that this is JS, you can remove this comment (it's only valid for TS)

source/examples/generated/node/v12/geospatial.test.snippet.define-geopoint-class.js
```diff
@@ -1,3 +1,5 @@
+// Implement `CanonicalGeoPoint` instead of extending
```
papafe, line 1: Same comment as up

source/sdk/node/model-data/data-types/geospatial.txt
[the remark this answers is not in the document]
papafe: What you wrote now is actually not correct. If you look at the link in the first comment you see this:
```TS
export type GeoPoint =
  | {
      latitude: number;
      longitude: number;
      altitude?: number;
    }
  | CanonicalGeoPoint
  | GeoPosition;
```

This means that, to express a `GeoPoint` is the union (` | `) of three types, so developers can choose one of those 3 forms:
- an object with latitude, longitute and optional altitude
- `CanonicalGeoPoint` (the same type that we suggest using when persisting)
- `GeoPosition`(that is an array with two or three elements)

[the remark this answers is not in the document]
dacharyc: Ahh, ok, got it! I misunderstood the prior feedback. I've updated the wording again. Thank you!

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,322 @@
...
+for persistance.
+
+To save geospatial data to realm, refer to :ref:`node-persist-geopoint`.
+
+.. _node-geocircle:
+
+GeoCircle
+~~~~~~~~~
+
+A :js-sdk:`GeoCircle <Realm.GeoCircle.html>` defines a circle on
```
krollins-mdb, line 53:
```suggestion
A :js-sdk:`GeoCircle <types/Realm.GeoCircle.html>` defines a circle on 
```

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,322 @@
...
+   :alt: Two GeoCircles
+   :width: 150
+   :lightbox:
+
+.. _node-geobox:
+
+GeoBox
+~~~~~~
+
+A :js-sdk:`GeoBox <Realm.GeoBox.html>` defines a rectangle on
```
krollins-mdb, line 89:
```suggestion
A :js-sdk:`GeoBox <types/Realm.GeoBox.html>` defines a rectangle on  
```

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,322 @@
...
+   :alt: 2 GeoBoxes
+   :width: 150
+   :lightbox:
+
+.. _node-geopolygon:
+
+GeoPolygon
+~~~~~~~~~~
+
+A :js-sdk:`GeoPolygon <Realm.GeoPolygon.html>` defines a polygon
```
krollins-mdb, line 120:
```suggestion
A :js-sdk:`GeoPolygon <types/Realm.GeoPolygon.html>` defines a polygon 
```

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,322 @@
...
+
+  - ``latitude``: a number value
+  - ``longitude``: a number value
+  - ``altitude``: an optional number value
+- ``CanonicalGeoPoint``
+- ``GeoPosition``
+
+A GeoPoint is used only as a building block of the other shapes: 
+GeoCircle, GeoBox, and GeoPolygon. GeoPoint is used only in queries, not 
+for persistance.
```
krollins-mdb, line 44: Perhaps I'm misreading, but it sounds like we're saying GeoPoint is only used for one thing twice, which meas it is used for two things: as a building block and in queries.

Or are we talking about an instance of GeoPoint being a building block and the GeoPoint type only being used in queries? If so, I think that nuance may be lost on some readers.

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,322 @@
...
+
+Geospatial Data Types
+---------------------
+
+.. _node-geopoint:
+
+GeoPoint
+~~~~~~~~
+
+A :js-sdk:`GeoPoint <Realm.GeoPoint.html>` defines a specific
```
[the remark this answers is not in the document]
krollins-mdb: Oh. I hadn't thought about this, but with the migration to the new TypeDoc API reference docs I might need to update _all_ the :js-sdk: links. 😱

krollins-mdb (approved):
Thanks for digging deeper on this feature, @dacharyc! The PR is looking good. Just a few small things and a note about links.

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,322 @@
...
+
+``GeoPoint`` can be one of three types:
+
+- An object with:
+
+  - ``latitude``: a number value
+  - ``longitude``: a number value
+  - ``altitude``: an optional number value
+- ``CanonicalGeoPoint``
+- ``GeoPosition``
```
papafe, line 40:
```suggestion
- ``GeoPosition``: an array with longitude, latitude, and an optional altitude
```

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,322 @@
...
+to define their location. 
+
+``GeoPoint`` can be one of three types:
+
+- An object with:
+
+  - ``latitude``: a number value
+  - ``longitude``: a number value
+  - ``altitude``: an optional number value
+- ``CanonicalGeoPoint``
```
papafe, line 39:
```suggestion
- ``CanonicalGeoPoint``:  an interface that satisfied the GeoJSON specifications for a point
```

examples/node/v12/__tests__/geospatial.test.js
```diff
@@ -0,0 +1,209 @@
...
+        location: new MyGeoPoint(-121.85, 47.9),
+      });
+    });
+    // :snippet-end:
+
+    // :snippet-start: geocircle
+    const smallCircle = {
+      center: [-121.9, 47.3],
+      // The GeoCircle radius is measured in radians
+      distance: 0.004363323,
```
papafe, line 62: Maybe we should at least say that this corresponds to 0.25 radians, so the image makes sense

papafe (approved):

source/sdk/node/model-data/data-types/geospatial.txt
dacharyc: The shapes are used for queries, so both things are true - it's a building block for queries. Let me tweak.

source/sdk/node/model-data/data-types/geospatial.txt
```diff
@@ -0,0 +1,322 @@
...
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. versionadded:: 12.0.0
+
+Geospatial data, or "geodata", specifies points and geometric objects on the Earth's 
+surface. With the geodata types, you can create queries that check whether a given  
+point is contained within a shape. For example, you can find all coffee shops within 
+15 km of a specified point.
```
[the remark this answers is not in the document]
dacharyc: Just tried it and it seems to render ok and doesn't throw an error!

PR #2940, opened 2023-08-09
DOCSP-30342: Update Read & Write Data page with info on managed and unmanaged objects

cbullinger:
## Pull Request Info

Add section describing managed vs. unmanaged objects to existing [Read and Write Data](https://www.mongodb.com/docs/realm/sdk/kotlin/realm-database/write-transactions/) page

### Jira

- https://jira.mongodb.org/browse/DOCSP-30342

### Staged Changes

- [Read & Write Data](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-30342-manage-section/sdk/kotlin/realm-database/write-transactions/) - updated page

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/kotlin/realm-database/write-transactions.txt
+++ b/source/sdk/kotlin/realm-database/write-transactions.txt
@@ -64,7 +64,13 @@ commits it or cancels it:
 Managed vs. Unmanaged Objects
 -----------------------------
 
-Realm objects written to a realm are either managed or unmanaged by Realm.
+Realm APIs may refer to objects as managed or unmanaged. When you create 
+a Realm object, it is unmanaged. You can then create a managed instance by 
+copying that object to a realm. 
+
+You can check if an object is managed with the 
+`isManaged() <{+kotlin-local-prefix+}io.realm.kotlin.ext/is-managed.html>`__  
+method.
 
 Managed Objects
 ~~~~~~~~~~~~~~~
@@ -72,10 +78,13 @@ Managed Objects
 **Managed objects** are Realm objects that persist in a realm. Managed objects 
 can only be accessed from an open realm. They can be automatically updated 
 with changes within write transactions as long as that realm remains open. 
+Managed objects are tied to the realm instance from which it originated 
+and cannot be written to another realm.
+
 You can use Realm APIs with managed objects, create relationships with 
 other objects, and observe managed Realm object for changes. 
 
-You create managed objects with 
+You create managed objects by copying unmanaged instances to a realm with 
 `Realm.copyToRealm() <{+kotlin-local-prefix+}io.realm.kotlin.dynamic/-dynamic-mutable-realm/copy-to-realm.html>`__. 
 See :ref:`Create Realm Objects <kotlin-create-a-new-object>` for more 
 information.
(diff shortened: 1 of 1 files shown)


source/sdk/kotlin/realm-database/write-transactions.txt
```diff
@@ -61,6 +61,45 @@ When you are done with your transaction, Realm either
   the transaction causes an error, all changes are discarded
   (or "rolled back").
 
+Managed vs. Unmanaged Objects
+-----------------------------
+
+Realm objects written to a realm are either managed or unmanaged by Realm.
```
dacharyc, line 67: This isn't entirely accurate as written; if a Realm object is written to a realm, it's managed. It's only unmanaged when it is _not_ written to a realm.
```suggestion
Realm APIs may refer to objects as managed or unmanaged. 
A Realm object becomes managed when you write it to a realm.
You can create unmanaged copies of managed objects.
```

source/sdk/kotlin/realm-database/write-transactions.txt
```diff
@@ -61,6 +61,45 @@ When you are done with your transaction, Realm either
...
+
+Managed Objects
```
dacharyc, line 69: I don't know if this is true in Kotlin, but in Swift, a managed object's lifecycle is tied to the realm instance from which it originated. So if you retrieve a managed object from a realm, make changes to it, and try to write it to a different instance of the realm, in Swift this throws an error because the object is managed by a different realm. Swift provides ways to get the originating realm if you need to mutate the object.

I know Kotlin's objects are frozen and I wonder if they work the same way (or if realms in Kotlin are singletons and there _is_ no separate instance of the realm) so I'm curious if there are implications around this that we should cover for Kotlin. I.e. do we need to talk about managed object lifecycle here, and the edge cases that can happen?

source/sdk/kotlin/realm-database/write-transactions.txt
```diff
@@ -61,6 +61,45 @@ When you are done with your transaction, Realm either
...
+can only be accessed from an open realm. They can be automatically updated 
+with changes within write transactions as long as that realm remains open. 
+You can use Realm APIs with managed objects, create relationships with 
+other objects, and observe managed Realm object for changes. 
+
+You create managed objects with 
+`Realm.copyToRealm() <{+kotlin-local-prefix+}io.realm.kotlin.dynamic/-dynamic-mutable-realm/copy-to-realm.html>`__. 
+See :ref:`Create Realm Objects <kotlin-create-a-new-object>` for more 
+information.
+
```
dacharyc, line 82: Should we mention that Realm objects have an `.isManaged` bool you can use to find out whether or not an object is managed?

source/sdk/kotlin/realm-database/write-transactions.txt
```diff
@@ -61,6 +61,45 @@ When you are done with your transaction, Realm either
...
+   access the managed data after it has been written.
+
+Unmanaged Objects
+~~~~~~~~~~~~~~~~~
+
+**Unmanaged objects** are instances of Realm objects that
+behave like normal Kotlin objects, but they are not persisted in a realm.
+You cannot use Realm APIs with unmanaged objects or observe them for changes.
+
+You can get an unmanaged object by manually constructing a Realm object
```
dacharyc, line 96: Should we say something here about _why_ you might want to make an unmanaged copy of a Realm object?

dacharyc (approved):
I think this is _mostly_ fine as written, although I have a nit about the opening line.

I do have a couple of non-blocking questions about whether we should add some additional details, though.

source/sdk/kotlin/realm-database/write-transactions.txt
cbullinger: Updated

source/sdk/kotlin/realm-database/write-transactions.txt
cbullinger: added

source/sdk/kotlin/realm-database/write-transactions.txt
cbullinger: added

rorbech (approved):

PR #2950, opened 2023-08-11
(DOCSP-30049) Set up React Native v12 testing suite

krollins-mdb:
## Pull Request Info

This PR creates a new test suite for React Native that uses v12 of the Realm.js SDK. The older test suite has been moved to a `legacy` directory.

No content changes. This is only to set up the infrastructure of the new test suite.

### Jira

- https://jira.mongodb.org/browse/DOCSP-30049

### Review Guidelines


examples/react-native/v12/TestApp/__tests__/App.test.jsx
```diff
@@ -0,0 +1,22 @@
+/**
+ * @format
+ */
```
nlarew, line 3: What's going on with this comment? I'm not familiar with `@format`.

examples/react-native/v12/TestApp/__tests__/App.test.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+import 'react-native';
+import React from 'react';
+import App from '../App';
+import {render} from '@testing-library/react-native';
+
+// Note: import explicitly to use the types shiped with jest.
+import {test, expect} from '@jest/globals';
```
nlarew, line 11:
```suggestion
// Note: import explicitly to use the types shipped with jest.
import {test, expect} from '@jest/globals';
```

Is there a reason we're going with explicit imports in each test file rather than using global variables? `@types/jest` provides globals out of the box but I think might have some lag behind the real jest types. Maybe we can import these and define some global vars in the jest setup file?

nlarew (approved):
One question about the jest globals but seems like a good base for the new tests. LGTM!

examples/react-native/v12/TestApp/__tests__/App.test.jsx
krollins-mdb: Ah, this test came along with the React Native template app. I hadn't cleaned it up yet. 😄

examples/react-native/v12/TestApp/__tests__/App.test.jsx
krollins-mdb: Oh, yeah. Good point! As mentioned above, this is the test file that came with the React Native template app and I hadn't cleaned it up yet.

I'll set up some global vars for testing types. Thanks for pointing this out!

PR #2956, opened 2023-08-16
DOCSP-32290 .NET Example Projects

lindseymoore:
### Jira

- https://jira.mongodb.org/browse/DOCSP-32290

### Staged Changes

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/dotnet.txt
+++ b/source/sdk/dotnet.txt
@@ -233,3 +233,41 @@ Recommended Reading
       :icon-alt: Realm Icon
 
       See the latest releases, new features, and bug fixes.
+
+Example Projects
+----------------
+
+.. card-group::
+   :columns: 3
+   :style: extra-compact
+
+   .. card::
+      :headline: Large File Storage
+      :icon: /images/icons/branding_2022/Technical_ATLAS_Functions3x.png
+      :icon-alt: Functions Icon
+
+      Store large files, such as images, on a remote server. Use Triggers and
+      Functions to store the references. Leverage Realm's offline-first design
+      to store large images, even when offline.
+
+      Available in `Example Project <https://github.com/realm/realm-dotnet-lfs>`__.
+
+   .. card::
+      :headline: Analytics/Telemetrics Data
+      :icon: /images/icons/branding_2022/Technical_REALM_SDK3x.png
+      :icon-alt: Realm Icon
+
+      Take sensor readings, push data to Atlas, and display the results in
+      Atlas Charts or a time series collection.
+
+      Available in `Example Project <https://github.com/realm/realm-dotnet-samples/tree/main/AnalyticsTelemetry>`__.
+
+   .. card::
+      :headline: Realm with REST APIs
+      :icon: /images/icons/branding_2022/Technical_REALM_SDK3x.png
+      :icon-alt: Realm Icon
+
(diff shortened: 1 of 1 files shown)


source/sdk/dotnet.txt
```diff
@@ -233,3 +233,41 @@ Recommended Reading
...
+Example Projects
+----------------
+
+.. card-group::
+   :columns: 3
+   :style: extra-compact
+
+   .. card::
+      :headline: Large File Storage
+      :icon: /images/icons/branding_2022/Technical_ATLAS_Functions3x.png
```
dacharyc, line 246:
```suggestion
      :headline: Large File Storage
      :cta: Example app
      :url: https://github.com/realm/realm-dotnet-lfs
      :icon: /images/icons/branding_2022/Technical_ATLAS_Functions3x.png
```

dacharyc: We should update these cards to use the `:cta:` and `:url:` fields instead of having the `Example Project` link below. This makes the whole card clickable, and gives it the on hover effect it's designed to have.

source/sdk/dotnet.txt
```diff
@@ -233,3 +233,41 @@ Recommended Reading
...
+
+      Store large files, such as images, on a remote server. Use Triggers and
+      Functions to store the references. Leverage Realm's offline-first design
+      to store large images, even when offline.
+
+      `Example Project <https://github.com/realm/realm-dotnet-lfs>`__
+
+   .. card::
+      :headline: Analytics/Telemetrics Data
+      :icon: /images/icons/branding_2022/Technical_REALM_SDK3x.png
```
dacharyc, line 257: Looks like this is the same `:icon:` and `:icon-alt:` as the Realm with Rest APIs card below. We should update one or both of these. For the Swift Analytics/Telemetrics card, I went with this one: https://github.com/mongodb/docs-realm/blob/69d0d79d56c2867c1d5b4773b4f7102993225a33/source/sdk/swift.txt#L266

It might make sense to use the same one. Or you could hunt through the brand icons to find something you prefer. Ping me on Slack if you need help finding out where those live!

dacharyc (changes requested):
Back to you for a few tweaks!

source/sdk/dotnet.txt
```diff
@@ -233,3 +233,41 @@ Recommended Reading
...
+      Store large files, such as images, on a remote server. Use Triggers and
+      Functions to store the references. Leverage Realm's offline-first design
+      to store large images, even when offline.
+
+   .. card::
+      :headline: Analytics/Telemetrics Data
+      :cta: Example App
+      :url: https://github.com/realm/realm-dotnet-samples/tree/main/AnalyticsTelemetry
+      :icon: /images/icons/branding_2023/Technical_ATLAS_AsymmetricSync10x.png
+      :icon-alt: Realm Icon
```
dacharyc, line 260: One tiny nit: this `:icon-alt:` text should probably be different than the one below since these are two different icons. Maybe something like this? (Fudging the naming a little as this is listed as an asymmetric sync icon but I don't think it matters for this use case"
```suggestion
      :icon-alt: Atlas Device Sync Icon
```

dacharyc (approved):
One tiny suggestion, but LGTM!

PR #2963, opened 2023-08-18
(DOCSP-27734): SwiftUI: Clarify adding and removing from ObservedResults

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-27734

### Staged Changes

- [Write Data - SwiftUI](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-27734/sdk/swift/swiftui/write/#add-or-remove-objects-in-an-observedresults-collection)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -15,7 +15,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: "14.3"
+          xcode-version: "14.3.1"
       - name: Build
         env:
           scheme: ${{ 'default' }}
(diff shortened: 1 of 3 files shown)


cbullinger (approved):
LGTM (trying to understand the ticket: we were already adding and removing in the example, but just not specifying/explaining it in the copy?)

dacharyc:
> we were already adding and removing in the example, but just not specifying/explaining it in the copy?

Yes, we were already adding and removing in the example, but I missed this section on the page when I was looking for it because it said "Update" and not "Add/Remove".

Then, when I started to look at the explanation, it didn't seem explicit enough for me so I reworded it a bit and added the emphasize lines to make it clearer where we were doing the thing we said we were doing.

The other part of the ticket was about ObservedSectionResults, which I'm not touching here as there is currently a bug in this functionality.

PR #2964, opened 2023-08-18
DOCSP-28146: Fix incorrect ObjectId link in Node Quick Start

cbullinger:
## Pull Request Info

Bug bash: Link should point to [BSON package](https://www.mongodb.com/docs/manual/reference/method/ObjectId/)

### Jira

- https://jira.mongodb.org/browse/DOCSP-28146

### Staged Changes

- [Quick Start - Node.js SDK](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/docsp-28146-js-objectid/sdk/node/quick-start/#define-your-object-model)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/node/quick-start.txt
+++ b/source/sdk/node/quick-start.txt
@@ -43,7 +43,7 @@ Object Properties <node-define-a-property>`.
 The following code shows how to define an object model for a ``Task`` object. In this example:
   
 - The ``primaryKey`` is the ``_id`` of type ``int``. Another common type used for 
-  primary keys is :js-sdk:`ObjectId <Realm.Object.html#objectId>`.
+  primary keys is :manual:`ObjectId </reference/method/ObjectId/>`.
 - The ``name`` field is required.
 - The ``status`` and ``owner_id`` fields are optional, denoted by the question 
   mark immediately after the data type.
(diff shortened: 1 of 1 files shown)


source/sdk/node/quick-start.txt
```diff
@@ -43,7 +43,7 @@ Object Properties <node-define-a-property>`.
 The following code shows how to define an object model for a ``Task`` object. In this example:
   
 - The ``primaryKey`` is the ``_id`` of type ``int``. Another common type used for 
-  primary keys is :js-sdk:`ObjectId <Realm.Object.html#objectId>`.
+  primary keys is :manual:`ObjectId </reference/method/ObjectId/>`.
```
dacharyc, line 46: I don't _think_ this is quite the right spot to link. The comment on the Jira ticket says that the ObjectId we usually use for Realm comes from the BSON package, which the React Native reference docs link out to here: https://www.npmjs.com/package/bson#documentation|the

We might want to add a sentence that `ObjectId` used in Realm comes from the BSON package, and either link to the above or link directly to the API reference for ObjectId here: https://mongodb.github.io/node-mongodb-native/Next/modules.html#ObjectId

I'm undecided what the correct fix is, but I don't think it's linking out to the manual for info on ObjectId, based on the user's feedback.

dacharyc (approved):
I don't _think_ this is quite the correct fix. But it's not 100% clear to me what the best fix is. So I'll leave it up to your judgement - I've provided a couple of alternative suggestions.

source/sdk/node/quick-start.txt
cbullinger: Kyle suggested this should point to the field types page. not great but at least fixes this.

PR #2970, opened 2023-08-23
(DOCSP-32426): Swift: Query MongoDB Test Fixes and Remove PBS from Examples

dacharyc:
## Pull Request Info

The linked Jira ticket requests fixing some failing tests, but we've got other tickets to remove PBS examples from the docs. Since I'm touching the page anyway, I've switched the PBS examples. While I'm in here, I've added IO code blocks which were not available when this page was created.

### Jira

- https://jira.mongodb.org/browse/DOCSP-32426

### Staged Changes

- [Query MongoDB](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-32426/sdk/swift/app-services/mongodb-remote-access/)

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/swift/app-services/mongodb-remote-access.txt
+++ b/source/sdk/swift/app-services/mongodb-remote-access.txt
@@ -114,18 +114,19 @@ Insert a Single Document
 
 You can insert a single document using :swift-sdk:`collection.insertOne() <Extensions/MongoCollection.html#/s:So18RLMMongoCollectionC10RealmSwiftE9insertOneyySDySSAC7AnyBSONOSgG_ys6ResultOyAFs5Error_pGctF>`.
 
-This snippet inserts a single document describing a "Bean of the Day" coffee 
+This snippet inserts a single document describing a "Colombian Blend" coffee 
 drink into a :ref:`collection of documents that describe coffee drinks for sale 
 in a group of stores <ios-mongodb-example-dataset>`:
 
-.. literalinclude:: /examples/generated/code/start/MongoDBRemoteAccess.snippet.insert-one.swift
-   :language: swift
+.. io-code-block::
 
-Running this snippet produces this output:
+   .. input:: /examples/generated/code/start/MongoDBRemoteAccess.snippet.insert-one.swift
+      :language: swift
 
-.. code-block:: text
+   .. output::
+      :language: text
 
-   Successfully inserted a document with id: objectId(6108...)
+      Successfully inserted a document with id: objectId(64e50cab7765243942cd04ce)
 
 .. _ios-mongodb-insertMany:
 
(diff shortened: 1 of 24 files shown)


source/sdk/swift/app-services/mongodb-remote-access.txt
```diff
@@ -197,26 +201,17 @@ This snippet finds all documents in a :ref:`collection of documents that
 describe coffee drinks for sale in a group of stores <ios-mongodb-example-dataset>`,
```
krollins-mdb, line 201: While you're cleaning this page up, maybe reconsider having the ref cover this much text. Hyperlink text is considered non-breaking in HTML, which can introduce weird layout issues when there's a lot of link text.

I see this happens in other parts of the page too.

source/sdk/swift/app-services/mongodb-remote-access.txt
```diff
@@ -241,26 +236,29 @@ describe coffee drinks for sale in a group of stores <ios-mongodb-example-datase
...
-      "name": Optional(RealmSwift.AnyBSON.string("Americano"))]
+.. io-code-block::
+
+   .. input:: /examples/generated/code/start/MongoDBRemoteAccess.snippet.find-and-sort.swift
+      :language: swift
+
+   .. output::
+      :language: text
+
+      Results:
```
krollins-mdb, line 247: The results bit feels a little odd because the test creates two Americano objects, but we don't indicate that anywhere in this code snippet. Readers have no idea how many Americano coffees there are.

This make the two lines of the same thing in the results like a typo. I think we should add a comment in the code to indicate there should be two coffees or add a third line to results to make it feel more intentional.

source/sdk/swift/app-services/mongodb-remote-access.txt
```diff
@@ -458,30 +462,49 @@ or apply a ``$match`` filter to incoming change events with :swift-sdk:`collecti
...
-   Change stream closed
+This code watches for changes to documents in the ``CoffeeDrinks`` 
+collection:
+
+.. io-code-block::
+
+   .. input:: /examples/generated/code/start/MongoDBRemoteAccess.snippet.watch-collection.swift
+      :language: swift
+
+   .. output::
```
krollins-mdb, line 479: I think we can use `:visible:` to control whether the output is visible by default. I think it may be a good idea to hide larger outputs by default - letting people expand them if they're interested.

source/sdk/swift/app-services/mongodb-remote-access.txt
```diff
@@ -458,30 +462,49 @@ or apply a ``$match`` filter to incoming change events with :swift-sdk:`collecti
...
-   Successfully inserted a document with id: objectId(642602546207d8b9ce866ec4)
-   Change stream closed
+This code watches for changes to documents in the ``CoffeeDrinks`` 
+collection:
```
krollins-mdb, line 472: In the "Watch for Changes in a Collection to Specific IDs" section, this introductory bit says something like this:
> The code below uses this delegate to watch for changes...

Should we add that here too? I was a little confused by the transition between the delegate example and the watch for changes example in this spot.

source/sdk/swift/app-services/mongodb-remote-access.txt
```diff
@@ -532,31 +581,59 @@ used as the query of a :manual:`$match operator
...
-   Successfully updated the document
-   Change stream closed
+To use the code below to watch for changes to documents in the
+``CoffeeDrink`` collection, but only trigger the provided callback for
+events corresponding to documents belonging to the ``storeNumber`` named
+``42``:
```
krollins-mdb, line 594: This sentence is a lot. Can we break it down?
What do you think about this?
```suggestion
The code below watches for changes to documents in the
``CoffeeDrink`` collection. It only triggers the provided callback for
events related to documents that belong to the ``storeNumber`` named
``42``:
```

krollins-mdb (approved):
LGTM! Just a few minor suggestions.

source/sdk/swift/app-services/mongodb-remote-access.txt
dacharyc: Yeah, you're right, good catch, thank you! I broke this down a little more.

PR #2896, opened 2023-07-18
(DOCSP-28906) Add core logging docs for Node.js SDK

krollins-mdb:
## Pull Request Info

DO NOT MERGE until v12 GA ships. After release:

- [x] Add links to generated API documentation for new APIs

### Jira

- https://jira.mongodb.org/browse/DOCSP-28906

### Staged Changes

- [Logging](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28906-nodejs/sdk/node/logging/). New page that outlines the new feature.
- [Set the Client Log Level](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-28906-nodejs/sdk/node/sync/log-level/). Add warning that points readers to the new core logger.

### Review Guidelines

[REVIEWING.md](https://github.com/mongodb/docs-realm/blob/master/REVIEWING.md)

PR #2981, opened 2023-08-30
(DOCSP-32575) Add new logging page for React Native

krollins-mdb:
## Pull Request Info

This PR adds info for the Realm core logger. Simlar info was added for the Node.js SDK in #2896.

I also made some minor changes to the React Native test suite.

### Jira

- https://jira.mongodb.org/browse/DOCSP-32575

### Staged Changes

- [Logging](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-32575/sdk/react-native/logging/)

### Review Guidelines


examples/react-native/v12/TestApp/App.tsx
```diff
@@ -1,20 +1,35 @@
...
 import {createDrawerNavigator} from '@react-navigation/drawer';
+import 'react-native-get-random-values';
 
 import {HomeScreen} from './src/screens/HomeScreen';
 import {Geospatial} from './src/components/data-types/Geospatial';
+import {Logger} from './src/components/Logger';
 
 const Drawer = createDrawerNavigator();
 
+/*
```
krollins-mdb, line 13: @takameyer, this is potentially out of scope for this PR, but do you have any insights here?

examples/react-native/v12/TestApp/src/components/Logger.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+export const Logger = () => {
+  const [logs, setLogs] = useState<Log[]>([]);
+
+  // Can't set logger in useEffect at same level as `RealmProvider` because
+  // useEffect runs afer `componentDidMount`. This means that `RealmProvider`
+  // has already opened a realm and the `setLogger` callback needs to be
+  // set up before the realm is opened.
+
+  Realm.setLogger((level, message) => {
```
krollins-mdb, line 44: @takameyer, as-is this will re-set the logger every time the component is re-rendered. Not ideal, but I couldn't think of a better way to do it. `useEffect` won't work (see the above comment, lines 39-42).

krollins-mdb: Ah, yeah, that makes a lot of sense. I'll move `logLevel` and `Realm.setLogger` outside of the React tree.

I'd forgotten about `useLayoutEffect`! I won't use it here, but I need to remember it in the future. 😄

source/sdk/react-native/sync-data/log-level.txt
```diff
@@ -1,4 +1,4 @@
-.. _react-native-set-the-client-log-level:
```
dacharyc, line 1: FWIW, it has generally been our policy not to remove ref targets. Because intersphinx supports linking to a ref target from other repositories, changing this target could potentially break links from other repos that we would not have visibility into from here. Instead, you could add the new ref target as a second line on this page, and leave this one in place.

dacharyc (approved):
One minor suggestion just to avoid potentially breaking links from other repos - otherwise, LGTM!

source/sdk/react-native/sync-data/log-level.txt
krollins-mdb: Ah, thank you for catching that! Added back in.

PR #2982, opened 2023-08-30
DOCSP-32483 programmatic tagging script for realm repo

sarahemlin:
Hey @cbush,

I ran the programmatic tagging script on the Realm repo and it looks like everything went through ok on my end as expected.  Please let me know if anything looks weird or if my process isn't quite right.

Thanks!
Sarah

### Jira

- https://jira.mongodb.org/browse/DOCSP-32483

### DIFF
--- a/source/examples/generated/dotnet-flexible-sync/facets.toml
+++ b/source/examples/generated/dotnet-flexible-sync/facets.toml
@@ -1,0 +1,7 @@
+[[facets.sub_facets]]
+category = "sub_product"
+value = "dotnet"
+
+[[facets]]
+category = "programming_language"
+value = "csharp"
--- a/source/examples/generated/dotnet/facets.toml
+++ b/source/examples/generated/dotnet/facets.toml
@@ -1,0 +1,7 @@
+[[facets.sub_facets]]
+category = "sub_product"
+value = "dotnet"
+
+[[facets]]
+category = "programming_language"
+value = "csharp"
--- a/source/examples/generated/flutter/facets.toml
+++ b/source/examples/generated/flutter/facets.toml
@@ -1,0 +1,7 @@
+[[facets]]
+category = "programming_language"
+value = "dart"
+
+[[facets.sub_facets]]
+category = "sub_product"
+value = "flutter"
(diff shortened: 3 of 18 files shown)


cbush:
@dacharyc @krollins-mdb

sarahemlin:
@dacharyc I removed the generated files and took out some redundancy so this one is ok to merge if it passes all checks.  I'll use this as a chance to test out updating the script and run it again to add in the missing sub_product values.

dacharyc (approved):
With the caveats that some of these are missing `sub_facets` / `sub_product` - this looks like a good starting point.

Although I am curious why some of the programming languages have two separate `programming_language` facets (Java has java & kotlin as separate facets, Swift has objective-c and swift as separate facets) while others have one facet with two programming languages separated by a slash (i.e. Node, RN, and Web all have one facet that is "javascript/typescript"). I'm wondering if JS and TS should be separate facets, or the others should be a combined facet, or if there's some internal logic driving this difference and it's fine as-is.

sarahemlin:
@dacharyc the language values of pushing JS/TS together was suggested/approved by the leads as being the same (the TS page just says use JS), so that's why it's different and the others they wanted spelled out.  That's something we could always change in the future if we get user feedback that they need to be different.

Seung is OOO until next week, so I can't get a fix to the script to update the missing sub_product values until then, so I think you can go ahead and merge.  Thank you!

dacharyc:
@sarahemlin Here in Realm-land, the languages are typically presented as two tabs on the same page, showing users how to work with each of the languages. Here are some examples of our [Quick Start for Java](https://www.mongodb.com/docs/realm/sdk/java/quick-starts/quick-start-local/) and our [Quick Start for Node.js](https://www.mongodb.com/docs/realm/sdk/node/quick-start/). You can see we are handling the languages the same way in each SDK, so distinguishing them in some cases but not others seems like an arbitrary distinction for our docs. But I don't know how these are being used ultimately, so it might make sense to do it this way in the bigger context of the purpose of these taxonomies.

Just wanted to flag it in case it's something that warranted further investigation or an artifact of the script, etc.

PR #2983, opened 2023-08-30
(DOCSP-30214) Flutter: Create new filter example with iterable argument

krollins-mdb:
### Jira

- https://jira.mongodb.org/browse/DOCSP-30214

### Staged Changes

- [Read and Write Data](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-30214/sdk/flutter/realm-database/read-and-write-data/#filter-results)

### Review Guidelines

### DIFF
--- a/examples/dart/test/read_write_data_test.dart
+++ b/examples/dart/test/read_write_data_test.dart
@@ -95,6 +95,18 @@ void main() {
       // :snippet-end:
       expect(team.name, 'Millennium Falcon Crew');
       expect(humanCrewMembers.length, 3);
+
+      // :snippet-start: filter-iterable
+      final listOfNames = ['Luke', 'Leia'];
+      final matchingRealmObjects =
+          realm.query<Person>(r'name IN $0', [listOfNames]);
+      // :snippet-end:
+      expect(matchingRealmObjects.length, 2);
+
+      for (var person in matchingRealmObjects) {
+        print(person.name);
+      }
+
       cleanUpRealm(realm);
     });
     test("Sort Results", () {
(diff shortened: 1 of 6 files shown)


examples/dart/test/read_write_data_test.dart
```diff
@@ -95,6 +95,18 @@ void main() {
       // :snippet-end:
       expect(team.name, 'Millennium Falcon Crew');
       expect(humanCrewMembers.length, 3);
+
+      // :snippet-start: filter-iterable
+      final listOfNames = ['Luke', 'Leia'];
+      final matchingRealmObjects =
+          realm.query<Person>(r'name IN $0', [listOfNames]);
```
dacharyc, line 102: I am assuming this is a typo? (Although if it is, I'm a little surprised we don't get an error.)
```suggestion
          realm.query<Person>('name IN $0', [listOfNames]);
```

source/examples/generated/flutter/read_write_data_test.snippet.filter-iterable.dart
```diff
@@ -0,0 +1,3 @@
+final listOfNames = ['Luke', 'Leia'];
+final matchingRealmObjects =
+    realm.query<Person>(r'name IN $0', [listOfNames]);
```
dacharyc, line 3: If this is a typo and you just want to batch these suggestions, offering this here for convenience so you don't have to regenerate w/Bluehawk.
```suggestion
    realm.query<Person>('name IN $0', [listOfNames]);
```

dacharyc (approved):
Minor typo (I think?) but mod that fix, LGTM 👍

examples/dart/test/read_write_data_test.dart
krollins-mdb: I actually am not sure exactly what it does. I should ask the Flutter team. They use it in their tests instead of escaping `$` for paramerterized queries.

However, since we use the `\$0` construction in the rest of the Flutter docs, I'm going to change to that.

PR #2985, opened 2023-08-31
React Native: Fix incorrect optional properties example

krollins-mdb:
### Staged Changes

- [Define a Realm Object Model](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/fix-rn-optional-property-example/sdk/react-native/model-data/define-a-realm-object-model/#declare-an-optional-property)

### Review Guidelines

### DIFF
--- a/examples/react-native/v12/TestApp/src/components/ObjectModels.tsx
+++ b/examples/react-native/v12/TestApp/src/components/ObjectModels.tsx
@@ -15,11 +15,11 @@ class Person extends Realm.Object<Person> {
       name: 'string',
       age: {
         type: 'int',
-        optional: true,
+        optional: true, // :emphasize:
       },
       birthday: {
         type: 'date',
-        optional: true,
+        optional: true, // :emphasize:
       },
     },
   };
--- a/examples/react-native/v12/TestApp/src/components/ObjectModels.tsx
+++ b/examples/react-native/v12/TestApp/src/components/ObjectModels.tsx
@@ -1,6 +1,6 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import Realm, {ObjectSchema} from 'realm';
-import {RealmProvider, useRealm} from '@realm/react';
+import {RealmProvider, useQuery, useRealm} from '@realm/react';
 import {View, Text, FlatList} from 'react-native';
 
 // :snippet-start: model-optional-properties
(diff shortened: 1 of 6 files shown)


source/sdk/react-native/model-data/define-a-realm-object-model.txt
```diff
@@ -69,30 +69,14 @@ The following schema defines a ``Car`` type that has these properties: ``_id``
 Declare an Optional Property
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To mark a property as optional, append a question mark ``?`` to its type. You can
-also add ``optional`` to more complicated properties and set it to ``true`` or
-``false``.
+To mark a property as optional, use object syntax and set ``optional`` to
```
dacharyc, line 72: I think both of these things are true. I think you can still use the `?` syntax, but you could _also_ use the more verbose/explicit `optional` annotation. The release notes show both of these syntaxes as "Good": https://github.com/realm/realm-js/releases/tag/v12.0.0

And LJ commented on my Relationships PR and suggested the shorthand declaration style was more idiomatic.

Maybe it would make sense to show one of these using the `?` syntax and one using the more verbose `optional`?

examples/react-native/v12/TestApp/src/components/ObjectModels.tsx
dacharyc, line 1: This seems fine, but if the purpose of the tests is to confirm our code is accurate, should we be testing the thing we say this is doing? i.e. should we be writing something using these optional properties and then confirming the object has the value we expect?

dacharyc (approved):
Couple of minor comments, but we should probably align internally and w/engineering on how we want to communicate around the preferred schema form for the JS SDKs.

source/sdk/react-native/model-data/define-a-realm-object-model.txt
krollins-mdb: Ah, great point. I was kind of following the "be more specific" philosophy behind the v12 changes. But it probably makes sense to include a shorthand example and note it as such.

examples/react-native/v12/TestApp/src/components/ObjectModels.tsx
krollins-mdb: You're absolutely right. I'll update. I had in my head that the only thing that needed testing was that the object model itself was valid and you could open a realm with it. But we should also test the optional nature of objects based on the model.

PR #2996, opened 2023-09-05
DOCSP-32483: add missing Realm subProducts

sarahemlin:
### Jira

- https://jira.mongodb.org/browse/DOCSP-32483

@dacharyc I got the facets.csv source file updated with the Realm subProduct values and re-ran the script, which should correct the missing subProduct values for each SDK that was previously missing.  Please review & merge if it looks good!

Thank you!
Sarah

### DIFF
--- a/source/sdk/react-native/facets.toml
+++ b/source/sdk/react-native/facets.toml
@@ -1,0 +1,11 @@
+[[facets]]
+category = "programming_language"
+value = "javascript/typescript"
+
+[[facets]]
+category = "target_product"
+value = "realm"
+
+[[facets.sub_facets]]
+category = "sub_product"
+value = "react-native"
--- a/source/sdk/react-native/facets.toml
+++ b/source/sdk/react-native/facets.toml
@@ -2,6 +2,10 @@
 category = "programming_language"
 value = "javascript/typescript"
 
+[[facets]]
+category = "target_product"
+value = "realm"
+
 [[facets.sub_facets]]
 category = "sub_product"
 value = "react-native"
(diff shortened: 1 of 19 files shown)


source/sdk/react-native/facets.toml
```diff
@@ -0,0 +1,11 @@
+[[facets]]
+category = "programming_language"
+value = "javascript/typescript"
+
+[[facets]]
+category = "target_product"
+value = "realm"
```
dacharyc, line 7: Hey @sarahemlin - just one question about this one in particular. The React Native SDK seems to be the only one tagged specifically with a `target_product` whose value is `realm`. 

The `source` directory itself has a `target_product` whose value is `realm`. Is it safe to assume this applies to all the SDKs within the source directory since it's the root dir of these products?

If that's correct, it would seem that adding `target_product`:`realm` to RN specifically is redundant and may signal something in the script needs tweaking. If not, it would seem _all_ the SDKs should have this `target_product`: `realm`.

Assuming there isn't something that needs to be done relative to this (removing this redundant `target_product` if having it appear twice will cause a problem, or adding it to all the other SDKs), everything else LGTM!

dacharyc (approved):
LGTM, potentially mod one Q if anything needs to be done as a result.

If nothing needs to be done, just let me know and I'm happy to merge this.

source/sdk/react-native/facets.toml
sarahemlin: Dang!  I thought I'd gotten them all.  You are correct: it's not supposed to be there, though it doesn't hurt anything.  I removed it and committed to my branch, so I think it's good to merge now? (fingers crossed!)

dacharyc: Alrighty, looks good! I'll go ahead and merge. 🤞

PR #3000, opened 2023-09-06
(DOCSP-32291): C++: Open a realm overload now takes a db::config

dacharyc:
## Pull Request Info

In updating to the current version of the C++ SDK, I also had to change `value()` to `detach()` as it has been renamed.

Prior to merging:

- [x] Update CMakeLists.txt `GIT_TAG` for C++ SDK to the release version instead of using a commit hash

### Jira

- https://jira.mongodb.org/browse/DOCSP-32291

### Staged Changes

- [Configure & Open a Realm](https://preview-mongodbdacharyc.gatsbyjs.io/realm/DOCSP-32291/sdk/cpp/realm-files/configure-and-open-a-realm/#provide-a-subset-of-classes-to-a-realm): New "Provide a Subset of Classes to a Realm" section

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/examples/cpp/beta/local/crud.cpp
+++ b/examples/cpp/beta/local/crud.cpp
@@ -223,7 +223,7 @@ TEST_CASE("Beta embedded object example", "[write]") {
         mongoDB.contactDetails->emailAddress = "info@example.com";
     });
     
-    std::cout << "New email address: " << mongoDB.contactDetails->emailAddress.value() << "\n";
+    std::cout << "New email address: " << mongoDB.contactDetails->emailAddress.detach() << "\n";
     // :snippet-end:
     REQUIRE(mongoDB.contactDetails->emailAddress == "info@example.com");
     // :snippet-start: beta-overwrite-embedded-object
@@ -294,7 +294,8 @@ TEST_CASE("Pass a subset of classes to a realm", "[write]") {
     std::cout << "dog: " << dog.name << "\n";
 
     // :snippet-start: beta-realm-specify-classes
-    auto realm = realm::experimental::open<Beta_Dog>(path);
+    auto config = realm::db_config();
+    auto realm = realm::experimental::open<Beta_Dog>(std::move(config));
     // :snippet-end:
 
     auto managedDog = realm.write([&] {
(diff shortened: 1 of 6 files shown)


source/sdk/cpp/realm-files/configure-and-open-a-realm.txt
```diff
@@ -236,3 +227,41 @@ To open a synced realm:
...
+
+   .. tab:: Current
+      :tabid: current
+
+      By default, the C++ SDK automatically adds all object types that have a
+      :ref:`cpp-object-schema` in your executable to your 
+      :ref:`cpp-realm-schema`. 
+
+      However, if you want the realm to manage only a subset of classes, you
+      can specify those models by passing them into the template parameter list
```
krollins-mdb, line 252: Perhaps there's a C++ SDK distinction here I'm not aware of, but classes are not necessarily 1:1 equivalent with models.

Here, in the section title, and on line 240, we're referring to "classes". However, we also talk about object models in several places. Nowhere have we established the relationship between classes and object models. We're assuming readers will understand that object models are classes, but that not all classes are object models.

Would it be possible to replace references to classes with "object model/s"?

krollins-mdb (approved):
LGTM, with one question/suggestion.

examples/cpp/beta/local/crud.cpp
```diff
@@ -223,7 +223,7 @@ TEST_CASE("Beta embedded object example", "[write]") {
         mongoDB.contactDetails->emailAddress = "info@example.com";
     });
     
-    std::cout << "New email address: " << mongoDB.contactDetails->emailAddress.value() << "\n";
+    std::cout << "New email address: " << mongoDB.contactDetails->emailAddress.detach() << "\n";
```
cbush, line 226: What does "detach" do vs. "value"?

cbush (approved):

examples/cpp/beta/local/crud.cpp
dacharyc: From the PR that changes this:

> * `managed<>::value()` has been renamed to `managed<>::detach()` to better convey that the returned value will be unmanaged. In the case where the value is a pointer type it is up to the consumer of the value to manage the lifetime of the object.

We have not really dived into object lifetimes in this SDK but it may be time to consider adding something to the docs about this.

It's also inconsistent with some of the design feedback we've gotten, where designs were chosen so the SDK didn't have to convey managed vs. unmanaged, but this is a change explicitly to convey that. So we may need some info around this, too.

I'll make a separate Jira ticket to add details around this.

source/sdk/cpp/realm-files/configure-and-open-a-realm.txt
dacharyc: Good catch! In Swift, where I borrowed this section from, models are always classes so I use the term interchangeably. In C++, models can be classes or structs, so models is probably more appropriate. Updating.

PR #3013, opened 2023-09-14
Add info about SubscribeAsync

MongoCaleb:
### Jira

https://jira.mongodb.org/browse/DOCSP-31591

This PR also includes a lot of cleanup work, including replacing missing code snippets and rearranging sections. Very little new content, in fact.

### Staged Changes

https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-31591/sdk/dotnet/sync/flexible-sync/


examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -236,9 +267,48 @@ public void WeeSnippetForDocs()
             });
             // :snippet-end:
         }
-    }
 
+        public async void MoreFlexSyncExamples()
```
krollins-mdb, line 271: Do we not typically have assertions, like the .NET team does in [their test files](https://github.com/realm/realm-dotnet/pull/3403/files#diff-d47317df0d71bb36af164af65b9a49d763dbb92382e671b51455dafec926b539)?

Not that we're testing the SDK itself - rather, asserting that our implementation of the SDK works as expected.

source/sdk/dotnet/sync/flexible-sync.txt
```diff
@@ -124,50 +93,70 @@ queries you want to use to bootstrap the realm, as shown in the following exampl
...
-Add a Subscription
-``````````````````
-To add a subscription to an existing ``FlexibleSyncConfiguration``, 
-You must perform all mutations on the subscriptions set within an update block.
-To create an update block, call the :dotnet-sdk:`SubscriptionSet.Update()
-<reference/Realms.Sync.SubscriptionSet.html#Realms_Sync_SubscriptionSet_Update_System_Action_>`.
-method.
-
-Within the delegate, call the :dotnet-sdk:`SubscriptionSet.Add()
+Add a Subscription to the Existing Subscription Set
```
krollins-mdb, line 96: The ideal path is initial subscriptions (broad data sync), then use `.SubscribeAsync()` to make changes where needed, right?

Would it make sense to mention that the existing subscription set would likely be the one set by initial subscriptions? I know this section is right after the initial subs section, but readers won't necessarily get to this section in a linear fashion.

examples/dotnet/Examples/FlexibleSyncExamples.cs
```diff
@@ -236,9 +267,48 @@ public void WeeSnippetForDocs()
...
+        {
+            var realm = Realm.GetInstance();
+            // :snippet-start: subasync
+            var query = realm.All<Team>().Where(t => t.Name == "MyTeam");
+            await query.SubscribeAsync();
 
+            // you can also pass a SubscriptionOptions object:
+            await query.SubscribeAsync(
+                new SubscriptionOptions() { Name = "myteam" });
```
krollins-mdb, line 280: Do you think it's clear to readers that `new SubscriptionOptions() { Name = "myteam" });` is naming the subscription? I was initially confused - though that's likely because of my inexperience with this SDK and not seeing the SubscriptionOptions section lower on the page.

I think part of my confusion was because the query uses the Name field for its logic. At a quick glance, they seem to be doing similar things - especially with their values being so similar.

krollins-mdb (approved):
This page looks great! I have some minor questions/suggestions.

PR #3062, opened 2023-10-30
Swift: Add docs for SyncSession.reconnect()

dacharyc:
## Pull Request Info

A few questions for a tech reviewer based on the similar docs in Kotlin:
- In Kotlin, we say "The SDK also automatically calls this method when a device toggles off airplane mode" - is this accurate in the Swift SDK, also? I didn't see anything special for this in the SDK PR so wondering if this is something in Core?
- In Kotlin, we say "the SDK forces all sync sessions to attempt to reconnect immediately" - is this also true in Swift, or is it forcing only the `SyncSession` you call it on to reconnect? Does this change with session multiplexing enabled/disabled?
- Kotlin has a callout about not being able to force a reconnection within the socket read timeout duration - is this also true in Swift?
- It seems the [RLMSyncSession got docs for the new method](https://www.mongodb.com/docs/realm-sdks/objc/latest/Classes/RLMSyncSession.html), but [Swift's SyncSession](https://www.mongodb.com/docs/realm-sdks/swift/latest/Extensions/SyncSession.html) didn't. Could we add this, or am I looking in the wrong place?

### Jira

- https://jira.mongodb.org/browse/DOCSP-33926

### Staged Changes

- [Manage Sync Sessions](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-33926/sdk/swift/sync/sync-session/)

The rest of the examples in the test file were still using Partition-Based Sync, so I also updated those examples to FS. Those examples are at:
- [Configure & Open a Synced Realm](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/DOCSP-33926/sdk/swift/sync/configure-and-open-a-synced-realm/#download-changes-before-open)

### Reminder Checklist

### Review Guidelines


source/sdk/swift/sync/sync-session.txt
```diff
@@ -160,3 +160,35 @@ Check Upload & Download Progress for a Sync Session
...
+Manually Reconnect All Sync Sessions
+------------------------------------
+
+.. versionadded:: 10.44.0
+
+Realm automatically detects when a device regains connectivity after being 
+offline and attempts to reconnect using an incremental backoff strategy.
+
+In Swift SDK version 10.44.0 and later, you can choose to manually trigger a 
+reconnect attempt with ``SyncSession.reconnect()``
```
cbullinger, line 175: Do we want to link to the API ref here?

cbullinger (approved):
Copy updates LGTM, though I'm curious if `.reconnect()` does apply to all sessions.
And FYI - the last example on the Configure & Open a Synced Realm page still uses PBS in the snippet: `/examples/generated/code/start/Authenticate.snippet.offline-login.swift`

dacharyc:
> And FYI - the last example on the Configure & Open a Synced Realm page still uses PBS in the snippet: `/examples/generated/code/start/Authenticate.snippet.offline-login.swift`

Yeah, that comes from a different test file - the `Authenticate` file - which I wasn't updating here. I was just updating the other examples in the file I was touching. At some point I should do a bigger sweep to update all the PBS examples in Swift.

tgoyne:
reconnect() will skip the reconnect delay for all sessions in the same App regardless of if multiplexing is enabled. I think saying that it'll force a reconnect might be misleading, as it won't cause a reconnection if we got a fatal error. It also doesn't do anything if the connection isn't currently waiting to reconnect, i.e. if it's already connected or is in the process of trying to connect.

On Apple platforms core listens for network change notifications and automatically triggers a reconnection immediately after receiving one. Some SDKs may implement similar logic for other platforms.

A longstanding problem we've had with jazzy is that part of our Swift API is just re-exporting obj-c types, and jazzy doesn't include those things in the generated Swift docs. Since we're no longer working on jazzy, this is unlikely to ever get fixed.

PR #3073, opened 2023-11-07
(DOCSP-33831): C++: Add docs for encryption and metadata encryption

dacharyc:
### Jira

- https://jira.mongodb.org/browse/DOCSP-33831

### Staged Changes

- [Encrypt a Realm](https://preview-mongodbdacharyc.gatsbyjs.io/realm/DOCSP-33831/sdk/cpp/realm-files/encrypt-a-realm/): New page based on similar pages in other SDKs
- [Connect to App Services](https://preview-mongodbdacharyc.gatsbyjs.io/realm/DOCSP-33831/sdk/cpp/app-services/connect-to-app/#encrypt-app-metadata): Add a section about encrypting metadata. Update custom HTTP section for deprecated `realm::App(...)`

### Reminder Checklist

### Review Guidelines

### DIFF
--- a/source/sdk/cpp/app-services/connect-to-app.txt
+++ b/source/sdk/cpp/app-services/connect-to-app.txt
@@ -47,13 +47,14 @@ Access the App Client
 Set Custom HTTP Headers
 -----------------------
 
-.. versionadded:: v0.3.0-preview
+.. versionchanged:: v0.4.0-preview replaces deprecated realm::App(...) with realm::App(const realm::App::configuration&)
 
 If you use App Services or Device Sync with a proxy setup, you may need
 to set custom HTTP headers. The Realm C++ SDK supports setting custom 
-HTTP headers on the ``App``, and on the ``flexible_sync_configuration()``.
+HTTP headers on the :cpp-sdk:`App <structrealm_1_1App_1_1configuration.html>`, 
+and on the ``flexible_sync_configuration()``.
 
-When you initialize the ``App``, pass in a map of string header keys
+When you initialize the App configuration, pass in a map of string header keys
 and values.
 
 .. literalinclude:: /examples/generated/cpp/app.snippet.set-custom-headers-for-app.cpp
@@ -65,3 +66,24 @@ these custom headers.
 If you want to use custom headers with Device Sync, you must 
 additionally :ref:`set the headers on the Flexible Sync configuration 
 <cpp-set-custom-headers-using-synced-realm>`.
+
+.. _cpp-encrypt-app-metadata:
+
+Encrypt App Metadata
+--------------------
+
+.. versionadded:: v0.4.0-preview
+
+When you connect to App Services, Realm creates additional metadata files 
+on a device. For more information about these metadata files, refer to 
+:ref:`cpp-realm-database`.
+
+You can encrypt the metadata that App Services stores on client devices, 
+similar to :ref:`encrypting the realm <cpp-encrypt-a-realm>`. 
+
+To encrypt the metadata, create a 64-byte encryption key. Then, set the 
+``metadata_encryption_key`` on your 
+:cpp-sdk:`realm::App::configuration <structrealm_1_1App_1_1configuration.html>`.
+
+.. literalinclude:: /examples/generated/cpp/app.snippet.encrypt-metadata.cpp
+   :language: cpp
--- a/source/sdk/cpp/app-services/connect-to-app.txt
+++ b/source/sdk/cpp/app-services/connect-to-app.txt
@@ -81,7 +81,10 @@ on a device. For more information about these metadata files, refer to
 You can encrypt the metadata that App Services stores on client devices, 
 similar to :ref:`encrypting the realm <cpp-encrypt-a-realm>`. 
 
-To encrypt the metadata, create a 64-byte encryption key. Then, set the 
+On Apple devices, the metadata is encrypted by default. To disable this, 
+add ``REALM_DISABLE_METADATA_ENCRYPTION`` to your environment variables.
+
+To enable metadata encryption on other platforms, you must set a 
 ``metadata_encryption_key`` on your 
 :cpp-sdk:`realm::App::configuration <structrealm_1_1App_1_1configuration.html>`.
 
--- a/source/sdk/cpp/app-services/connect-to-app.txt
+++ b/source/sdk/cpp/app-services/connect-to-app.txt
@@ -51,7 +51,7 @@ Set Custom HTTP Headers
 
 If you use App Services or Device Sync with a proxy setup, you may need
 to set custom HTTP headers. The Realm C++ SDK supports setting custom 
-HTTP headers on the :cpp-sdk:`App <structrealm_1_1App_1_1configuration.html>`, 
+HTTP headers on the :cpp-sdk:`App <structrealm_1_1App_1_1configuration.html>`
 and on the ``flexible_sync_configuration()``.
 
 When you initialize the App configuration, pass in a map of string header keys
--- a/source/sdk/cpp/app-services/connect-to-app.txt
+++ b/source/sdk/cpp/app-services/connect-to-app.txt
@@ -90,3 +90,4 @@ To enable metadata encryption on other platforms, you must set a
 
 .. literalinclude:: /examples/generated/cpp/app.snippet.encrypt-metadata.cpp
    :language: cpp
+   :emphasize-lines: 20-21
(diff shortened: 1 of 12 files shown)


source/sdk/cpp/app-services/connect-to-app.txt
```diff
@@ -47,13 +47,14 @@ Access the App Client
 Set Custom HTTP Headers
 -----------------------
 
-.. versionadded:: v0.3.0-preview
+.. versionchanged:: v0.4.0-preview replaces deprecated realm::App(...) with realm::App(const realm::App::configuration&)
 
 If you use App Services or Device Sync with a proxy setup, you may need
 to set custom HTTP headers. The Realm C++ SDK supports setting custom 
-HTTP headers on the ``App``, and on the ``flexible_sync_configuration()``.
+HTTP headers on the :cpp-sdk:`App <structrealm_1_1App_1_1configuration.html>`,
```
cbullinger, line 54:
```suggestion
HTTP headers on the :cpp-sdk:`App <structrealm_1_1App_1_1configuration.html>`
```

source/sdk/cpp/app-services/connect-to-app.txt
```diff
@@ -65,3 +66,27 @@ these custom headers.
...
+.. _cpp-encrypt-app-metadata:
+
+Encrypt App Metadata
+--------------------
+
+.. versionadded:: v0.4.0-preview
+
+When you connect to App Services, Realm creates additional metadata files 
+on a device. For more information about these metadata files, refer to 
+:ref:`cpp-realm-database`.
```
cbullinger, line 79: should we link to the direct section since it's a long page (I'm assuming the Realm Files section)?

source/sdk/cpp/app-services/connect-to-app.txt
```diff
@@ -65,3 +66,27 @@ these custom headers.
...
+
+You can encrypt the metadata that App Services stores on client devices, 
+similar to :ref:`encrypting the realm <cpp-encrypt-a-realm>`. 
+
+On Apple devices, the metadata is encrypted by default. To disable this,
```
cbullinger, line 84: potentially stupid Q but if it's encrypted by default, what encryption key does it use?

source/sdk/cpp/app-services/connect-to-app.txt
```diff
@@ -65,3 +66,27 @@ these custom headers.
...
+
+When you connect to App Services, Realm creates additional metadata files 
+on a device. For more information about these metadata files, refer to 
+:ref:`cpp-realm-database`.
+
+You can encrypt the metadata that App Services stores on client devices, 
+similar to :ref:`encrypting the realm <cpp-encrypt-a-realm>`. 
+
+On Apple devices, the metadata is encrypted by default. To disable this, 
+add ``REALM_DISABLE_METADATA_ENCRYPTION`` to your environment variables.
```
cbullinger, line 85: this disables it even if you set a `metadata_encryption_key`? (is `metadata_encryption_key` completely ignored for Apple devices?)

source/sdk/cpp/realm-files/encrypt-a-realm.txt
```diff
@@ -0,0 +1,59 @@
...
+first 256 bits of the given 512-bit encryption key. Realm
+uses the other 256 bits of the 512-bit encryption key to validate
+integrity using a :wikipedia:`hash-based message authentication code
+(HMAC) <HMAC>`.
+
+.. include:: /includes/encrypt-use-strong-cryptographic-hash.rst
+
+.. versionadded:: v0.4.0-preview
+
+Encrypt a realm by calling the ``set_encryption_key()`` function on
```
cbullinger, line 28: Apple devices don't encrypt this by default too?

source/sdk/cpp/realm-files/encrypt-a-realm.txt
```diff
@@ -0,0 +1,59 @@
...
+apps cannot read the key.
+
+Performance Impact
+------------------
+
+Reads and writes on encrypted realms can be up to 10% slower than unencrypted realms.
+
+Encryption and Atlas Device Sync
+--------------------------------
+
```
cbullinger, line 51: Do we want to mention/link out to encrypting app metadata here, as well?

source/sdk/cpp/realm-files/encrypt-a-realm.txt
```diff
@@ -0,0 +1,59 @@
...
+Encrypt a Realm - C++ SDK Preview
+=================================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can encrypt the realm file on disk with AES-256 +
```
cbullinger, line 13: maybe dumb Q: can you encrypt an existing unencrypted realm (i.e. add the key to an existing config)? or does it have to be encrypted from the start? i know you can create a new realm, then copy an existing unencrypted realm to the new encrypted one...

source/sdk/cpp/realm-files/encrypt-a-realm.txt
```diff
@@ -0,0 +1,59 @@
...
+------------------
+
+You **must** pass the same encryption key every time you open the encrypted realm. 
+If you don't provide a key or specify the wrong key for an encrypted
+realm, the Realm SDK throws an error.
+
+Apps should store the encryption key securely on the device so that other 
+apps cannot read the key.
+
+Performance Impact
```
cbullinger, line 44: probably not, but are there any impacts or considerations to encrypting the app metadata

cbullinger (approved):
random Qs from me, but overall LGTM

source/sdk/cpp/app-services/connect-to-app.txt
dacharyc: We have logic in Realm Core that creates a metadata key and stores it in the Keychain. It abstracts the need for the user to do this themselves, I suppose.

source/sdk/cpp/app-services/connect-to-app.txt
dacharyc: Yes, if you set `REALM_DISABLE_METADATA_ENCRYPTION`, it ignores setting a `metadata_encryption_key`. If you set an encryption key and you do not disable encryption using this variable, it uses the encryption key that you set instead of randomly generating one for you. So you _can_ create a custom one if you want, but it's not necessary. We'll handle it for you.

I kind of think this level of detail is unnecessary. But since you're asking, maybe I'm wrong. Do you think we ought to explicitly state this?

source/sdk/cpp/realm-files/encrypt-a-realm.txt
dacharyc: I believe the answer to this is yes - it _looks_ to me like you can add the key to an existing config.

source/sdk/cpp/realm-files/encrypt-a-realm.txt
dacharyc: Correct. We have code in Core that encrypts the metadata by default on Apple devices. But if you want realms to be encrypted, you have to do that manually.

source/sdk/cpp/realm-files/encrypt-a-realm.txt
dacharyc: Good question! I'm going to have to ask an expert on this one.

source/sdk/cpp/app-services/connect-to-app.txt
cbullinger: Nah, it's not needed. Was mostly curious since I don't know much about Apple's encryption

source/sdk/cpp/realm-files/encrypt-a-realm.txt
cbullinger: I thought the config was cached: "Attempting to "close" an App and then re-open it with a changed configuration within the same process has no effect. The client continues to use the cached configuration."

dacharyc: I think that just pertains to the App, not the realm itself.

dacharyc: Actually, you are correct here! Well spotted. I'll add something to the page about this.

source/sdk/cpp/realm-files/encrypt-a-realm.txt
dacharyc: Confirmed that the volume of metadata access is so small, encrypting it has no measurable impact.

PR #3090, opened 2023-11-27
Add tagging checklist item

krollins-mdb:
## Pull Request Info

This PR adds a reminder about content tags to the PR checklist.

### Review Guidelines

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,7 @@
 ## Pull Request Info
 
 Jira ticket: https://jira.mongodb.org/browse/DOCSP-NNNNN
-Staged changes:
-
-- [PAGE_NAME](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/BRANCH_NAME/)
+Staged changes: [PAGE_NAME](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/BRANCH_NAME/)
 
 ### Description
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,22 +1,25 @@
...
 - List any required dependencies for this change
--->
 
 ### Reminder Checklist
 
 Before merging your PR, make sure to check a few things.
 
+- [ ] Did you tag pages appropriately?
+  - genre
+  - programming_language
```
dacharyc, line 18: Do we need `programming_language` in the tags, since that is handled for all of the SDKs with the facets at the root of the SDK directory? I think the only page we might need to tag manually is the RQL page, which lives outside of an SDK. We could just make a ticket to tag it, I think, and leave this out of the tag list for this repo?

.github/pull_request_template.md
```diff
@@ -1,22 +1,25 @@
 ## Pull Request Info
 
 Jira ticket: https://jira.mongodb.org/browse/DOCSP-NNNNN
-Staged changes:
+Staged changes: [PAGE_NAME](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/BRANCH_NAME/)
 
-- [PAGE_NAME](https://docs-mongodbcom-staging.corp.mongodb.com/realm/docsworker-xlarge/BRANCH_NAME/)
+### Description
```
dacharyc, line 6: I'm undecided how I feel about adding a separate Description section. I tend to just put a colon after my staged page link and provide context for the reviewer about what has changed there. I think that's easier when multiple pages are affected, because it keeps the description of changes tied to the page where the changes occur. I have just been deleting the below comment when making my PRs, and continuing to provide info in line with the staging links. I wonder if we might want to discuss this as a team?

.github/pull_request_template.md
krollins-mdb: Yeah, it's a good point. After thinking it over, I don't think it's necessary here.

.github/pull_request_template.md
krollins-mdb: I think I was trying to give people a place to talk about higher-level things than page-specific stuff. The linked Jira ticket explains what, staged link descriptions explain where/how, but we often don't have info about why the PR author did things the way they did.

I don't think it's necessary for every PR, but can help communicate context to reviewers.

I'll remove the dedicated section for now. I'd like to get these PRs merged. 😄

PR #3114, opened 2023-12-15
DOCSP-35056 Dart upgrade dependencies

lindseymoore:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-35056

The following tests are not working:

- test/sync_multiple_processes_test.dart
- test/authenticate_users_test.dart
- test/manage_sync_session_test.dart
- test/open_flexible_sync_realm_test.dart


examples/dart/test_sdk_example/pubspec.lock
lindseymoore, line 1: Should I add back this deleted file?

dacharyc: Generally speaking, I would expect changes to this file if we do the load dependency command _inside_ this directory. But I know we had some challenges while we were on the call dealing with multiple pubspec.yaml files. It's probably not super important for this PR. We can always regenerate it in the future when we're working in this directory.

dacharyc (changes requested):
It looks like there are a couple of things added here that we should remove - a few generated files, but more importantly, the `flutter` git submodule. After those changes, this should be good to go!

dacharyc (approved):
LGTM! Thanks for taking care of this, Lindsey!

PR #3160, opened 2024-01-24
(DOCSP-35852): Swift: Base URL can change in App config

dacharyc:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-35852

- [Connect to an App Services Backend](https://preview-mongodbdacharyc.gatsbyjs.io/realm/DOCSP-35852/sdk/swift/app-services/connect-to-app-services-backend/#configuration): Note that starting in 10.46.0, you can change a `baseURL` and the App client no longer uses the cached value.

### Reminder Checklist

- [x] Did you tag pages appropriately?
  - genre
  - meta.keywords
  - meta.description
- [x] Describe your PR's changes in the Release Notes section
- [x] Create a Jira ticket for related docs-app-services work, if any

### Release Notes

- **Swift SDK**
  - Application Services/Connect to an App Services App: Note that starting in 10.46.0, you can change a `baseURL` and the App client no longer uses the cached value.

### Review Guidelines

### DIFF
--- a/source/sdk/swift/app-services/connect-to-app-services-backend.txt
+++ b/source/sdk/swift/app-services/connect-to-app-services-backend.txt
@@ -41,7 +41,18 @@ You can pass a configuration object to ``App``:
 .. literalinclude:: /examples/generated/code/start/RealmApp.snippet.realm-app-config.swift
    :language: swift
 
-.. include:: /includes/multiple-app-client-details-and-app-config-cache.rst
+You can create multiple App client instances to connect to multiple
+Apps. All App client instances that share the same App ID use the same 
+underlying connection.
+
+.. important:: Changing an App Config After Initializing the App
+
+   Starting with Swift SDK version 10.46.0, you can change a ``baseURL``
+   in the App config and the App client uses the new ``baseURL``. In Swift
+   SDK versions 10.45.3 and earlier, when you initialize the App client, 
+   the configuration is cached internally. Attempting to "close" an App and 
+   then re-open it with a changed configuration within the same process has 
+   no effect. The client continues to use the cached configuration.
 
 Sync Connection Sharing
 ~~~~~~~~~~~~~~~~~~~~~~~


source/sdk/swift/app-services/connect-to-app-services-backend.txt
```diff
@@ -41,7 +41,18 @@ You can pass a configuration object to ``App``:
...
 
-.. include:: /includes/multiple-app-client-details-and-app-config-cache.rst
+You can create multiple App client instances to connect to multiple
+Apps. All App client instances that share the same App ID use the same 
+underlying connection.
+
+.. important:: Changing an App Config After Initializing the App
+
+   Starting with Swift SDK version 10.46.0, you can change a ``baseURL``
+   in the App config and the App client uses the new ``baseURL``. In Swift
```
lindseymoore, line 51:
```suggestion
   in the App config, so the App client uses the new ``baseURL``. In Swift
```
To emphasize the first is action and second is result. Was considering future tense, but I guess we want to avoid based on style guide.

source/sdk/swift/app-services/connect-to-app-services-backend.txt
```diff
@@ -41,7 +41,18 @@ You can pass a configuration object to ``App``:
...
+   SDK versions 10.45.3 and earlier, when you initialize the App client, 
+   the configuration is cached internally. Attempting to "close" an App and
```
lindseymoore, line 53:
```suggestion
   the configuration is cached internally. Attempting to close an App and 
```
why the quotations?

lindseymoore (approved):
Small comments, LGTM!

source/sdk/swift/app-services/connect-to-app-services-backend.txt
dacharyc: Because this is a change in behavior, and the "result" does not _necessarily_ follow from the "action," I think the "and" is more appropriate. In prior versions, you could change the ``baseURL`` in the app config and the app client would not use the new baseURL. The logically-following "so" did not apply.

source/sdk/swift/app-services/connect-to-app-services-backend.txt
dacharyc: This is the language we've chosen in the existing `.. include` (which I couldn't modify directly because it still applies to the other SDKs where this functionality hasn't been released yet). It's not literally `close` because there are a lot of ways a developer might attempt to do this depending on the SDK/language/platform and their app. They might attempt to start a new process, they might attempt to deinit the app client, or they might try other things to force the app to initialize with new values. So close is in quotation marks here because it's a catch-all proxy for all the ways a developer might attempt to accomplish this - we're not literally suggesting that a developer close the app.

PR #3174, opened 2024-02-07
DOCSP-35846: Update BaseUrl in Flutter SDK

cbullinger:
## Pull Request Info

Jira ticket: https://jira.mongodb.org/browse/DOCSP-35846

- [Connect to App Services - Flutter SDK](https://preview-mongodbcbullinger.gatsbyjs.io/realm/docsp-35846-dart-baseurl/sdk/flutter/app-services/connect-to-app/): Replaced the sharedInclude with new note indicating that, starting in v1.8.0, the ``baseUrl`` is no longer cached in the ``AppConfig``

### Reminder Checklist

- [x] Did you tag pages appropriately?
  - genre
  - meta.keywords
  - meta.description
- [x] Describe your PR's changes in the Release Notes section
- [x] Create a Jira ticket for related docs-app-services work, if any

### Release Notes

- **Flutter** SDK
  - Atlas App Services/Connect to App Services: Note that, starting in v1.8.0, you can update the baseUrl in the AppConfiguration because the App client no longer caches it.

### Review Guidelines

### DIFF
--- a/source/sdk/flutter/app-services/connect-to-app.txt
+++ b/source/sdk/flutter/app-services/connect-to-app.txt
@@ -56,7 +56,21 @@ information.
 .. literalinclude:: /examples/generated/flutter/app_services_test.snippet.access-app-client.dart
    :language: dart
 
-.. include:: /includes/multiple-app-client-details-and-app-config-cache.rst
+You can create multiple App client instances to connect to multiple
+Apps. All App client instances that share the same App ID use the same 
+underlying connection.
+
+.. important:: Changing an App Config After Initializing the App
+
+   When you initialize the App client, the configuration is cached internally. 
+   Attempting to "close" and then re-open an App with a changed configuration
+   within the same process has no effect. The client continues to use the 
+   cached configuration. 
+
+   In Flutter SDK version 1.9.0 and later, the :flutter-sdk:`baseUrl <realm/AppConfiguration/baseUrl.html>`
+   configuration property is *not* cached in the App configuration. This 
+   means that you can change the ``baseUrl`` at runtime, and the App client will use the updated configuration. In earlier SDK versions, changes to the ``baseUrl`` in a cached App configuration have no 
+   effect.
 
 .. _flutter-app-client-configuration:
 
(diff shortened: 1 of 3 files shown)


source/sdk/flutter/app-services/connect-to-app.txt
```diff
@@ -56,7 +56,23 @@ information.
 .. literalinclude:: /examples/generated/flutter/app_services_test.snippet.access-app-client.dart
    :language: dart
 
-.. include:: /includes/multiple-app-client-details-and-app-config-cache.rst
+You can create multiple App client instances to connect to multiple
+Apps. All App client instances that share the same App ID use the same 
+underlying connection.
+
+.. important:: Changing an App Config After Initializing the App
```
dacharyc, line 63: This is a very subjective nit, but I might lead with the second paragraph - or a variation on it. We generally want to encourage people to use the most recent versions, and specifically for this ability to change the base URL, we want to present the more recent option as the "default". Similar to the Swift page: https://www.mongodb.com/docs/realm/sdk/swift/app-services/connect-to-app-services-backend/

source/sdk/flutter/app-services/connect-to-app.txt
```diff
@@ -66,14 +82,17 @@ Advanced Configuration
 .. deprecated:: 1.6.0
    ``App.localAppName`` and ``App.localAppVersion`` are no longer used.
 
+.. versionchanged:: 1.8.0
```
dacharyc, line 85: Since we don't show baseURL in this example, I think we could probably omit this versionchanged directive here. What do you think?

dacharyc (approved):
LGTM - just a couple of small subjective nits/Qs.

source/sdk/flutter/app-services/connect-to-app.txt
cbullinger: I kind of don't like presenting it all in a note, but i don't love how it's presented on the Swift page. Unless I'm misreading the scope doc, the app config caching outside the baseUrl remains the same. But the Swift note reads as though the entire app config is only cached in previous versions.

source/sdk/flutter/app-services/connect-to-app.txt
cbullinger: it seemed easy to miss in the note, so I added here as well. overkill?

PR #3273, opened 2024-05-23
(DOCSP-39541) Consolidate logging page

krollins-mdb:

Jira ticket: https://jira.mongodb.org/browse/DOCSP-39541

*Staged Page*

- [Logging](https://preview-mongodbkrollinsmdb.gatsbyjs.io/realm/DOCSP-39541/sdk/test-and-debug/log/)

*Page Source*

Add links to every SDK's pages where you got the SDK-specific information:

- [Logging - C++ SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/cpp/logging/)
- [Logging - Flutter SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/flutter/logging/)
  - Added new Logger examples because the current examples no longer work
- [Log Realm Events - Java SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/java/test-and-debug/log-realm-events/)
- [Logging - Kotlin SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/kotlin/logging/)
- [Logging - .NET SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/dotnet/logging/)
- [Logging - Node.js SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/node/logging/)
- [Logging - Swift SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/swift/logging/)
- [Log Realm Events - Java SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/java/test-and-debug/log-realm-events/)
- [Set the Sync Client Log Level - C++ SDK](https://www.mongodb.com/docs/atlas/device-sdks/sdk/cpp/sync/log-level/)
  - All of the other SDKs' Sync Client Log Level page content will not be migrated. It's a deprecated feature that makes talking about logging awkward at best.

### PR Author Checklist

Before requesting a review for your PR, please check these items:

- [x] Open the PR against the `feature-consolidated-sdk-docs` branch instead of `master`
- [x] Tag the consolidated page for:
  - genre
  - meta.keywords
  - meta.description

#### Naming
- [x] Update Realm naming and the language around persistence layer/local/device per [this document](https://docs.google.com/document/d/126OczVxBWAwZ4P5ZsSM29WI3REvONEr1ald-mAwPtyQ/edit?usp=sharing)
- [x] Include `.rst` files comply with [the naming guidelines](https://docs.google.com/document/d/1h8cr66zoEVeXytVfvDxlCSsUS5IZwvUQvfSCEXNMpek/edit#heading=h.ulh8b5f2hu9)

#### Links and Refs
- [x] Create new consolidated SDK ref targets starting with "_sdks-" for relevant sections
- [x] Remove or update any SDK-specific refs to use the new consolidated SDK ref targets
- [x] [Update any Kotlin API links](https://jira.mongodb.org/browse/DOCSP-32519) to use the new Kotlin SDK roles

#### Content
- [x] Shared code boxes have snippets or placeholders for all 9 languages
- [x] API description sections have API details or a generic placeholder for all 9 languages
- [x] Check related pages for relevant content to include
- [x] Create a ticket for missing examples in each relevant SDK: Consolidation Gaps epic

### Reviewer Checklist

As a reviewer, please check these items:

- [ ] Shared code example boxes contain language-specific snippets or placeholders for every language
- [ ] API reference details contain working API reference links or generic content
- [ ] Realm naming/language has been updated
- [ ] All relevant content from individual SDK pages is present on the consolidated page


examples/dart/test/logger.dart
```diff
@@ -1,167 +1,85 @@
...
-      Realm.logger.level = RealmLogLevel.trace;
-      executeComplexCodeToDebug();
-      // :snippet-end:
+    realm.write(() => realm.add(Person(ObjectId(), 'Anakin', 'Skywalker')));
 
-      // Because we changed logging from .info to .trace, this should log a message.
-      RealmInternal.logMessageForTesting(RealmLogLevel.trace,
-          "This message should be logged because the log level was changed");
+    // Hack to wait for stream to process
+    await delay(200);
```
krollins-mdb, line 37: This is a bit of a hack to wait for the stream to and listeners to do their thing.

I would love to make this code a more realistic use case and incorporate it into a Flutter app, but there's no time for that - especially as part of the consolidation project.

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,240 @@
 Logging
 =======
 
+.. meta::
+   :description: Use logging to debug your Atlas Device SDK app.
+   :keywords: Realm, C++ SDK, Flutter SDK, Kotlin SDK, Java SDK, Node.js SDK, Swift SDK, code example
```
dacharyc, line 9: Sorry, we're missing .NET here in the copyable template. I've added it on one of my PRs that is pending review, but it's still pending review.

```suggestion
   :keywords: Realm, C++ SDK, Flutter SDK, Kotlin SDK, Java SDK, .NET SDK, Node.js SDK, Swift SDK, code example
```

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,240 @@
...
+
+.. facet::
+  :name: genre
+  :values: reference
+
+.. facet::
+   :name: programming_language
+   :values: cpp, csharp, dart, java, javascript/typescript, kotlin, objective-c, swift
```
dacharyc, line 17: I don't see Objective-C examples here, so I think we should omit it.

```suggestion
   :values: cpp, csharp, dart, java, javascript/typescript, kotlin, swift
```

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,240 @@
...
+.. _sdks-turn-off-logging:
+
+Turn Off Logging
+----------------
+
+You can turn off logging by setting the log level:
+
+.. include:: /includes/sdk-examples/logger/logging-turn-off-example.rst
+
+Java SDK: Deprecated Sync Logger
```
dacharyc, line 204: I think, instead of having two separate "Deprecated Sync Logger" sections for different SDKs, I'd probably rename this generically to:

"Deprecated Sync Logger"

And then use a tab group here to display the C++ and Java info.

For the tabs that aren't C++ or Java, I'd probably add a quick note saying something like:

> The <Foo> SDK deprecated the Sync logger in favor of the Realm logger in version <version number>. For versions <version number immediately before this was deprecated> and earlier, refer to the <link to SyncManager API reference> for details about the deprecated APIs.

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,240 @@
...
+--------------------------------
+
+The Java SDK's only logger is a deprecated sync logger that behaves differently
```
dacharyc, line 207: I'm not sure this is technically true? The RealmLogger in Java doesn't say anything about being restricted to a Sync logger - it's not in the Sync package, and the API reference docs say:

> Global logger used by all Realm components. Custom loggers can be added by registering classes implementing [RealmLogger](https://www.mongodb.com/docs/realm-sdks/java/latest/io/realm/log/RealmLogger.html).

So - I think this can be used similar to the other loggers, and it just doesn't have any examples other than setting the log level.

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,240 @@
...
+--------------------------------
+
+The C++ SDK uses a sync logger that behaves differently than the rest of the
+examples on this page. You can set or change your sync client's log level to
+develop or debug your application. You might want to change the log level to log different 
+amounts of data depending on the app's environment.
+
+.. _cpp-set-the-client-log-level:
+
+Set the Sync Client Log Level
```
dacharyc, line 233: We probably don't need a separate subheader here. It's taking more real estate/attention than I think it needs in the OtP. And if we make this a generic "Deprecated Sync Logger" header with tab groups for SDK-specific info, then I think we wouldn't want this header to show on the page anyway since it would only be in the C++ tab.

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,240 @@
...
+.. include:: /includes/sdk-examples/logger/logging-set-custom-logger-example.rst
+
+After setting a custom logger, you may need to initialize the logger.
+
+.. tabs-drivers::
+
+   .. tab::
+      :tabid: cpp
+
+      .. include:: /includes/api-details/cpp/logger/logging-customize-logger-description-2.rst
```
dacharyc, line 154: So - you don't actually have to include empty files for each tab. You can omit content within a tab if there is no SDK-specific content we need to call out, as long as you still include the tab itself. So I'd probably remove these empty includes.

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,240 @@
...
+      .. include:: /includes/api-details/swift/logger/logging-customize-logger-description.rst
+
+   .. tab::
+      :tabid: typescript
+
+      .. include:: /includes/api-details/javascript/logger/logging-customize-logger-description.rst
+
+.. include:: /includes/sdk-examples/logger/logging-set-custom-logger-example.rst
+
+After setting a custom logger, you may need to initialize the logger.
```
dacharyc, line 147: This reads a bit weird to me on languages where there is no example of initializing the logger. For example, on C#, there's a "missing example" placeholder below this copy. I wasn't sure if that's because the example is _actually_ missing, or if you don't have to do a separate step of initializing the logger, since the copy here says "you may need to."

I think I would probably move this content for SDKs that require it up into the tab group above, and just omit it from here.

Then, I would include the "last" code example as the code example in the `/includes/sdk-examples/logger/logging-set-custom-logger-example.rst` file.

For example, for C++ - I see there's an example of initializing the logger below. I would make _that_ the example in the `/includes/sdk-examples/logger/logging-set-custom-logger-example.rst` file, and then move the C++ example of actually defining the custom logger into the `/includes/api-details/cpp/logger/logging-customize-logger-description.rst` file.

Let me know if this makes sense or if you have Qs!

temp/swift/app-services/connect-to-app-services-backend.txt
```diff
@@ -1,4 +1,3 @@
-.. _ios-init-appclient:
```
dacharyc, line 1: We should probably omit this file from the PR so we're not getting into merge conflicts in the `temp` directory.

source/sdk/atlas/connect.txt
```diff
@@ -1,3 +1,4 @@
+.. _ios-init-appclient:
```
dacharyc, line 1: One of the units of work for this project is to remove SDK-specific ref targets, so we shouldn't add this here. We should just use the generic `sdks-connect-to-atlas` ref target if we need to link to this page.

source/includes/api-details/swift/logger/logging-set-change-level-description.rst
```diff
@@ -0,0 +1,12 @@
+For the Swift SDK, set the log level for the default logger with
+``Logger.shared.level``. The :objc-sdk:`RLMLogLevel <Enums/RLMLogLevel.html>`
+enum represents the different levels of logging you can configure.
+
+.. tip:: Set the Logger Before Initializing an App Client
+
+  When you initialize an :ref:`App client <ios-init-appclient>`, the Swift SDK
```
dacharyc, line 7: One of the units of work for this project is to replace the language and SDK-specific ref targets with generic ones. So we can just use `sdks-connect-to-atlas` here, instead, since that's the one on the page.
```suggestion
  When you initialize an :ref:`App client <sdks-connect-to-atlas>`, the Swift SDK
```

source/includes/api-details/kotlin/logger/logging-set-change-level-description.rst
```diff
@@ -0,0 +1,16 @@
...
+singleton.
+
+By default, all logs go to a default system logger that varies by system:
+
+- Android logs to Logcat.
+- JVM logs to stdout.
+- MacOS logs to NSLog.
+- iOS logs to NSLog.
+
+.. include:: /includes/tip-sync-log-levels.rst
```
dacharyc, line 16: One of the benefits of consolidating the docs pages is that we generally shouldn't need generic includes, because we don't need to split content across a bunch of pages. Instead of nesting this includes here, I'd try taking the content out of the include and putting it directly on the main page wherever it needs to be.

dacharyc (changes requested):
This is a great start! There are some things that come across kind of weird on on the consolidated page that I think we can improve, and there are some updates we should make to the metadata.

dacharyc (commented):
Just to avoid blocking this, changing my review status from "request changes" to "comment."

dacharyc (commented):
Trying again to change the review state to "comment" just so I don't block this while I'm out next week.

source/includes/api-details/kotlin/logger/logging-set-change-level-description.rst
krollins-mdb: Thanks for catching that. I do have this on the main page already. 😄

source/sdk/atlas/connect.txt
krollins-mdb: Hmm. Yeah, I can't remember why I added it here. It's not used anywhere in the consolidated content. 🤷 

Thanks for catching this!

temp/swift/app-services/connect-to-app-services-backend.txt
krollins-mdb: Ah, right. I think I had a build error about ambiguous refs, so deleted this one. Before I realized the right way is to use the consolidated refs.

source/sdk/test-and-debug/log.txt
krollins-mdb: Got it! Thanks. 😄

source/sdk/test-and-debug/log.txt
krollins-mdb: I renamed the section to "Other Loggers" and have just two tabs: C++ and Java. I did this because, as you say, the Java logger isn't really the same thing as the deprecated Sync Logger. But the C++ SDK currently needs that Sync Logger. So, they're both different from the rest of the examples on the page.

source/sdk/test-and-debug/log.txt
krollins-mdb: I renamed the section to "Other Loggers" and have just two tabs: C++ and Java. I did this because, as you say, the Java logger isn't really the same thing as the deprecated Sync Logger. But the C++ SDK currently needs that Sync Logger. So, they're both different from the rest of the examples on the page.

source/sdk/test-and-debug/log.txt
krollins-mdb: Hmm, yeah. I went back and forth trying to figure out how much structure to impose. Including initialization as an include in the SDK description hides the expected structure a bit (set -> initialize). But doing the way I currently am definitely makes it awkward for a couple of the SDKs.

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,216 @@
...
+.. _sdks-turn-off-logging:
+
+Turn Off Logging
+----------------
+
+You can turn off logging by setting the log level:
+
+.. include:: /includes/sdk-examples/logger/logging-turn-off-example.rst
+
+Other Loggers
```
dacharyc, line 158: lol, I didn't think about this, but for languages _other_ than C++ or C#, this section header now shows with no content. (i.e. Kotlin has an "Other Loggers" heading with no content.) I guess in this case, we should add something to the tabs. Maybe this would be a place to mention the deprecated Sync logger? i.e.

> In SDK versions {FOO} and earlier, you could configure a Sync logger using `methodName`. That method has been deprecated in favor of the global logger.

One for each SDK with the version number in which the Sync logger was deprecated, and the Sync logger method name. I'd say we should avoid linking to the deprecated Sync logger API reference, because at some point those links will break. But it may be useful for folks using the older methods, if nothing else than to have the method name show up if they do a search in the docs so they can see it has been deprecated.

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,216 @@
...
+
+Set or Change the Log Level
+---------------------------
+
+You can set the level of detail reported by Atlas Device SDK.
+
+.. tabs-drivers::
+
+   .. tab::
+      :tabid: cpp
```
dacharyc, line 48: The new Realm-specific C++ ID has been added, so we should update all of these `cpp` tab references to use our ID. (This one will display "C++" in the language drop down _without_ the "11" that the Driver version uses.)
```suggestion
      :tabid: cpp-sdk
```

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,216 @@
...
+--------------------
+
+You can create a custom logger. You might want to customize logging to add
+specific tags or set specific log levels during development, testing, or
+debugging.
+
+.. tabs-drivers::
+
+   .. tab::
+      :tabid: cpp
```
dacharyc, line 106:
```suggestion
      :tabid: cpp-sdk
```

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,216 @@
...
+
+.. include:: /includes/sdk-examples/logger/logging-turn-off-example.rst
+
+Other Loggers
+-------------
+
+.. tabs-drivers::
+
+   .. tab::
+      :tabid: cpp
```
dacharyc, line 164:
```suggestion
      :tabid: cpp-sdk
```

source/includes/sdk-examples/logger/logging-set-change-level-example-2.rst
```diff
@@ -0,0 +1,44 @@
+.. tabs-drivers::
+
+   tabs:
+     - id: cpp
```
dacharyc, line 4:
```suggestion
     - id: cpp-sdk
```

source/includes/sdk-examples/logger/logging-set-change-level-example.rst
```diff
@@ -0,0 +1,44 @@
+.. tabs-drivers::
+
+   tabs:
+     - id: cpp
```
dacharyc, line 4:
```suggestion
     - id: cpp-sdk
```

source/includes/sdk-examples/logger/logging-set-custom-logger-example.rst
```diff
@@ -0,0 +1,47 @@
+.. tabs-drivers::
+
+   tabs:
+     - id: cpp
```
dacharyc, line 4:
```suggestion
     - id: cpp-sdk
```

source/includes/sdk-examples/logger/logging-turn-off-example.rst
```diff
@@ -0,0 +1,47 @@
+.. tabs-drivers::
+
+   tabs:
+     - id: cpp
```
dacharyc, line 4:
```suggestion
     - id: cpp-sdk
```

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,216 @@
...
+   .. tab::
+      :tabid: dart
+
+   .. tab::
+      :tabid: kotlin
+
+   .. tab::
+      :tabid: csharp
+
+      The Java SDK uses an older Logger that may not behave like the other SDKs.
```
dacharyc, line 194: This content is in a `csharp` tab (maybe it was copied from above?), but it pertains to the Java SDK. We'll need to update this ID to `java` for this info, which means we'll have to add it to _all_ of the tab sets on the page _and_ the ones in the `includes/sdk-examples/logger` or we'll get Snooty errors about a missing tab. But since we don't have Java content that is relevant for those other sections, you'll have to use your judgement about what to put there. Maybe the missing API snippet for the code blocks, and I don't know if it's relevant to add text copy for each Java tab saying that the Java SDK uses a separate logger described below, or if we can just let the missing API snippet serve in those sections.

dacharyc (changes requested):
Looking good! Just a couple of things with tab IDs, and I think we need to add some generic placeholder content to the "Other Loggers" section for languages _other_ than C++ or Java or else it looks like the page is missing text for that section.

source/sdk/test-and-debug/log.txt
krollins-mdb: It seems like a lot of effort for two minor, relatively unsupported, things. I think I'm going to go back to dedicated sections, but have them be under "Other Loggers".

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,195 @@
...
 
-Placeholder page for information about logging.
+.. tabs-selector:: drivers
+
+You can set or change your app's log level when developing or debugging 
+your app. You might want to change the log level to log different 
+amounts of data depending on your development needs. You can specify 
+different log levels or custom loggers.
+
+.. warning:: Java SDK Logging Support
```
dacharyc, line 32: I don't love this bright red warning tab at the beginning of the page for an SDK we don't particularly want people to use. I think we should remove this warning block. It's not relevant if I'm here for any of the languages that we _do_ support. There's a ToC entry in the OTP about Java SDK, so if I'm confused about why Java isn't in the language selector, I can just jump right to that section.

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,195 @@
...
+and pass a :cpp-sdk:`realm::logger::level <structrealm_1_1logger.html>`
+to the ``set_log_level()`` member function:
+
+.. literalinclude:: /examples/generated/cpp/quick-start.snippet.set-sync-log-level.cpp
+   :language: cpp
+
+Java SDK
+~~~~~~~~
+
+The Java SDK uses an older Logger that may not behave like the other SDKs.
```
dacharyc, line 185: Just a general Q - should we be more specific about how this logger differs from the other SDKs?

source/sdk/test-and-debug/log.txt
```diff
@@ -4,10 +4,195 @@
...
+The SDK logs events to the Android system log automatically. You can
+view these events using :android:`Logcat </studio/debug/am-logcat>`.
+
+To access newer logging features, use the Kotlin SDK.
```
dacharyc, line 189: And again, if we want to drive people to the Kotlin SDK, should we be more specific about what new features they'll get access to if they use the Kotlin SDK?

dacharyc (approved):
Nice work! I know this has been a big rethink about how the docs work versus the old way. I've got just a few more minor comments, but those are more subjective and not blockers. Thanks for all your effort on this!

source/sdk/test-and-debug/log.txt
krollins-mdb: Makes sense to me! Should be a better experience without that warning.

source/sdk/test-and-debug/log.txt
krollins-mdb: Maybe. But I don't think it's worth spending more time on. I think I added this sentence to make it clear that the Java SDK wasn't using the same logger type as the other SDKs. But that's accomplished by putting it in the new "Other Loggers" section. I've removed this sentence as it doesn't add anything now and only invites questions. 😄

PR #3276, opened 2024-05-24
DOCSP-39647 Add docs release notes for w/e May 24

lindseymoore:

Jira ticket: https://jira.mongodb.org/browse/DOCSP-39647

### DIFF
--- a/docs-release-notes.md
+++ b/docs-release-notes.md
@@ -1,3 +1,17 @@
+# May 24, 2024
+
+## Java SDk
+
+- Throughout the SDK: Add a note that Realm Java cannot use ``RealmAny``.
+
+## .NET SDK
+
+- Application Services/Connect to an App Services App: Add a section on updating the ``baseURL`` during runtime.
+
+## Swift SDK
+
+- Internal: Fix broken ``baseURL`` and SwiftUI property wrapper behavior.
+
 # May 17, 2024
 
 ## C++ SDK
--- a/docs-release-notes.md
+++ b/docs-release-notes.md
@@ -1,16 +1,12 @@
 # May 24, 2024
 
-## Java SDk
+## Java SDK
 
-- Throughout the SDK: Add a note that Realm Java cannot use ``RealmAny``.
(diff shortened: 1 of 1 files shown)


docs-release-notes.md
```diff
@@ -1,3 +1,17 @@
...
+
+- Throughout the SDK: Add a note that Realm Java cannot use ``RealmAny``.
+
+## .NET SDK
+
+- Application Services/Connect to an App Services App: Add a section on updating the ``baseURL`` during runtime.
+
+## Swift SDK
+
+- Internal: Fix broken ``baseURL`` and SwiftUI property wrapper behavior.
```
dacharyc, line 13: I think we should probably remove this section - I left out release notes for this PR on purpose. I don't think it adds any value for release notes consumers to include this. I guess I should have added a release notes section and noted to omit it.

docs-release-notes.md
```diff
@@ -1,3 +1,17 @@
+# May 24, 2024
+
+## Java SDk
+
+- Throughout the SDK: Add a note that Realm Java cannot use ``RealmAny``.
```
dacharyc, line 5: This isn't really accurate. The Java SDK _can_ use `RealmAny` - just not in new App Services Apps created after a certain date. (I'm realizing now we should have added the date to this note.) We also wouldn't call it "Realm Java" in the release notes.

Also, FYI, the double ticks are reStructuredText syntax for monospace - Markdown uses single backticks for monospace.

```suggestion
## Java SDK

- Throughout the SDK: Note that new App Services Apps cannot use `RealmAny` with the Java SDK due to an incompatible protocol version.
```

dacharyc (changes requested):
I think we should remove the Swift SDK section and provide some more details around the Java SDK changes.

docs-release-notes.md
lindseymoore: I included because I think it serves as a good indicator of work done! I think release notes help us show both the quality of work (as far as documenting specific features) and volume of work that gets done (internal fixes nobody but us knows about). But I will leave out if you'd like!

PR #3329, opened 2024-07-19
Snyk dependency updates 07 19 24

dacharyc:
Resolve Snyk dependency updates:
- Closes #3324 
- Closes #3325 
- Closes #3327


dacharyc:
The test that is failing is failing because the username is already in use - it's not related to the changes here.

PR #3349, opened 2024-07-31
(DOCSP-39519): Port Swift Filter Data page for consolidated docs

dacharyc:

Jira ticket: https://jira.mongodb.org/browse/DOCSP-39519

This is a port of the existing Swift page - we decided to maintain these SDK-specific query engines as one-off pages because they are too dissimilar to consolidate. I have updated the Realm naming and refs to use the `sdk` ref targets.

I switched from the Realm languages tab group to the Drivers tab group, so if someone has selected "Swift" or "Objective-C" from another page, I believe the selection should carry over. There are some examples that don't exist or apply in the other language, so if you have selected "Swift" you will still see some Objective-C examples in the NSPredicate part of the page, and if you have selected Objective-C, you will still see the entire Swift Type-Safe Query API and some Swift examples in the NSPredicate part of the page. I don't plan to go back and address these or add "missing example" placeholders, because it's the same behavior as the existing page. But I could be convinced otherwise if a reviewer feels strongly.

*Staged Page*

- [Type-Safe and NS Predicate Queries](https://preview-mongodbdacharyc.gatsbyjs.io/realm/DOCSP-39519/sdk/crud/query-engines/filter-data-swift-sdk/)

*Page Source*

Add links to every SDK's pages where you got the SDK-specific information:

- [Swift: Filter Data](https://www.mongodb.com/docs/atlas/device-sdks/sdk/swift/crud/filter-data/)

### PR Author Checklist

Before requesting a review for your PR, please check these items:

- [x] Open the PR against the `feature-consolidated-sdk-docs` branch instead of `master`
- [x] Tag the consolidated page for:
  - genre
  - meta.keywords
  - meta.description

#### Naming
- [x] Update Realm naming and the language around persistence layer/local/device per [this document](https://docs.google.com/document/d/126OczVxBWAwZ4P5ZsSM29WI3REvONEr1ald-mAwPtyQ/edit?usp=sharing)

#### Links and Refs
- [x] Create new consolidated SDK ref targets starting with "_sdks-" for relevant sections
- [x] Remove or update any SDK-specific refs to use the new consolidated SDK ref targets

### Reviewer Checklist

As a reviewer, please check these items:

- [x] Realm naming/language has been updated
- [x] All relevant content from individual SDK pages is present on the consolidated page

### DIFF
--- a/source/includes/tip-realm-query-language-reference-examples.rst
+++ b/source/includes/tip-realm-query-language-reference-examples.rst
@@ -2,5 +2,5 @@
 
    You can also find useful Realm Query Language examples on the following pages:
 
-   - :ref:`iOS Query Examples <ios-client-query-engine>`
+   - :ref:`iOS Query Examples <sdks-filter-data-swift>`
    - :ref:`Node.js Query Engine <node-client-query-engine>`
--- a/source/sdk/crud/query-engines/filter-data-swift-sdk.txt
+++ b/source/sdk/crud/query-engines/filter-data-swift-sdk.txt
@@ -1,8 +1,8 @@
 .. _sdks-filter-data-swift:
 
-=======================
-Filter Data - Swift SDK
-=======================
+=============================================
+Type-Safe and NSPredicate Queries - Swift SDK
+=============================================
 
 .. meta:: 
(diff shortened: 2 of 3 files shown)


krollins-mdb (approved):
I think you took the right approach with this page. No need to add missing example placeholders.

And no suggestions from me on this one! 🥇

PR #3352, opened 2024-07-31
Fix a slew of Node.js tests and regenerate some snippets

krollins-mdb:

This PR mostly fixes fixes tests, but made some snippet changes for a couple of pages too.

- [Counters](https://deploy-preview-3352--device-sdk.netlify.app/sdk/node/model-data/data-types/counters/)
- [Quickstart](https://deploy-preview-3352--device-sdk.netlify.app/sdk/node/quick-start/)

### Release Notes

- Mostly internal: Node SDK
  - Fixed failing tests


examples/node/v12/__tests__/quickstart.test.js
```diff
@@ -0,0 +1,203 @@
...
+    const realm = await Realm.open(config);
+    // :snippet-end:
+
+    expect(realm.isClosed).toBe(false);
+
+    const assignedTasks = realm
+      .objects(QuickstartTask)
+      .filtered(`owner_id == "${anonymousUser.id}"`);
+
+    expect(assignedTasks.length).toBe(0);
```
dacharyc, line 199: In some of the other sync quick starts, we actually write stuff, verify it uploaded, and then delete it.

Should we do that in this test?

source/sdk/node/model-data/data-types/counters.txt
```diff
@@ -89,38 +87,14 @@ To initialize a counter, create your object using the ``realm.create()`` method.
...
-    
-      .. literalinclude:: /examples/generated/node/data-types.test.snippet.initialize-counter.ts 
-          :language: typescript
-    
-    .. tab::
-      :tabid: javascript
-
-      .. literalinclude:: /examples/generated/node/data-types.test.snippet.initialize-counter.js
-          :language: javascript
+.. include:: /examples/generated/node/v12/formatted/data-types.test.snippet.initialize-counter.ts.rst
```
dacharyc, line 90: We may want to discuss how to handle these things going forward, but for the consolidated world, seeing only a `.ts` file here would make me say the JS examples are missing. I would put in a "missing placeholder" for the JS.

If we choose not to maintain two separate snippets because the code would be the same in JS *and* TS, can we use some sort of naming convention to make it clear to folks who don't regularly do JS and TS dev stuff that this snippet works for both?

dacharyc (approved):
Handful of Qs here, but none of them are blocking. Thanks for fixing up these tests!

examples/node/v12/__tests__/models/models.js
```diff
@@ -0,0 +1,32 @@
+import Realm, { BSON } from "realm";
+
+// :snippet-start: declare-counter-schema
+export class ClassWithCounter extends Realm.Object {
```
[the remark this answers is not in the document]
krollins-mdb: Yeah, I agree. I was struggling with determining how much to change the stuff Omar added. I think I ended up in a weird middle ground. I'll update the class and property names.

examples/node/v12/__tests__/models/models.js
```diff
@@ -0,0 +1,32 @@
...
+      // or myCounter: "counter"
+      nullableCounter: { type: "int", presentation: "counter", optional: true },
+      // or nullableCounter: "counter?"
+    },
+  };
+}
+// :snippet-end:
+
+// :snippet-start: define-an-object-model
+export class QuickstartTask extends Realm.Object {
```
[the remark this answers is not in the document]
krollins-mdb: Generally, yes. Best practice with the JS SDK is to define models in a separate file from where you use them. So, the default should export/import.

Worth questioning here though. In the Reactive Native docs, the standard is pretty well established. I don't think as clear in the Node docs. I'll think about this a bit more.

examples/node/v12/__tests__/quickstart.test.js
```diff
@@ -0,0 +1,203 @@
...
+    realm.close();
+
+    expect.hasAssertions();
+  });
+});
+
+describe("Quickstart Sync", () => {
+  test("should open a Flexible Sync realm with initial subscriptions", async () => {
+    // :snippet-start: open-realm-with-subscriptions
+    // :replace-start: {
```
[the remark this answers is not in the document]
krollins-mdb: Happy to standardize - though I don't think any of the JS docs have ever done that. 😅

examples/node/v12/__tests__/quickstart.test.js
krollins-mdb: Probably. If I don't have time to do it in this PR, I'll create a Jira ticket so that the work is documented.

source/sdk/node/model-data/data-types/counters.txt
krollins-mdb: Good point. Yeah, we'll need something. JS tests should be TypeScript first, even when there isn't any TypeScript code. The reasoning being that the compiler can catch all sorts of errors with plain ol' JS.

I'll do some thinking on this one too.

PR #3353, opened 2024-08-01
(DOCSP-42045): Fix wrong C++ tab ID

dacharyc:

Jira ticket: https://jira.mongodb.org/browse/DOCSP-42045

*Staged Page*

- [Manage Sync Sessions](https://preview-mongodbdacharyc.gatsbyjs.io/realm/DOCSP-42045/sdk/sync/manage-sync-sessions/): Page correctly changes code examples when you switch language tabs.
- [Help](https://preview-mongodbdacharyc.gatsbyjs.io/realm/DOCSP-42045/help/): The Help page was using realm-sdks tab set - I changed it to Drivers tab set so it would preserve the language selection when the user switched to the page. I also added a couple of notes related to naming because there are some naming things here that still reflect the old name, and we can't change them.

### DIFF
--- a/source/help.txt
+++ b/source/help.txt
@@ -4,16 +4,33 @@
 Get Help
 ========
 
+.. meta::
+   :description: Get help for Atlas Device SDK through community forums, sharing feedback, or professional support.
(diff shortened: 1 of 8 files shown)


krollins-mdb (approved):
LGTM! Thanks for fixing the tabset and updating the help page. 🥇

PR #3354, opened 2024-08-01
Updates Dart tests to use new notification API

MongoCaleb:

Jira ticket: hhttps://jira.mongodb.org/browse/DOCSP-40843

### DIFF
--- a/examples/dart/test/open_flexible_sync_realm_test.dart
+++ b/examples/dart/test/open_flexible_sync_realm_test.dart
@@ -26,8 +26,8 @@ class _Car {
 
 void main() {
   group('Open Flexible Sync Realm', () {
-    const APP_ID = "flutter-flexible-luccm";
-    final appConfig = AppConfiguration(APP_ID);
+    const appId = "flutter-flexible-luccm";
+    final appConfig = AppConfiguration(appId);
     final app = App(appConfig);
     test("Open Flexible Sync Realm", () async {
       final credentials = Credentials.anonymous();
@@ -71,31 +71,24 @@ void main() {
     test('Track download progress', () async {
       final credentials = Credentials.anonymous();
       final currentUser = await app.logIn(credentials);
-      late int transferred;
-      late int transferable;
+      late double progress = -1;
       final config = Configuration.flexibleSync(currentUser, [Tricycle.schema]);
       // :snippet-start: async-open-track-progress
-      final realm =
-          await Realm.open(config, onProgressCallback: (syncProgress) {
-        if (syncProgress.transferableBytes == syncProgress.transferredBytes) {
-          print('All bytes transferred!');
-          // :remove-start:
-          transferred = syncProgress.transferredBytes;
-          transferable = syncProgress.transferableBytes;
-          // :remove-end:
-        }
+      final realm = await Realm.open(config, onProgressCallback: (syncProgress) {
+            progress = syncProgress.progressEstimate;
+            // Percent complete == progress * 100
+          if (syncProgress.progressEstimate == 1.0) {
+             //transfer is complete
+          }
       });
       // :snippet-end:
       expect(realm.isClosed, false);
-      expect(transferred, transferable);
-      expect(transferred, greaterThanOrEqualTo(0));
+      expect(progress, greaterThanOrEqualTo(0));
       cleanUpRealm(realm, app);
     });
     test('Cancel download in progress', () async {
       final credentials = Credentials.anonymous();
       final currentUser = await app.logIn(credentials);
-      late int transferred;
-      late int transferable;
       final config = Configuration.flexibleSync(currentUser, [Tricycle.schema]);
       // :snippet-start: async-open-cancel
       final token = CancellationToken();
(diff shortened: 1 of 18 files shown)


examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -71,31 +71,24 @@ void main() {
...
+            // Percent complete == progress * 100
+          if (syncProgress.progressEstimate == 1.0) {
+             //transfer is complete
+          }
       });
       // :snippet-end:
       expect(realm.isClosed, false);
-      expect(transferred, transferable);
-      expect(transferred, greaterThanOrEqualTo(0));
+      expect(progress, greaterThanOrEqualTo(-1));
```
krollins-mdb, line 86: This expect statement will always evaluate to true because we set `progress` to `-1` on line 74. So, sync progress could be nothing or it could be complete. I think we'd want to test that the sync is complete, so should be looking for greater than `-1`, not equal to.

I _think_ the `await` on realm open will wait for sync to finish, but I'm not certain of the default behavior for the Flutter SDK. Either way, we'll need to make sure we're waiting for sync to finish so that we can get the expected result.

Alternatively, we could change the type and value of `progress`. Within the `if` block on line 80, we could set `progress` to "Transfer complete" or something. Then assert equality between the expected string and what is actually in `progress`.

examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -71,31 +71,24 @@ void main() {
...
-          // :remove-start:
-          transferred = syncProgress.transferredBytes;
-          transferable = syncProgress.transferableBytes;
-          // :remove-end:
-        }
+      final realm = await Realm.open(config, onProgressCallback: (syncProgress) {
+            progress = syncProgress.progressEstimate;
+            // Percent complete == progress * 100
+          if (syncProgress.progressEstimate == 1.0) {
+             //transfer is complete
```
krollins-mdb, line 81:
```suggestion
             // Transfer is complete
```

krollins-mdb (changes requested):
Thanks for working on this, @MongoCaleb! I have a couple suggestions for the updated test. And a couple of things beyond my in-line comments:

- We may need to update the "Monitor sync progress", in the `manage_sync_session_test.dart` file. You may have already looked at this, but wanted to make sure.
  - It looks like we're already using the correct API for the [monitor progress section](https://deploy-preview-3354--device-sdk.netlify.app/sdk/flutter/sync/manage-sync-session/#monitor-sync-upload-and-download-progress) of the Manage Sync Sessions page. However, we need to confirm that the content around the example describes the behavior and API correctly. Looks like this section was updated for v2.0.0 and probably doesn't account for the changes in the latest version.
- Can you add a Release Notes section to the PR? You're doing release notes this time, so it should make your life a bit easier later. 😄

examples/dart/test/open_flexible_sync_realm_test.dart
MongoCaleb: General comment above: I believe the page content around each of the code snippets is correct. However, the one you link (Monitor Progress) shows using [getProgressStream](https://pub.dev/documentation/realm/latest/realm/Session/getProgressStream.html). This is *not* deprecated, so I think it's good to leave in here. Do you think it should be replaced with the other way to do it (as shown in this PR)?

MongoCaleb:
Note: we are keeping information about both `getProgressStream` and `onProgressCallback` because the latter is really a fix to make things simplier for the user, but is, behind the scenes, just a wrapper over the former. As per Kasper Nielsen: 

getProgressStream is the most important I think. It can be used throughout the apps lifetime, and can be customized with both direction and mode.
In a sense onProgressCallback exists due to arguably bad design choice of combining setup and open of a realm file in one operation, preventing the user from calling getProgressStream before the realm is opened. But, alas history is history.. and as such it is also important.

examples/dart/test/open_flexible_sync_realm_test.dart
krollins-mdb: Per the new info from Kaspar, we're good to have both.

Can we add a mention of `getProgressStream` to the section where the `onProgressCallback` example is used? We don't talk about the fact that they're really the same thing anywhere. It would be nice to have info about when devs might use one instead of the other. But at least acknowledging that there are two ways to add progress notifications would be great.

krollins-mdb: Also, thanks for digging into this! I had no idea there were two ways to do this. 😆

examples/dart/test/open_flexible_sync_realm_test.dart
```diff
@@ -68,34 +67,63 @@ void main() {
...
+    realm2.add(Tricycle(1001, "a1001"));
+    realm2.add(Tricycle(2002, "a2002"));
+    realm2.add(Tricycle(3003, "a3003"));
+  });
+
+  await realm2.syncSession.waitForUpload();
+  await realm2.syncSession.waitForDownload();
+  realm2.close();
+  return config1;
+}
```
krollins-mdb, line 105: Is there a reason to have these functions outside of the "Track upload progress" test that starts on line 106? They're only for that test, right?

krollins-mdb (approved):
Thanks for your work on this, @MongoCaleb! LGTM.

I have a couple of comments you should consider before merging, but they're not blocking comments.

This one in particular could help devs navigate this feature a bit: https://github.com/mongodb/docs-realm/pull/3354#discussion_r1711856424

MongoCaleb:
@krollins-mdb added notes.

PR #3399, opened 2025-08-11
Update "npm install realm" to use v12 explicitly

kraenhansen:
## Pull Request Info

The latest (non-sync) "community" edition of the Realm JS SDK was published on the "latest" channel to help migrate more off the sync-enabled v12. This means running `npm install realm` will download the community edition and not v12 as before.

With this PR I propose updating MongoDB related docs to explicitly install `realm@12` to avoid breaking existing content.


cbullinger (approved):
thanks @kraenhansen! LGTM

